PR #1, opened 2018-07-10
Make app more idempotent to rails app:update

benthorner:
https://trello.com/c/CPWJzL9y/5-create-skeleton-rails-app

### DIFF
--- a/bin/setup
+++ b/bin/setup
@@ -20,6 +20,11 @@ chdir APP_ROOT do
   # Install JavaScript dependencies if using Yarn
   # system('bin/yarn')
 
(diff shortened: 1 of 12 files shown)


emmabeynon:
@benthorner Could you add some more context in the PR description for this?  I'm not sure I understand what this is for.

benthorner:
When a new version of rails comes out, we should be able to run `rails app:update`, which will update rails and attempt to make changes to certain files. The more we deviate from a standard rails config, the more difficult this update process becomes; the changes in the PR were those made by running `rails app:update` and will help keep future diffs shorter when we need to run this command.

PR #2, opened 2018-07-10
Remove gems until we need them, simplify specs

benthorner:
https://trello.com/c/CPWJzL9y/5-create-skeleton-rails-app

The preference from Tijmen and Kevin is to keep the spec_helper short and avoid having a rails_helper and a spec_helper.


emmabeynon:
The card for this specifies that we should be adding in sidekiq, rspec, gds-sso and govuk_app_config gems.  Have I missed a conversation that has changed this?

benthorner:
Oops - one too many. We arrived at only including the gems we need right now, so we can lose Sidekiq as we don't have any jobs yet and `rake` still runs without it. The app_config one we should keep (I'll put it back) as the next thing to do will be to check logs 'n stuff work.

Gemfile
```diff
@@ -2,30 +2,21 @@ ruby File.read(".ruby-version").strip
...
+gem "pg", "~> 1"
+gem "plek", "~> 2"
+
+group :development do
+  gem "byebug"
   gem "govuk-lint"
-  gem "poltergeist"
-  gem "pry"
+  gem "listen"
   gem "rspec-rails"
```
emmabeynon: Shall we move this to the test group?

PR #6, opened 2018-07-11
More cleanup for the skeleton rails app

benthorner:
* Remove app/jobs, vendor/, bin/update, package.json, config/puma.rb
  * Upgrade to ruby 2.5.1
  * Enable rubocop check for FrozenStringLiteralComment
  * Version all gems (including dev/test)
  * Add designated port to Procfile


Procfile
```diff
@@ -1 +1 @@
-web: bundle exec unicorn -c ./config/unicorn.rb -p ${PORT}
+web: bundle exec unicorn -c ./config/unicorn.rb -p ${PORT:-3221}
```
kevindew: Great port 🥇 💯

db/migrate/20160622154200_create_users.rb
```diff
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
```
kevindew: Not really related to this file. But how did we get such an old migration?

Gemfile
```diff
@@ -11,15 +13,15 @@ gem "pg", "~> 1"
...
-  gem "govuk-lint"
-  gem "listen"
+  gem "byebug", "~> 10"
+  gem "govuk-lint", "~> 3"
+  gem "listen", "~> 3"
 end
 
 group :test do
-  gem "simplecov", require: false
+  gem "simplecov", "~> 0.16"
```
kevindew: do you mean to strip the require from this?

benthorner: Yes - doesn't seem to affect anything...

db/migrate/20160622154200_create_users.rb
benthorner: Another gift from that script repo...

kevindew:
Shall we get a quick PR in so we have CI?

benthorner:
I think @tijmenb was doing something there...

PR #18, opened 2018-07-16
Create the smallest working document workflow

tijmenb:
This PR is pretty much the least we can do to get a working workflow where a user chooses a format (out of a choice of 1), updates the title and saves the document.

Nothing in this PR is intended to be set in stone - it's likely that most of it will change in the coming months. It's here to get something started and we can start on more stories to build out parts of the interface.

- After a document is created, we redirect to the edit page. I think this would a nice thing to keep, because we don't have to build a separate "create" workflow
- There's just 1 format to choose from now
- Note that the routes are a mess, and should probably use `resources`, but I wanted to keep an open mind on how RESTful the application will be.
- We'll probably want to use better form helpers to create our forms.
- We'll extract creating and updating of models into service objects - doing it like this is recipe for a mess

https://trello.com/c/i3wG2KjR


erino:
UTF-8 works as I was able to make a story with an emoji

app/controllers/content_controller.rb
```diff
@@ -0,0 +1,35 @@
+# frozen_string_literal: true
+
+class ContentController < ApplicationController
```
emmabeynon: Just a small thing, but I wonder if it would make sense to rename this `DocumentController` as all the actions we have here are to do with Documents?  I feel like some of the difficulties with getting to grips with Whitehall centre on lack of consistency around Document vs. Edition vs. Content Item so getting consistency right early on in this app could be helpful.

config/routes.rb
```diff
@@ -1,7 +1,14 @@
 # frozen_string_literal: true
 
 Rails.application.routes.draw do
-  root to: 'development#index'
+  root to: redirect('/content')
+  get '/dev' => 'development#index'
+
+  get '/content' => 'content#index'
+  get '/content/new' => 'content#choose_format', as: :new_document
+  post '/content/create' => 'content#create', as: :create_content
```
emmabeynon: Similar story here - maybe we need to make a decision as to whether we call this "content" or "document"?

tijmenb:
@emmabeynon good points! All renamed.

db/schema.rb
```diff
@@ -10,11 +10,21 @@
...
+ActiveRecord::Schema.define(version: 2018_07_16_161306) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
 
+  create_table "documents", force: :cascade do |t|
+    t.string "content_id", null: false
+    t.string "locale", null: false
+    t.string "document_type"
+    t.string "title"
```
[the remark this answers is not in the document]
tijmenb: That's no longer the plan! @erino will be writing an ADR to explain that decision.

PR #25, opened 2018-07-20
ADR 2 - Use local datastore

erino:


docs/adr/0002-use-local-datastore.md
```diff
@@ -0,0 +1,59 @@
...
+Date: 2018-07-20
+
+## Context
+We're building a new publishing architecture and MVP workflow in private beta. We have taken the decision to build this in a new application and therefore we need to decide how we use the publishing-api.
+
+One of the goals of the migration effort from 2016 was to for all publishing applications to be clients of the publishing-api and this API would means that these applications would not need local datastores. This is written up in the [GOV.UK Platform Architecture Goals 2017-18+][goals]
+
+[goals]: https://docs.google.com/document/d/1Oft4akc6dZfhhOjosNPbFpcLUOUjz7YG7QPcVZi8hww/edit#heading=h.5uytjxbfoe58
+
+Since writing these goals we have built the specialist-publisher application using this pattern. The practicalities of using the pattern
```
emmabeynon: Looks like this sentence is unfinished.

docs/adr/0002-use-local-datastore.md
```diff
@@ -0,0 +1,59 @@
...
+## Context
+We're building a new publishing architecture and MVP workflow in private beta. We have taken the decision to build this in a new application and therefore we need to decide how we use the publishing-api.
+
+One of the goals of the migration effort from 2016 was to for all publishing applications to be clients of the publishing-api and this API would means that these applications would not need local datastores. This is written up in the [GOV.UK Platform Architecture Goals 2017-18+][goals]
+
+[goals]: https://docs.google.com/document/d/1Oft4akc6dZfhhOjosNPbFpcLUOUjz7YG7QPcVZi8hww/edit#heading=h.5uytjxbfoe58
+
+Since writing these goals we have built the specialist-publisher application using this pattern. The practicalities of using the pattern 
+
+We are about to build another publishing-api client and we should now use the learnings we have from building specialist-publisher to ensure that the not using a local datastore is the best decision.
```
emmabeynon: I think there's an errant "the" here => "to ensure that the not using"

docs/adr/0002-use-local-datastore.md
```diff
@@ -0,0 +1,59 @@
...
+- Expanded Links
+- Combination of local and publisher API workflows
+- Hard to feedback to use on asynchronous API updates
+- Content-Schema may not matching Rails validations causing unhelpful error messages
+
+## Decision
+We will use a local datastore which syncs to the publishing api.
+
+## Status
+
```
emmabeynon: Looks like an errant empty line here.

docs/adr/0002-use-local-datastore.md
```diff
@@ -0,0 +1,58 @@
...
+In this option we continue with the platform architecture goals.
+
+#### Pros
++ Single source of truth - all data is in the publishing api
++ Multiple apps can edit the same content
++ Single implementation of scheduling and publishing logic
++ Single abstraction layer
+
+#### Cons
+- Effectively building an Active Resource layer
```
davbo: I don't have enough Rails experience to see how this is a con? perhaps if it's expanded a bit that would help

docs/adr/0002-use-local-datastore.md
```diff
@@ -0,0 +1,58 @@
...
+- Harder to resolve performance problems
+- Current workflow is opinionated and may not match needs of content
+- Validation can be difficult to explain to user
+- Hard to deploy
```
davbo: Deploying an app which holds less state seems like it would be easier to deploy?

docs/adr/0002-use-local-datastore.md
```diff
@@ -0,0 +1,58 @@
+# 2: Use local datastore for publisher workflow
+
+Date: 2018-07-20
+
+## Context
+We're building a new publishing architecture and MVP workflow in private beta. We have taken the decision to build this in a new application and therefore we need to decide how we use the publishing-api.
+
+One of the goals of the migration effort from 2016 was to for all publishing applications to be clients of the publishing-api and this API would means that these applications would not need local datastores. This is written up in the [GOV.UK Platform Architecture Goals 2017-18+][goals]
```
cbaines: > not need local datastores

I don't think this was ever a serious aim. The Publishing API is just for content, not everything an application might need to store.

I've always interpreted this as a unclear way of saying "only store content in the Publishing API". That's my reading of the referenced goals document as well, as it says "all content stored within the Publishing API, not relying on their own local datastores".

docs/adr/0002-use-local-datastore.md
```diff
@@ -0,0 +1,58 @@
...
+#### Pros
++ Single source of truth - all data is in the publishing api
++ Multiple apps can edit the same content
++ Single implementation of scheduling and publishing logic
++ Single abstraction layer
+
+#### Cons
+- Effectively building an Active Resource layer
+- Harder to resolve performance problems
+- Current workflow is opinionated and may not match needs of content
```
cbaines: How does this related to where the data is stored? Even if you were to keep a copy of the content locally, you'd still be using the Publishing API right, and thus still be involved and bound somewhat by it's supported workflow.

docs/adr/0002-use-local-datastore.md
```diff
@@ -0,0 +1,58 @@
...
+- Validation can be difficult to explain to user
```
cbaines: Assuming this related to the validation the Publishing API performs, I only see storing the content locally as delaying the validation, not avoiding it. Do you have a different view?

docs/adr/0002-use-local-datastore.md
```diff
@@ -0,0 +1,58 @@
...
+- Hard to deploy
+
+### Option 2: Use a local datastore
+#### Pros
++ It’s standard Rails so quick to build and easy to understand
++ Easy to test, testing over an API is not easy and slow
```
cbaines: None of the current publishing application test suites use the Publishing API, so I'm not sure how this applies? There is the Publishing End to end tests, but they explicitly are meant to use the Publishing API.

docs/adr/0002-use-local-datastore.md
```diff
@@ -0,0 +1,58 @@
...
++ Easy to iterate workflow, changing the API workflow is much harder
```
cbaines: Referring to my comments here https://github.com/alphagov/content-publisher/pull/25/files#r204017078 , there has been previous discussion on how much "workflow" the Publishing API should support, as at the moment, it only provides some minimal things (e.g. draft, published), and leaves it up to applications to add any additional layers (e.g. fact check).

In my mind, iterating the workflow in this way, building on top of the Publishing API is the way to go, at least initially.

cbaines: Well, for most features I can imagine, like fact checking.

docs/adr/0002-use-local-datastore.md
```diff
@@ -0,0 +1,58 @@
...
++ Only final state of content can be put into publishing-api - some formats will need interim data which isn’t needed for publishing
```
cbaines: Using the Publishing API for storing content doesn't exclude storing interim data locally. If it isn't needed for publishing, it probably isn't necessary to go in to the Publishing API at all.

docs/adr/0002-use-local-datastore.md
```diff
@@ -0,0 +1,58 @@
...
++ No new features needed for meta data - notes, and other publisher info
+
+#### Cons
+- Syncing between local and publishing api
+- Local data is hard to query by other applications
+- Expanded Links
+- Combination of local and publisher API workflows
```
cbaines: `s/publisher API/Publishing API/`?

Also, I don't really see this as a bad thing. Can you elaborate on your thoughts on dividing responsibility regarding the workflow between the Publishing API, and the publishing application?

docs/adr/0002-use-local-datastore.md
```diff
@@ -0,0 +1,58 @@
...
+- Hard to feedback to use on asynchronous API updates
+- Content-Schema may not matching Rails validations causing unhelpful error messages
```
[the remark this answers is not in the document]
erino: Fixed in 3cc865d3a11f029d95e4d960a38635a2599cc616

docs/adr/0002-use-local-datastore.md
erino: Hmm, now you mention it that makes sense. I had it in my notes but I can't think why. Removed in 0a94c9b67d71fbd76a9d03bd85612d352fca03ae

docs/adr/0002-use-local-datastore.md
erino: Maybe I misunderstood the document, but I was under the impression that it was the aim to avoid using a local datastore. Which is why specialist publisher is built the way it is.

docs/adr/0002-use-local-datastore.md
erino: If all of the data is stored in the publishing API then it's bound by it's workflow. Where's if you're building you're own data locally then it's possible to deviate from the workflow to better meet the needs of the content. Admittedly this workflow still needs to fit into the publishing-apis, but with the current functionality we can build support for 2i or fact check in the local app.

docs/adr/0002-use-local-datastore.md
erino: This relates to the errors from validating content against the govuk-content-schemas. I think you're right that storing content later only delays the (potentially cryptic) validation error.

docs/adr/0002-use-local-datastore.md
erino: This was in the context of the specialist publisher where the publishing api has to be mocked for a large part of the test suite.

What about changing the item to be:
> Easy to test, testing over a remote API involves a lot of stubbing or running a version of the API as part of the test suite.

erino: Updated in 4ce879a7b40632a614e2dff2c386aab6263dc8df

docs/adr/0002-use-local-datastore.md
erino: Yes, maybe all features for the Publishing API should be developed locally as candidates for addition.

docs/adr/0002-use-local-datastore.md
erino: This goes back to the intent of the Architecture Goals, does it mean absolutely no datastore or only interim data.

docs/adr/0002-use-local-datastore.md
erino: If a format needs a different workflow to the publishing api then you'll need to keep two state machines in sync. This may be a subset of the above point - Syncing between local and publishing api. Let me try rewording it

Also f1370cd

docs/adr/0002-use-local-datastore.md
cbaines: > Maybe I misunderstood the document, but I was under the impression that it was the aim to avoid using a local datastore. Which is why specialist publisher is built the way it is.

In my mind, this is only for content. Specialist Publisher has a local datastore, it uses it for storing user information. We should try and clarify this, do you see it as applying to any and all data?

docs/adr/0002-use-local-datastore.md
cbaines: I guess I don't see the Publishing API being limiting. Particularly in the case of things like 2i or fact check, I can imagine ways to implement that without needing to keep a copy of the content locally in the publishing application.

docs/adr/0002-use-local-datastore.md
cbaines: My unauthoritative view, is that it's mostly about having a single source of truth for content, not storing this locally in the Publishing application, but we should get consensus on the interpretation of the Architecture Goals somehow....

erino: The first paragraph Section 1 from the Architecture Goals is as follows

> The original goal was that the Publishing API was the “source of truth” for all content, and that individual publishing applications should effectively be stateless, relying on the Publishing API as their “database.”

Given the above, especially the point about statelessness I would say that it means that new publishing apps should not use a local datastore. Also @timblair is back from holiday on Wednesday so we can ask him then.

docs/adr/0002-use-local-datastore.md
erino: I think you're correct that we could build a review process into the Publishing API. But the additional complexity added to the API and additional time needed for development would out weigh the benefits. In this quarters mission we need to iterate and build the new workflow quickly, trying to build this in the Publishing API is probably not the best use of our time.

docs/adr/0002-use-local-datastore.md
erino: I see it as being applied to all data, see here for an [expansion](https://github.com/alphagov/content-publisher/pull/25#discussion_r204361821) . Although I would say that user data is a special case as that's required by OmniAuth.

docs/adr/0002-use-local-datastore.md
cbaines: Personally, I actually dislike the idea of adding any functionality around review processes in to the Publishing API.

In my last comment, what I was probably unclear about is that I can imaging keeping the content in the Publishing API, and keeping the details of fact check/2i review in the publishing app, so building _on_ the Publishing API, instead of building _in_ the Publishing API.

docs/adr/0002-use-local-datastore.md
```diff
@@ -0,0 +1,59 @@
...
+- Expanded Links
+- Combination of local and Publishing API workflows
+- Hard to feedback to use on asynchronous API updates
+- Content-Schema may not match Rails validations causing unhelpful error messages
+
+## Decision
+We will use a local datastore which syncs to the Publishing API.
+
+## Status
+Accepted
```
kevindew: Consistency heads up - status is the first heading in the initial ADR: https://github.com/alphagov/content-publisher/blob/master/docs/adr/0001-record-architecture-decisions.md

docs/adr/0002-use-local-datastore.md
```diff
@@ -0,0 +1,59 @@
...
+
+We are about to build another [Publishing API] client and we should now use the learnings we have to ensure that not using a local datastore is the best decision.
+
+### Option 1: No local datastore
+In this option we continue with the platform architecture goals.
+
+#### Pros
++ Single source of truth - all data is in the Publishing API
++ Multiple apps can edit the same content
++ Single implementation of scheduling and publishing logic
```
kevindew: Nitpick - Publishing API doesn't support scheduling.

docs/adr/0002-use-local-datastore.md
erino: Looking at the document linked to in the first ADR then status is in the correct place here, and not in the other ADRs

erino:
@kevindew I've added you're comments. Let me know what you think

docs/adr/0002-use-local-datastore.md
```diff
@@ -0,0 +1,72 @@
...
+
+To achieve the aims of rapidly building a new publishing application to be ready for beta we're able to do this faster and more effectively by modelling these concepts in the local database.
+
+## Decision
+We will use a local datastore which syncs to the Publishing API.
+
+## Status
+Accepted
+
+## Consequences
```
timblair: I agree with this decision. Given that, we also need to review and update the Architecture Goals document to be consistent with it.

kevindew:
Thanks @erino I noticed a couple of typos in my bit so I'll just fix them.

kevindew:
I've updated the commit with fixes and opened https://github.com/alphagov/content-publisher/pull/74 for the other ADRs

docs/adr/0002-use-local-datastore.md
erino: @timblair I've added this as a consequence in 757f1068092548ea77cf702c4f6cd4abb08a56ab

PR #30, opened 2018-07-23
Validate the format configuration

tijmenb:
These tests validate that our format config is internally consistent and all the required fields have been set.

Also fixes 2 mistakes in the current config from https://github.com/alphagov/content-publisher/pull/26: `detailed_guides ` was written plural, and `publication` is not a document type (it's a schema).

https://trello.com/c/egKyMGRP


spec/formats/configuration_spec.rb
```diff
@@ -0,0 +1,49 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+RSpec.describe "Format configuration" do
+  supertypes = YAML.load_file("app/formats/supertypes.yml")
+  document_types = YAML.load_file("app/formats/document_types.yml")
+
+  describe "Supertypes" do
```
benthorner: Could we collapse these two describes into one to make it a bit more readable?

benthorner: (Or the inner describe and the it - it's very block-tastic at the moment ;-).

spec/formats/configuration_spec.rb
```diff
@@ -0,0 +1,49 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+RSpec.describe "Format configuration" do
+  supertypes = YAML.load_file("app/formats/supertypes.yml")
```
benthorner: Would a 'let' be more conventional here?

benthorner:
With the format files, I'm confused as to why we've put these config files in the app/ directory - usually config files go in config/ and get loaded into the environment so we don't have to keep doing 'YAML.load' everywhere. Could we look at doing that?

spec/formats/configuration_spec.rb
tijmenb: Good point, I merged the two `describes`!

spec/formats/configuration_spec.rb
tijmenb: As it turns out, you can only use `let` inside an `it` block (there are other problems with `let` but we'll talk about those later :)).

tijmenb:
> With the format files, I'm confused as to why we've put these config files in the app/ directory - usually config files go in config/ and get loaded into the environment so we don't have to keep doing 'YAML.load' everywhere. Could we look at doing that?

@benthorner I would have thought you'd seen it in https://github.com/alphagov/content-publisher/pull/26/commits/7b1d315af7d57943a0b9cd30510bdde3d4d26da0 😄 

> Added the YAML files to `/app/formats` which may be controversial. Not sure where to put them though - they're so core to the app that `config` feels wrong.

and

I'm expecting follow up PRs to:

- Refactor YAML handling. I've resisted the temptation to refactor the `YAML.load_file` stuff into classes at this time, because I want to avoid making the decisions about naming / placing stuff now. For what it's worth, I'm thinking about something like `DocumentType.all` and `DocumentType.find("press_release")`.

Basically, the YAML files feel more business logic than config.

How do you get to the contents of a file from the contents?

spec/formats/configuration_spec.rb
tijmenb: Clarification: you can only use `let` in a `describe` block, and the resulting variable inside an `it` block. Otherwise it will raise an error like:

> `foo` is not available on an example group (e.g. a `describe` or `context` block). It is only available from within individual examples (e.g. `it` blocks) or from constructs that run in the scope of an example (e.g. `before`, `let`, etc).

PR #31, opened 2018-07-23
Send people who select "Not sure" to guidance page

tijmenb:
When somebody selects "I'm not sure this should be on GOV.UK", they'll now be redirected to the guidance page.

To achieve this, we've:

- Added a guidance page with the content from https://github.com/alphagov/whitehall-prototype/commit/503d77eb49d0c4e36457039dff23fa7a246ab65f (which is wrong in parts, but that will be fixed later)
- Removed the "I'm not sure this should be on GOV.UK" supertype from the supertypes YAML, because it's unlike the others.
- Created a temporary `SupertypeThing` abstraction that captures some of the logic to generate the radio button input. This will be refactored in a future pull request (likely together with the form
validation).

Paired with @benthorner.

https://trello.com/c/egKyMGRP

### DIFF
--- a/app/controllers/new_document_controller.rb
+++ b/app/controllers/new_document_controller.rb
@@ -2,10 +2,15 @@
 
 class NewDocumentController < ApplicationController
   def choose_supertype
-    @supertypes = YAML.load_file("app/formats/supertypes.yml")
+    @supertype_radio_options = SupertypeThing.new.options
   end
 
   def choose_document_type
+    if params[:supertype] == SupertypeThing::NOT_SURE_OPTION.fetch(:value)
+      redirect_to guidance_path
+      return
+    end
+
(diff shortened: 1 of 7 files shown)


app/controllers/new_document_controller.rb
```diff
@@ -2,10 +2,15 @@
 
 class NewDocumentController < ApplicationController
   def choose_supertype
-    @supertypes = YAML.load_file("app/formats/supertypes.yml")
+    @supertype_radio_options = SupertypeThing.new.options
   end
 
   def choose_document_type
+    if params[:supertype] == SupertypeThing::NOT_SURE_OPTION.fetch(:value)
```
benthorner: We should make this a question method for the SupertypeThing.

PR #32, opened 2018-07-23
Add nomenclature based on namorama session

benthorner:


README.md
```diff
@@ -2,6 +2,13 @@
 
 A unified publishing application for content on GOV.UK
 
+## Nomenclature
+
+  * Content - Some text (and related fields) a user wants to publish
+  * Revision - A version of a piece of content in a particular locale
+  * Edition - A revision that is currently in a content store (draft/live)
```
kevindew: Should this be a revision that is in the Publishing API? e.g. we have all the past editions in Publishing API

README.md
```diff
@@ -2,6 +2,13 @@
 
 A unified publishing application for content on GOV.UK
 
+## Nomenclature
+
+  * Content - Some text (and related fields) a user wants to publish
```
kevindew: I'm not sure if we need this term - I don't think it's a concrete enough concept that we'd use.

README.md
benthorner: Any better now?

README.md
benthorner: I was thinking that, since I mentioned the word 'content' in these definitions, it makes sense to explain that it's not just a block of text. I added a link to content-schemas to clarify further - does that help?

README.md
kevindew: Yes - though this implies that Publishing API exists on draft and live stack when that happens when you get to Content Store. I'd just drop off the stack part.

README.md
kevindew: Yup cool 👍

README.md
```diff
@@ -1,7 +1,16 @@
...
+
 # Content Publisher
 
 A unified publishing application for content on GOV.UK
 
+## Nomenclature
+
+  * Content - Some text ([and related fields][content-schemas]) a user wants to publish
+  * Revision - A version of a piece of content in a particular locale
+  * Edition - A revision that is in the Publishing API (draft or live stack)
```
erino: Should we remove draft? as didn't we talk about pushing each revision to the draft stack which would then mean we'd end up with an edition for each revision.

Then if an edition is a revision which exists in the live stack then we could say then an edition is explicitly created when the revision is published.

README.md
```diff
@@ -1,7 +1,16 @@
+[content-schemas]: https://github.com/alphagov/govuk-content-schemas
+
 # Content Publisher
 
 A unified publishing application for content on GOV.UK
 
+## Nomenclature
+
+  * Content - Some text ([and related fields][content-schemas]) a user wants to publish
```
erino: What about adding images and attachements

README.md
benthorner: I think we should keep the reference to draft/live, as we know people already refer to 'draft edition' as a thing. Can you think of a better word to use for the revision that exists in the draft content store?

README.md
benthorner: I think we should stick to text+related fields, unless we're planning to incorporate images/attachments into our revisions/editions. This is a good question - we should be consistent with the word we use to define our other definitions ;-)

benthorner:
@kevindew Are you doing a PR this should be waiting for?

kevindew:
@benthorner no my PR is already merged: https://github.com/alphagov/content-publisher/pull/34

README.md
erino: Also at the moment we don't know how we're going to handle images/attachments so lets leave it out for now

PR #37, opened 2018-07-24
Refactor Document type / Super type concepts

tijmenb:
This refactors the app to use the `DocumentTypeSchema` and `SupertypeSchema` classes. Also fixes the bug that we tried to fix in https://github.com/alphagov/content-publisher/pull/36.

Best reviewed per-commit, and with reading the commit message about ActiveModel.

https://trello.com/c/PxtaaObC/69-how-can-we-define-format-rules-l

### DIFF
--- a/app/controllers/new_document_controller.rb
+++ b/app/controllers/new_document_controller.rb
@@ -11,7 +11,8 @@ class NewDocumentController < ApplicationController
       return
     end
 
-    @document_types = YAML.load_file("app/formats/document_types.yml").select { |s| s["supertype"] == params[:supertype] }
+    supertype_schema = SupertypeSchema.find(params[:supertype])
+    @document_types = supertype_schema.document_types
   end
 
   def create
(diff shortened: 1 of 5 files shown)


app/controllers/new_document_controller.rb
```diff
@@ -48,11 +49,11 @@ def options
   private
 
     def supertype_options
-      YAML.load_file("app/formats/supertypes.yml").map do |supertype|
+      SupertypeSchema.all.map do |supertype|
```
benthorner: Not related to this PR, but could we put this in the view for now, to be consistent `choose_document_type.html.erb`?

tijmenb: Which parts would you put in the view? The entire class?

benthorner: It's the map part - for doc types we're doing a map in the view, whereas for supertypes we're doing a map in the controller. It seems better to do it in the view, as it's view-specific.

tijmenb: Not sure how I would put this in a view though? We're doing the mapping inside a class:

https://github.com/alphagov/content-publisher/blob/903b6f29decdd3429dac01248d7d56c62339516b/app/controllers/new_document_controller.rb#L35-L61

benthorner: Oh, I meant just this bit (assigning @supertype_schemas in the controller).

```
      @supertype_schemas.all.map do |supertype|
        {
          value: supertype.id,
          text: supertype.label,
          hint_text: supertype.description,
          bold: true,
        }
      end
    end
```

Compare with line 7 of choose_document_types.html.erb.

benthorner: Ah, sorry, just understood. I guess it'll get refactored anyway as part of the SupertypeThing.

PR #38, opened 2018-07-24
Add database cleaner for testing

benthorner:
https://trello.com/c/Vbwd7cye/68-allow-sending-news-to-publishing-api-m

This stops the test database filling up with data, which was making the
Capybara page diffs (even) more difficult to read when a feature test
fails. I've left the default strategy (transaction) as-is on the basis
that my tests continue to run fast locally.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -20,6 +20,7 @@ group :development do
 end
 
 group :test do
+  gem "database_cleaner"
   gem "simplecov", "~> 0.16"
 end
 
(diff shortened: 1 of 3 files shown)


benthorner:
They might do, but in general it's a bit weird if the test database is full of unused data that's never cleaned up.

kevindew:
@benthorner yeah I thought by default activerecord tests run in a transaction so they don't commit to the database meaning you don't get stuff left and don't need database cleaner. I could be wrong though and/or may not apply to feature tests. From googling `use_transactional_tests` seems to be the option

benthorner:
Ah yes, I thought Tijmen mentioned something about that but assumed we'd turned it on already. Closing in favour of #42.

PR #40, opened 2018-07-24
Publish documents

benthorner:
https://trello.com/c/Vbwd7cye/68-allow-sending-news-to-publishing-api-m

This PR makes it possible to send Documents to the Publishing API in draft and live states. Most of the payload is hard-coded, including the schema_name ("news_article"). A future PR should refactor the payload logic to better relate to the content of the Document.

Draft content is sent to the Publishing API whenever a document is updated (the "Save") action at the end of the new document form. We decided to put the live "Publish" button on a new document-specific page, to better represent the intended interaction.

The calls to the Publishing API are handled by a DocumentPublishingService proxy class, which is a deliberate break from the convention of "Service.publishing_api" and will allow us to implement custom config (e.g. timeouts) and error handling, without affecting other uses of the API.

For now, the calls to the Publishing API are synchronous and we handle any errors (and successes) by showing a flash message. Depending on the latency, it may be better to make some publishing actions asynchronous, although we have yet to thoroughly consider how we handle errors.


Gemfile
```diff
@@ -20,13 +20,16 @@ group :development do
...
 
 group :test do
+  gem "govuk-content-schema-test-helpers"
   gem "simplecov", "~> 0.16"
 end
 
 group :development, :test do
   gem "byebug", "~> 10"
   gem "capybara"
+  gem "factory_bot_rails", "~> 4.0"
```
kevindew: Any particular reason for the 4.0 here? though not sure it's worth my energies trying to keep this file consistent.

app/controllers/documents_controller.rb
```diff
@@ -9,11 +9,24 @@ def edit
...
+
   def update
     document = Document.find(params[:id])
     allowed_field_names_in_contents = document.document_type_schema.fields.map(&:id)
     document_update_params = params.require(:document).permit(:title, contents: allowed_field_names_in_contents)
     document.update_attributes(document_update_params)
-    redirect_to edit_document_path(document)
+    DocumentPublishingService.new.publish_draft(document)
+    redirect_to edit_document_path(document), notice: "Preview creation successful"
+  rescue StandardError
```
kevindew: Is this just here because this is a proof of concept? As we'd lose all exception notifications if we kept this here

app/services/document_publishing_service.rb
```diff
@@ -0,0 +1,45 @@
...
+
+  def publish(document)
+    publishing_api.publish(document.content_id, "major")
+  end
+
+private
+
+  def payload(document)
+    {
+      base_path: "/test/government/foo",
```
kevindew: Shall we generate a uuid or something here for proof of concept? otherwise once you'll keep overriding same content / get blocked.

app/services/document_publishing_service.rb
```diff
@@ -0,0 +1,45 @@
...
+      routes: [
+        { path: "/test/government/foo", type: "exact" },
+      ]
+    }
+  end
+
+  def publishing_api
+    GdsApi::PublishingApiV2.new(
+      Plek.new.find('publishing-api'),
+      disable_cache: true,
```
kevindew: any particular reason to disable cache?

spec/integration/create_a_document_api_down_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+    choose "News"
+    click_on "Continue"
+
+    choose "Press release"
+    click_on "Continue"
+  end
+
+  def and_the_publishing_api_is_down
+    @request = stub_publishing_api_put_content(Document.last.content_id,
+                                               hash_including(title: "A great title"))
```
kevindew: Is the linter suggesting you indent like this? doesn't seem very GOV.UK ruby (even though I quite like it) and actually surprised if it's not telling you off.

benthorner: My editor indents like that automatically. I prefer massive indentation over hanging '('s.

spec/spec_helper.rb
```diff
@@ -1,19 +1,29 @@
...
 
-require "byebug"
-require "simplecov"
-
 ENV["RAILS_ENV"] ||= "test"
 require File.expand_path("../../config/environment", __FILE__)
 require "rspec/rails"
 
+require "byebug"
+require "govuk-content-schema-test-helpers/rspec_matchers"
```
tijmenb: We should use `govuk_schemas` for this (it's already in the Gemfile). It exposes the same API but doesn't require the setup.

This is all that's needed:

```rb
require 'govuk_schemas/rspec_matchers'
```

app/services/document_publishing_service.rb
benthorner: As there's no way to see what the URL is at the moment, I don't think there's much point. We should still be able to publish a single document, though 😄 .

benthorner: Is that OK with you?

kevindew: d'oh yeah good point on not knowing what the URL is heh. Yeah sure.

app/controllers/documents_controller.rb
benthorner: We were thinking there should be some minimal error handling in all external actions at the point we put them in. I've updated these to be more specific. In future we may want to show the specific message, but for now I've also added a logger line. Any better?

Gemfile
benthorner: Not really - it's the latest one. Do we have a more established way to select the version?

kevindew: Just means the same as `~> 4` but it's written different to the ones pinning to those.

spec/integration/create_a_document_api_down_spec.rb
kevindew: Yeah - it is more common in ruby to be like that. I thought our rubocop moaned at that but presumably not.

PR #46, opened 2018-07-25
Add a summary field to documents

tijmenb:
This field contains a summary of the document. All of the formats will have this field.

https://trello.com/c/PxtaaObC

### DIFF
--- a/app/controllers/documents_controller.rb
+++ b/app/controllers/documents_controller.rb
@@ -27,6 +27,6 @@ private
 
   def document_update_params(document)
     contents_params = document.document_type_schema.fields.map(&:id)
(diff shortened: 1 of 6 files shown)


tijmenb:
@benthorner It's called "summary" in the designs, so I'd like to stick to that because it has a more meaning than "description". That doesn't preclude using the summary for the page description though (the reverse wouldn't work).

tijmenb:
@benthorner @kevindew naming this summary, because that will make this app internally consistent. The only inconsistency will exist when we publish to the publishing-api. I think that's preferable to having an inconsistency between what we call something in the UI (summary) and in the database (description). We could at some point before beta change it if it really turns out to be confusing though?

PR #48, opened 2018-07-25
Allow user to specify base_path for now

benthorner:
https://trello.com/c/Vbwd7cye

In order to effectively test publishing, we need to be able to generate
unique URLs for each document. Rather than preempt the intended work of
generating these automatically, for now the user can specify them
manually, without any validation.

### DIFF
--- a/app/controllers/documents_controller.rb
+++ b/app/controllers/documents_controller.rb
@@ -14,20 +14,19 @@ class DocumentsController < ApplicationController
   end
 
   def update
-    document = Document.find(params[:id])
-    allowed_field_names_in_contents = document.document_type_schema.fields.map(&:id)
-    document_update_params = params.require(:document).permit(:title, contents: allowed_field_names_in_contents)
-    document.update_attributes(document_update_params)
-    DocumentPublishingService.new.publish_draft(document)
-    redirect_to document, notice: "Preview creation successful"
+    @document = Document.find(params[:id])
+    @document.update_attributes(document_update_params)
+    DocumentPublishingService.new.publish_draft(@document)
+    redirect_to @document, notice: "Preview creation successful"
   rescue GdsApi::HTTPErrorResponse, SocketError => e
     Rails.logger.error(e)
-    redirect_to document, alert: "Error creating preview"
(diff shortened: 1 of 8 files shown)


app/controllers/documents_controller.rb
```diff
@@ -14,20 +14,19 @@ def show
...
   rescue GdsApi::HTTPErrorResponse, SocketError => e
     Rails.logger.error(e)
-    redirect_to document, alert: "Error creating preview"
+    redirect_to @document, alert: "Error creating preview"
   end
 
 private
 
   def document_update_params
-    params.require(:document).permit(:title)
```
tijmenb: Can we pass in the `document` instead of the instance variable? Instance variables imply that it's used inside a view, which isn't the case here.

spec/factories/document_factory.rb
```diff
@@ -4,6 +4,8 @@
   factory :document do
     content_id { SecureRandom.uuid }
     locale { I18n.available_locales.sample }
+    base_path { "/#{SecureRandom.alphanumeric(8)}" }
```
tijmenb: What would you think about only specifying things that are necessary to create a valid record in the factories?

benthorner: Fine with me :-) Updated.

app/controllers/documents_controller.rb
benthorner: Also fine - updated.

kevindew:
@benthorner should we add a unique index on base path?

benthorner:
I'm assuming we're going to do this properly as part of the card to address URL generation and the associated user feedback. For now, it'll just come back with a message saying the it failed. It's mainly to allow us to fiddle with publishing 'for now' until we have something better.

kevindew:
@benthorner sure - we will have to do something in app to handle the conflicts as the publishing-api path reservation works on a per app rather than per document approach.

benthorner:
Oh really - that's news to me and a surprise it's implemented that way. I'm interested to know why - perhaps we can talk about it sometime :-).

PR #49, opened 2018-07-25
Replace hard-coded payload fields with config

benthorner:
https://trello.com/c/Vbwd7cye

### DIFF
--- a/app/services/document_type_schema.rb
+++ b/app/services/document_type_schema.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 class DocumentTypeSchema
-  attr_reader :fields, :document_type, :name, :supertype, :managed_elsewhere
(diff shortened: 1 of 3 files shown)


app/services/document_type_schema.rb
```diff
@@ -1,14 +1,16 @@
 # frozen_string_literal: true
 
 class DocumentTypeSchema
-  attr_reader :fields, :document_type, :name, :supertype, :managed_elsewhere
+  attr_reader :fields, :document_type, :name, :supertype, :managed_elsewhere, :schema_name, :rendering_app
```
tijmenb: Maybe we can make these required in the spec: https://github.com/alphagov/content-publisher/blob/master/spec/formats/configuration_spec.rb#L20

benthorner: Except they're actually optional e.g. consultation.

tijmenb: Ah yes, that makes sense.

PR #53, opened 2018-07-25
Add the component guide to the application

tijmenb:
This adds the component guide to the application. This means we can start developing components that are only in this application.

Because we've disabled Static in this app, the normal GOV.UK styling won't work:

https://github.com/alphagov/content-publisher/blob/fee9dbc6c25f4b9f71903daa3cc3284be6599cae/config/application.rb#L30-L33

This is not a big issue, because the components preview correctly. Alex will look at tidying up the component guide so it's better usable in admin apps as well.

https://trello.com/c/YWkUAjtN/45-make-sure-everything-is-documented


benthorner:
Useful reading for people if you're new to this repo like me: https://github.com/alphagov/govuk_publishing_components/blob/master/docs/generate-a-new-component.md.

PR #55, opened 2018-07-26
Add change_history to API Payload for email alerts

benthorner:
https://trello.com/c/Vbwd7cye

This is a required field for email alerts, which gets extracted by the
Publishing API and added to any existing change history i.e. the history
is set and stored incrementally.


benthorner:
Updated :-)

PR #61, opened 2018-07-26
Beautify the pages again

benthorner:
This is another attempt to beautify the app - feel free to modify the branch and discuss :-)

   * We should have a way of showing label+value for the show page (currently abusing the label component); similarly for govspeak
   * Some of the buttons and text have the wrong font, except for the Publish button - presumably because it's type=submit
   *  The Publish button doing a post request is a bit of a hack - should be able to do this by setting the method like rails button_to
   * The document list indentation for the index page is messed up, and we should be translating the document_type to look nice
   * The flash messages are hard-coded and not very informative in the case of an error - we should expose more detailed info
   * The error_message component has an orange and a red outline when it first appears - probably shouldn't allow it to be focusable

### DIFF
--- a/app/views/documents/edit.html.erb
+++ b/app/views/documents/edit.html.erb
@@ -1,36 +1,39 @@
+<%= content_for :back_link, document_path(@document) %>
 <% content_for :title, @document.title || "Untitled document" %>
+
 <div class="govuk-grid-row">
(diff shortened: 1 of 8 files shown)


tijmenb:
I’ve updated this PR to focus on the forms and a few other things we’re fairly certain about. We’ll address other layout issues in seperate PRs.

PR #70, opened 2018-07-27
Generate paths and show preview of the URL to publishers

davbo:
Trello: https://trello.com/c/SSWJLaP2/70-how-do-we-create-and-reserve-a-basepath-l

This PR updates the app to generate base paths e.g. /news/my-news-story from titles. URLs are generated as users enter the title to give feedback of how the URL will look on GOV.UK.

We do this by introducing a new endpoint `/documents/:id/generate-path` which is called by JavaScript when a user edits the title. The endpoint creates a slug from the title and checks with the publishing-api if that slug exists. *In progress* If a content item exists at that path already then we append an integer until we arrive at a unique path.

Initially we planned to reserve the path with the publishing-api, this introduces a couple of edge cases which we've chosen to avoid for now. Such as the possiblilty of unpublished content having several reserved paths which cannot be used, introducing the need for some way to 'unreserve' a path. See https://github.com/alphagov/content-publisher/pull/70/commits/eff681aeb9605d4023bc763d684b9fd18df3bdd1 for more details.


app/services/document_publishing_service.rb
```diff
@@ -40,6 +49,14 @@ def payload(document)
   end
 
   def publishing_api
+    GdsApi::PublishingApi.new(
```
tijmenb: This should be nicer after https://github.com/alphagov/gds-api-adapters/pull/841 is released.

app/controllers/documents_controller.rb
```diff
@@ -16,15 +16,44 @@ def show
   def update
     document = Document.find(params[:id])
     document.update_attributes(document_update_params(document))
+    save_base_path!(document)
```
tijmenb: Would it be possible to do this in the `update_attributes`? That saves a database call.

app/controllers/documents_controller.rb
```diff
@@ -16,15 +16,44 @@ def show
...
 
+  def save_base_path!(document)
+    base_path = generate_base_path(document, document.title)
+    # TODO: Generate a base_path which isn't in use
+    return false if path_in_publishing_api?(base_path)
+    document.base_path = base_path
+    document.save!
+  end
+
+  def generate_base_path(document, title)
```
tijmenb: Perhaps this method and the things in `DocumentPublishingService` can go into a separate class? Something like `PathGenerator.new(document).path` that returns a unique path?

app/formats/document_types.yml
```diff
@@ -4,6 +4,7 @@
   rendering_app: government-frontend
   name: Press release
   supertype: news
+  prefix: /news
```
tijmenb: Perhaps `path_prefix` for clarity.

config/routes.rb
```diff
@@ -15,6 +15,7 @@
   get '/documents/:id/edit' => 'documents#edit', as: :edit_document
   patch '/documents/:id' => 'documents#update', as: :document
   get '/documents/:id' => 'documents#show'
+  post '/documents/:id/generate-path' => 'documents#generate_path'
```
tijmenb: Can this be simplified to a `GET` now that we don't perform writes in the action?

spec/integration/create_a_document_spec.rb
```diff
@@ -12,45 +13,66 @@
     and_the_preview_creation_succeeded
   end
 
+  scenario "User create a document and the chosen base path is already in use" do
```
tijmenb: I wonder if we should pull this out into its own feature test (file). We currently have to do setups in this test (`when_i_click_on_create_a_document` / `and_i_choose_news `) that are unrelated to the thing that we're attempting to test (the base path).

app/controllers/documents_controller.rb
emmabeynon: Good shout - done here https://github.com/alphagov/content-publisher/pull/70/commits/ef414afe615cdd0f23a5c70071f135cbc852431f

app/controllers/documents_controller.rb
```diff
@@ -15,16 +15,40 @@ def show
...
+  end
+
+  def path_in_publishing_api?(base_path)
+    publishing_service = DocumentPublishingService.new
+    publishing_service.path_exists?(base_path)
+  end
+
   def document_update_params(document)
     contents_params = document.document_type_schema.fields.map(&:id)
     params.require(:document).permit(:title, :summary, :base_path, contents: contents_params)
```
tijmenb: Don't forget to remove `base_path` from here, otherwise an intrepid content designer can fiddle with the POST body and change the base path.

app/controllers/documents_controller.rb
emmabeynon: Makes sense - added here https://github.com/alphagov/content-publisher/pull/70/commits/016ba76ca8127115f008d8454b4b6a72ae8cda65

app/controllers/documents_controller.rb
emmabeynon: 👍

Gemfile
```diff
@@ -25,10 +25,10 @@ end
 
 group :development, :test do
   gem "byebug", "~> 10"
-  gem "capybara", "~> 3"
   gem "factory_bot_rails", "~> 4"
   gem "govuk-lint", "~> 3"
   gem "govuk_schemas", "~> 3.2"
+  gem "govuk_test", "~> 0.1"
```
benthorner: Can this be split out into a separate PR? (It seems to be independent of this one.)

startup.sh
```diff
@@ -1,4 +1,4 @@
 #!/bin/bash
 
 bundle install
-bundle exec rails s -p 3221 
+bundle exec rails s webrick -p 3221
```
benthorner: Also this - should it be in a separate PR?

spec/integration/create_a_document_spec.rb
```diff
@@ -12,45 +13,66 @@
...
     expect(@request).to have_been_requested
+    expect(@lookup_base_path_request).to have_been_requested
     expect(page).to have_content "Preview creation successful"
 
     expect(a_request(:put, /content/).with { |req|
       expect(req.body).to be_valid_against_schema("news_article")
     }).to have_been_requested
   end
+
+  def and_i_fill_in_the_title_but_base_path_is_already_in_use
```
benthorner: I agree with @tijmenb this is really a separate feature - could we put this in a different file? (The idea was to try and have one feature per file, to keep them short.)

spec/integration/create_a_document_spec.rb
```diff
@@ -12,45 +13,66 @@
...
   def and_i_choose_a_press_release
-    choose "Press release"
+    choose "Press release", allow_label_click: true
     click_on "Continue"
   end
 
   def and_i_fill_in_the_form_fields
     fill_in "document[title]", with: "A great title"
-    fill_in "document[base_path]", with: "/government/foo"
+    @lookup_base_path_request = publishing_api_has_lookups("/news/a-great-title": nil)
```
benthorner: This variable doesn't seem to be used anywhere - should we be checking if it's requested?

spec/integration/create_a_document_spec.rb
```diff
@@ -12,45 +13,66 @@
...
+  end
+
   def when_i_click_on_create_a_document
     visit "/"
     click_on "New document"
   end
 
   def and_i_choose_news
-    choose "News"
+    choose "News", allow_label_click: true
```
benthorner: What's the `allow_label_click` for here? It doesn't seem to be necessary elsewhere...

spec/integration/create_a_document_spec.rb
emmabeynon: Now removed 👍

Gemfile
emmabeynon: We added it so that we could make use of the selenium-webdriver gem in order to test using Javascript. @alex-ju is looking into getting the Capybara tests to play nicely with Javascript as having it enabled is breaking them at the moment.

spec/integration/create_a_document_spec.rb
emmabeynon: Now extracted into its own file, but currently disabled because of the aforementioned JS issues.

app/services/document_publishing_service.rb
```diff
@@ -3,6 +3,8 @@
 require 'gds_api/publishing_api_v2'
 
 class DocumentPublishingService
+  PUBLISHING_APP = "content-publisher"
```
erino: This constant isn't referenced anywhere else in the PR, is it needed?

app/assets/javascripts/application.js
```diff
@@ -0,0 +1,34 @@
+'use strict'
+
+document.onreadystatechange = function () {
```
kevindew: Do we need to think about a way to have this javascript initialise only when this form is available? as it stands now I think it'll run on every page.

spec/services/path_generator_service_spec.rb
```diff
@@ -0,0 +1,23 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+RSpec.describe PathGeneratorService do
+  describe '.path' do
+    it "converts a title into a base path when there is no duplicate" do
+      service = PathGeneratorService.new
+      @document = create(:document, :with_body)
```
kevindew: This doesn't need an @ since it's only used in this scope.

app/assets/javascripts/application.js
```diff
@@ -0,0 +1,34 @@
...
+        return
+      }
+      url.searchParams.append('title', documentTitle.value)
+      window.fetch(url, {
+        credentials: 'include',
+        headers: {
+          'Content-Type': 'application/json; charset=utf-8',
+          'X-CSRF-Token': csrf
+        },
+      }).then(function(response) {
```
alex-ju: Would be good to check the response status and examine the response only in case of 200
```
     }).then(
        function (response) {
          if (response.status !== 200) {
            console.error('Looks like there was a problem. Status Code: ' +
              response.status)
            return
          }
          // Examine the response
          response.json().then(function (result) {
            var basePath = document.getElementById('base-path-id')
            if (result.reserved) {
              basePath.innerHTML = 'www.gov.uk' + result['base_path']
            } else {
              basePath.innerHTML = 'Path is taken, please edit the title.'
            }
          })
        }).catch(function (err) {
        console.error(err)
      })
```

app/assets/javascripts/application.js
```diff
@@ -0,0 +1,36 @@
+'use strict'
+
+document.onreadystatechange = function () {
+  var csrfElement = document.querySelector('meta[name="csrf-token"]')
+  var csrf = csrfElement ? csrfElement.content : ""
+  if (document.readyState === 'interactive') {
+    var basePath = document.getElementById('base-path-id')
+    var documentTitle = document.getElementById('document-title-id')
+    if (basePath==null || documentTitle==null) {return;}
```
kevindew: I don't particularly know our JS standards (@alex-ju do we have some?) but this feels a bit unorthadox with the lack of spacing on the nulls and the inline return

spec/factories/document_factory.rb
```diff
@@ -4,13 +4,19 @@
...
+    base_path { "#{DocumentTypeSchema.find(document_type).path_prefix}/#{title.parameterize}" }
     document_type { DocumentTypeSchema.all.reject(&:managed_elsewhere).sample.id }
 
     trait :with_body_in_schema do
       document_type do
         DocumentTypeSchema.all.select { |schema| schema.contents.any? { |field| field.id == "body" } }.sample.id
       end
     end
+
+    trait :without_title do
```
benthorner: This seems wrong as we're doing something and then negating it. I thought we were favouring a minimal set of defaults, which seems to have worked so far. Can we change it to :with_contents or something and lose the defaults, instead?

spec/features/create_a_document_api_down_spec.rb
```diff
@@ -23,9 +23,8 @@ def given_i_start_to_create_a_document
   end
 
   def and_the_publishing_api_is_down
-    @request = stub_publishing_api_put_content(Document.last.content_id,
-                                               hash_including(title: "A great title"))
-
+    @put_content_request = stub_publishing_api_put_content(Document.last.content_id,
```
benthorner: Why has this been renamed? There doesn't seem to be a conflicting variable?

spec/features/edit_a_document_spec.rb
```diff
@@ -23,7 +23,7 @@ def given_there_is_a_document
     def when_i_go_to_edit_the_document
       visit document_path(Document.last)
       click_on "Edit document"
-      @request = stub_publishing_api_put_content(Document.last.content_id, {})
+      @put_content_request = stub_publishing_api_put_content(Document.last.content_id, {})
```
benthorner: Why has this been renamed - there doesn't seem to be a conflicting variable?

startup.sh
```diff
@@ -1,4 +1,4 @@
 #!/bin/bash
 
 bundle install
-bundle exec unicorn -p 3221 
+bundle exec unicorn -p 3221
```
benthorner: Can we check this out from master to make the PR smaller?

spec/spec_helper.rb
```diff
@@ -17,7 +17,12 @@
 GovukTest.configure
 WebMock.disable_net_connect!(allow_localhost: true)
 Capybara.automatic_label_click = true
+ActiveRecord::Migration.maintain_test_schema!
+
 Capybara::Chromedriver::Logger.raise_js_errors = true
+Capybara::Chromedriver::Logger.filters = [
+  /the server responded with a status of 409/i
```
benthorner: 409 (Conflict) seems OK to mask here 👍 .

app/services/path_generator_service.rb
```diff
@@ -0,0 +1,39 @@
...
+  end
+
+private
+
+  def document_exists_with_path?(base_path, document)
+    doc = Document.where(base_path: base_path).first
+    doc && doc != document
+  end
+
+  def create_path(prefix, slug, count = nil)
```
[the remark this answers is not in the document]
davbo: Setting the default makes the parameter optional, the count isn't always passed as a param.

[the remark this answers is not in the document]
benthorner: Oh yeah - was thinking of JS :-|

app/controllers/documents_controller.rb
```diff
@@ -15,18 +15,27 @@ def show
 
   def update
     document = Document.find(params[:id])
-    document.update_attributes(document_update_params(document))
+    base_path = PathGeneratorService.new.path(document, params[:document][:title])
```
benthorner: Could we put this in document_update_params? I'm a fan of making the action methods as small as possible, as they're meant to be the orchestrators.

app/controllers/documents_controller.rb
```diff
@@ -15,18 +15,27 @@ def show
...
     redirect_to document, notice: "Preview creation successful"
   rescue GdsApi::HTTPErrorResponse, SocketError => e
     Rails.logger.error(e)
     redirect_to document, alert: "Error creating preview"
   end
 
+  def generate_path
+    document = Document.find(params[:id])
+    base_path = PathGeneratorService.new.path(document, params[:title])
+    render json: { base_path: base_path, available: true } if base_path.present?
```
benthorner: What happens if !base_path.present?

spec/factories/document_factory.rb
```diff
@@ -4,7 +4,8 @@
   factory :document do
     content_id { SecureRandom.uuid }
     locale { I18n.available_locales.sample }
-    base_path { "/#{SecureRandom.alphanumeric(8)}" }
+    title { SecureRandom.alphanumeric(8) }
```
benthorner: Could we extract these to a trait? Technically it raises the likelihood of false positives for some of the feature tests where we're testing using @document.title.

spec/services/content_validator_spec.rb
```diff
@@ -5,7 +5,7 @@
 RSpec.describe ContentValidator do
   describe 'title validation' do
     it 'raises issue if the title is not set' do
-      document = build(:document, title: nil)
+      document = build(:document, title: nil, base_path: nil)
```
benthorner: Why are we unsetting the base_path here? (The test description only mentions the title.)

spec/services/path_generator_service_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+    it "generates a base path which is unique to our database" do
+      service = PathGeneratorService.new
+      original_document = create(:document)
+      new_document = build(:document, document_type: original_document.document_type)
+      publishing_api_has_lookups("#{original_document.base_path}": nil)
+      expect(service.path(new_document, original_document.title)).to eq("#{original_document.base_path}-1")
+    end
+
+    it "raises a 'Path unable to be generated' when many variations of that path are in use" do
+      service = PathGeneratorService.new(5)
```
benthorner: Could we make this '2' and make the test shorter? I don't feel like we're proving anything special with a bigger number.

app/assets/javascripts/application.js
```diff
@@ -0,0 +1,45 @@
...
+        return
+      }
+      url.searchParams.append('title', documentTitle.value)
+      window.fetch(url, {
+        credentials: 'include',
+        headers: {
+          'Content-Type': 'application/json; charset=utf-8'
+        }
+      }).then(function (response) {
+        if (!response.ok) {
```
benthorner: This should be redundant as we're using .then.

spec/integration/previews_a_url_when_editing_title_spec.rb
```diff
@@ -0,0 +1,49 @@
+# frozen_string_literal: true
+
+require "spec_helper"
+
+RSpec.describe "Shows a preview of the URL", type: :feature, js: true do
```
benthorner: These are now in /features - can you move it? Also, you can lose the 'type: :feature' bit and change 'RSpec.describe' to 'RSpec.feature'.

app/views/documents/edit.html.erb
```diff
@@ -3,18 +3,26 @@
...
-        value: @document.title
+        value: @document.title,
+        id: "document-title-id"
       } %>
 
+      <% url_preview = capture do %>
+        <p class="govuk-heading-s">The web address for this content will look like:</p>
+        <span id="no-title-id" class="govuk-body <%= 'app-hidden' unless @document.base_path.nil? %>">You haven’t entered a title yet.</span>
+        <div id="url-preview-id" <%= 'class="app-hidden"' if @document.base_path.nil? %>>
+            <span class="govuk-body">www.gov.uk</span><span id="base-path-id" class="govuk-body" ><%= "#{@document.base_path}" %></span>
```
benthorner: Could we use Plek here to make it environment-specific?

app/assets/javascripts/application.js
benthorner: So I had a closer look at this, and it's only used when the controller throws an unexpected error and returns an HTML response instead of valid JSON. If we want to handle unexpected errors, then it would be shorter to just put a '.catch' on the end of response.json().

Thanks for prompting me to look at this - I had no idea you can do this. On reflection, though, the great thing about promises is the ability to separate happy/sad code, so on balance I'd prefer to avoid throwing errors and use '.catch' as it keeps that logical separation.

app/assets/javascripts/application.js
```diff
@@ -0,0 +1,45 @@
...
+    var errorGeneratingPath = document.getElementById('error-generating-path-id')
+    var showErrorMessage = function () {
+      urlPreview.setAttribute('class', 'app-hidden')
+      noTitle.setAttribute('class', 'app-hidden')
+      errorGeneratingPath.removeAttribute('class')
+    }
+    documentTitle.onblur = function () {
+      var url = new URL(document.location.origin + editDocumentForm.getAttribute('data-generate-path-path'))
+      if (!documentTitle.value) {
+        noTitle.removeAttribute('class')
```
benthorner: Could we replace this code with a call to 'showErrorMessage'?

app/assets/javascripts/application.js
```diff
@@ -0,0 +1,45 @@
...
+        credentials: 'include',
+        headers: {
+          'Content-Type': 'application/json; charset=utf-8'
+        }
+      }).then(function (response) {
+        if (!response.ok) {
+          throw Error('Unable to generate response.')
+        }
+        response.json().then(function (result) {
+          noTitle.setAttribute('class', 'app-hidden')
```
benthorner: Since we have a showErrorMessage function, can we have a hideErrorMessage function as well? I think that's what this is doing.

davbo: this is showing the URL preview, it only happens the one time so I don't feel like extracting it benefits us by much? Happy to compromise if you disagree.

app/controllers/documents_controller.rb
davbo: `document_update_params` seems to have a different purpose ie. whitelist the params from the POST to what we expect. I asked about it last week and it seems like it's a rails convention to have these kind of methods of whitelist the POST attributes.

Feels like a separate thing to the path generation.

app/controllers/documents_controller.rb
davbo: Ah don't think that scenario is possible now, I'll remove the nil check.

spec/factories/document_factory.rb
davbo: Discussed in person and me and @benthorner agreed to leaving this in. Reasoning being that now we're making use of Title for something (URL path generation) it's good to set a default. At the same time we'd like to keep the factory minimal.

app/assets/javascripts/application.js
davbo: This shows a different message for the case where a user deletes the title in the box. It's not the same as the error message.

app/assets/javascripts/application.js
davbo: We discussed this and we'll leave it unchanged because the non-200 case doesn't hit the catch() call unless an error is raised explicitly. 

Seems like this confusion may be because non-200 would go to catch() when implemented like this in jQuery.

PR #78, opened 2018-07-31
Basic flow of associations

kevindew:
Trello: https://trello.com/c/4sTwhdPN/87-make-associations-available-to-select-when-creating-a-document-l

This puts together a basic flow of storing a topical event association for a press release. I'm putting it up as a PR to try gather some feedback and pose a few of the questions/decisions to a wider audience before they're implemented and harder to reverse/adjust.

Unfortunately as this uses Publishing API for linkables it won't be heroku previewable, which does raise a wider question as to whether we should set up something to mock aspects of it to maintain it's usability.

It also has an annoying bug where strong params fails if you don't submit anything (since there are no params submitted to require) - but that should be easily fixed.

## Associations field and naming

I made the call to go with an `associations` field on a document rather than using the `contents` field. I figured that this field is going to be very common so it's not like adding data just for a particular format. I expect the contents of it to be edition links rather than linkset links (I'm hoping there aren't cases outside taxonomy where this is impractical). I named it associations rather than links as that seemed to be the name in the UI and isn't burdened with the same confused meaning of links - maybe this is a bad idea?

Regarding having it separate I figured we might use contents as a namespace for things that are on the content form and then if we have other sections which have dedicated bespoke forms they can also have their own JSON data rather than one big object that is quite nested. 

## Linkset service/caching

I'm working off some of the stuff Emma & David did with Linkset from the Publishing API. I'm imagining that we'll gradually build caching into it so it doesn't need to make a request each instance. I guess it'll be extra cool if we can make the cache be updated asynchronously. But I'm considering the cache part of this service classes responsibility rather than something the wider application should worry about.

### Moving existing fields into a contents section

We could move the field section into a contents section of the document_type YAML file:

```
- document_type: press_release
  schema_name: news_article
  rendering_app: government-frontend
  name: Press release
  supertype: news
  contents:
    fields:
      - id: body
        label: Body
        type: govspeak
        validations:
          min_length: 10
  associations:
    fields:
      - id: topical_event
        label: Topical event the press release relates to
        type: topical_event_linkable_selector
```

Not sure whether it's good to keep having the fields within a "fields" section if we do this. On the plus it's clearer they are fields reading the file and leaves a space for any specific options. On the minus it can make for some pretty deep object traversal.

But anyway doing this would mean we'd just need to loop through the section of fields for the appropriate part of the form.

### Field type as a class

There's a number of aspects of logic associated with linkables that could be encapsulated inside a class for the field type. These are:

- Converting model data to go into the form (if we assume we store the name of a link in the db then we need to get back to just an id for the form)
- Checking the data that has been submitted is actual content_ids before we store them
- Converting the submitted data to format for the DB
- What type of data can be permitted in the controller (may not be needed)
- Presenting the data to summary page
- Presenting the data to the Publishing API

So I imagined we could have some options like this:

```
  associations:
    fields:
      - id: topical_event
        label: Topical event the press release relates to
        type: linkables_select
        options: 
             document_type: "topical_event"
```
Which could create a Fields::LinkablesSelect.new(document_type: "topical_event") field 

or could be more explicit (but less YAML like)
```
  associations:
    fields:
      - id: topical_event
        label: Topical event the press release relates to
        type: linkables_select
        class: Fields::LinkablesSelect
        class_arguments: 
             document_type: "topical_event"
```

By having some options/arguments we could do multiple linkable selects with the same type, as otherwise we'll have to duplicate it.

Thoughts/ideas on this very welcome.

### Other things

- We probably need a name field in the YAML fields for the summary esque pages unless we plan to keep label short and use that everywhere
- Not too sure how to write good feature tests for this without getting bogged into a specific format or being rather generic 😕


app/services/linkables_service.rb
```diff
@@ -0,0 +1,33 @@
+# frozen_string_literal: true
+
+require "gds_api/publishing_api_v2"
+
+class LinkablesService
```
tijmenb: It might be good to take a look at content tagger and publisher for prior art:

https://github.com/alphagov/content-tagger/blob/master/app/models/linkables.rb
https://github.com/alphagov/publisher/blob/master/lib/tagging/linkables.rb

They do some additional things:

- There are some document type specific transformations (draft/no-draft filtering
- Use `internal_name` instead of the `title` (`internal_name` is meant for disambiguation in select boxes like this) 
- Showing "(draft)" in the title

app/views/document_associations/edit.html.erb
```diff
@@ -0,0 +1,17 @@
+<%= content_for :back_link, document_path(@document) %>
+<% content_for :title, "Document associations" %>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <%= form_tag(document_associations_path(@document)) do %>
+
+      <% @document.document_type_schema.associations.each do |field| %>
+        <%= render "document_associations/fields/#{field.type}_input", name: field.id, label: field.label, associations: @document["associations"] %>
```
tijmenb: Should we pass in `associations: @document["associations"][name]` here instead?

spec/integration/add_document_associations_spec.rb
```diff
@@ -0,0 +1,65 @@
...
+  scenario "A document has associations added" do
+    given_there_is_a_document_with_associations
+    when_i_visit_the_document_page
+    then_i_can_navigate_to_associations
+    when_i_add_some_associations
+    then_i_can_view_the_associations
+  end
+
+  def given_there_is_a_document_with_associations
+    @document = create :document, document_type: "press_release"
```
tijmenb: Perhaps create a factory that is specific about what kind of document type we want here, instead of the document type:

```rb
create(:document, :with_a_topical_event_association)
```

tijmenb:
This all looks very nice and sensible!

> Moving existing fields into a contents section

I'd vote for renaming less nesting, and rename the existing `fields` to `contents_fields`.

> Field type as a class

Unlike the contents field types, there will be a finite and manageable number of associations (orgs, ministers, topical events, needs, a few more). Perhaps we use a specific input type for each association?

```
  associations:
    fields:
      - id: topical_event
        label: Topical event the press release relates to
        type: topical_event_select
```

> Unfortunately as this uses Publishing API for linkables it won't be heroku previewable, which does raise a wider question as to whether we should set up something to mock aspects of it to maintain it's usability.

We should also use Heroku as the example for what happens if the publishing-api is unavailable. Can we show a nice error message on the page if that's the case?

app/controllers/document_associations_controller.rb
```diff
@@ -0,0 +1,45 @@
+# frozen_string_literal: true
+
+class DocumentAssociationsController < ApplicationController
+  attr_reader :document
```
tijmenb: I had to look at this a couple of times to understand what was happening here (`check_document_supports_associations` loads `@document` as a side effect, reader exposes `document` to subsequent actions). Do we need to check if the document has associations -  I think all formats would have to be tagged to organisations? The code might be simplified if we don't need to check.

app/services/linkables_service.rb
kevindew: Good shout 👍

app/views/document_associations/edit.html.erb
kevindew: yeah that'd probably be better. I got caught in a halfway house between how it's done elsewhere.

spec/integration/add_document_associations_spec.rb
kevindew: Yup sure, guess that'd mean feature tests for each link type but that's probably cool - albeit a touch verbose.

kevindew:
> Unlike the contents field types, there will be a finite and manageable number of associations (orgs, ministers, topical events, needs, a few more). Perhaps we use a specific input type for each association?

We certainly could but it feels like we'll have quite a lot of repetition between them and need somewhere to put the logic. Would work for now though. I'd be happier if we could find ways of encapsulating field logic rather than lots of conditionals in the business logic.

app/controllers/document_associations_controller.rb
kevindew: Yeah it's never great for readability using before_action stuff. I could set up a set_document one first if that'd improve readability for you? - I find the before_action stuff all a bit icky

As for associations. We have a bunch of specialist formats that have implicit associations e.g. aaib reports and wouldn't have any others. We wouldn't need to check for now though while we're just dealing with Whitehall doc types.

app/formats/document_types.yml
```diff
@@ -10,6 +10,11 @@
       type: govspeak
       validations:
         min_length: 10
+  associations:
```
benthorner: Do we need to have a 'fields' subhash? I noticed later on we write 'document_type_schema.associations', implying we're not expecting the contents/associations to have any other members.

benthorner:
Could we use 'edition_links' instead of 'associations'? This would make it clearer what they're for from a technical perspective.

Related to that, have we thought about how we're going to deal with linkset links, or even if we're going to store them at all? They are also a kind of association, after all.

tijmenb:
@kevindew I think I changed my mind a bit on this. I was pondering the associations that we'll support, and I'm starting to suspect that having a "Association" abstraction won't work because they are all different in UI, display, and data handling.

Organisations linking:

- need to show 2 boxes for lead and supporting organisations
- default the first lead organisations to the organisation of the currently logged in user
- we need to send the first of the lead organisations to the publishing-api as `primary_publishing_organisation`

Topical events:

- We might want to make a distinction between current and past topical events

People:

- Needs a lot of data processing (https://trello.com/c/3GT17tkH)

In addition, the associations will be on/off per format, and there might not be custom validations and labels.

Perhaps we treat associations less as a field, and more as a "feature toggle"?

```yaml
- document_type: press_release
  schema_name: news_article
  rendering_app: government-frontend
  name: Press release
  supertype: news
  contents:
- id: body
    label: Body
    type: govspeak
    validations:
      min_length: 10
  associations:
     - topical_events
     - ministers
```

kevindew:
> Could we use 'edition_links' instead of 'associations'? This would make it clearer what they're for from a technical perspective.
> 
> Related to that, have we thought about how we're going to deal with linkset links, or even if we're going to store them at all? They are also a kind of association, after all.

@benthorner using the naming of associations was an attempt to try avoid loading this with the meanings of linkset and edition links. I'm hoping these will all be edition links, but it could well be that for legacy reasons or other complications we end up with some involved. I thought use of association abstracted it a bit and kept the options open there (as well as being consistent with frontend).

I appreciate it adds the load of it being a different name but I think using the Publishing API naming is also not ideal and definitely not naming we'd share with a user.

> @kevindew I think I changed my mind a bit on this. I was pondering the associations that we'll support, and I'm starting to suspect that having a "Association" abstraction won't work because they are all different in UI, display, and data handling.

Yeah you raise some good points here and I like the idea of defining them succinctly.

I'm not sure if they're that different to fields though, I think they're just a case of common fields in the same way govspeak body is. Do we want instead just a succinct way of naming fields altogether like feature flags and then just room for verbosity on the more snowflake ones?

```
- document_type: press_release
  schema_name: news_article
  rendering_app: government-frontend
  name: Press release
  supertype: news
  contents:
     - govspeak_body
     - type: date
        id: opened_on
  associations:
     - topical_events
     - ministers
```

tijmenb:
@kevindew yeah, that might be a good idea for the content fields too. I would need to see more formats before knowing if it fits though. Maybe we can do a spike next sprint into adding as many fields from other formats into the app to see what sticks?

spec/services/linkables_service_spec.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+RSpec.describe LinkablesService do
+  describe "#select_options" do
+    it "returns an array of titles with content_ids" do
+      linkable = { "content_id" => SecureRandom.uuid, "internal_name" => "Linkable 1" }
```
davbo: Could we replace this with the new hash syntax?

app/views/document_associations/associations/_multi_association_input.html.erb
```diff
@@ -0,0 +1,13 @@
+<p>
```
davbo: We open a `<p>` tag here but I don't see any close? not certain we even need the `<p>` ?

app/views/document_associations/edit_api_down.html.erb
```diff
@@ -0,0 +1,4 @@
+<%= content_for :back_link, document_path(@document) %>
+<% content_for :title, "Document associations" %>
+
+This content can't be edited right now. We're having trouble getting the data we need for you to make changes on this page.
```
davbo: Maybe this should be in a `<p>` ?

app/views/documents/show_api_down.html.erb
```diff
@@ -0,0 +1,36 @@
...
+<% end %>
+</ul>
+
+<% @document.document_type_schema.contents.each do |schema| %>
+  <%= render "documents/fields/#{schema.type}", schema: schema, document: @document %>
+<% end %>
+
+<h2>Associations</h2>
+
+This content isn't available right now. We're having trouble getting the data we need to show you this content.
```
davbo: Another `<p>`?

spec/services/linkables_service_spec.rb
benthorner: We need these to be string keys, to match the API response.

davbo: aha, thanks! I forget the new syntax only applies to symbols

benthorner:
@davbo Fixed - please check again :-)

app/controllers/document_associations_controller.rb
```diff
@@ -0,0 +1,31 @@
+# frozen_string_literal: true
+
+class DocumentAssociationsController < ApplicationController
+  rescue_from GdsApi::BaseError do
+    render 'edit_api_down', status: 503
```
davbo: We should probably log this error, if we don't I'm not sure Sentry will pick it up.

benthorner: Done done :-)

PR #80, opened 2018-08-01
Disable active_storage

cbaines:
It doesn't look to be used yet, and including it means that you have
to provide a SECRET_KEY_BASE when running the assets:precompile Rake
task (which breaks creating mini environments in the GOV.UK Mini
Environment Admin).


cbaines:
> We're actually quite likely to use active storage in this app.

Interesting, I would have thought Asset Manager would be the thing to use?

> Can we solve your problem differently?

Maybe, I can specify some value for the SECRET_KEY_BASE, but it's a bit odd that it's required for asset precompilation, no other GOV.UK app requires this. If a value was provided, I don't know how it would be used? It could be that Rails/active storage is being a bit eager, and doesn't really need this... 

This is the stack trace in case you're interested, and you can reproduce it locally by running something like `GOVUK_APP_DOMAIN=dev.gov.uk RAILS_ENV=production bundle exec rake assets:precompile`:

```
ArgumentError: Missing `secret_key_base` for 'production' environment, set this string with `rails credentials:edit`
railties-5.2.0/lib/rails/application.rb:585:in `validate_secret_key_base'
railties-5.2.0/lib/rails/application.rb:432:in `secret_key_base'
railties-5.2.0/lib/rails/application.rb:176:in `key_generator'
railties-5.2.0/lib/rails/application.rb:205:in `message_verifier'
activestorage-5.2.0/lib/active_storage/engine.rb:67:in `block (2 levels) in <class:Engine>'
activesupport-5.2.0/lib/active_support/lazy_load_hooks.rb:69:in `block in execute_hook'
activesupport-5.2.0/lib/active_support/lazy_load_hooks.rb:62:in `with_execution_control'
activesupport-5.2.0/lib/active_support/lazy_load_hooks.rb:67:in `execute_hook'
activesupport-5.2.0/lib/active_support/lazy_load_hooks.rb:52:in `block in run_load_hooks'
activesupport-5.2.0/lib/active_support/lazy_load_hooks.rb:51:in `each'
activesupport-5.2.0/lib/active_support/lazy_load_hooks.rb:51:in `run_load_hooks'
railties-5.2.0/lib/rails/application/finisher.rb:75:in `block in <module:Finisher>'
railties-5.2.0/lib/rails/initializable.rb:32:in `instance_exec'
railties-5.2.0/lib/rails/initializable.rb:32:in `run'
railties-5.2.0/lib/rails/initializable.rb:61:in `block in run_initializers'
railties-5.2.0/lib/rails/initializable.rb:60:in `run_initializers'
railties-5.2.0/lib/rails/application.rb:361:in `initialize!'
config/environment.rb:5:in `<main>'
bootsnap-1.3.1/lib/bootsnap/load_path_cache/core_ext/kernel_require.rb:21:in `require'
bootsnap-1.3.1/lib/bootsnap/load_path_cache/core_ext/kernel_require.rb:21:in `block in require_with_bootsnap_lfi'
bootsnap-1.3.1/lib/bootsnap/load_path_cache/loaded_features_index.rb:65:in `register'
bootsnap-1.3.1/lib/bootsnap/load_path_cache/core_ext/kernel_require.rb:20:in `require_with_bootsnap_lfi'
bootsnap-1.3.1/lib/bootsnap/load_path_cache/core_ext/kernel_require.rb:29:in `require'
activesupport-5.2.0/lib/active_support/dependencies.rb:283:in `block in require'
activesupport-5.2.0/lib/active_support/dependencies.rb:249:in `load_dependency'
activesupport-5.2.0/lib/active_support/dependencies.rb:283:in `require'
railties-5.2.0/lib/rails/application.rb:337:in `require_environment!'
railties-5.2.0/lib/rails/application.rb:520:in `block in run_tasks_blocks'
sprockets-rails-3.2.1/lib/sprockets/rails/task.rb:62:in `block (2 levels) in define'
rake-12.3.1/exe/rake:27:in `<top (required)>'
```

tijmenb:
@cbaines found it - all of the apps actually do set this: https://github.com/alphagov/govuk-jenkinslib/pull/14.

erino:
> We're actually quite likely to use active storage in this app.
>> Interesting, I would have thought Asset Manager would be the thing to use?

@cbaines We're planning on allowing users to crop their uploaded images. The current plan is to use ActiveStorage to manage the initial image upload and crop. Then use Asset Manager for final version which is served from draft or live.

cbaines:
> @cbaines found it - all of the apps actually do set this: alphagov/govuk-jenkinslib#14.

Interesting. I'm not sure the motivation is right, as testing Content Publisher locally shows that asset precompilation works without the SECRET_KEY_BASE, if active storage is removed:

>  This is because starting with Rails 5.2, SECRET_KEY_BASE is required when running any command, and CI doesn't have access to this value when running with RAILS_ENV=production.

I think it's just some Rails components, active storage, and active mailer it seems too https://github.com/alphagov/release/pull/246 .

Anyway, I've added a workaround in to govuk-guix ( https://github.com/alphagov/govuk-guix/commit/c8501108b13d00b84d0d730dfe1c19967e6add4f ), so I'll close this now. Thanks for taking a look :)

kevindew:
With the end-to-end test work it was such a pain all the environment variables you need to pass in to compile assets in Dockerfiles :(

Really hope Rails sorts it out so asset compilation can be abstracted from environment configuration.

/grievings/musings

cbaines:
> Really hope Rails sorts it out so asset compilation can be abstracted from environment configuration.

Some of the environment variables do actually make a difference though. One of the slow parts of building mini environments is precompiling all the assets with the appropriate configuration, the Plek related environment variables for example. I can't remember the exact cases unfortunately, otherwise I would be more specific....

PR #82, opened 2018-08-01
Reorganise document types to improve readability

benthorner:
https://trello.com/c/4sTwhdPN/87-make-associations-available-to-select-when-creating-a-document-l

This PR renames 'fields' to 'contents' to support future data keys at
this level, such as the upcoming 'associations' work.

This PR also makes the document types more consistent with fields and
supertype schemas by renaming 'document_type' to 'id', as well as nested
schema_name and rendering_app under a 'publishing_metadata' hash to
reduce the number of top-level keys and thereby improve readability.

To make this change easier, I refactored the configuration_spec to make
use of the schema models, as opposed to accessing the YAML files
directly. I also replaced several instances of SupertypeSchema.find with
document_type.supertype, so the lookup is done automatically.

This PR also contains a few minor changes to make the tests more
consistent and remove a redundant managed_elsewhere? question method.
I can move these into a separate PR if they're not to people's tastes.

### DIFF
--- a/app/controllers/documents_controller.rb
+++ b/app/controllers/documents_controller.rb
@@ -26,7 +26,7 @@ class DocumentsController < ApplicationController
 private
 
   def document_update_params(document)
-    contents_params = document.document_type_schema.fields.map(&:id)
+    contents_params = document.document_type_schema.contents.map(&:id)
     params.require(:document).permit(:title, :summary, :base_path, contents: contents_params)
   end
 end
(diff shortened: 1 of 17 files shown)


tijmenb:
Cool! We talked IRL about a few things that we'll fix later, like using `fetch` consistently to make typos easier to spot. Also I already miss `managed_elsewhere?`, but that's my loss I guess :)

benthorner:
I already made a little nested model class :-).

PR #84, opened 2018-08-01
Remove ApplicationRecord

tijmenb:
As discussed at the start of the quarter, we're going to prefer [composition over inheritance](https://docs.google.com/document/d/142IslEIKXSSilNb5EVlM2l7usRfaQlhQDVHWhputGFs/edit) in this application.

### DIFF
--- a/app/models/application_record.rb
+++ b/app/models/application_record.rb
@@ -1,5 +1,0 @@
-# frozen_string_literal: true
-
-class ApplicationRecord < ActiveRecord::Base
-  self.abstract_class = true
-end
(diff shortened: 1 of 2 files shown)


erino:
What about `ApplicationController`? :trollface:

benthorner:
I think we should leave this as-is - it's standard Rails and although it's cruft, it's standard cruft and people will expect it to be there. We can still follow the rails practices document without touching this class - just don't touch the class ;-).

benthorner:
Closing as this would break #97.

PR #87, opened 2018-08-02
Replace supertype thing as a first-class supertype

benthorner:
Previously a static not-sure supertype was defined in the new documents
controller, which required a proxy SupertypeThing class to handle the
generation and processing of options. My original intention was simply
to move this class into /services to highlight its existance, but
instead this PR is an attempt to remove it altogether by adding a 'path'
option to supertypes, behaving similarly to managed_elsewhere for docs.

I also added a few more configuration_specs after the required spec to check
a document_type has not been associated with the new "not-sure" supertype.

### DIFF
--- a/app/formats/supertypes.yml
+++ b/app/formats/supertypes.yml
@@ -17,5 +17,10 @@
 
 - id: consultations
   label: Consultations
-  description: To request users views or evidence on an issue and to build collective
-    agreement.
+  description: To request users views or evidence on an issue and to build collective agreement.
+
+- id: not-sure
+  label: I'm not sure this should be on GOV.UK
+  description: View this guide to see what should go on GOV.UK and where else you can publish content.
+  managed_elsewhere:
+    path: /documents/publishing-guidance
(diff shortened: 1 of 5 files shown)


app/formats/supertypes.yml
```diff
@@ -17,5 +17,10 @@
 
 - id: consultations
   label: Consultations
-  description: To request users views or evidence on an issue and to build collective
-    agreement.
+  description: To request users views or evidence on an issue and to build collective agreement.
+
+- id: not-sure
+  label: I'm not sure this should be on GOV.UK
+  description: View this guide to what should go on GOV.UK and where else we can publish content.
```
kevindew: I know this copied verbatim but is it missing a word "View this guide to _see_ what should" perhaps? and maybe "where else _~~we~~you_ can publish content"

benthorner: Good spot - corrected.

benthorner:
The managed_elsewhere thing is indeed new and may not be the best thing to call it, but it's certainly preferable to having the hard-coded constant in a class somewhere :-).

PR #93, opened 2018-08-03
Lint Javascript with StandardJS

kevindew:
This adds the StandardJS linter as advised in https://github.com/alphagov/styleguides/blob/master/js.md#linting

This uses NPM to install the linter rather than Yarn as our CI servers
do have NPM installed and don't have Yarn. Since Yarn plays nicer with
Rails that is probably a better option to take moving forward if we make
use of node-modules in the app, but this seems a reasonable approach for
now.

This sets up a lint and a lint:fix command, the latter can be used to
fix issues automatically.

Slightly annoying thing with this is not caching the node_modules between builds :-/

@alex-ju you might have opinions on this and the usefulness of it. It also seems like we've jumped from a no semi-colon land to a semi-colon land in GDS JS standards in the switch from standardJs to XO.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,6 +15,7 @@
 
 /node_modules
 /yarn-error.log
+/npm-debug.log
 
 /public/assets
 .byebug_history
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -4,6 +4,12 @@ library("govuk")
 
 node {
   govuk.buildProject(
+    beforeTest: {
+      stage("Lint Javascript") {
+        sh("npm install")
+        sh("npm run lint --silent")
+      }
+    },
     brakeman: true
   )
 }
--- a/README.md
+++ b/README.md
@@ -18,6 +18,7 @@ This is a Ruby on Rails application.
 ### Dependencies
 
 - [postgresql]() - provides a backing database
+- [npm]() - package manager for JavaScripts
 
 ### Running the application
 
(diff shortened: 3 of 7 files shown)


alex-ju:
The [GDS Way recommendations](https://gds-way.cloudapps.digital/manuals/programming-languages/nodejs/#source-formatting-and-linting) refer to NodeJS and it's not aligned with the [GOV.UK coding standards](https://github.com/alphagov/styleguides/blob/master/js.md) nor with the [GOV.UK Frontend/Design System coding standards](https://github.com/alphagov/govuk-frontend/blob/master/docs/contributing/coding-standards/js.md).

kevindew:
> The GDS Way recommendations refer to NodeJS and it's not aligned with the GOV.UK coding standards nor with the GOV.UK Frontend/Design System coding standards.

Dang! I set this all up with Standard then discovered the GDS way advice was a lot newer than the other linting guidlines so switched to that - any particular reason why we've we got different standards for JS in node vs client side?

Oh well - I've switched this back to standard now.

alex-ju:
> Dang! I set this all up with Standard then discovered the GDS way advice was a lot newer than the other linting guidlines so switched to that - any particular reason why we've we got different standards for JS in node vs client side?

No strong reasons. I think it's about different people preferring different styles of writing JavaScript. For example GOV.UK Frontend and GOV.UK Design System were both written using Standard for both node and client side because the GDS Way didn't exist at that point and we used the GOV.UK coding standards. I'll ask around and try to understand the reason for the change.

PR #104, opened 2018-08-07
Revert "Revert "Lint Javascript with StandardJS""

benthorner:
Reverts alphagov/content-publisher#100

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,6 +15,7 @@
 
 /node_modules
 /yarn-error.log
+/npm-debug.log
 
 /public/assets
 .byebug_history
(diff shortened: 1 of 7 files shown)


benthorner:
The original merge of this PR caused the builds for other PRs to fail. After some more investigation, it looks like the cause of the failures was due to the pipelines for the other PRs not having the new 'Lint Javascript' stage, which is added in this PR. It looks like the pipelines are only initialised when the branch is first built, so we should be OK to merge this, as long as we don't rebase the existing PRs.

benthorner:
https://ci.integration.publishing.service.gov.uk/job/content-publisher/job/associations/
https://ci.integration.publishing.service.gov.uk/job/content-publisher/job/npm/

PR #108, opened 2018-08-08
Add basic govspeak editor and contextual guidance

alex-ju:
This PR:
 - Sets up a basic markdown editor using GitHub's toolbar web component. For the moment the editor doesn't contain the govspeak syntax and will be extended in a separate PR
 - Add contextual guidance component with dummy content for now. It will be updated when the content is ready and the content will be separated from the view.

basic-markdown-editor

[Trello card](https://trello.com/c/yKPf7Kd6)

### DIFF
--- a/app/assets/javascripts/components/contextual-guidance.js
+++ b/app/assets/javascripts/components/contextual-guidance.js
@@ -1,0 +1,60 @@
+function ContextualGuidance ($module) {
+  if (!$module) $module = document
+  this.$module = $module
+  this.$fields = $module.querySelectorAll('[data-contextual-guidance]')
+}
+
+ContextualGuidance.prototype.handleFocus = function (event) {
+  // Get the target element
+  var target = event.target
+  var guidanceId = ContextualGuidance.prototype.getGuidanceId(target)
+
+  // If we have guidance for the field
+  if (guidanceId) {
+    var guidance = document.querySelector('#' + guidanceId)
+    ContextualGuidance.prototype.hideAllGuidance()
+    ContextualGuidance.prototype.showGuidance(guidance)
+  }
+}
+
+ContextualGuidance.prototype.showGuidance = function (element) {
+  if (element) {
+    element.classList.remove('govuk-visually-hidden')
+  }
+}
+
+ContextualGuidance.prototype.hideAllGuidance = function () {
+  var $guidances = document.querySelectorAll('.app-c-contextual-guidance-wrapper')
+  $guidances.forEach(function ($guidance) {
+    $guidance.classList.add('govuk-visually-hidden')
+  })
+}
+
+ContextualGuidance.prototype.getGuidanceId = function (element) {
+  var guidanceId = element.getAttribute('data-contextual-guidance')
+  return guidanceId
+}
+
+ContextualGuidance.prototype.init = function () {
+  var $fields = this.$fields
+
+  this.hideAllGuidance()
+
+  /**
+  * Loop over all items with [data-contextual-guidance]
+  * Check if they have a matching contextual guidance
+  * If they do, add event listener on focus
+  **/
+  $fields.forEach(function ($field) {
+    var guidanceId = ContextualGuidance.prototype.getGuidanceId($field)
+
+    if (!guidanceId || !document.querySelector('#' + guidanceId)) {
+      return
+    }
+    $field.addEventListener('focus', ContextualGuidance.prototype.handleFocus)
+  })
+}
+
+// Initialise guidance at document level
+var guidance = new ContextualGuidance()
+guidance.init(document)
--- a/app/assets/javascripts/components/markdown-editor.js
+++ b/app/assets/javascripts/components/markdown-editor.js
@@ -1,0 +1,2 @@
+//= require @webcomponents/webcomponentsjs/webcomponents-loader.js
+//= require @github/markdown-toolbar-element/dist/index.umd.js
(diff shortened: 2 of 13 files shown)


app/assets/stylesheets/application.scss
```diff
@@ -1,3 +1,12 @@
+// Include all of the GOV.UK Frontend styles. This includes fonts, and individual components.
```
benthorner: This comment doesn't seem to be referring to the code it's directly above. Do we need this comment?

app/assets/stylesheets/components/_markdown-editor.scss
```diff
@@ -0,0 +1,25 @@
+// TODO: require govuk-frontend
```
benthorner: Is this still TODO?

app/views/documents/fields/_govspeak_input.html.erb
```diff
@@ -1,12 +1,38 @@
...
-    html_for: "document[contents][#{schema.id}]"
-  %>
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <%= render "govuk_publishing_components/components/textarea", {
+      label: {
+        text: schema.label
+      },
+      data: {
+        module: "markdown-editor",
```
benthorner: Could we make the key syntax consistent here?

app/assets/javascripts/components/markdown-editor.js
```diff
@@ -0,0 +1,2 @@
+//= require @webcomponents/webcomponentsjs/webcomponents-loader.js
```
benthorner: I was expecting lots of JS here - a nice surprise 😄.

app/assets/javascripts/components/contextual-guidance.js
```diff
@@ -0,0 +1,69 @@
...
+      $guidance.classList.add('govuk-visually-hidden')
+    })
+  }
+}
+
+ContextualGuidance.prototype.init = function () {
+  var $fields = this.$fields
+
+  // Hide all guidance
+  this.hideGuidance()
```
benthorner: Could we rename this to hideAllGuidance and (therefore) lose the comment?

app/assets/javascripts/components/contextual-guidance.js
```diff
@@ -0,0 +1,69 @@
...
+  }
+  for (var i = 0; i < nodes.length; i++) {
+    callback.call(window, nodes[i], i, nodes)
+  }
+}
+
+ContextualGuidance.prototype.handleFocus = function (event) {
+  // Get the target element
+  var target = event.target
+  var guidanceId = target.getAttribute('data-contextual-guidance')
```
benthorner: Could we refactor uses of guidanceId into a method that returns a guidance, given a target? I notice the same getAttribute/querySelector code is being repeated here.

app/assets/javascripts/components/contextual-guidance.js
```diff
@@ -0,0 +1,69 @@
...
+  }
+}
+
+ContextualGuidance.prototype.showGuidance = function (element) {
+  if (element) {
+    element.classList.remove('govuk-visually-hidden')
+  }
+}
+
+ContextualGuidance.prototype.hideGuidance = function (element) {
```
benthorner: I can't see where this function is being called with a param - I think we always go into the 'else' clause. Do we need the param/if?

app/assets/javascripts/components/contextual-guidance.js
```diff
@@ -0,0 +1,69 @@
+function ContextualGuidance ($module) {
+  if (!$module) $module = document
+  this.$module = $module
+  this.$fields = $module.querySelectorAll('[data-contextual-guidance]')
+}
+
+function nodeListForEach (nodes, callback) {
+  if (window.NodeList.prototype.forEach) {
```
benthorner: I think we've got the 'same' code written twice here - shall we just have the second version?

app/assets/javascripts/components/contextual-guidance.js
```diff
@@ -0,0 +1,69 @@
+function ContextualGuidance ($module) {
+  if (!$module) $module = document
+  this.$module = $module
+  this.$fields = $module.querySelectorAll('[data-contextual-guidance]')
```
benthorner: In other places I've seen '[data-module="blah"]'. Is this one better do you think?

app/assets/javascripts/components/contextual-guidance.js
```diff
@@ -0,0 +1,69 @@
+function ContextualGuidance ($module) {
```
benthorner: Beautiful, beautiful classes 😄 .

app/assets/javascripts/components/contextual-guidance.js
alex-ju: This function is polyfilling the `forEach` for browsers that don't support it. If exists makes use of it.

app/assets/javascripts/components/contextual-guidance.js
alex-ju: I'm not able to use data-module for the contextual guidance as this component spans across the page so I'm using [data-contextual-guidance] to query the fields. I can use a `.js-field-with-guidance` class if we thing it can cause confusion with `[data-modules]` selectors.

alex-ju:
@benthorner Updated addressing all suggestions except for two for which I added comments.

app/assets/javascripts/components/contextual-guidance.js
benthorner: OK. So I understand that a polyfill is a good thing, but we shouldn't be putting them in our actual code - they should be defined globally. Another way of looking at it is that I could delete lines 8-10 and the code would still work...so let's delete lines 8-10 (or write the actual polyfill so it's avaialble globally.)

PR #111, opened 2018-08-09
Add markdown preview

alex-ju:
This PR:
 - Makes the markdown editor a component
 - Allows markdown preview
 - Sets `$govuk-global-styles: true;` to style anchors, paragraphs and headings in markdown preview

### DIFF
--- a/app/assets/javascripts/components/markdown-editor.js
+++ b/app/assets/javascripts/components/markdown-editor.js
@@ -1,2 +1,75 @@
 //= require vendor/@webcomponents/webcomponentsjs/webcomponents-loader.js
 //= require vendor/@github/markdown-toolbar-element/dist/index.umd.js
+//= require vendor/marked/lib/marked.js
+
+function MarkdownEditor ($module) {
+  this.$module = $module
+  this.$input = $module.querySelector('textarea')
+  this.$preview = $module.querySelector('.js-preview-body .govuk-textarea')
+  this.$previewButton = $module.querySelector('.js-preview-button')
+  this.$editButton = $module.querySelector('.js-edit-button')
+  this.$editorInput = $module.querySelector('.js-editor-input')
+  this.$previewBody = $module.querySelector('.js-preview-body')
+}
+
+MarkdownEditor.prototype.init = function () {
+  var $module = this.$module
+
+  // Save bounded functions to use when removing event listeners during teardown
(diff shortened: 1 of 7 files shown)


app/assets/javascripts/components/markdown-editor.js
```diff
@@ -1,2 +1,70 @@
...
+
+  var $preview = this.$preview
+  var text = this.$input.value
+
+  // Mirror textarea's height
+  $preview.style.height = this.$input.offsetHeight + 'px'
+
+  // Render markdown
+  window.marked(text, function (err, content) {
+    if (err) throw err
```
benthorner: What kind of errors are we anticipating here - should we be handling them?

app/assets/javascripts/components/markdown-editor.js
```diff
@@ -1,2 +1,70 @@
...
+
+MarkdownEditor.prototype.toggleElements = function () {
+  this.toggle(this.$editButton)
+  this.toggle(this.$previewButton)
+  this.toggle(this.$editorInput)
+  this.toggle(this.$previewBody)
+}
+
+MarkdownEditor.prototype.toggle = function (element) {
+  if (element.ownerDocument.defaultView.getComputedStyle(element, null).display === 'none') {
```
benthorner: Would element.style.display not work? Might be good to document why we need the computed version.

app/assets/javascripts/components/markdown-editor.js
alex-ju: marked can return parsing errors and if they happen I'm thinking we don't want them to fail silently.

benthorner: Totally agree we shouldn't forget about the error, but how will the user know they've done something wrong?

app/assets/javascripts/components/markdown-editor.js
alex-ju: we need `getComputedStyle` to evaluate the styles coming from CSS. `element.style.display` only reads styles applied with JavaScript.

app/assets/javascripts/components/markdown-editor.js
alex-ju: They're evaluating the preview coming from content, but they may not realise that indeed. I added 'Error previewing content' as a placeholder if marked throws an error.

PR #112, opened 2018-08-09
Add at least one format for the policy supertype

benthorner:
The lack of any format for the policy supertype is leading to Sentry
errors where the user click 'Continue' but no format is selected.

We should separately also validate the supertype form and the format
form to check that an option has been selected.


benthorner:
@emmabeynon We'll want to add another 'managed_elsewhere' format spec for this.

PR #113, opened 2018-08-09
Extend markdown toolbar

alex-ju:
This PR:
- moves markdown-toolbar in a component
- groups toolbar items in sections using the `details` component and add links to the relevant guidance (the content of these sections will be reviewed with a content designer and potentially read from a separate file)
- ~prevents markdown contextual guidance to overlap footer or other components on small screens~
- extends markdown-toolbar with the following snippets:
  - heading 2 (`##`), heading 3 (`###`), heading 4 (`####`)
  - ~call to action~ (`$CTA [Start](url) %CTA`)
  - link email (`<john.doe@mail.tld>`)
  - ~address~ (`$A 10 Whitechapel $A`)
- ~configures marked renderer to~
  - support line breaks
  - render tables
  - sanitise user input
- removes the absolute position of contextual guidance

[Trello card](https://trello.com/c/yKPf7Kd6)

### DIFF
--- a/app/assets/javascripts/components/markdown-editor.js
+++ b/app/assets/javascripts/components/markdown-editor.js
@@ -22,6 +22,9 @@ MarkdownEditor.prototype.init = function () {
   // Handle events
   this.$previewButton.addEventListener('click', $module.boundPreviewButtonClick)
   this.$editButton.addEventListener('click', $module.boundEditButtonClick)
+
+  // Prevent toolbar item reveals to overlap with footer
+  this.setContextualGuidancePosition()
 }
 
 MarkdownEditor.prototype.handlePreviewButton = function (event) {
@@ -33,16 +36,26 @@ MarkdownEditor.prototype.handlePreviewButton = function (event) {
   // Mirror textarea's height
   $preview.style.height = this.$input.offsetHeight + 'px'
 
-  // Render markdown
-  window.marked(text, function (err, content) {
-    if (err) {
-      $preview.innerHTML = 'Error previewing content'
-      throw err
-    } else {
-      $preview.innerHTML = content
-    }
-  })
-
+  if (text) {
+    // Render markdown
+    window.marked(
+      text,
+      {
+        gfm: true,
+        breaks: true,
+        tables: true,
+        sanitize: true
+      },
+      function (err, content) {
+        if (err) {
+          $preview.innerHTML = 'Error previewing content'
+          throw err
+        } else {
+          $preview.innerHTML = content
+        }
+      }
+    )
+  }
   this.toggleElements()
 }
 
@@ -65,6 +78,16 @@ MarkdownEditor.prototype.toggle = function (element) {
     element.style.display = 'block'
   } else {
     element.style.display = 'none'
+  }
+}
+
+MarkdownEditor.prototype.setContextualGuidancePosition = function () {
+  var guidanceId = this.$input.getAttribute('data-contextual-guidance')
+  if (guidanceId) {
+    var $guidance = document.querySelector('#' + guidanceId + ' .app-c-contextual-guidance')
+    if ($guidance) {
+      $guidance.style.position = 'relative'
+    }
   }
 }
 
--- a/app/assets/javascripts/components/markdown-editor.js
+++ b/app/assets/javascripts/components/markdown-editor.js
@@ -22,9 +22,6 @@ MarkdownEditor.prototype.init = function () {
   // Handle events
   this.$previewButton.addEventListener('click', $module.boundPreviewButtonClick)
   this.$editButton.addEventListener('click', $module.boundEditButtonClick)
-
-  // Prevent toolbar item reveals to overlap with footer
-  this.setContextualGuidancePosition()
 }
 
 MarkdownEditor.prototype.handlePreviewButton = function (event) {
@@ -72,16 +69,6 @@ MarkdownEditor.prototype.toggle = function (element) {
     element.style.display = 'block'
   } else {
     element.style.display = 'none'
-  }
-}
-
-MarkdownEditor.prototype.setContextualGuidancePosition = function () {
-  var guidanceId = this.$input.getAttribute('data-contextual-guidance')
-  if (guidanceId) {
-    var $guidance = document.querySelector('#' + guidanceId + ' .app-c-contextual-guidance')
-    if ($guidance) {
-      $guidance.style.position = 'relative'
-    }
   }
 }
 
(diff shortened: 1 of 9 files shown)


app/views/components/_markdown-toolbar.html.erb
```diff
@@ -0,0 +1,71 @@
+<markdown-toolbar class="app-c-markdown-toolbar" for="<%= html_for %>">
+  <%= render "govuk_publishing_components/components/details", {
+      title: "Headings and sub-headings"
```
benthorner: Which ones are sub-headings?

app/views/components/_markdown-toolbar.html.erb
```diff
@@ -0,0 +1,71 @@
...
+
+  <%= render "govuk_publishing_components/components/details", {
+      title: "Blockquotes"
+  } do %>
+  <ul class="app-c-markdown-toolbar__list">
+    <li class="app-c-markdown-toolbar__item">
+      <md-quote class="app-c-markdown-toolbar__button">Blockquotes</md-quote>
+    </li>
+  </ul>
+  <a href="https://www.gov.uk/guidance/how-to-publish-on-gov-uk/markdown#blockquotes" class="govuk-link" target="_blank"> Guidance on blockquotes</a>
```
benthorner: Interestingly this one seems to toggle them on or off, whereas the other tool buttons just toggle-on. Is that expected?

app/views/components/_markdown-toolbar.html.erb
```diff
@@ -0,0 +1,71 @@
...
+  <% end %>
+
+  <%= render "govuk_publishing_components/components/details", {
```
benthorner: I feel like we shouldn't add this feature until it works end-to-end - if it's in the toolbar then I'd expect it to work in preview mode, otherwise it's liable to make the user feel like they did something wrong. Can we leave it out of this PR and add it separately?

app/views/components/_markdown-toolbar.html.erb
```diff
@@ -0,0 +1,71 @@
...
+  } do %>
+  <ul class="app-c-markdown-toolbar__list">
+    <li class="app-c-markdown-toolbar__item">
+      <md-link class="app-c-markdown-toolbar__button">Link</md-link>
+    </li>
+    <li class="app-c-markdown-toolbar__item">
+      <md-link-email class="app-c-markdown-toolbar__button">Email link</md-link-email>
+    </li>
+    <li class="app-c-markdown-toolbar__item">
+      <md-link-cta class="app-c-markdown-toolbar__button">Call to action</md-link-cta>
```
benthorner: I feel like we shouldn't add this feature until it works end-to-end - if it's in the toolbar then I'd expect it to work in preview mode, otherwise it's liable to make the user feel like they did something wrong. Can we leave it out of this PR and add it separately?

app/assets/stylesheets/components/_contextual-guidance.scss
```diff
@@ -1,10 +1,18 @@
...
+  @include govuk-responsive-margin(6, "bottom");
+
+  @include mq ($from: desktop) {
+    padding-top: govuk-spacing(6);
+  }
 }
 
 .app-c-contextual-guidance {
-  position: absolute;
+  @include mq ($from: desktop) {
```
benthorner: Could you explain briefly what the change here is doing?

app/assets/javascripts/components/markdown-editor.js
```diff
@@ -68,6 +81,16 @@ MarkdownEditor.prototype.toggle = function (element) {
   }
 }
 
+MarkdownEditor.prototype.setContextualGuidancePosition = function () {
+  var guidanceId = this.$input.getAttribute('data-contextual-guidance')
+  if (guidanceId) {
+    var $guidance = document.querySelector('#' + guidanceId + ' .app-c-contextual-guidance')
```
benthorner: How come this is overriding the style 'position: absolute' - could we not just change the style?

app/views/documents/fields/_govspeak_input.html.erb
```diff
@@ -11,7 +11,7 @@
         id: "govspeak-editor",
         name: "document[contents][#{schema.id}]",
         value: document.contents[schema.id],
-        rows: 30,
+        rows: 25,
```
benthorner: Why has this changed?

app/assets/javascripts/components/markdown-editor.js
```diff
@@ -33,16 +36,26 @@ MarkdownEditor.prototype.handlePreviewButton = function (event) {
...
-      $preview.innerHTML = content
-    }
-  })
-
+  if (text) {
+    // Render markdown
+    window.marked(
+      text,
+      {
+        gfm: true,
```
benthorner: I'm not really sure what this one's doing. I tried 'hello<newline>there' and 'hello<space><space><newline>there' and in the latter the editor inserted a '<dot>' in edit mode. Perhaps this bit (together with these other options), would be better in a separate PR which gives more focus to the features they enable (and maybe adds it to the toolbar?).

app/views/components/_markdown-toolbar.html.erb
alex-ju: This is the current copy. I will update with the rest of the guidance.

app/views/components/_markdown-toolbar.html.erb
alex-ju: Sure, we can leave them aside for now.

app/assets/stylesheets/components/_contextual-guidance.scss
alex-ju: On mobile, the guidance shouldn't be positioned absolute, as they will overlap with the next element, so switched to position them absolute only on desktop when we don't want the height of the guidance to push the next components down the page.

app/assets/javascripts/components/markdown-editor.js
alex-ju: We're overwriting the position only for the contextual guidance that relates to the markdown editor as it contains the expandable toolbar which can overlap the footer.

app/views/documents/fields/_govspeak_input.html.erb
alex-ju: There isn't a fixed value for this. I initially set this arbitrary at 30, but it seems to be to tall. Will discuss with the designer to figure out what will be a 'good' height for the editor.

app/assets/javascripts/components/markdown-editor.js
alex-ju: Yes, I can remove them from this PR and add them when we decide on what we want to show in the markdown preview.

app/views/components/_markdown-toolbar.html.erb
alex-ju: Not sure I get this one. The 'Blockquotes' button should behave the same with the rest.

PR #116, opened 2018-08-10
Brakeman now runs automatically for Rails apps

erino:
As per https://github.com/alphagov/govuk-jenkinslib/pull/19

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -16,6 +16,7 @@ gem "plek", "~> 2"
 gem "uglifier", "~> 4"
 
 group :development do
+  gem "brakeman", "~> 4"
   gem "listen", "~> 3"
 end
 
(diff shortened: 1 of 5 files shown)


erino:
@benthorner Isn't the current behaviour that brakeman doesn't run locally? as it's not included in the Gemfile?

benthorner:
@erino Yes, but that's not optimal as it means we can't debug brakeman errors without manually setting it up locally - it should really be part of the local development environment, like the linting is. I'm suggesting that we add it to the Gemfile and ensure it runs after the linting, so we're not dependent on CI for debugging output.

erino:
@benthorner Cool, just checking that I wasn't missing anything

benthorner:
@erino 👍 but also need to change the `Rakefile`.

benthorner:
End of Friday PR cleanout ;-).

erino:
thanks for merging!

PR #118, opened 2018-08-10
Separate testing features and formats

emmabeynon:
For https://trello.com/c/b5RLbDyJ/118-users-can-add-a-primary-organisation
Paired with @davbo 

This refactors our test suite so that we have a clear distinction between format-specific tests and feature tests where format-specific behaviour shouldn't have an impact.  In doing so, we create a document_type_schema factory which allows us to better test features by using a non-format-specific schema.


spec/features/edit_a_document_spec.rb
```diff
@@ -3,62 +3,46 @@
...
+  def when_i_go_to_edit_the_document
+    visit document_path(Document.last)
+    click_on "Edit document"
+    expect(page).to have_content("Existing body")
+  end
 
-    def then_i_see_the_document_is_saved
-      expect(page).to have_content "A summary of the release."
+  def and_i_fill_in_the_content_fields
+    fill_in "document[summary]", with: "A summary of the release."
```
benthorner: The summary is making a sneaky first appearance here. Is it necessary for this test?

spec/factories/document_factory.rb
```diff
@@ -5,19 +5,7 @@
...
-        DocumentTypeSchema.all.select { |schema| schema.contents.any? { |field| field.id == "body" } }.sample.id
-      end
-    end
-
-    trait :with_associations_in_schema do
-      document_type do
-        DocumentTypeSchema.all.select { |schema| schema.associations.any? }.sample.id
-      end
-    end
+    base_path { title ? "/prefix/#{title.parameterize}" : nil }
```
benthorner: If we set base_path to a random string, would the tests still pass?

spec/features/edit_a_document_spec.rb
```diff
@@ -3,62 +3,46 @@
...
+    expect(@request).to have_been_requested
+    expect(page).to have_content "Preview creation successful"
 
-      expect(a_request(:put, /content/).with { |req|
-        expect(req.body).to be_valid_against_schema(@schema.publishing_metadata.schema_name)
-        expect(JSON.parse(req.body)["description"]).to eq "A summary of the release."
-      }).to have_been_requested
-    end
+    expect(a_request(:put, /content/).with { |req|
+      expect(JSON.parse(req.body)["description"]).to eq "A summary of the release."
```
benthorner: Could we use the technique in the new format specs to DRY this up a bit?

spec/features/edit_document_associations_spec.rb
```diff
@@ -3,94 +3,69 @@
...
-
-    def and_i_click_on_edit_associations
-      click_on "Edit associations"
-    end
+  let(:initial_association) { { "content_id" => SecureRandom.uuid, "internal_name" => "Initial association" } }
+  let(:association_to_select_1) { { "content_id" => SecureRandom.uuid, "internal_name" => "Association to select 1" } }
+  let(:association_to_select_2) { { "content_id" => SecureRandom.uuid, "internal_name" => "Association to select 2" } }
+
+  scenario "User edits associations to a document" do
+    given_there_is_a_document_with_associations
```
benthorner: Slight inconsistency with this step and the API down version. I'd be happy with 'given_there_is_a_document'.

spec/features/save_document_associations_api_down_spec.rb
```diff
@@ -13,11 +13,12 @@
...
   def given_there_is_a_document_with_associations
-    @document = create(:document, :with_associations_in_schema)
-
-    @document.document_type_schema.associations.each do |schema|
-      publishing_api_has_linkables([], document_type: schema.document_type)
-    end
+    association_schema = attributes_for(:association, type: "multi_association")
+    document_type_schema = build(:document_type_schema, associations: [association_schema])
+    association = { association_schema[:id] => ["a-content-id"] }
+    linkable = { "content_id" => ["a-content-id"], "internal_name" => "Association" }
```
benthorner: We don't actually need this - can just pass `[]` to the stub, as we're going to take it down anyway and doing the request doesn't depend on the select box having any items.

spec/features/show_a_document_api_down_spec.rb
```diff
@@ -11,18 +11,10 @@
...
-      publishing_api_has_linkables([], document_type: schema.document_type)
-    end
-
-    associations = @document.document_type_schema.associations.map do |schema|
-      [schema.id, [SecureRandom.uuid]]
-    end
-
-    @document.update(associations: Hash[associations])
-    publishing_api_isnt_available
+    association_schema = attributes_for(:association, type: "multi_association")
```
benthorner: We should probably all this factory 'association_schema' to avoid confusion with an actual association.

spec/features/show_a_document_api_down_spec.rb
```diff
@@ -11,18 +11,10 @@
...
+    document_type_schema = build(:document_type_schema, associations: [association_schema])
+    association = { association_schema[:id] => ["a-content-id"] }
```
benthorner: Could we change this to `associations`?

spec/features/edit_document_associations_spec.rb
```diff
@@ -3,94 +3,69 @@
...
-    end
-
-    def when_i_visit_the_document_page
-      visit document_path(@document)
-    end
-
-    def and_i_click_on_edit_associations
-      click_on "Edit associations"
-    end
+  let(:initial_association) { { "content_id" => SecureRandom.uuid, "internal_name" => "Initial association" } }
```
benthorner: We're using SecureRandom.uuid here and hard-coded strings in other specs - maybe we could have an association factory to build these things, as they have a defined form.

spec/services/content_validator_spec.rb
```diff
@@ -57,7 +57,9 @@
 
   describe 'custom validation' do
     it 'raises issue if the summary is not set' do
-      document = build(:document, :with_body_in_schema, contents: { body: "Too short" })
+      body_field = { id: "body", label: "Body", type: "govspeak", validations: { "min_length" => 10 } }
```
benthorner: The number of times we have a field hash is making me think we should have a Field factory and use attributes_for - what do you think?

spec/formats/create_a_statistical_data_set_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+        "organisations" => [linkable["content_id"]]
+      },
+      "title" => "A great title",
+      "document_type" => "statistical_data_set",
+      "description" => "A great summary",
+    }
+  end
+
+  def linkable
+    @linkable ||= { "content_id" => SecureRandom.uuid, "internal_name" => "Linkable" }
```
benthorner: The number of times we're making this linkable hash makes me think we should have a factory and use attributes_for - what do you think?

spec/factories/document_factory.rb
benthorner: We discussed this a bit in person and agreed this is a good compromise that is both random and roughly realistic.

spec/features/edit_document_associations_spec.rb
benthorner: This was already an issue in the previous specs, so let's solve it separately.

spec/formats/create_a_statistical_data_set_spec.rb
benthorner: This was already in issue in the previous specs - let's solve it separately.

benthorner:
💘

PR #123, opened 2018-08-13
Update contextual guidance component

alex-ju:
This PR:
 - defines typography styles for the component instead of relying on the defaults
 - defines a default example for the component


app/assets/stylesheets/components/_contextual-guidance.scss
```diff
@@ -1,4 +1,6 @@
 .app-c-contextual-guidance-wrapper {
+  @include govuk-font(19);
```
benthorner: This looks the same to me - what's wrong with the defaults?

PR #124, opened 2018-08-13
Update markdown editor component

alex-ju:
This PR:
 - makes the preview bar hidden by default (when JavaScript is disabled the preview bar will not be shown) and enable it via JavaScript
 - moves the "Nothing to preview" message from markup to JavaScript to be consistent with "Error previewing content" - it won't be shown/used anyway without having JavaScript enabled.
 - adds a `js-markdown` prefix for element queries to prevent clashes with other components
 - adds a default example for the markdown editor component

### JavaScript enabled
screen shot 2018-08-13 at 14 34 11

### JavaScript disabled
screen shot 2018-08-13 at 14 33 52

### DIFF
--- a/app/views/components/_markdown-editor.html.erb
+++ b/app/views/components/_markdown-editor.html.erb
@@ -1,17 +1,23 @@
+<%
+  textarea ||= {}
+  edit_button_text ||= 'Edit'
+  preview_button_text ||= 'Preview'
+%>
 <div class="app-c-markdown-editor" data-module="markdown-editor">
   <%= render "govuk_publishing_components/components/label", {
-      text: textarea[:label][:text],
+      text: label[:text],
       html_for: textarea[:id]
   } %>
-  <% textarea[:label] = nil %>
-  <div class="app-c-markdown-editor__preview-toggle">
-    <button class="app-c-markdown-editor__button app-c-markdown-editor__button--edit js-edit-button">Edit markdown</button>
-    <button class="app-c-markdown-editor__button app-c-markdown-editor__button--preview js-preview-button">Preview markdown</button>
+  <div class="app-c-markdown-editor__preview-toggle js-markdown-toggle-bar">
+    <button class="app-c-markdown-editor__button app-c-markdown-editor__button--edit js-markdown-edit-button"><%= edit_button_text %></button>
+    <button class="app-c-markdown-editor__button app-c-markdown-editor__button--preview js-markdown-preview-button"><%= preview_button_text %></button>
   </div>
-  <div class="app-c-markdown-editor__input js-editor-input">
+  <div class="app-c-markdown-editor__input js-markdown-editor-input">
     <%= render "govuk_publishing_components/components/textarea", textarea %>
   </div>
-  <div class="app-c-markdown-editor__preview js-preview-body">
-    <div class="govuk-textarea">Nothing to preview</div>
+  <div class="app-c-markdown-editor__preview js-markdown-preview-body">
+  <%= render "govuk_publishing_components/components/govspeak", { } do %>
+    <div class="govuk-textarea"></div>
+  <% end %>
   </div>
 </div>
--- a/app/views/components/docs/markdown-editor.yml
+++ b/app/views/components/docs/markdown-editor.yml
@@ -1,7 +1,11 @@
 name: Markdown editor
 description: Allows editing and previewing markdown
 body: |
-  Optional markdown providing further detail about the component
+  If JavaScript is disabled this component falls back to a textarea component
+  without the preview functionality
+part_of_admin_layout: true
+shared_accessibility_criteria:
+  - link
 accessibility_criteria: |
   The component must:
 
@@ -12,4 +16,11 @@ accessibility_criteria: |
   - indicate when they have focus
   - have correctly associated labels
 examples:
-  default: {}
+  default:
+    data:
+      label:
+        text: Body
+      textarea:
+        name: markdown-editor
+        id: markdown-editor
+
--- a/app/assets/javascripts/components/markdown-editor.js
+++ b/app/assets/javascripts/components/markdown-editor.js
@@ -4,12 +4,13 @@
 
 function MarkdownEditor ($module) {
   this.$module = $module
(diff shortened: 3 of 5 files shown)


app/views/components/_markdown-editor.html.erb
```diff
@@ -1,17 +1,26 @@
+<%
+  textarea||= {}
+  edit_button_text||= 'Edit'
+  preview_button_text||= 'Preview'
+%>
 <div class="app-c-markdown-editor" data-module="markdown-editor">
+  <% if textarea[:label] %>
   <%= render "govuk_publishing_components/components/label", {
```
benthorner: Could you indent these, so it's easier to follow the start/end of the blocks?

app/views/components/_markdown-editor.html.erb
```diff
@@ -1,17 +1,26 @@
+<%
+  textarea||= {}
```
benthorner: Missing spaces here (to be consistent with the Ruby style guide).

app/views/components/_markdown-editor.html.erb
```diff
@@ -1,17 +1,26 @@
...
+  edit_button_text||= 'Edit'
+  preview_button_text||= 'Preview'
+%>
 <div class="app-c-markdown-editor" data-module="markdown-editor">
+  <% if textarea[:label] %>
   <%= render "govuk_publishing_components/components/label", {
       text: textarea[:label][:text],
       html_for: textarea[:id]
   } %>
   <% textarea[:label] = nil %>
```
benthorner: I know this isn't part of the PR, but I'm not sure what this line is for. Do we need it?

app/views/components/_markdown-editor.html.erb
```diff
@@ -1,17 +1,26 @@
...
-    <button class="app-c-markdown-editor__button app-c-markdown-editor__button--edit js-edit-button">Edit markdown</button>
-    <button class="app-c-markdown-editor__button app-c-markdown-editor__button--preview js-preview-button">Preview markdown</button>
+  <% end %>
+  <div class="app-c-markdown-editor__preview-toggle js-markdown-toggle-bar">
+    <button class="app-c-markdown-editor__button app-c-markdown-editor__button--edit js-markdown-edit-button"><%= edit_button_text %></button>
+    <button class="app-c-markdown-editor__button app-c-markdown-editor__button--preview js-markdown-preview-button"><%= preview_button_text %></button>
   </div>
-  <div class="app-c-markdown-editor__input js-editor-input">
+  <div class="app-c-markdown-editor__input js-markdown-editor-input">
+    <% if textarea[:name] %>
```
benthorner: Can you explain what this new conditional is for?

app/views/components/_markdown-editor.html.erb
alex-ju: Indented

app/views/components/_markdown-editor.html.erb
alex-ju: Space added

app/views/components/_markdown-editor.html.erb
alex-ju: The minimum we need to render a textarea component is its name, hence we make sure we have one set before trying to render it

app/views/components/_markdown-editor.html.erb
alex-ju: The markdown editor tries to emulate the textarea component in terms of how we structure the attributes when rendering a component (as shown in the example below).
```
    <%= render "components/markdown-editor", {
      textarea: {
        label: {
          text: schema.label
        }
      }
    } %>
```
The markdown editor is different from a textarea component by having the preview toolbar between the label element and the textarea element. Because of this, we don't render the label as part of the textarea component, but separately (L8-L11) then we remove the label from the textarea object (that stands as parameter for the textarea component).

Hope that makes sense.

benthorner: I think I see the aim, but I still don't understand if we actually need that line or not. Does the code work without it?

app/views/components/_markdown-editor.html.erb
benthorner: Hmm...this sounds like a validation in disguise to me. What would be the impact of not having this conditional, assuming people use the component as documented in the manual?

app/views/components/_markdown-editor.html.erb
alex-ju: We need it otherwise we'll have two labels. We need to stop the label going to the textarea component and render the 2nd time.

app/views/components/_markdown-editor.html.erb
alex-ju: The textarea component will complain (error) about not having a name

benthorner: What I'm getting at is this: is there a use case for this component where we don't specify a name?

app/views/components/_markdown-editor.html.erb
alex-ju: Dropped the approach when we reuse the label parameter passed to the textarea component and moved into defining a separate label for the markdown editor component.

app/views/components/_markdown-editor.html.erb
alex-ju: Removed this condition.

app/views/components/_markdown-editor.html.erb
```diff
@@ -1,17 +1,23 @@
+<%
+  textarea||= { }
```
benthorner: Indentation :-(

benthorner: Oops, I meant spacing :-(

alex-ju: Fixed

app/views/components/docs/markdown-editor.yml
```diff
@@ -12,4 +16,11 @@ accessibility_criteria: |
   - indicate when they have focus
   - have correctly associated labels
 examples:
-  default: {}
+  default:
+    data:
+      textarea:
```
benthorner: I think these examples might be out of date...

benthorner:
One important thing I forgot - there's also a 'show_api_down.html.erb' view.

PR #126, opened 2018-08-14
Display the "publication state" to the user

tijmenb:
This PR makes the app keep track of the "publication state". There are 8 states, and they're documented in the locale file.

In this PR we only display the state. In upcoming PRs we'll change the view depending on the state - for example, only display the preview button if the content has been sent to draft.

This is part of a [number of states](https://docs.google.com/drawings/d/14h2MU5rN1NF8O6rP17Ca47IxEpuKJvVKSnOYIG2i6UY/edit) that we'll have in this app.

Easiest to review per commit.

https://trello.com/c/i6C3pY03


app/services/publication_state.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+class PublicationState
```
kevindew: Maybe move this into Document? it doesn't feel like a service.

Potentially can use: https://til.hashrocket.com/posts/b492fb26ec-rails-enums-and-postgresql-enums ?

config/locales/en.yml
```diff
@@ -31,3 +31,29 @@
...
+    sent_to_draft:
+      name: Saved as draft
+      description: The page is now visible in draft.
+    error_sending_to_draft:
+      name: Error sending to draft
+      description: We've encountered a problem making your preview available.
+    publishing:
+      name: Publishing
+      description: We're currently publishing a new edition to GOV.UK.
+    published:
```
kevindew: Maybe "Live on GOV.UK"?

tijmenb: Done.

app/services/publication_state.rb
tijmenb: 😞 Doesn't look like custom types work very well with Rails. Normal schema dump won't work with custom types:

screen shot 2018-08-14 at 17 38 59

And when I switch to dump as `schema.sql`, I get this error:

```
psql:/var/govuk/content-publisher/db/structure.sql:20: ERROR:  must be owner of extension plpgsql
rake aborted!
failed to execute:
psql -v ON_ERROR_STOP=1 -q -f /var/govuk/content-publisher/db/structure.sql content_publisher_test
``` 

I've moved the constant into `Document` instead.

kevindew:
I had a thought about this and us tracking of error state here. We hit an area of ambiguity for things that have once been draft or once been live. E.g a state of error_sending_to_draft could mean that we have no draft on gov.uk or that we have an out of date draft on gov.uk - I guess there's a similar angle with live if we've not managed to sync a live transition (e.g. to unpublishing). Could be that we have just 3 states of pre_draft, draft, live and then track separately whether this item is syncing with publishing_api or failed to do so, what do you think?

Ah yes, that's annoying about the structure.sql file and that looks familiar - I think that's what blocked us last time we tried. I assume we could get that file to actually work but it's a bit of a wtf having that file rather than the familiar schema.sql oh well!

You could switch to a rails enum with strings as per https://stackoverflow.com/a/26698885 as these appear to now not need integers, but I've never tried it - You'd be able to kill of the validation as you'd get an ArgumentError instead. It'd also add methods like Document#newly_created?

spec/services/document_publishing_service_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+  end
+
+  describe '#publish' do
+    it "keeps track of the publication state" do
+      stub_any_publishing_api_publish
+      document = create(:document)
+      allow(document).to receive(:update!)
+
+      DocumentPublishingService.new.publish(document)
+
```
benthorner: Could we use subject here instead? It’s exactly the same.

spec/features/create_a_document_api_down_spec.rb
```diff
@@ -42,5 +42,6 @@ def then_i_see_the_document_exists
   def and_the_preview_creation_failed
     expect(@request).to have_been_requested
     expect(page).to have_content "Error creating preview"
+    expect(Document.last.publication_state).to eq "error_sending_to_draft"
```
benthorner: Could we check this using the page content instead - something visible to the user?

spec/features/edit_a_document_spec.rb
```diff
@@ -41,5 +41,7 @@ def and_the_preview_creation_succeeded
     expect(a_request(:put, /content/).with { |req|
       expect(JSON.parse(req.body)["details"]["body"]).to eq "<p>Edited body.</p>\n"
     }).to have_been_requested
+
+    expect(Document.last.publication_state).to eql("sent_to_draft")
```
benthorner: Could we check this using the content instead?

spec/features/publish_a_document_api_down_spec.rb
```diff
@@ -28,5 +28,6 @@ def when_i_try_to_publish_the_document
   def then_i_see_the_publish_failed
     expect(@request).to have_been_requested
     expect(page).to have_content "Error publishing"
+    expect(@document.reload.publication_state).to eql("error_in_publishing")
```
benthorner: Could we check this using the content instead?

spec/features/publish_a_document_spec.rb
```diff
@@ -31,5 +31,6 @@ def and_i_confirm_the_publishing
   def then_i_see_the_publish_succeeded
     expect(@request).to have_been_requested
     expect(page).to have_content "Publish successful"
+    expect(@document.reload.publication_state).to eql("live_on_govuk")
```
benthorner: Can we check this using the content?

spec/services/document_publishing_service_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+RSpec.describe DocumentPublishingService do
+  describe '#publish_draft' do
+    it "keeps track of the publication state" do
+      stub_any_publishing_api_put_content
+      document = create(:document)
+      allow(document).to receive(:update!)
+
+      DocumentPublishingService.new.publish_draft(document)
+
+      expect(document).to have_received(:update!).with(publication_state: "sending_to_draft")
```
benthorner: This is very sneaky :-)

spec/features/publish_a_document_spec.rb
benthorner: Also, is there a reason to use eql here? I think we mostly use eq elsewhere.

benthorner:
Also wondering if we need a validation - we’ve got tests for all the states and we’re the only people who can set those states. What are we afraid of?

spec/features/create_a_document_api_down_spec.rb
tijmenb: I would like to wait with that. There's no design for things like the error states yet, and I expect it to be different from what we do at the moment.

spec/services/document_publishing_service_spec.rb
benthorner: OK. I think this has come up before, but I don't think we resolved it.

spec/features/publish_a_document_spec.rb
tijmenb: Not sure this is important, we use both on GOV.UK. If you feel strongly you could make a PR to govuk-lint to enforce a default.

tijmenb:
@benthorner the validation is there to enforce data consistency. It prevents us from typos and makes it obvious if we have incorrect data in the database.

spec/features/create_a_document_api_down_spec.rb
kevindew: I feel like this test is a bit strange as a feature test anyway - should we have a controller test that just checks that controller responds correctly and appropriate states change? 

Feels quite specific for a high level feature test.

benthorner:
@kevindew Yes, go ahead and merge. That said, in future I think our cards should make it clearer what the user sees as a result of the work we're doing, so that we can have a design discussion before we do the work. In this case, it would have been good to test the state at a feature level by displaying it to the user, but we haven't had the design discussion to know if we want to display it.

PR #130, opened 2018-08-15
Add preview & shareable preview links

tijmenb:
This temporarily adds 3 links to the action menu:

- View on GOV.UK - for the live version of the page
- Preview - the draft page on the draft stack, Signon required
- Shareable preview - the draft page with secret token to bypass Signon

I'm assuming we'll redesign the placing of these links, so I haven't added integration tests for this. This PR is just to add the logic of the links.

The "bypass" code is mostly taken from mainstream publisher and collections-publisher (who introduced the same feature recently - https://github.com/alphagov/collections-publisher/pull/428). It's a complicated system, and I hope I've explained it a bit in the comments.

For this to work on production, we'll need to make `JWT_AUTH_SECRET` available to content publisher: https://github.com/alphagov/govuk-puppet/pull/7968.

https://trello.com/c/ECYqAHfX

### DIFF
--- a/app.json
+++ b/app.json
@@ -9,6 +9,9 @@
     },
     "GOVUK_APP_DOMAIN": {
       "required": true
+    },
+    "JWT_AUTH_SECRET": {
+      "generator": "secret"
     }
   },
   "addons": [
--- a/app/services/document_url.rb
+++ b/app/services/document_url.rb
@@ -1,0 +1,54 @@
+# frozen_string_literal: true
+
+class DocumentUrl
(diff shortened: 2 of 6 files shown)


app.json
```diff
@@ -9,6 +9,9 @@
     },
     "GOVUK_APP_DOMAIN": {
       "required": true
+    },
+    "JWT_AUTH_TOKEN": {
```
emmabeynon: Should this be `JWT_AUTH_SECRET`?

PR #131, opened 2018-08-15
Replace hard copy with translations

benthorner:
Using the Rails translations framework allows us to centralise the text in 
our views, so that it's easier to find and edit. This includes format-specific 
text, for which we would otherwise require additional modelling/lookup 
behaviour. Using the Rails translations framework for everything means 
we can also test that all of the text has been defined for the views.

Translations are organised into a hierarchy in one or more YAML files. There 
are multiple files to make finding things a bit easier, with each file following 
the same conventions for the upper levels of the hierarchy, which will make 
it easier to add new groups and avoids conflict. Here are some examples.

```
# formats/supertypes/en.yml (choosing a supertype)
en:
  formats:
    supertypes:
      news:
        label: News
        description: Some news

# formats/document_types/en.yml (guidance and labels)
en:
  formats:
    document_types:
      press_release:
        label: Press release
        guidance: ...

# layouts/en.yml (layout-specific text)
# not very interesting

pages/new_document/en.yml (new_document page-specific text)
en:
  pages:
    documents:
      index:
        title: No title
        actions:
          new: New document
        search_results:
          untitled: No title
```

This PR also makes sure we're using Plek for the URL preview, which 
could be split out. Other PRs that we should do include: lifting the text 
out of components (excluded here because the approach is unclear); 
and changing the tests to use translations instead of hard-coded text.


benthorner:
Oh, and another PR is to do extract the text from the content validator. I also excluded this here because I think it doesn't seem to fit into any of the categories so far (formats, layouts and pages). Perhaps a new 'models' or 'services' category would make sense, so that it's clearer where the translations are being used.

app/formats/document_types.yml
```diff
@@ -1,102 +1,83 @@
 ---
 - id: press_release
```
tijmenb: The extraction of copy from `document_types.yml` is not an improvement I think. Everything used to be in one file, now it's spread out. If we want to keep things together that change together, than this would be better. Can we do this without putting it into a separate file?

spec/factories/document_type_schema_factory.rb
```diff
@@ -5,6 +5,46 @@
     supertype { "not-sure" }
     id { SecureRandom.hex(4) }
     document_type { SecureRandom.alphanumeric(8) }
-    initialize_with { DocumentTypeSchema.create(attributes.stringify_keys) }
+    contents { [] }
+    associations { [] }
+
+    transient do
+      translations do
```
tijmenb: What does this do? It rings a lot of alarm bells that we have to do this.

config/locales/formats/associations/en.yml
```diff
@@ -0,0 +1,11 @@
+en:
```
tijmenb: To avoid having lots of `en.yml` files, you can put `config.i18n.load_path += Dir[Rails.root.join('config', 'locales', 'en', '*.yml')]` in your `application.rb`. Then you can put everything in `/locales/en/*.yml`.

app/views/documents/edit.html.erb
```diff
@@ -17,27 +17,28 @@
...
     </div>
+
     <div class="govuk-grid-column-one-third">
       <%= render "components/contextual-guidance", {
         id: "document-title-guidance",
-        title: "Writing a news title"
+        title: t("formats.document_types.#{@document.document_type}.guidance.title.title")
       } do %>
-        <p class="govuk-body">The title must make clear what the content offers users. Use the words your users do to help them find this. Avoid wordplay or teases.</p>
+        <p class="govuk-body"><%= t("formats.document_types.#{@document.document_type}.guidance.title.body") %></p>
```
davbo: I understand guidance on a field could vary on document_type but I don't believe we have this need right now? maybe we can place the guidance further up the hierarchy, that could go some way to avoiding the need to define translations as part of the factory.

app/formats/document_types.yml
davbo: I agree with @tijmenb  - even if it means we're not able to load these strings using the `t()` function I think it's fine to have the copy in `document_types.yml`

benthorner: I've moved the format-specific text back into supertypes.yml and document_types.yml. Doing it this way requires some additional modelling, so we've now got a Guidance model and a lookup method to avoid too many [][]'s.

spec/factories/document_type_schema_factory.rb
benthorner: This used to make up for the fact that guidance was enforced, but the factory-generated schemas had none. It was the extra bit of dynamic config to make format-specific translations work in test, but it's no longer needed.

app/views/documents/edit.html.erb
benthorner: I understand it's going to vary on most things. Hopefully this is now less of a concern now the format-specific guidance is in document_types.yml

config/locales/formats/associations/en.yml
benthorner: Good idea - done.

benthorner:
@tijmenb @davbo This should be ready to review again now.

config/locales/en/documents/show.yml
```diff
@@ -0,0 +1,51 @@
+en:
+  documents:
+    show:
+      title: No title
```
tijmenb: Should this be `fallback_title` or something?

benthorner: I've renamed it - I was trying to be consistent with the other pages, but yeah it's not always the title.

PR #132, opened 2018-08-15
Use differnt Plek method to avoid errors

benthorner:
The preview use of website_root would require us to configure
    another Heroku environment variable. Using the external_url_for
    method seems to achieve the same thing and on Heroku falls back
    to the GOVUK_APP_DOMAIN (which is incorrect, but we don't care).

### DIFF
--- a/app/services/document_url.rb
+++ b/app/services/document_url.rb
@@ -6,7 +6,7 @@ class DocumentUrl
   end
 
   def public_url
-    Plek.new.website_root + document.base_path
+    Plek.new.find('www') + document.base_path
   end
 
   def preview_url


benthorner:
@tijmenb You can see if you try to create a document. One error is about the API adapters (think we introduced that last week). There's also another error when Plek can't generate the preview link. I've added a description above, now that I'm satisfied this PR fixes the problem.

tijmenb:
@benthorner do you mean https://sentry.io/govuk/app-content-publisher/issues/648262925?

benthorner:
Yeah, but this time it's on Heroku. We could define another environment variable to fix it, but this seems easier.

kevindew:
@benthorner Just curious - did we resolve this a different way?

PR #133, opened 2018-08-15
Enforce GOVUK_WEBSITE_ROOT (for preview URL)

benthorner:


benthorner:
@kevindew This replaces the previous PR - I got confused about the difference between .find('www') and 'website_root'.

PR #134, opened 2018-08-15
Fix renamed state name in translations

tijmenb:
The `publishing` state was renamed to `sending_to_live` at some point. This fixes that and adds tests.

https://trello.com/c/i6C3pY03

### DIFF
--- a/spec/models/document_spec.rb
+++ b/spec/models/document_spec.rb
@@ -1,0 +1,14 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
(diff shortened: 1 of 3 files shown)


spec/models/document_spec.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
```
kevindew: quotes 😉

kevindew: quotes 😉

PR #135, opened 2018-08-15
Start exposing the inner workings of the application

tijmenb:
This is a first stab at a documentation approach for this app. The target of these docs are the people working on it - the PM, DM, user researchers, designers, content designers, and developers.

My dream for documentation is that content publisher will be able to "explain itself", rather than needing developers to explain it. It's mainly inspired by the difficulty we have in understanding the current workings of Whitehall.

This PR adds two things to `/documentation`:

- All the formats. This is to answer the question "What formats are currently available in this application?". Instead of reading the code, or clicking through the entire app, team members have access to a single list.
- All the publication states. This is especially important because some will be rare or transient, so having a central place to show all the states and translations can make it clearer to the team how the app works.

https://content-publisher-revie-pr-135.herokuapp.com/documentation

I expect this section to grow as the app becomes more complicated.

### DIFF
--- a/spec/features/documentation_spec.rb
+++ b/spec/features/documentation_spec.rb
@@ -1,0 +1,18 @@
+# frozen_string_literal: true
+
+require "spec_helper"
+
+RSpec.feature "Documentation" do
+  scenario "GDS staff visits the docs page" do
+    when_i_visit_the_documentation_page
+    then_i_see_the_documentation_page
+  end
+
+  def when_i_visit_the_documentation_page
+    visit "/documentation"
+  end
+
+  def then_i_see_the_documentation_page
+    expect(page).to have_content "App documentation"
+  end
+end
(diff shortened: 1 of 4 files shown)


kevindew:
Neat - this looks great 👍

spec/features/documentation_spec.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+require "spec_helper"
+
+RSpec.feature "Documentation" do
+  scenario "GDS staff visits the docs page" do
+    when_i_visit_the_documentation_page
+    then_i_see_the_documentation_page
```
erino: Is the plan to add a spec that loops through the formats defined in the yaml and asserts that they are mentioned here in the documentation?

tijmenb: Not really - I think if the page becomes more complicated there will be more tests, but this seems okay-ish for now.

PR #136, opened 2018-08-15
Run full rake on CI so we match dev

benthorner:
Running the CI pipeline currently has a different result than running
rake locally, as the default rake task has more strict checks than those
on CI. Rather than trying to sync the two, which would affect other
apps, we can go back to running both and benefit from their combination.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -11,11 +11,6 @@ node {
       }
     },
     rubyLintDiff: false,
-    rubyLintRails: true,
-    overrideTestTask: {
-      stage("Run tests") {
-        sh("bundle exec rake spec")
-      }
-    }
+    rubyLintRails: true
   )
 }
--- a/db/migrate/20180814134709_add_publication_state_to_documents.rb
+++ b/db/migrate/20180814134709_add_publication_state_to_documents.rb
@@ -1,6 +1,10 @@
+# frozen_string_literal: true
+
 class AddPublicationStateToDocuments < ActiveRecord::Migration[5.2]
-  def change
+  def up
     execute "TRUNCATE documents"
-    add_column :documents, :publication_state, :string, null: false
+    add_column :documents, :publication_state, :string, null: false # rubocop:disable Rails/NotNullColumn
   end
+
+  def down; end
 end


kevindew:
Does this mean we'll be linting twice - but slightly differently?

benthorner:
@kevindew Yes. We're doing more linting locally than CI does, but it's default behaviour is also to run 'rake', rather than the override I put in. The other option is of course to change the CI behaviour, but I don't have the time to go and fix every app that breaks from it. We could also introduce yet another option into the Jenkins library, but again that seems to be just increasing the variability.

kevindew:
@benthorner yeah it seems a pain. I want to try avoid us running linting twice, but don't expect fixing other apps.

if we set a [testTask](https://github.com/alphagov/govuk-jenkinslib/blob/master/vars/govuk.groovy#L614) option of `rspec` do we get same test behaviour?

I guess the problem we're also caught on is whether to lint all the dirs as we're currently doing in rake or the gov.uk convention of "app lib spec test" I think if we wanted to lint more than those it'd be better to provide another option for the dirs (and maybe we can provide an asterisk)

benthorner:
@kevindew It sounds like you're aiming for the 'configurable option' approach. I guess this makes sense - every project should really be using '*' and a .rubocop.yml file for the actual configuration.

benthorner:
I still feel we should be running a full 'rake' somewhere on CI. It doesn't actually take long to lint twice, even if it's unintuitive. In general, we may have more checks in a full 'rake' than CI has pre-programmed.

kevindew:
The thing I want to avoid is this app going off on it's tangent and doing things differently. That's very much the purpose of the `govuk.buildProject()` that for better or worse it provides a common workflow and it sounds like if we're doing things like linting twice and breaking out of pipelines we're actually fighting that flow.

Maybe our rake should just do the tests by default if that's the common behaviour for govuk rails apps and linting be a separate task?

benthorner:
OK. How about this https://github.com/alphagov/govuk-jenkinslib/pull/24?

benthorner:
Replaced with https://github.com/alphagov/content-publisher/pull/138.

PR #138, opened 2018-08-15
Lint all directories (subject to rubocop.yml)

benthorner:


Jenkinsfile
```diff
@@ -12,6 +12,7 @@ node {
     },
     rubyLintDiff: false,
     rubyLintRails: true,
+    rubyLintDirs: "*",
     overrideTestTask: {
```
kevindew: Sorry I know not introduced here, but same world. Can this be switched to be `testTask: "spec"` ?

benthorner: Oh yeah - I wasn't sure how to do that when I wrote it. Done.

kevindew: I actually thought we might have had a testTask option so thought we could get it much more succinct but I was wrong - so apologies if this seemed a bit more minor than it was.

PR #140, opened 2018-08-15
Add primary publishing organisation

davbo:
Trello: https://trello.com/c/b5RLbDyJ/118-users-can-add-a-primary-organisation

Action takes place at a couple of points in the journey:

1. Creating a new document
When the user picks their document type and an initial Document is created the primary_publishing_organisation is set to the organisation associated with their signon user.

2. Putting content to the publishing-api
When a draft is saved to the publishing-api the value set in primary_publishing_organisation is copied into original_primary_publishing_organisation. This original_primary_publishing_organisation association isn't visible or editable for the user.

Otherwise it behaves similarly to other associations but renders with a typical single select box.


app/views/documents/associations/_single_association.html.erb
```diff
@@ -0,0 +1,3 @@
+<% service = LinkablesService.new(schema.document_type) %>
+<% content = values ? service.by_content_id(values.first) : nil %>
+<%= content ? content["internal_name"] : "None" %>
```
benthorner: Could we add a translation for this?

app/services/publishing_api_payload.rb
```diff
@@ -34,6 +34,15 @@ def payload
 
 private
 
+  def links
+    associations = document.associations
+    if associations.include?("primary_publishing_organisation") && !associations.include?("original_primary_publishing_organisation")
```
benthorner: We can simplify this for now as the original org will never be there.

app/services/publishing_api_payload.rb
```diff
@@ -34,6 +34,15 @@ def payload
 
 private
 
+  def links
+    associations = document.associations
+    if associations.include?("primary_publishing_organisation") && !associations.include?("original_publishing_organisation")
```
emmabeynon: It might be nice to extract this out into a method to make it a bit easier to read.

app/services/publishing_api_payload.rb
```diff
@@ -34,6 +34,15 @@ def payload
 
 private
 
+  def links
+    associations = document.associations
+    if associations.include?("primary_publishing_organisation") && !associations.include?("original_primary_publishing_organisation")
+      associations["original_primary_publishing_organisation"] = associations["primary_publishing_organisation"]
+      document.update(associations: associations)
```
benthorner: I don’t think we should be changing the document in what looks like an accessory method. We don’t actually need to change it either - the original org is something we could derive each time we send the payload.

spec/features/edit_document_associations_spec.rb
```diff
@@ -18,11 +21,16 @@
...
-    association_schema = build(:association_schema, type: "multi_association", id: "association_id")
-    document_type_schema = build(:document_type_schema, associations: [association_schema])
+    multi_association_schema = build(:association_schema, type: "multi_association", id: "multi_association_id")
+    single_association_schema = build(:association_schema, type: "single_association", id: "single_association_id")
+    document_type_schema = build(:document_type_schema, associations: [multi_association_schema, single_association_schema])
     multi_association_linkables = [initial_association, association_to_select_1, association_to_select_2]
-    publishing_api_has_linkables(multi_association_linkables, document_type: association_schema["document_type"])
-    initial_associations = { association_schema["id"] => [initial_association["content_id"]] }
+    publishing_api_has_linkables(multi_association_linkables, document_type: multi_association_schema["document_type"])
+    publishing_api_has_linkables([selected_association, other_association], document_type: single_association_schema["document_type"])
```
benthorner: Could we reuse the mock associations we already had? It would reduce the number of variables to track, is what I’m thinking.

spec/features/edit_document_associations_spec.rb
```diff
@@ -36,22 +44,28 @@ def and_i_click_on_edit_associations
...
 
   def when_i_edit_the_associations
-    select "Association to select 1", from: "associations[association_id][]"
-    select "Association to select 2", from: "associations[association_id][]"
-    unselect "Initial association", from: "associations[association_id][]"
+    select "Association to select 1", from: "associations[multi_association_id][]"
+    select "Association to select 2", from: "associations[multi_association_id][]"
+    unselect "Initial association", from: "associations[multi_association_id][]"
 
+    select "Other association", from: "associations[single_association_id][]"
```
benthorner: Again, this reads strangely from top to bottom, as it looks like ‘Other’ somehow relates to ‘Association to select 1’, etc.

spec/features/edit_document_associations_spec.rb
```diff
@@ -36,22 +44,28 @@ def and_i_click_on_edit_associations
...
     click_on "Save"
   end
 
   def then_i_can_view_the_associations
     expect(page).to have_content "Association to select 1"
     expect(page).to have_content "Association to select 2"
     expect(page).not_to have_content "Initial association"
+
+    expect(page).to have_content "Other association"
+    expect(page).not_to have_content "Selected association"
```
benthorner: Could we use a ‘within’ block here (or more specific text) to make it clearer which select box we’re testing?

spec/features/primary_publishing_org_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+RSpec.feature "Setting a primary publishing organisation" do
+  scenario "User creates a new document" do
+    given_a_user_belonging_to_an_org_exists
+    when_they_create_a_document
+    and_save_a_draft
+    then_the_primary_publishing_org_should_default_to_their_own
+  end
+
+  def given_a_user_belonging_to_an_org_exists
+    User.first.update(organisation_content_id: "some-org")
```
benthorner: I think you could add this to the factory - this should be true for most users.

spec/features/primary_publishing_org_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+    click_on "Continue"
+
+    choose "Test schema"
+    click_on "Continue"
+  end
+
+  def and_save_a_draft
+    @request = stub_publishing_api_put_content(Document.last.content_id, {})
+    publishing_api_has_linkables([{ "internal_name" => "My cool org", "content_id" => "some-org" }], document_type: "organisation")
+    fill_in "document[title]", with: "A great title"
```
benthorner: As this isn’t necessary, it would be good to leave it out so we reduce the overlap with other tests.

spec/features/primary_publishing_org_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+    click_on "Save"
+  end
+
+  def then_the_primary_publishing_org_should_default_to_their_own
+    expect(Document.last.associations["primary_publishing_organisation"]).to eq(["some-org"])
```
benthorner: We avoid this in other specs by just checking the API call.

spec/features/primary_publishing_org_spec.rb
```diff
@@ -0,0 +1,46 @@
+# frozen_string_literal: true
+
+require "spec_helper"
+
+RSpec.feature "Setting a primary publishing organisation" do
+  scenario "User creates a new document" do
+    given_a_user_belonging_to_an_org_exists
```
benthorner: These steps don’t quite tally with what’s happening in them (create isn’t actually creating). Could we tweak them and make them a bit more succinct?

app/services/publishing_api_payload.rb
davbo: well if someone updates the primary_publishing_organisation we don't want to overwrite the original? this way it'll save the original once the document is first sent off to publishing-api.

benthorner: I think you're right, but I also think we should leave that for now until we properly support a history. Also, if the user accidentally chose the wrong org initially, then it's valid to overwrite it.

spec/features/edit_document_associations_spec.rb
davbo: Unfortunately there isn't a good selector to use for a within - had a look at adding one but there isn't really an obvious attribute to use.

spec/services/publishing_api_payload_spec.rb
```diff
@@ -4,14 +4,73 @@
...
+      payload_hash = {
+        "links" => {
+          "primary_publishing_organisation" => ["my-org-id"],
+          "organisations" => ["other-org-id", "my-org-id"],
+        },
+      }
+      expect(payload).to match a_hash_including(payload_hash)
+    end
+
+    it "only sends an organisation link once if a document is linked by organisations links and primary_publishing_organisation" do
```
benthorner: I think this could be more concise - e.g. 'ensures the organisation links are unique'.

kevindew:
@benthorner after a fun ol' rebase I think this is now good

spec/services/publishing_api_payload_spec.rb
```diff
@@ -4,14 +4,73 @@
...
+        "links" => {
+          "primary_publishing_organisation" => ["my-org-id"],
+          "organisations" => ["my-org-id", "other-org-id"],
+        },
+      }
+      expect(payload).to match a_hash_including(payload_hash)
+    end
+
+    it "ensures the organisation links are unique" do
+      primary_publishing_org = build(:association_schema, type: "single_association", id: "primary_publishing_organisation")
```
benthorner: Tiny thing, but it caused me to read it twice: can we add 'schema' on the end of this?

kevindew: Updated - the variable name got a bit long so I switched it to organisation_schema

spec/services/publishing_api_payload_spec.rb
```diff
@@ -4,14 +4,73 @@
...
```
benthorner: Here as well...

spec/spec_helper.rb
```diff
@@ -33,8 +33,8 @@
   config.include GdsApi::TestHelpers::PublishingApiV2
   config.include GovukSchemas::RSpecMatchers
 
-  config.before(:suite) do
-    User.create!(permissions: ["signin"])
```
tijmenb: @kevindew @benthorner why was this removed? The tests don't work without a user (`undefined method undefined method organisation_content_id for nil:NilClass`.

kevindew: There was some confusion around it and the usage of seeds - as db:setup will create the user.

It changed around to be a create user before each test to ensure there was always a user, but that didn't really work out as it wasn't the user under test.

I'm fine to have this be a before(:suite) find_or_create on User to create one if it doesn't exist, although I think @benthorner is keener to always used seed user.

PR #142, opened 2018-08-16
Use webpacker for Javascript

kevindew:
This switches our javascript processing to use the [rails/webpacker](https://github.com/rails/webpacker) gem

This allows us to:

- Use node_modules from a package.json without having to install them into the app hardcoded
- Make use of modern javascript with features such as imports and es2015 as part of Rails asset-pipeline handling
- Have webpacker functionality like auto reloading and maybe even hot module reloading
- Talking to Alex we already currently have problems extending open source libraries that are written in modern javascript because we need to pull in transpiled ones

This has inflicted some degree of pain:

- There seems to be a lot of configuration files as Rails kind of throws the kitchen sink at this
- If we write es2015 modules here we'll have to transpile them to move them into govuk_publishing_components until that reaches the point of supporting modern js
- Running this on the dev VM is less easy than on a single machine as we lose file change events over NFS meaning webpacker has to poll which makes it a bit slower to update for a file change
- Also to get this working on dev vm we've done the bad thing of installing foreman to Gemfile, but I'm hoping I can sort this out in puppet so we can then remove this gem
- Yarn itself seems to have been a bit of a pain with it's integrity checks, this may need changing.

Hopefully though this is a net positive change, I expect we might have some growing pains with Yarn/Webpacker since it's new - but hopefully this'll be outweighted by bringing our approach to JS up to date and making it simpler for us to use current tooling.


package.json
```diff
@@ -5,15 +5,23 @@
   "author": "Government Digital Service",
   "license": "MIT",
   "scripts": {
-    "lint": "standard app/assets/javascripts/**/*",
-    "lint:fix": "standard --fix app/assets/javascripts/**/*"
+    "lint": "standard app/javascript/{*,**/*}",
```
benthorner: When I do `ls **/*` I get the top-level files. Was there a problem with this here?

kevindew: Yeah there was unfortunately - I think it ties into what exactly does the globbing.

e.g. if I do:

```
vm  content-publisher git:(yarn) echo "withSemiColon();\n" >> app/javascript/restrict.js
```

then with os globbing:

```
vm  content-publisher git:(yarn) ✗ yarn run standard app/javascript/**/*
yarn run v1.9.4
$ /var/govuk/content-publisher/node_modules/.bin/standard app/javascript/components app/javascript/components/contextual-guidance.js app/javascript/components/markdown-editor.js app/javascript/documents app/javascript/documents/edit.js app/javascript/packs app/javascript/packs/application.js app/javascript/restrict.js
standard: Use JavaScript Standard Style (https://standardjs.com)
standard: Run `standard --fix` to automatically fix some problems.
  /var/govuk/content-publisher/app/javascript/restrict.js:6:1: 'withSemiColon' is not defined.
  /var/govuk/content-publisher/app/javascript/restrict.js:6:16: Extra semicolon.
  /var/govuk/content-publisher/app/javascript/restrict.js:7:1: Too many blank lines at the end of file. Max of 0 allowed.
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

but relying on the config file:

```
vm  content-publisher git:(yarn) ✗ yarn run lint
yarn run v1.9.4
$ standard app/javascript/**/*
Done in 1.39s.
```

however with the `{*,*/*}` change:

```
vm  content-publisher git:(yarn) ✗ yarn run lint
yarn run v1.9.4
$ standard app/javascript/{*,**/*}
standard: Use JavaScript Standard Style (https://standardjs.com)
standard: Run `standard --fix` to automatically fix some problems.
  /var/govuk/content-publisher/app/javascript/restrict.js:6:1: 'withSemiColon' is not defined.
  /var/govuk/content-publisher/app/javascript/restrict.js:6:16: Extra semicolon.
  /var/govuk/content-publisher/app/javascript/restrict.js:7:1: Too many blank lines at the end of file. Max of 0 allowed.
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

I'll update the commit though to make it clearer that this was actually a wtf

PR #144, opened 2018-08-16
Model validations with their messages

benthorner:
https://trello.com/c/7XfrouK7/90-allow-content-designers-to-edit-the-application-microcopy-m

After having this discussion for document types and supertypes, we
concluded the copy related to an item should be co-located with the
definition of the item, so that they can both change together. This also
has the advantage of better documenting what each validation means.

This also solves a hierarchy issue with the translations for
validation messages: the convention so far has been to place each
translation in a file based on the page it's being displayed on;
validations buck this trend, as they're not associated with a page.

By making this change, we're treating the definition of domain concepts
(document types, supertypes, validations) consistently.


benthorner:
Replaced with https://github.com/alphagov/content-publisher/pull/145.

PR #145, opened 2018-08-16
Replace hard-coded text with translations in specs

benthorner:
https://trello.com/c/7XfrouK7/90-allow-content-designers-to-edit-the-application-microcopy-m

This helps to decouple the tests from the exact text being shown to the
user, such that the tests are robust against content-only changes.

This comes at the expense of making the tests a little harder to read
due to the calls to the translation framework. I've chosen to keep the
lookups inline rather than creating variables, as this would distract
from the more important state being created within each test method.

### DIFF
--- a/app/services/content_validator.rb
+++ b/app/services/content_validator.rb
@@ -19,13 +19,11 @@ private
 
   def perform_validations_that_apply_to_all_formats(messages)
     if document.title.to_s.size < 10
-      # TODO: extract string into locale file
-      messages << "The title needs to be at least 10 characters long"
+      messages << I18n.t("validations.title", min_length: 10)
     end
 
     if document.summary.to_s.size < 10
-      # TODO: extract string into locale file
-      messages << "The summary needs to be at least 10 characters long"
+      messages << I18n.t("validations.summary", min_length: 10)
     end
   end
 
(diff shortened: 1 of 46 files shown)


kevindew:
Yeah I see what you mean about making the tests hard to read - I think it actually makes it too hard to read so I wonder if we can look at different options.

Some options I'd consider:

1) Should we consider the button parts of the UI either uneditable or not supposed to be changed easily - they're not really parts I'd consider to be the copy of the app and could really cause some surprises editing those
2) I think easier than looking up translations would be to look up selectors

I don't having any issues with it for text that's quite likely to be changed e.g.https://github.com/alphagov/content-publisher/pull/145/files#diff-779ff60c4536045cc038527a4768b176R39 but then I do think we really lose a lot when we consider this file: https://github.com/alphagov/content-publisher/pull/145/files#diff-4c14ca93cc0d887032130d0368396e05L14

benthorner:
@kevindew Thanks for the suggestions. I think keeping all copy out of the tests makes it easier to explain the situation to a copy designer, as well as making it clearer to future developers writing future tests. To that end, I think going with your second suggestion would be a good compromise.

emmabeynon:
I think I'm more of a fan of option 1, as I think someone new to the app reading these tests for the first time might find it quite hard to understand what's going on without having to make constant references to the translations file.  I think the likelihood of button text changing often is quite low, as would be the overhead of changing the tests to reflect such changes.

benthorner:
@kevindew @emmabeynon This should be ready to review again now - choosing Option 1 :-).

benthorner:
Yay 🍾 .

PR #152, opened 2018-08-17
Fix documents not showing due to nil base_path

benthorner:
If the user aborts creating a document on the edit form, the document
will exist but with a nil base_path. Having a migration to ensure the
column has a realistic value will prevent this. It also makes sense to
default the other textual columns to a textual value.


kevindew:
We've got a problem with base_path nil here as we have a unique index on it and thus they'll collide if they are two empty strings but they're fine for nil values.

Seems good for title and summary to default to empty strings though.

benthorner:
Replaced with https://github.com/alphagov/content-publisher/pull/153.

PR #156, opened 2018-08-20
Integrate application with raven-js

kevindew:
Trello: https://trello.com/c/0pzXlC9R/144-track-javascript-errors-s

Raven-js is a JS module that is used to send errors to sentry.io. To
enable us to do this we expose our SENTRY_DSN and SENTRY_CURRENT_ENV
environment variables to HTML meta tags. We then read those attributes
in as part of our Javascript initialisation and install Sentry.

The code for the application is then wrapped in a Raven.context function
as recommended by: https://docs.sentry.io/clients/javascript/#configuring-the-client

Finally we set Raven as a global variable on window so that it can be
available in the console for testing purposes.

I've configured sentry to just accept errors from *.publishing.service.gov.uk hosts, ~I'm going to expand this to include heroku too.~ - actually I can't do that since we don't have a sentry dsn set up for heroku currently.

### DIFF
--- a/app/javascript/packs/application.js
+++ b/app/javascript/packs/application.js
@@ -11,6 +11,17 @@ import restrict from '../restrict'
 import editDocumentForm from '../documents/edit'
 import MarkdownEditor from '../components/markdown-editor'
 import ContextualGuidance from '../components/contextual-guidance'
+import Raven from 'raven-js'
+
+var $sentryDsn = document.querySelector('meta[name=sentry-dsn]')
+var $sentryCurrentEnv = document.querySelector('meta[name=sentry-current-env]')
+
+if ($sentryDsn && $sentryCurrentEnv) {
+  // set as global variable for console usage
+  window.Raven = Raven.config($sentryDsn.getAttribute('content'), {
+    environment: $sentryCurrentEnv.getAttribute('content')
+  }).install()
+}
 
 restrict('edit-document-form', editDocumentForm)
 restrict('markdown-editor', ($el) => new MarkdownEditor($el).init())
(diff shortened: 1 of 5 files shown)


app/javascript/components/contextual-guidance.js
```diff
@@ -1,6 +1,7 @@
 function ContextualGuidance ($module) {
   if (!$module) $module = document
   this.$module = $module
+  throw new Error('Testing raven')
```
benthorner: I'm guessing this is temporary :-).

kevindew: Ha yup. I thought I should test it properly before shouting about this PR.

app/javascript/packs/application.js
```diff
@@ -11,10 +11,23 @@ import restrict from '../restrict'
...
-var guidance = new ContextualGuidance()
-guidance.init(document)
+if ($sentryDsn && $sentryCurrentEnv) {
+  // set as global variable for console usage
+  window.Raven = Raven.config($sentryDsn.getAttribute('content'), {
+    environment: $sentryCurrentEnv.getAttribute('content')
+  }).install()
+}
+
+Raven.context(() => {
```
benthorner: So the docs say this is intended to wrap our 'app start' to capture additional errors, but we don't have an app here. Could we leave it out for now, to avoid confusion?

kevindew:
@benthorner Raven.context is used to wrap anything in a try / catch. As, like you, I'm not sure what advantage it gives us I've removed it for now and we can look at adding it if we find our JS tracking seems to be lacking. It's removed in 88585f0

PR #157, opened 2018-08-20
Update summary

alex-ju:
This PR updates the summary page as per design and content
   - Adds possible actions in a side box
   - Adds document metadata component
   - Updates main heading to match gov.uk style (for summary page: document format and document title)

Contains CSS refactor:
   - Import only the needed bits in each SCSS file
   - Structure CSS following ITCSS

Updates tests to use updated content and states.

### Before
government estate strategy 2018 - gov uk 1

### After
government estate strategy 2018 - gov uk

### DIFF
--- a/app/assets/stylesheets/components/_contextual-guidance.scss
+++ b/app/assets/stylesheets/components/_contextual-guidance.scss
@@ -1,3 +1,6 @@
+@import "../../../node_modules/govuk-frontend/settings/all";
+@import "../../../node_modules/govuk-frontend/helpers/all";
+
 .app-c-contextual-guidance-wrapper {
   @include govuk-font(19);
 
--- a/app/views/documents/show/_actions.html.erb
+++ b/app/views/documents/show/_actions.html.erb
@@ -4,10 +4,10 @@ view_live_button = { text: t("documents.show.actions.view_live"), href: Document
 share_draft_button = { text: t("documents.show.actions.share_draft"), href: DocumentUrl.new(@document).secret_preview_url }
 preview_button = { text: t("documents.show.actions.view_draft"), href: DocumentUrl.new(@document).preview_url }
 create_draft_button = { text: t("documents.show.actions.create_draft"), href: edit_document_path(@document) }
-submit_2i_button = { text: t("documents.show.actions.submit_2i"), href: "submit_2i" }
-delete_draft_button = { text: t("documents.show.actions.delete_draft"), href: "delete" }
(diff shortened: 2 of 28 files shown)


app/views/documents/show/_actions.html.erb
```diff
@@ -1,24 +1,71 @@
...
 
-<div class='govuk-tag' title="<%= t("publication_states.#{@document.publication_state}.description") %>">
-  <%= t("publication_states.#{@document.publication_state}.name") %>
+case @document.publication_state
+when "sent_to_draft" || "sending_to_live" || "error_sending_to_live" # Draft
+  primary_button = submit_2i_button
+  secondary_button = preview_button
+  ternary_button = publish_button
+  destructive_button = delete_draft_button
+when "sent_to_review" # Ready for review
```
kevindew: We're not anticipating sent_to_review to be part of publication state (see https://trello.com/c/NVjuv8Fs/170-the-way-we-encode-states-needs-to-be-more-flexible-so-we-can-add-more) so this case statement won't ever execute this block.

It's probably worth us having a chat to try get design/dev/product in similar mindsets regarding state.

app/assets/stylesheets/components/_contextual-guidance.scss
```diff
@@ -1,3 +1,6 @@
+@import "../../../node_modules/govuk-frontend/settings/all";
+@import "../../../node_modules/govuk-frontend/helpers/all";
+
```
kevindew: I wonder if we can do anything govuk_publishing_components to make these be in a more logical place - they caught me out for a while the other day when I realised it was from the gems node_modules directory rather than ours, and I think they only work off a bit of a hack working up from the gems app/assets directory.

alex-ju: Yes, they come from the gem and is not clear. I'll think about it

app/views/documents/show/_actions.html.erb
alex-ju: Would be good to discuss them. This is based on the design states Alex T. prepared for design. I can remove this case for now.

alex-ju:
@kevindew: the component preview doesn't properly load component assets. Wasn't a trivial fix so I didn't got the chance to look into it, but I'm aware and I'll address it this week.

alex-ju:
@kevindew updated: commented the `sent_to_review` case and added screenshots until the component guide is working.

PR #158, opened 2018-08-21
Add ActiveStorage infrastructure

benthorner:
https://trello.com/c/63nq4r30/138-setup-an-s3-bucket-to-store-assets

This enables image attachments for a Document via ActiveStorage, using a
local 'Disk' store for test/dev and an AWS S3 bucket for production.


Gemfile
```diff
@@ -6,6 +6,7 @@ source "https://rubygems.org"
 
 gem "rails", "~> 5.2"
 
+gem "aws-sdk-s3"
```
kevindew: Do you want to set require: false on this as suggested in https://edgeguides.rubyonrails.org/active_storage_overview.html#amazon-s3-service ?

config/storage.yml
```diff
@@ -4,31 +4,11 @@ test:
...
 
-# mirror:
-#   service: Mirror
-#   primary: local
-#   mirrors: [ amazon, google, microsoft ]
+amazon:
+  service: S3
+  access_key_id: <%= ENV["AWS_ACCESS_KEY_ID"] %>
+  secret_access_key: <%= ENV["AWS_SECRET_ACCESS_KEY"] %>
+  region: us-east-1
```
kevindew: We have an AWS_REGION env var for this

Gemfile
kevindew: Also good to pin it to version 1 - even though Rails is a bit less cautious

benthorner: I don't think it's necessary - it's always loaded anyway. Good point with the versioning - apparently it's not in my internal linter yet.

config/storage.yml
benthorner: Oops - good spot :-)

PR #159, opened 2018-08-21
Update edit document page

alex-ju:
This PR:
- Disables the autocomplete on the edit form
- Updates edit page content
- Replace toggle link with tabs for markdown preview
  - use `webcomponents-bundle` instead of `webcomponents-loader`, as the loader doesn't contain the polyfills (Firefox and IE fail now because of the lack of this polyfill)
- Replace Markdown toolbar elements with the extended version (this was causing the Heading injectors to fail, because they're not part of the original Github Toolbar)
  - Github doesn't doesn't publish the source code for the markdown toolbar element in its package so we need to publish our own fork to npm. Until then we'll bring the transpiled code in the project.

### Before
ethical standards for public service 1

### After
ethical standards for public service

### DIFF
--- a/app/assets/stylesheets/components/_markdown-editor.scss
+++ b/app/assets/stylesheets/components/_markdown-editor.scss
@@ -3,35 +3,53 @@
 @import "../../../node_modules/govuk-frontend/tools/all";
 @import "../../../node_modules/govuk-frontend/core/links";
 
-.app-c-markdown-editor__preview-toggle {
+.app-c-markdown-editor__head {
   @include govuk-font(19);
 
   display: none;
+  position: relative;
+  z-index: 20;
   margin-bottom: -$govuk-border-width-form-element;
-  padding: govuk-spacing(1);
   border: $govuk-border-width-form-element solid $govuk-input-border-colour;
-  border-bottom: $govuk-border-width-form-element solid $govuk-border-colour;
+  border-bottom: 0;
   background-color: govuk-colour("grey-3");
+}
+
+.app-c-markdown-editor__preview-toggle {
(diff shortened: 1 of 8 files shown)


app/assets/stylesheets/components/_markdown-editor.scss
```diff
@@ -1,30 +1,50 @@
-.app-c-markdown-editor__preview-toggle {
+.app-c-markdown-editor__head {
   @include govuk-font($size: 19);
 
   display: none;
+  position: relative;
+  z-index: 20;
```
benthorner: This causes the focus outline to be hidden and looks a bit weird - is that intentional?

kevindew:
Looks good - I noticed we lose focus indicator on the top of the markdown textarea

screen shot 2018-08-21 at 12 57 45

app/assets/stylesheets/components/_markdown-editor.scss
benthorner: Also, the outline for the toggle buttons is slightly misaligned with the button borders.

alex-ju: It is actually intentional, to position the toolbar over the textarea. I'll try to think of ways of enhancing it.

alex-ju:
@kevindew @benthorner 
Changed the approach now the focus contains the toolbar while preserving tab-index order for accessibility.
screen shot 2018-08-21 at 14 37 45

benthorner:
@alex-ju That looks good, although the button outlines are still weird on my machine.

screen shot 2018-08-21 at 14 47 54

screen shot 2018-08-21 at 14 47 50

alex-ju:
@benthorner just pushed an update to fix that

PR #160, opened 2018-08-21
Fix tests failing locally due to missing user

benthorner:
The tests assume a user exists in order to run, but we never explicitly
create one - this commit fixes that using a new factory.


benthorner:
Closing as we can use seeds instead...

PR #166, opened 2018-08-21
Add character count on edit document page

alex-ju:
This PR adds a character count to the Title (65 chars) and Summary (160 chars) fields on the edit document page. The code for this PR is based on the upcoming [Character count component](https://github.com/alphagov/govuk-frontend/pull/959) in govuk-frontend.

[Trello card](https://trello.com/c/28Ol82Ky)

screen shot 2018-08-21 at 16 21 52

### DIFF
--- a/app/javascript/components/character-count.js
+++ b/app/javascript/components/character-count.js
@@ -1,0 +1,187 @@
+function CharacterCount ($module) {
+  this.$module = $module
+}
+
+CharacterCount.prototype.defaults = {
+  characterCountAttribute: 'data-maxlength',
+  wordCountAttribute: 'data-maxwords'
+}
+
+// Initialize component
+CharacterCount.prototype.init = function (options) {
+  // Check for module
+  var $module = this.$module
+  if (!$module) {
+    return
+  }
+
+  // Read options set using dataset ('data-' values) and merge them with options
+  var dataset = this.getDataset($module)
+  this.options = this.merge(options || {}, dataset)
+
+  // Determine the limit attribute (characters or words)
+  var countAttribute = (this.options.maxwords) ? this.defaults.wordCountAttribute : this.defaults.characterCountAttribute
+
+  // Set the element limit
+  if ($module.getAttribute) {
+    this.maxLength = $module.getAttribute(countAttribute)
+  }
+
+  // Generate and reference message
+  var boundCreateCountMessage = this.createCountMessage.bind(this)
+  this.countMessage = boundCreateCountMessage()
+
+  // If there's a maximum length defined and the count message was successfully applied
+  if (this.maxLength && this.countMessage) {
+    // Replace maxlength attribute with data-maxlength to avoid hard limit
+    $module.setAttribute('maxlength', '')
+    $module.setAttribute('data-maxlength', $module.maxLength)
+
+    // Bind event changes to the textarea
+    var boundChangeEvents = this.bindChangeEvents.bind(this)
+    boundChangeEvents()
+
+    // Update count message
+    var boundUpdateCountMessage = this.updateCountMessage.bind(this)
+    boundUpdateCountMessage()
+  }
+}
+
+// Fills in default values
+CharacterCount.prototype.merge = function (obj) {
+  for (var i = 1; i < arguments.length; i++) {
+    var def = arguments[i]
+    for (var n in def) {
+      if (obj[n] === undefined) obj[n] = def[n]
+    }
+  }
+  return obj
+}
+
+CharacterCount.prototype.getDataset = function (element) {
+  var dataset = {}
+  var attributes = element.attributes
+  if (attributes) {
+    for (var i = 0; i < attributes.length; i++) {
+      var attribute = attributes[i]
+      var match = attribute.name.match(/^data-(.+)/)
+      if (match) {
+        dataset[match[1]] = attribute.value
+      }
+    }
+  }
+  return dataset
+}
+
+// Counts characters or words in text
+CharacterCount.prototype.count = function (text, options) {
+  var length
+  if (options.maxwords) {
+    var tokens = text.match(/\S+/g) || [] // Matches consecutive non-whitespace chars
+    length = tokens.length
+  } else {
+    length = text.length
+  }
+  return length
+}
+
+// Generate count message and bind it to the input
+// returns reference to the generated element
+CharacterCount.prototype.createCountMessage = function () {
+  var countElement = this.$module
+  var elementId = countElement.id
+  // Check for existing info count message
+  var countMessage = document.getElementById(elementId + '-info')
+  // If there is no existing info count message we add one right after the field
+  if (elementId && !countMessage) {
+    countElement.insertAdjacentHTML('afterend', '<span id="' + elementId + '-info" class="govuk-hint govuk-character-count__message" aria-live="polite"></span>')
+    this.describedBy = countElement.getAttribute('aria-describedby')
+    this.describedByInfo = this.describedBy + ' ' + elementId + '-info'
+    countElement.setAttribute('aria-describedby', this.describedByInfo)
+    countMessage = document.getElementById(elementId + '-info')
+  }
+  return countMessage
+}
+
+// Bind input propertychange to the elements and update based on the change
+CharacterCount.prototype.bindChangeEvents = function () {
+  var $module = this.$module
+  $module.addEventListener('keyup', this.updateCountMessage.bind(this))
+
+  // Bind focus/blur events to start/stop polling
+  $module.addEventListener('focus', this.handleFocus.bind(this))
+  $module.addEventListener('blur', this.handleBlur.bind(this))
+}
+
+// Speech recognition software such as Dragon NaturallySpeaking will modify the
+// fields by directly changing its `value`. These changes don't trigger events
+// in JavaScript, so we need to poll to handle when and if they occur.
+CharacterCount.prototype.checkIfValueChanged = function () {
+  if (!this.oldValue) this.oldValue = ''
+  if (this.$module.value !== this.oldValue) {
+    this.oldValue = this.$module.value
+    var boundUpdateCountMessage = this.updateCountMessage.bind(this)
+    boundUpdateCountMessage()
+  }
+}
+
+// Update message box
+CharacterCount.prototype.updateCountMessage = function () {
+  var countElement = this.$module
+  var options = this.options
+  var countMessage = this.countMessage
+
+  // Determine the remaining number of characters/words
+  var currentLength = this.count(countElement.value, options)
+  var maxLength = this.maxLength
+  var remainingNumber = maxLength - currentLength
+
+  // Set threshold if presented in options
+  var threshold = 0
+  if (options.threshold) {
+    threshold = options.threshold
+  }
+  var thresholdValue = maxLength * threshold / 100
+  if (thresholdValue > currentLength) {
+    countMessage.classList.add('govuk-character-count__message--disabled')
+  } else {
+    countMessage.classList.remove('govuk-character-count__message--disabled')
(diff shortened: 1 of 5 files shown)


alex-ju:
@kevindew yes, I rely much on the Github PR message instead of commit message. I should mirror these because different people have different ways of reviewing. Using a branch from npm is a good idea, unfortunately content-publisher doesn't yet consume frontend directly, but via govuk_publishing_components gem, but after updating it to consume govuk-frontend directly I'll take the branch approach.

benthorner:
@kevindew Should this have a small spin-off story for us to update the validations?

app/views/documents/edit.html.erb
```diff
@@ -55,6 +57,8 @@
         value: @document.summary,
         rows: 4,
         data: {
+          "module": "character-count",
```
benthorner: It looks like this kind of technique doesn’t extend if we wanted to add more modules to this element - would it be better to attach it in a similar way as contextual guidance?

app/javascript/components/character-count.js
```diff
@@ -0,0 +1,187 @@
...
+
+  // Read options set using dataset ('data-' values) and merge them with options
+  var dataset = this.getDataset($module)
+  this.options = this.merge(options || {}, dataset)
+
+  // Determine the limit attribute (characters or words)
+  var countAttribute = (this.options.maxwords) ? this.defaults.wordCountAttribute : this.defaults.characterCountAttribute
+
+  // Set the element limit
+  if ($module.getAttribute) {
```
benthorner: Should we just quit early here instead of having another if later on?

app/javascript/components/character-count.js
```diff
@@ -0,0 +1,187 @@
...
+  if (!$module) {
+    return
+  }
+
+  // Read options set using dataset ('data-' values) and merge them with options
+  var dataset = this.getDataset($module)
+  this.options = this.merge(options || {}, dataset)
+
+  // Determine the limit attribute (characters or words)
+  var countAttribute = (this.options.maxwords) ? this.defaults.wordCountAttribute : this.defaults.characterCountAttribute
```
benthorner: It looks like we’re doing this on characters - why are we supporting words?

app/javascript/components/character-count.js
```diff
@@ -0,0 +1,187 @@
...
+    boundChangeEvents()
+
+    // Update count message
+    var boundUpdateCountMessage = this.updateCountMessage.bind(this)
+    boundUpdateCountMessage()
+  }
+}
+
+// Fills in default values
+CharacterCount.prototype.merge = function (obj) {
```
benthorner: Would it be easier to just manually merge attributes? I think there are only two of them ;-)

app/javascript/components/character-count.js
```diff
@@ -0,0 +1,187 @@
...
+  for (var i = 1; i < arguments.length; i++) {
+    var def = arguments[i]
+    for (var n in def) {
+      if (obj[n] === undefined) obj[n] = def[n]
+    }
+  }
+  return obj
+}
+
+CharacterCount.prototype.getDataset = function (element) {
```
benthorner: Again, would it be easier to just fish these out manually if there aren’t many?

benthorner: Or perhaps extract this to somewhere more general as it looks like it could be generally useful :-)

app/javascript/components/character-count.js
```diff
@@ -0,0 +1,187 @@
...
+  }
+  return dataset
+}
+
+// Counts characters or words in text
+CharacterCount.prototype.count = function (text, options) {
+  var length
+  if (options.maxwords) {
+    var tokens = text.match(/\S+/g) || [] // Matches consecutive non-whitespace chars
+    length = tokens.length
```
benthorner: You can just return here rather than have the variable - would be a bit easier to follow then :-)

app/javascript/components/character-count.js
```diff
@@ -0,0 +1,187 @@
...
+  } else {
+    length = text.length
+  }
+  return length
+}
+
+// Generate count message and bind it to the input
+// returns reference to the generated element
+CharacterCount.prototype.createCountMessage = function () {
```
benthorner: Again, wouldn’t it be easier to just the character count as actual HTML like contextual guidance?

app/javascript/components/character-count.js
```diff
@@ -0,0 +1,187 @@
...
+}
+
+// Speech recognition software such as Dragon NaturallySpeaking will modify the
+// fields by directly changing its `value`. These changes don't trigger events
+// in JavaScript, so we need to poll to handle when and if they occur.
+CharacterCount.prototype.checkIfValueChanged = function () {
+  if (!this.oldValue) this.oldValue = ''
+  if (this.$module.value !== this.oldValue) {
+    this.oldValue = this.$module.value
+    var boundUpdateCountMessage = this.updateCountMessage.bind(this)
```
benthorner: Why do we need a variable here?

app/javascript/components/character-count.js
```diff
@@ -0,0 +1,187 @@
...
+  var options = this.options
+  var countMessage = this.countMessage
+
+  // Determine the remaining number of characters/words
+  var currentLength = this.count(countElement.value, options)
+  var maxLength = this.maxLength
+  var remainingNumber = maxLength - currentLength
+
+  // Set threshold if presented in options
+  var threshold = 0
```
benthorner: You could make this a ternary - it seems to be the pattern used in the code above.

app/javascript/components/character-count.js
```diff
@@ -0,0 +1,187 @@
...
+  if (options.threshold) {
+    threshold = options.threshold
+  }
+  var thresholdValue = maxLength * threshold / 100
```
benthorner: Er, where’s the magic 100 coming from?

app/javascript/components/character-count.js
```diff
@@ -0,0 +1,187 @@
...
+  // Determine the limit attribute (characters or words)
+  var countAttribute = (this.options.maxwords) ? this.defaults.wordCountAttribute : this.defaults.characterCountAttribute
+
+  // Set the element limit
+  if ($module.getAttribute) {
+    this.maxLength = $module.getAttribute(countAttribute)
+  }
+
+  // Generate and reference message
+  var boundCreateCountMessage = this.createCountMessage.bind(this)
```
benthorner: How come we need a variable here?

app/javascript/components/character-count.js
```diff
@@ -0,0 +1,187 @@
...
+  this.countMessage = boundCreateCountMessage()
+
+  // If there's a maximum length defined and the count message was successfully applied
+  if (this.maxLength && this.countMessage) {
+    // Replace maxlength attribute with data-maxlength to avoid hard limit
+    $module.setAttribute('maxlength', '')
+    $module.setAttribute('data-maxlength', $module.maxLength)
+
+    // Bind event changes to the textarea
+    var boundChangeEvents = this.bindChangeEvents.bind(this)
```
benthorner: How come we need a variable here?

app/views/documents/edit.html.erb
alex-ju: Yes, that is true, this method allows only one module per container, but generally I would avoid initialising multiple scripts on the same component as they may hijack each other's events. One script per component is consistent with how JavaScript is attached to all the components that are coming from `govuk-frontend`. If there are scripts acting at a document level, as it was the case of the contextual guidance, it makes sense for that script to use a `data-[component-name]` attribute selector or a `.js-[component-name]` class selector as we don't want to overlap with a component model and we want to be able to identify which script manipulates/affects the component. This is one of those things that I should document and explain at the project level.

app/javascript/components/character-count.js
alex-ju: Yes, we can quit early here. Maybe with a console message to mention an attribute is missing. Will update.

app/javascript/components/character-count.js
alex-ju: It was part of the original script which allows both characters and words. I pulled this in 'as is' because this is how it will come from `govuk-frontend`. Can remove the maxwords for now.

app/javascript/components/character-count.js
alex-ju: `threshold` is a percentage value

app/javascript/components/character-count.js
alex-ju: We need to store and execute the bounded function so no matter how and when it is called (on which component's instance), is called with a particular this (in our case our component's instance).

benthorner: Hmm...so is `this.updateCountMessage.bind(this)()` not equivalent?

alex-ju: nope

alex-ju:
> So this seems like quite a lot of code to generate single line of text.

I wouldn't judge or measure the code by how much the users sees 😉 

It may look a bit complex because it deals with both scenarios (character and word count). This is already a cleaned-up version without the highlight functionality which in my opinion contains all the complexity.

Updated:
- Refactored some parts (i.e. return early when possible, use ternary operator consistently)
- Removed word count option as we don't need it right now

Regarding `CharacterCount.prototype.merge` and  `CharacterCount.prototype.getDataset ` they are generic utility functions, but I wouldn't extract them until they are needed in other components.

Tried to reply the other questions inline, please let me know if I missed any.

app/javascript/components/character-count.js
```diff
@@ -0,0 +1,166 @@
...
+  if (!$module) {
+    return
+  }
+
+  // Read options set using dataset ('data-' values) and merge them with options
+  var dataset = this.getDataset($module)
+  this.options = this.merge(options || {}, dataset)
+
+  // Read the limit attribute
+  var countAttribute = 'data-maxlength'
```
benthorner: Do we need this variable now?

app/javascript/components/character-count.js
```diff
@@ -0,0 +1,166 @@
...
+  } else {
+    return
+  }
+
+  // Generate and reference message
+  var boundCreateCountMessage = this.createCountMessage.bind(this)
+  this.countMessage = boundCreateCountMessage()
+
+  // If there's a maximum length defined and the count message was successfully applied
+  if (this.maxLength && this.countMessage) {
```
benthorner: Again, can we return early here instead of having the block?

app/javascript/components/character-count.js
```diff
@@ -0,0 +1,166 @@
...
+  var countMessage = this.countMessage
+
+  // Determine the remaining number of characters/words
+  var currentLength = countElement.value.length
+  var maxLength = this.maxLength
+  var remainingNumber = maxLength - currentLength
+
+  // Set threshold if presented in options
+  var thresholdPercent = options.threshold ? options.threshold : 0
+  var thresholdValue = maxLength * thresholdPercent / 100
```
benthorner: I still don't understand what this threshold percent thing is for - how does it relate to the UI?

app/javascript/components/character-count.js
```diff
@@ -0,0 +1,166 @@
...
+    countMessage.classList.remove('govuk-hint')
+    countMessage.classList.add('govuk-error-message')
+  } else {
+    countElement.classList.remove('govuk-textarea--error')
+    countMessage.classList.remove('govuk-error-message')
+    countMessage.classList.add('govuk-hint')
+  }
+
+  // Update message
+  var charVerb = 'remaining'
```
benthorner: Do we need these variables?

alex-ju:
Closing this as I consider unproductive to extract and branch code from a component that will replace this one. Thanks @benthorner for reviewing! For now we can use a hint text to state the maximum limit and rely on server side for validation. Will bring the character count component from `govuk-frontend` when is ready.

PR #169, opened 2018-08-22
Enforce double quotes and trailing commas

benthorner:
This is brings us closer to the origin GDS style guide and removes the
ambiguity of which quote and list syntax to use, which has been a
persistent pain. Note that the quoting isn't enforced where it would be
semantically invalid (e.g. '"hello"').

https://github.com/alphagov/styleguides/blob/7a9045bf653e6f12d614217b30d70b7798ece5bd/ruby.md

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -17,3 +17,13 @@ Style/FrozenStringLiteralComment:
   Enabled: true
   Exclude:
     - config/**/*
+
+Style/StringLiterals:
+  Enabled: true
+
+Style/TrailingCommaInLiteral:
+  Enabled: true
+
+Style/TrailingCommaInArguments:
+  Enabled: true
+  EnforcedStyleForMultiline: comma
--- a/Rakefile
+++ b/Rakefile
@@ -3,7 +3,7 @@
 # Add your own tasks in files placed in lib/tasks ending in .rake,
 # for example lib/tasks/capistrano.rake, and they will automatically be available to Rake.
 
-require_relative 'config/application'
+require_relative "config/application"
 
 Rails.application.load_tasks
 task default: %i(spec lint brakeman)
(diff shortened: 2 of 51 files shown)


benthorner:
@kevindew Happy to drop this if you think it could be contentious or done in a wider forum :-).

kevindew:
@benthorner Yeah, I think it's worth trialling ones here that we intend to see if they can be part of the govuk linting one. I think both of these seem like things that can be sorted with auto-correct so don't seem like they'd be painful to embrace, is that right?

Quotes seems a good candidate - and by god would I be happy that was linted - and can see from the [commit](https://github.com/alphagov/govuk-lint/commit/67e3ea748c55015818e310e6cc8d9ac6a9b148c8) that the reason that was disabled was for dealing with diffs and we seem to be moving away from that.

Trailing comma wise - that seems to have disabled years ago due to [buggy implementation](https://github.com/boffbowsh/styleguides/commit/2312a7e7519c89cea256deb085a1340b69058152) which I'd presume is fixed now since it seems separated for arguments and lists. For consistency would it make sense enabling all 3 of these https://github.com/rubocop-hq/rubocop/blob/master/config/default.yml#L1449-L1480 ? I also wonder if that might be somewhat contentious but if they're consistent with the old styleguide that's a better start than just coming up with a personal opinion.

Seems good 👍

kevindew:
Oh, only loose related but I think we can probably kill all the require "spec_helper" lines as we have that in rspecrc

benthorner:
@kevindew Yeah, it's just a '-a' flag and it works fine. And definitely these would be good to export more widely if we find we're happy to work with them.

Only one of the additional checks you mentioned still exists, while the other two are now combined into the one I already added. And good point with rspec - I've removed them as they changed anyway.

benthorner:
@emmabeynon How do you feel about this?

emmabeynon:
Whoops I missed your tag until just now - seems a reasonable trial to me 👍

benthorner:
@kevindew Done :-).

PR #170, opened 2018-08-23
[DO NOT MERGE] Basic image flow spike

kevindew:
Trello: https://github.com/alphagov/content-publisher/pull/170

This PR serves as a demonstration of work done so far in regards to image handling. I don't anticipate this will be merged but instead used as a basis for the next part of image work

## What this does

It provides a route "/documents/{id}/images" which has a file upload. When a file is uploaded it is validated that it is: 1) a jpg/gif/png 2) less than 20 MB 3) greater in dimensions than 960x640. We then process the file to strip exif data and correct orientation, the file is then stored on S3. We then show the form again which includes the newly uploaded file which has been cropped and resized automatically to 960 x 640.

## Trying this out

This branch is currently on integration and an example can be found at http://content-publisher.dev.gov.uk/documents/1/images. At later times this may need re-deploying.

## Why it's done like it's done

There are some aspects of this that deviate from the [suggested use of ActiveStorage](https://edgeguides.rubyonrails.org/active_storage_overview.html)

### Detecting mime type and calculating dimensions our self

ActiveStorage does not provide a means to validate files and the process used to calculate dimensions involves re-downloading the file from S3.

It seemed more efficient / logical to validate the uploaded file directly to check it was as anticipated and then upload it to active storage than to upload and then go through this process. It seems that the next release of ActiveStorage is geared more towards this behaviour with the introduction of [skipping some of it's steps](https://github.com/rails/rails/pull/32833)

### Initialising ActiveStorage::Blob directly

This is done mostly to experiment with the ease of this. I'm anticipating we won't be using the ActiveStorage::Attachment model as a many-to-many relationship would be tricky to maintain with our approach to versioning. The Attachment doesn't actually have a whole lot of stuff all the meat is in the Blob model.

## What next?

The next step for this is to try turn this into something that would be useful for the app. Having chatted with @alex-ju yesterday the proposal is to create two endpoints that could be used by JavaScript: 

- One to uploads an image and stores it on a document. It'd return data such as errors, path, suggested crop dimensions
- The other to update those crop dimensions for the image

This will involve working out how to store the image on a document - for which I am anticipating a JSON blob on document in the meantime, though maybe/could should evolve to be a model in it's own right particularly if the data stored gets hairy. Opinions?

The bits here are all a bit hacky and could use some refactoring and testing.


app/controllers/document_images_controller.rb
```diff
@@ -0,0 +1,168 @@
...
+    if !upload.valid?
+      redirect_to document_images_path(document),
+                   flash: { notice: "Image upload failed", errors: upload.errors }
+    else
+      blob = ActiveStorage::Blob.build_after_upload(
+        io: upload.file,
+        content_type: upload.mime_type,
+        filename: upload.filename,
+      )
+      blob.metadata[:width] = upload.dimensions[:width]
```
benthorner: Can we not give it a metadata hash when we build it?

app/controllers/document_images_controller.rb
```diff
@@ -0,0 +1,168 @@
...
+    def degrees_to_rotate(image)
+      case image["%[orientation]"]
+      when "BottomRight" then 180
+      when "RightTop" then 90
+      when "LeftBottom" then -90
+      else 0
+      end
+    end
+
+    def suggested_crop
```
benthorner: Would be nice to comment rename to mention it’s a crop around the centre.

app/controllers/document_images_controller.rb
```diff
@@ -0,0 +1,168 @@
...
+
+private
+
+  def image_params
+    params.fetch(:document, {}).permit(:image)
+  end
+
+  class UploadedImageProcessorService
+    InvalidImage = Struct.new(:valid?, :errors)
+    ValidImage = Struct.new(
```
benthorner: It probably shouldn’t be possible for ValidImage.valid? to be false ;-)

app/controllers/document_images_controller.rb
```diff
@@ -0,0 +1,168 @@
...
+
+      errors
+    end
+
+    def normalise_image(image)
+      rotate = degrees_to_rotate(image)
+      image.combine_options do |file|
+        file.rotate(rotate) unless rotate.zero?
+
+        # remove exif data
```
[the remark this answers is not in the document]
kevindew: Exif data can store information like when and where a photo was taken, these things feel like potentially sensitive information that could accidentally reveal things that are secret or embarrassing.

app/controllers/document_images_controller.rb
kevindew: Nah, it doesn't accept one as an argument. I don't think we should use it for our metadata really since it feels like it could be semi-private to Blob, next round of this has us storing this information in our own model.

app/controllers/document_images_controller.rb
kevindew: Sure. I'm extracting this out to a class of it's own and will take that on board.

app/controllers/document_images_controller.rb
kevindew: good point!

kevindew:
Thanks for having a look over @benthorner. You're right that soft delete seems something really important - I'm not clear though how that might fit into UI and thus what sort of endpoint it might need so that's unlikely to be part of next iteration. Shouldn't be complex to implement though.

PR #171, opened 2018-08-23
Additional content changes

emmabeynon:
For https://trello.com/c/Gm1BOO1O/151-review-pages-with-updated-content-in-content-publisher-s

#### Before - display the document title
edit-title-before

#### After - display "Create" + document_type when the document is new
new-create-press-release-after

#### After - display "Edit" + document_type when editing an existing document
screen shot 2018-08-23 at 16 18 23

#### Before
tags-before

#### After
_Note:_ It's unclear where Worldwide Organisations should go. I've tagged Ben H in the Google doc to check; in the meantime it seems logical to place it near to Organisations.

tags-after

### DIFF
--- a/app/formats/document_types.yml
+++ b/app/formats/document_types.yml
@@ -23,14 +23,10 @@
       validations:
         min_length: 10
   associations:
-    - id: topical_events
-      label: Topical events
-      type: multi_association
-      document_type: topical_event
-    - id: world_locations
-      label: World locations
-      type: multi_association
-      document_type: world_location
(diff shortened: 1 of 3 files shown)


emmabeynon:
@soniaturcotte It definitely looks better with the document type downcased, and actually it looks like it should be that way [here](https://docs.google.com/document/d/1RHvAf1MEjwOuR4POpcLAkhBQWLejYhe4WOyVCAdCb0I/edit#heading=h.nw6klflsjtt6) so now amended 👍

PR #172, opened 2018-08-24
Add Google Tag Manager script

alex-ju:
This PR adds the Google Tag Manager component and it needs [Add Google Tag Manager env vars for Content Publisher](https://github.com/alphagov/govuk-puppet/pull/8041) to get merged in order to render the component. Raised this PR to have it ready.

[Trello card](https://trello.com/c/wh7K3ul2)

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -4,6 +4,14 @@
     <meta name="sentry-dsn" content="<%= ENV["SENTRY_DSN"] %>">
     <meta name="sentry-current-env" content="<%= ENV["SENTRY_CURRENT_ENV"] %>">
   <% end %>
+
+  <% if ENV["GOOGLE_TAG_MANAGER_ID"] %>
+    <%= render "govuk_publishing_components/components/google_tag_manager_script", {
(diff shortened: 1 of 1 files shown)


app/views/layouts/application.html.erb
```diff
@@ -4,6 +4,14 @@
     <meta name="sentry-dsn" content="<%= ENV["SENTRY_DSN"] %>">
     <meta name="sentry-current-env" content="<%= ENV["SENTRY_CURRENT_ENV"] %>">
   <% end %>
+
+  <% if ENV["GOOGLE_TAG_MANAGER_ID"] && ENV["GOOGLE_TAG_MANAGER_AUTH"] && ENV["GOOGLE_TAG_MANAGER_PREVIEW"] %>
```
kevindew: Are we fine excluding a scenario where we've only got tag_manager_id ? I know it's not that realistic since Andy plans to use all 3 but I'd have thought we'd only need to check that env var and be happy if some of the others are nil.

alex-ju: You're right @kevindew, we only need to check for `GOOGLE_TAG_MANAGER_ID`. Updated.

PR #174, opened 2018-08-24
Add components for flash messages

alex-ju:
This PR:
 - updates the component names in yml and erb to follow the ruby convention (`_` as separator)
 - adds error-alert component (as we don't have one in the gem library except for the error summary)
 - uses success-alert (from gem) and error-alert for flash messages

This doesn't address the error message summary, only simple success and error messages.

[Trello card](https://trello.com/c/aHbkreSI)

success-alert
error-alert


alex-ju:
@benthorner: Thanks!. I realised the file names were not following the convention when I ran into an issue with `error-alert` that's why I did it with this PR. Will keep them separated in future.

PR #179, opened 2018-08-28
Set app time zone to London

emmabeynon:
We store dates in the database in UTC, which means that when we're displaying
dates and times in the views they're an hour out of date during daylight saving
time. This sets the app's time zone to 'London' so that ActiveRecord
automatically converts dates and times to the UK's current time zone, whether
it's GMT or BST.


benthorner:
Would this be better...

https://stackoverflow.com/questions/22581235/rails-4-want-to-have-rails-use-my-local-time-zone-for-created-at-updated-at

kevindew:
On our servers we use UTC as the timezone so we can't use :local

PR #180, opened 2018-08-28
Endpoint for attaching images to document via Javascript

kevindew:
Trello: https://trello.com/c/8tYjBHaX/159-create-endpoint-to-upload-image-via-javascript

This supersedes https://github.com/alphagov/content-publisher/pull/170 as the initial work for attaching images to a document.

It provides a service for image uploads that:

- Validates the file is a valid image meeting:
  - accepting JPG, GIF or PNG (SVG has a different set of concerns)
  - minimum 960 x 640 dimensions (as per Whitehall)
  - a max file size of 20 MB (an abstract number picked out thinking we really don't want to deal with images bigger than that) 
- Removes EXIF data and fixes orientation (which could be lost in exif data) - which can remove any location/equipment info from an image
- Determines a centre cropping for the image automatically based on it's dimensions which can be suggested to the user.

It has images as a separate model, this may well be a pain point for handling document history but seemed simplest route for now. I somewhat experimented with storing them as JSON within a document but it didn't look great. Maybe a gem like https://github.com/sufleR/embedson could help?

Some new things that this introduces:

## Controller tests

Since this isn't a complete feature it seemed appropriate that we test the behaviour of the endpoint using a different testing mechanism to feature tests. Controller tests seemed to be more appropriate than request tests as they're mostly unit testing the controller. 

## Presenters 

I know, I know. They're not the nicest things to add to a project. I wanted to avoid manipulating `as_json` on an Image model and felt a presenter class was a lesser evil. I thought maybe we could use [Active Model Serializers](https://github.com/rails-api/active_model_serializers) instead but that seems a bit neglected. https://github.com/Netflix/fast_jsonapi seems a better thing if we wanted to go that route.


app/controllers/document_images_controller.rb
```diff
@@ -0,0 +1,37 @@
...
+      image = create_image_from_upload(upload, document)
+      render json: ImageJsonPresenter.new(image).present, status: 201
+    else
+      render json: { errors: upload.errors }, status: 422
+    end
+  end
+
+private
+
+  def create_image_from_upload(upload, document)
```
tijmenb: Would it make sense for this to live in the `UploadedImageService`?

kevindew: Potentially - I hadn't thought about putting it there. I did debate putting it into the model but it felt a bit strange to pass the ValidImage models there.

I feel a bit happier with it outside the UploadedImageService as I think it confuses that a bit since it's designed for a single purpose. If we have to move other bits into there though this would be a good candidate.

app/models/image.rb
```diff
@@ -0,0 +1,17 @@
...
+class Image < ApplicationRecord
+  belongs_to :document
+  belongs_to :blob, class_name: "ActiveStorage::Blob" # rubocop:disable Rails/InverseOf
+
+  WIDTH = 960
+  HEIGHT = 640
+
+  def crop_variant
+    crop = "#{crop_width}x#{crop_height}+#{crop_x}+#{crop_y}"
+    blob.variant(
```
erino: I think we should add `colorspace: "sRGB"` here to normalise any colour profiles.

kevindew: Thanks - that feels like something that'd be better served [pre-cropping](https://github.com/alphagov/content-publisher/blob/e6cd962cae1cf79b5c764933da20b52754bf6293/app/services/uploaded_image_service/image_normaliser.rb#L20-L27) 

I don't know much/anything about colour profiles - is that stored in exif data or some different way? and I'm not too sure how to test whether adding something will have an effect.

erino: That does look like a better place, but I wasn't sure that the it could be added there.

PR #183, opened 2018-08-29
Set anchors in the preview container to open in a new tab

alex-ju:
This PR adds target blank to anchors generated in the preview container so the user doesn't loose the current page when clicking them.

### Context
screen shot 2018-08-29 at 12 24 46

### DIFF
--- a/app/javascript/components/markdown-editor.js
+++ b/app/javascript/components/markdown-editor.js
@@ -56,6 +56,8 @@ MarkdownEditor.prototype.handlePreviewButton = function (event) {
         }
       }
     )
+    // Open preview links in a new tab
+    this.setTargetBlank(this.$preview)
   } else {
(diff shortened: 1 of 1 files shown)


kevindew:
This looks good to me. Should we being writing some JS tests for this sort of thing? 

There's a risk this will conflict with the work Ben is doing with https://trello.com/c/77pWuPzi/170-render-govspeak-server-side so might be worth checking in with him in case we'll need to fix it a different way with that.

benthorner:
@kevindew That's true, although we're going to have to do this somewhere, and it looks like it's easier to do in JavaScript. What do you think - Ruby v. JS?

kevindew:
Yeah, whichever way seems easier for this since it'll be inserted via JS.

PR #188, opened 2018-08-30
Endpoint to update the crop of an image

kevindew:
Trello: https://trello.com/c/jY14RKov/182-create-endpoint-for-cropping-image-via-javascript

This creates an endpoint: `POST /documents/:document_id/images/:id/crop` which takes parameters: x, y, width and height and updates the stored crop rule of an image (as long as the parameters are valid).

The bulk of the code is probably the means to validate that a crop is of a valid aspect ratio, this is a slightly non-trivial task as an aspect-ratio often results in a non-integer for one side. To counter this we consider an aspect ratio valid if it is either the ceiling or floor of a non-integer.

I got a bit stuck in this about where to apply the validation. I followed the approach discussed in https://docs.google.com/document/d/142IslEIKXSSilNb5EVlM2l7usRfaQlhQDVHWhputGFs/edit#heading=h.mr9gdsek2gvz where model validation and whats returned to user are kept separate. It could well be though that this is all simpler if we moved the validations to the `Image` model and changed the crop endpoint to be just a generic PATCH of an image - does anyone have thoughts on this?

### DIFF
--- a/app/controllers/document_images_controller.rb
+++ b/app/controllers/document_images_controller.rb
@@ -13,7 +13,23 @@ class DocumentImagesController < ApplicationController
     end
   end
 
+  def crop
+    image = Image.find_by!(id: params[:id], document_id: params[:document_id])
+    crop = UpdateImageCropService.new(image, crop_params)
+
+    if crop.valid?
+      crop.update_image
+      render json: ImageJsonPresenter.new(image).present, status: :ok
+    else
+      render json: { errors: crop.errors }, status: :unprocessable_entity
+    end
+  end
+
 private
+
+  def crop_params
+    params.require(:crop).permit(:x, :y, :width, :height)
+  end
 
   def create_image_from_upload(upload, document)
     blob = ActiveStorage::Blob.create_after_upload!(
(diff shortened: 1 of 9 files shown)


benthorner:
I like the way the validations are structured, but I think a separate validation service or the Image class is a better place for them. I can think of a few reasons for choosing a different approach:

   - We're now duplicating some of the validation code from UploadedImageService
   - ActiveModel::Validations is already provided on the Image class - no need for anther model
   - We should centralise our model validations so that we always apply them consistently
   - It's normal to see validations defined inside the model class for ActiveRecord models
   - Some of these validations (e.g. crop_x > 0) are definitely essential concepts for the model

Although I initially agreed with the validation app preference, it was on the assumption that we would only have one code path to input the data - here we have two and it's starting to get a bit messy :-|. That said, a central validation service might be a compromise that satisfies both points of view.

emmabeynon:
I agree with Ben on the validations taking place on the model, as they seem an essential part of Image. Taking cropping as an example, ultimately what is required for a valid crop doesn’t change whether the crop was done when the image was uploaded or updated at a later time, so it makes sense to me for this validation to be managed from one place.

kevindew:
Closing this in favour of https://github.com/alphagov/content-publisher/pull/191

PR #189, opened 2018-08-30
Associate document with user who created it

emmabeynon:
For https://trello.com/c/Th0t8cUb/165-associate-a-document-with-the-user-who-created-it

In this PR we associate a document with a creator so we're able to look up who created it.  This allows us to display information about the person who created the document; for now it will just be the creator's name on the document summary page, but in future could include their organisation for example.

screen shot 2018-08-30 at 14 13 36

We also handle instances where a document has no creator by displaying it as "Unknown".

### DIFF
--- a/app/models/document.rb
+++ b/app/models/document.rb
@@ -2,6 +2,7 @@
 
 class Document < ApplicationRecord
   has_many :images, dependent: :destroy
+  belongs_to :creator, class_name: "User", optional: true, foreign_key: :creator_id, inverse_of: :documents
 
   PUBLICATION_STATES = %w[
     changes_not_sent_to_draft
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -2,6 +2,6 @@
 
 class User < ApplicationRecord
   include GDS::SSO::User
-
+  has_many :documents, foreign_key: :creator_id, inverse_of: :creator, dependent: :restrict_with_exception
   serialize :permissions, Array
 end
--- a/config/locales/en/documents/show.yml
+++ b/config/locales/en/documents/show.yml
@@ -32,6 +32,7 @@ en:
         status: Status
         updated_at: Last updated
         created_at: Created
+        created_by: Document created by
       flashes:
         draft_success: Preview creation successful
         draft_error: Error creating preview
--- a/config/locales/en/documents/show.yml
+++ b/config/locales/en/documents/show.yml
@@ -33,6 +33,7 @@ en:
         updated_at: Last updated
         created_at: Created
         created_by: Document created by
+        unknown_data: Unknown
       flashes:
         draft_success: Preview creation successful
         draft_error: Error creating preview
--- a/spec/factories/document_factory.rb
+++ b/spec/factories/document_factory.rb
@@ -8,5 +8,6 @@ FactoryBot.define do
     base_path { title ? "/prefix/#{title.parameterize}" : nil }
     document_type { build(:document_type_schema, path_prefix: "/prefix").id }
     publication_state { "changes_not_sent_to_draft" }
+    creator_id { User.last.id }
   end
 end
--- a/spec/features/show_a_document_spec.rb
+++ b/spec/features/show_a_document_spec.rb
@@ -18,7 +18,11 @@ RSpec.feature "Showing a document summary" do
   def then_i_see_the_document_summary
     expect(page).to have_content(@document.title)
     expect(page).to have_content(@document.summary)
-    expect(page).to have_content(@document.created_at.strftime("%l:%M%P on %d %B %Y"))
-    expect(page).to have_content(@document.updated_at.strftime("%l:%M%P on %d %B %Y"))
+
+    within("div.app-c-metadata") do
+      expect(page).to have_content(@document.created_at.strftime("%l:%M%P on %d %B %Y"))
+      expect(page).to have_content(@document.updated_at.strftime("%l:%M%P on %d %B %Y"))
+      expect(page).to have_content(@document.creator.name)
+    end
   end
 end
--- a/spec/features/show_a_document_with_no_creator_spec.rb
+++ b/spec/features/show_a_document_with_no_creator_spec.rb
@@ -1,0 +1,24 @@
+# frozen_string_literal: true
+
+RSpec.feature "Showing a document summary" do
+  scenario "User views a document that has no creator" do
+    given_there_is_a_document_with_no_creator
+    when_i_visit_the_document_page
+    the_document_creator_is_shown_as_unknown
+  end
+
+  def given_there_is_a_document_with_no_creator
+    @document = create(:document, creator_id: nil)
+  end
+
+  def when_i_visit_the_document_page
+    visit document_path(@document)
+  end
+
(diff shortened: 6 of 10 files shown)


db/migrate/20180830104245_add_creator_id_to_document.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class AddCreatorIdToDocument < ActiveRecord::Migration[5.2]
+  def change
+    add_column :documents, :creator_id, :integer
```
kevindew: It might be better to do this through a [`add_reference`](https://apidock.com/rails/v4.0.2/ActiveRecord/ConnectionAdapters/SchemaStatements/add_reference) since we're referencing the user table, and will match the data type (which I believe is a bigint)

app/models/document.rb
```diff
@@ -22,4 +22,8 @@ def document_type_schema
   def newly_created?
     self.created_at == self.updated_at
   end
+
+  def created_by
+    User.find_by(id: self.creator_id)
+  end
```
kevindew: It might be better to use a belongs_to to reference this association?

spec/factories/document_factory.rb
```diff
@@ -8,5 +8,6 @@
     base_path { title ? "/prefix/#{title.parameterize}" : nil }
     document_type { build(:document_type_schema, path_prefix: "/prefix").id }
     publication_state { "changes_not_sent_to_draft" }
+    creator_id { User.last.id }
```
kevindew: I think if we switch this to be a one to one association we can simplify this rather than needing to pull an id out of the db.

config/locales/en/documents/show.yml
```diff
@@ -32,6 +32,7 @@ en:
         status: Status
         updated_at: Last updated
         created_at: Created
+        created_by: Document created by
```
kevindew: We could pass in the name as a variable as per https://github.com/alphagov/content-publisher/blob/b62190d6873105e2cc244dc5e502e980aaa724b7/config/locales/en/validations.yml#L5 and then you wouldn't need to concatenate the string after?

app/models/document.rb
```diff
@@ -22,4 +22,8 @@ def document_type_schema
   def newly_created?
     self.created_at == self.updated_at
   end
+
+  def created_by
```
tijmenb: Have you considered setting up a `belongs_to :creator` relationship for this? It would prevent us from having to manually define the `created_by`.

db/migrate/20180830104245_add_creator_id_to_document.rb
tijmenb: Can we add a foreign key to the users table here, so that we won't be able to accidentally delete users that have documents attached?

emmabeynon:
@tijmenb @kevindew I was originally going to do a belongs_to association, but not quite knowing how history is going to work in future I thought it might be simpler and clearer to do it this way and hopefully easy to get rid of once we know what history will look like.  I also wanted to avoid creating a separate model for a creator; it's how it's done in Whitehall and it seemed a bit unnecessary.  Happy to chat through further in any case 👍

config/locales/en/documents/show.yml
```diff
@@ -32,6 +32,8 @@ en:
...
+        unknown_data: Unknown
```
benthorner: Tiny thing, but could we call this "unknown_creator", just so it's obvious where it's being shown?

kevindew:
@emmabeynon Yeah hopefully we'll crack history in the not too distant future. The person who changes a document shouldn't change as part of history.

I don't think you'll need to create a creator model, just an association between document and user that is named creator?

config/locales/en/documents/show.yml
emmabeynon: Sure - my thinking on this was it could be reused for other future metadata fields for which we don't have the data.

spec/features/show_a_document_spec.rb
```diff
@@ -20,5 +20,6 @@ def then_i_see_the_document_summary
     expect(page).to have_content(@document.summary)
     expect(page).to have_content(@document.created_at.strftime("%l:%M%P on %d %B %Y"))
     expect(page).to have_content(@document.updated_at.strftime("%l:%M%P on %d %B %Y"))
+    expect(page).to have_content(I18n.t("documents.show.metadata.created_by") + ": " + @document.creator.name)
```
benthorner: I think we're not bothering with the label here - just the value is fine.

spec/features/show_a_document_with_no_creator_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+  def given_there_is_a_document_with_no_creator
+    @document = create(:document, creator_id: nil)
+  end
+
+  def when_i_visit_the_document_page
+    visit document_path(@document)
+  end
+
+  def the_document_creator_is_shown_as_unknown
+    expect(page).to have_content(I18n.t("documents.show.metadata.unknown_data"))
```
benthorner: This one could be a bit flakey, as there may be other 'Unknown's on the page in future. Could we make the copy include something about users, so the test should always pass for the right reason?

app/models/user.rb
```diff
@@ -2,6 +2,6 @@
 
 class User < ApplicationRecord
   include GDS::SSO::User
-
+  has_many :documents, foreign_key: :creator_id, inverse_of: :user, dependent: :restrict_with_exception
```
benthorner: I think we should be OK without the inverse_of, as it's only required when we change the user, which isn't very likely.

spec/features/show_a_document_spec.rb
emmabeynon: Ah so the reason I had to include this was that without it the test would pass as the creator's name would already appear in the page in the top right hand corner.  This helps make sure we're testing the right thing.

spec/features/show_a_document_with_no_creator_spec.rb
emmabeynon: Good point, I could add the `documents.show.metadata.created_by` microcopy as I did in the other test? What do you think?

app/models/user.rb
emmabeynon: Rubocop gets upset and demands the `:inverse_of` option if I leave it out 🙈

spec/features/show_a_document_spec.rb
benthorner: So I think you can use a 'within' block (which could go around all of those expectations) to scope the have_content matcher to look within the status box.

app/models/user.rb
benthorner: Ah, OK. In that case, I think the documentation says it the inverse_of for user should be 'creator' rather than 'user'.

spec/features/show_a_document_spec.rb
emmabeynon: Good idea - will implement that 👍

app/models/user.rb
emmabeynon: Ah yes good spot, thanks.

emmabeynon:
@benthorner Those changes should be sorted now 👍

PR #191, opened 2018-08-30
Update image endpoint for cropping

kevindew:
Trello: https://trello.com/c/jY14RKov/182-create-endpoint-for-cropping-image-via-javascript

This is a refactor of https://github.com/alphagov/content-publisher/pull/188 based on feedback. 

This switches the validations to be done on model layer and then, since we're doing that, sets up an update image endpoint rather than a crop endpoint.

Because we're updating the image directly we then need to accept the parameters in the form of the db structure (or we have to do some mapping and swapping). 

Having put this together I'm feeling that this is potentially the worse of the routes, it may be way more consistent with how Rails does things in a RESTful context but since we don't really seem to be operating in this manner for the rest of the app it feels like an exception. It also fears like it forces us to have a generic REST update endpoint rather than anything specialist and that we should update presentation to match db so format so that's consistent with input

Not sure if there's any sort of half way approach that might be nicer?

### DIFF
--- a/spec/validators/image_aspect_ratio_validator_spec.rb
+++ b/spec/validators/image_aspect_ratio_validator_spec.rb
@@ -1,0 +1,50 @@
+# frozen_string_literal: true
+
+RSpec.describe ImageAspectRatioValidator do
+  subject(:instance) do
+    anon_class = Class.new do
+      include ActiveModel::Validations
+      attr_accessor :crop_width, :crop_height
+      validates_with ImageAspectRatioValidator
+    end
+    anon_class.new
+  end
+
+  it "is valid for a 3:2 aspect ratio" do
+    instance.crop_width = 3000
+    instance.crop_height = 2000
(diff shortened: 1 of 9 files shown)


spec/validators/image_aspect_ratio_validator_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+    expect(instance).to be_valid
+
+    instance.crop_width = 664
+    instance.crop_height = 442
+    expect(instance).to be_valid
+
+    # 443 wide at 3:2 equates to 664.5
+    instance.crop_height = 443
+    instance.crop_width = 664
+    expect(instance).to be_valid
```
davbo: is this expectation any different to the one on line 20-23?

kevindew:
@davbo yup - whoops. Had a typo in the comment above it. Fixed in 906a659

PR #192, opened 2018-08-31
Show raw markdown on the summary page

benthorner:
https://trello.com/c/tOx0eK0R/176-show-raw-markdown-and-truncate-output-on-summary-page

We think users will find it confusing to see styled markdown content
next to static page content (e.g. titles), so this commit changes the
preview to show the raw markdown using a new 'markdown_raw' component.

The component truncates the raw markdown by lines, for which 3 lines
seems to give a reasonable preview (two paragraphs, or title and a
paragraph).

### DIFF
--- a/app/views/documents/fields/_govspeak.html.erb
+++ b/app/views/documents/fields/_govspeak.html.erb
@@ -1,3 +1,4 @@
-<%= render "govuk_publishing_components/components/govspeak", {} do %>
-  <%= raw Govspeak::Document.new(@document.contents[schema.id]).to_html %>
-<% end %>
+<%= render "components/formatted_text", {
+  text: @document.contents[schema.id].to_s,
+  lines: 3
+} %>
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -8,6 +8,7 @@ $govuk-global-styles: true;
 
 @import "components/contextual-guidance";
 @import "components/error-alert";
+@import "components/formatted-text";
 @import "components/markdown-editor";
 @import "components/markdown-toolbar";
 @import "components/metadata";
--- a/app/assets/stylesheets/components/_formatted-text.scss
+++ b/app/assets/stylesheets/components/_formatted-text.scss
@@ -1,0 +1,6 @@
+.app-c-formatted-text {
+  @include govuk-font(19);
+
+  margin: 0;
+  white-space: pre-wrap;
+}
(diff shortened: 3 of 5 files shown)


app/assets/stylesheets/components/_markdown-raw.scss
```diff
@@ -0,0 +1,5 @@
+.app-c-markdown-raw {
+  font-family: inherit;
```
alex-ju: If this is a component it should be isolated and not dependent on other styles. In our case we need to declare it explicitly and we can use the following mixin `@include govuk-font(19);`

app/views/documents/fields/_govspeak.html.erb
```diff
@@ -1,3 +1,4 @@
-<%= render "govuk_publishing_components/components/govspeak", {} do %>
-  <%= raw Govspeak::Document.new(@document.contents[schema.id]).to_html %>
-<% end %>
+<%= render "components/markdown_raw", {
+  text: @document.contents[schema.id].to_s,
+  lines: 3
```
alex-ju: This gives us the first 4 lines, which depending on how the editor (the person who's editing the text) is using empty lines can mean any of the following:
 - a title and a paragraph
 - a title a paragraph and a list item
 - only the first paragraph

Do you think we should skip counting the empty lines?

app/assets/stylesheets/components/_markdown-raw.scss
benthorner: Good point - done.

app/views/documents/fields/_govspeak.html.erb
benthorner: The first and second are true, but I'm not sure how the lat one would happen unless they manually entered lots of newlines within the paragraph. I've tried it out with a few documents and think it looks alright (it'll probably never be perfect), but I could get Humin to look at it if you like?

alex-ju: Yep, we can even go forward with this and test it with users.

emmabeynon: A small thing, but since we're not rendering govspeak here anymore and instead displaying the raw markdown, should we rename this partial to reflect that?

benthorner: That's a good point - it is just arbitrary text we're passing over. I believe textbox inputs have the same issue when the content is being previewed. @alex-ju do you have an opinion on the name? I was thinking something like 'formatted_text' to go with the use of the <pre> tag.

alex-ju: don't have strong opinions on naming. I would be tempted to mimic the label/name of the field, but in our case body is too generic and can refer to many things. `formatted_text ` sounds good.

benthorner: Cool - @emmabeynon that OK with you?

emmabeynon: Works for me 👍

PR #193, opened 2018-08-31
Add separate preview page

tijmenb:
This adds the first bit of the [preview page](https://trello.com/c/KO7s5XJG/152-design-the-preview-page-s). 

It only implements the desktop view (see other todo's on the Trello ticket).

https://trello.com/c/i2Fi5qNQ

### DIFF
--- a/app/assets/stylesheets/components/_page_preview.scss
+++ b/app/assets/stylesheets/components/_page_preview.scss
@@ -1,0 +1,5 @@
+.app-c-preview__desktop-preview {
+  width: 100%;
+  height: 1000px;
+  border: 0;
+}
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -12,6 +12,7 @@ $govuk-global-styles: true;
 @import "components/markdown-toolbar";
 @import "components/metadata";
 @import "components/summary";
+@import "components/page_preview";
 
 @import "utilities/display";
 @import "utilities/overwrites";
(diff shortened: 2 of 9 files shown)


app/assets/stylesheets/components/_page_preview.scss
```diff
@@ -0,0 +1,6 @@
+.app-c-preview {
```
alex-ju: There's no need to nest these as the second selector `app-c-preview__desktop-preview` is already unique. The following will be enough:
```
.app-c-preview__desktop-preview {
  width: 100%;
  height: 1000px;
}
```

If we'll have specific styles for `.app-c-preview` we can define them separately
```
.app-c-preview { ... }
```

app/assets/stylesheets/components/_page_preview.scss
```diff
@@ -0,0 +1,6 @@
...
+  .app-c-preview__desktop-preview {
+    width: 100%;
+    height: 1000px;
```
alex-ju: I would add `border: 0;` to get rid of the iframe border.

I get `1000px` is an arbitrary number which is ok for now; if we want to not have an inner scroll for the page we'll need some JS on both this component and the inner document/page. We can achieve that with mutation observers, did that for the component preview in the Design System, happy to add it afterwards if we want it do behave this way.

alex-ju:
One more thing. Can we add the preview link to the preview button on the summary page?

tijmenb:
@alex-ju updated, thanks! I'll add the link to this page later, once it's more complete.

PR #199, opened 2018-09-03
Add mobile preview

tijmenb:
This implements the mobile preview in the component.

screen shot 2018-09-03 at 10 55 25

https://trello.com/c/i2Fi5qNQ

### DIFF
--- a/app/views/components/_page_preview.html.erb
+++ b/app/views/components/_page_preview.html.erb
@@ -5,6 +5,12 @@
 <% end %>
 
 <div class="app-c-preview">
+  <% mobile_preview = capture do %>
+    <%= tag.div class: "app-c-preview__mobile" do %>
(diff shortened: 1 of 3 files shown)


benthorner:
Do we want people to be able to scroll inside the iframes? I'm just thinking it'll interfere with the scrolling of the main page, especially (and ironically) on a mobile.

app/views/components/_page_preview.html.erb
```diff
@@ -5,22 +5,28 @@
 <% end %>
 
 <div class="app-c-preview">
+  <% mobile_preview = capture do %>
+    <%= tag.div class: "app-c-preview__mobile" do %>
```
benthorner: What's the benefit of using tag. here? I think we're using explicit HTML everywhere else.

PR #201, opened 2018-09-03
Link to the separate preview page

tijmenb:
This changes the action link to go to the new separate preview page, so the user can see the preview on mobile, desktop and Google.

https://trello.com/c/i2Fi5qNQ

### DIFF
--- a/spec/features/preview_document_spec.rb
+++ b/spec/features/preview_document_spec.rb
@@ -15,7 +15,7 @@ RSpec.feature "Preview a document" do
 
   def when_i_visit_the_preview_page
     visit document_path(@document)
-    click_on I18n.t("documents.show.actions.view_draft")
+    click_on "Preview"
   end
 
   def then_i_see_the_previews
--- a/spec/features/preview_document_spec.rb
+++ b/spec/features/preview_document_spec.rb
@@ -10,11 +10,12 @@ RSpec.feature "Preview a document" do
   end
 
   def given_there_is_a_document_in_draft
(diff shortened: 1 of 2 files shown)


spec/features/preview_document_spec.rb
```diff
@@ -10,11 +10,12 @@
...
 
   def given_there_is_a_document_in_draft
-    @document = create(:document, base_path: "/foo/foo", content_id: "d2547c42-8ed3-49f5-baeb-6112f98c2bf9")
+    @document = create(:document, publication_state: "sent_to_draft", base_path: "/foo/foo", content_id: "d2547c42-8ed3-49f5-baeb-6112f98c2bf9")
   end
 
   def when_i_visit_the_preview_page
-    visit preview_document_path(@document)
+    visit document_path(@document)
+    click_on I18n.t("documents.show.actions.view_draft")
```
benthorner: We decided not to use translations for clickable things, as people felt it made the tests too hard to read. Looks like the show.yml file still has the actions section, though - would you mind removing it?

tijmenb: I've hardcoded it here. Not sure what you mean by show.yml, it's used by actions.html: https://github.com/alphagov/content-publisher/blob/8cb2311f26425a9bc093bdc90e14d294956b346f/app/views/documents/show/_actions.html.erb#L4

benthorner: Indeed they are, but they should have been removed there as well. We adopted the convention that we wouldn't use translations for clickable things, but it looks like these got missed when we applied the convention over the app. If you like I'll do a separate PR to make it consistent.

tijmenb: Yeah, you can do a separate PR for that.

PR #203, opened 2018-09-03
Update Whitehall import script

emmabeynon:
For https://trello.com/c/LVpjWa42/173-update-whitehall-import-script-to-support-organisations

This updates the Whitehall import script to handle a document summary and its various associations.

### DIFF
--- a/lib/tasks/whitehall_news_importer.rb
+++ b/lib/tasks/whitehall_news_importer.rb
@@ -43,6 +43,7 @@ module Tasks
       associations["organisations"] = organisations(edition)
       associations["worldwide_organisations"] = edition["worldwide_organisations"]
       associations["topical_events"] = edition["topical_events"]
+      associations["world_locations"] = edition["world_locations"]
       associations
     end
 
--- a/lib/tasks/whitehall_news_importer.rb
+++ b/lib/tasks/whitehall_news_importer.rb
@@ -41,6 +41,7 @@ module Tasks
       associations = {}
       associations["primary_publishing_organisation"] = primary_publishing_organisation(edition)
       associations["organisations"] = organisations(edition)
+      associations["worldwide_organisations"] = edition["worldwide_organisations"]
       associations
     end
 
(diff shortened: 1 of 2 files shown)


lib/tasks/whitehall_news_importer.rb
```diff
@@ -35,7 +37,39 @@ def create_document(translation, edition, document)
...
+        summary: translation["summary"],
+        associations: associations(edition),
       )
     end
+
+    def associations(edition)
+      associations = {}
+      associations["primary_publishing_organisation"] = primary_publishing_organisation(edition)
+      associations["organisations"] = organisations(edition)
+      ASSOCIATIONS_WITH_ONLY_CONTENT_IDS.each do |type|
```
benthorner: I think it would be a bit clearer to just have the repeated lines here, which would actually decrease the total number of lines in the file. Can we do that for now?

spec/tasks/whitehall_news_importer_spec.rb
```diff
@@ -15,12 +15,45 @@
...
     expect { Tasks::WhitehallNewsImporter.new(JSON.parse(import_json)).import }
       .to change { Document.count }.by(1)
+
+    imported_edition = JSON.parse(import_json)[0]["editions"][0]
+    document_associations = Document.last.associations
+
+    expect(Document.last.summary)
+      .to eq(imported_edition["translations"][0]["summary"])
+    expect(document_associations["primary_publishing_organisation"][0])
+      .to eq(imported_edition["lead_organisations"][0]["content_id"])
```
benthorner: I'm not sure why the org JSON is different to the other things - can we check with @kevindew and look at simplifying the export JSON so it's easier to write the import code?

kevindew: Yeah I think that could be simplified - it was wrote while we still anticipated storing the organisation name.

The ministers one is also complex but for different reasons to store roles/people

lib/tasks/whitehall_news_importer.rb
emmabeynon: Fair enough, will do 👍

spec/tasks/whitehall_news_importer_spec.rb
emmabeynon: I've got a PR here to simplify this and just return content_ids for orgs https://github.com/alphagov/whitehall/pull/4316

emmabeynon:
@benthorner your comments should be addressed now 👍

PR #207, opened 2018-09-04
Rework choosing a document type

tijmenb:
This PR implements the new flow for choosing a document type [as designed in this Google Doc](https://docs.google.com/document/d/1ivZ34_SBeIRllwi31MVewyxudE7F0QnW3U3p5QoHZ6E/edit#).

It differs from the current implementation:

- Some supertypes now redirect instantly to Whitehall, without the second screen
- The "publication" document type appears on the second page for multiple supertypes. This means that we've had to flip the relationship from a "document type belongs to supertype" to a "supertype has many document types".
- Supertypes and document types can have conditionally revealed content
- Some document types will not be official document types like "publication"

https://trello.com/c/5GSUif0g


spec/formats/consultation_spec.rb
```diff
@@ -1,15 +1,15 @@
 # frozen_string_literal: true
```
benthorner: This looks good. I think the purpose of this spec is mainly now to test the redirect capability in general, so could we rename it to indicate this?

tijmenb: Not sure about that because the [detailed guide spec is exactly the same as this test](https://github.com/alphagov/content-publisher/blob/518e7046b386c6256c10c9287f14ab45fd1c4ead/spec/formats/detailed_guide_spec.rb#L3-L22), so I'm don't know what to do about that one. 

However, maybe we could talk about testing another time? I don't think the current system works very well, but I don't want to derail this PR with that discussion.

PR #209, opened 2018-09-04
Fix encoding issue with govspeak preview body

benthorner:
https://trello.com/c/ORACsrIh/172-create-a-jenkins-job-to-update-integration-content-publisher-with-whitehall-content

It looks like Rack is interpreting the body text as ASCII-8BIT rather
than UTF-8, so we need to force this encoding.

### DIFF
--- a/app/controllers/govspeak_preview_controller.rb
+++ b/app/controllers/govspeak_preview_controller.rb
@@ -4,6 +4,6 @@ class GovspeakPreviewController < ApplicationController
   skip_before_action :verify_authenticity_token
 
   def to_html
-    render plain: GovspeakService.new.to_html(request.raw_post)
+    render plain: GovspeakService.new.to_html(params[:govspeak])
   end
 end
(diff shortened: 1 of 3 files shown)


app/controllers/govspeak_preview_controller.rb
```diff
@@ -4,6 +4,8 @@ class GovspeakPreviewController < ApplicationController
   skip_before_action :verify_authenticity_token
 
   def to_html
-    render plain: GovspeakService.new.to_html(request.raw_post)
```
tijmenb: I think it'd be better if we change `request.raw_post` into something Rails-y, that will probably solve the encoding issues as well. Can we change the JS so that Rails can access it as `params[:govspeak]`?

benthorner: Ah, I think this was a gap in my knowledge about Rails forms. Thanks for prompting me to look at it again - I'll have a go at fixing it.

PR #211, opened 2018-09-04
Implement 2i and publish workflow

tijmenb:
This reworks the publishing flow, and implements the 2i system.

https://trello.com/c/9szY348f


app/models/document.rb
```diff
@@ -14,7 +14,15 @@ class Document < ApplicationRecord
     error_sending_to_live
   ].freeze
 
+  REVIEW_STATES = %w[
+    unreviewed
+    submitted_for_review
+    force_published
+    reviewed
+  ].freeze
```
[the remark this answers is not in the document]
tijmenb: Yes, there's a difference. When a document is `submitted_for_review`, the primary action is a publish button (because if you think the document is okay you can publish it). When the document is `published_without_review`, the primary action is the approve button.

I expect that user research might give us some new insights here, but this is how it's current designed.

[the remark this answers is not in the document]
tijmenb: I've added the user facing state to this PR anyway, it probably makes it easier to understand the whole system.

app/services/user_facing_state.rb
```diff
@@ -0,0 +1,24 @@
...
+  delegate :review_state, :publication_state, to: :document
+
+  def initialize(document)
+    @document = document
+  end
+
+  def to_s
+    if review_state == "submitted_for_review"
+      "submitted_for_review"
+    elsif publication_state.in?(%w[changes_not_sent_to_draft sent_to_draft sending_to_draft error_sending_to_draft sending_to_live])
```
davbo: Wondering if having `error_sending_to_draft` here will cause problems ? would it be easier for now to have a user facing error state to prevent other actions on the document e.g. to stop someone 'publishing' something which hasn't been sent to draft yet.

app/services/user_facing_state.rb
```diff
@@ -0,0 +1,24 @@
...
+class UserFacingState
+  attr_reader :document
+  delegate :review_state, :publication_state, to: :document
+
+  def initialize(document)
+    @document = document
+  end
+
+  def to_s
+    if review_state == "submitted_for_review"
```
davbo: Maybe a case statement or a lookup on a hash of states here would be a bit easier to read? If we match on both review_state and publication_state we're more likely to catch errors where they go out of sync too.

Right now I could imagine a documents review_state being`submitted_for_review` but having an `error_sending_to_draft` publication_state which seems like a state we'd like to raise an exception for.

app/services/user_facing_state.rb
tijmenb: That's a really good point. I've put in a story here https://trello.com/c/VkVrVGeB, because this is something Ben and Sonia need to think about. I'll pick that up directly after this.

app/services/user_facing_state.rb
tijmenb: I agree it's a bit iffy now, but I wanted to hold off refactoring until we have the full logic - this is currently missing the scheduling, withdrawal, history mode states & the [error state work](https://trello.com/c/VkVrVGeB). Does that make sense?

davbo: Makes sense 👍 thanks

spec/features/force_publishing_spec.rb
```diff
@@ -0,0 +1,52 @@
...
+  def then_i_see_that_the_document_was_published_without_review
+    expect(page).to have_content("Content has been published but still needs 2i review")
+  end
+
+  def and_i_click_the_approval_button
+    click_on "Approve"
+  end
+
+  def then_i_see_that_its_reviewed
+    expect(page).to have_content "Content has been reviewed and approved"
```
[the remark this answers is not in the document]
tijmenb: I'll do this in a future PR.

spec/features/publish_a_document_spec.rb
```diff
@@ -21,14 +23,22 @@ def and_i_click_on_the_publish_button
     click_on "Publish"
   end
 
+  def and_i_say_that_the_document_has_been_reviewed
+    choose "This content has been reviewed and approved for publication"
```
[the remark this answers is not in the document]
tijmenb: I'll do this in a future PR.

binaryberry:
Is there a reason why the user action "withdraw" and the user state "retired" are using different words? Retire & retired or Withdraw & withdrawn would be more explicit.

benhazell:
You're right, we should be consistent and this may be an oversight.

However, I wanted us to adopt the naming recommendations from the archiving
team, but Humin has heard that they weren't based on any research and so
this might not be a good idea. Waiting to hear more.

binaryberry:
Hi Ben! Thanks for answering - sorry I retroactively deleted my comment when I realised the term "withdraw" was only used in the mapping above and not in the code - it seems the code uses 'retire' and 'retired' only. My concern was about developers (and users) potentially calling the same thing different names, which could have led to confusion.

benhazell:
Ok. Right to watch out for that.
Carry on as before then.

PR #213, opened 2018-09-04
Make API down rescue mechanism generic

benthorner:
This allows us to have multiple views containing GdsApi code and still
handle the API down case gracefully via a dedicated view.


kevindew:
I'm a bit confused on the problem this is solving? do we have different views for each of the actions? 

Might it be better to just have a single generic view and pass in different messages? feels like we could have a lot of very similar views to manage?

benthorner:
It's not solving a current problem, it's just a poor design decision I noticed I'd made and wanted to fix. For now I think it's a net improvement, and perhaps we should make time to think about our error handling strategy more generally; at the moment, the api_down views are more complex than just a single message.

PR #217, opened 2018-09-05
Remove webpack

kevindew:
This pulls out the integration of webpack and re-instates the process of storing Javascript files we need in a vendor directory.

The reason for pulling this out is problems caused by integration with govuk_publishing_components and that's wider dependency on govuk_frontend_toolkit.

A problem I didn't realise when setting up webpack is that there is no means to import javascript that is provided by gems. It turns out the Rails approach to this is to also release a NPM module - which is a process that we (GOV.UK) can't really adapt to rapidly.

This problem can be overcome with this somewhat hacky approach (added here for prosperity):

config/webpack/environment.js
```
const { environment } = require('@rails/webpacker')
const { execSync } = require('child_process')

const config = environment.toWebpackConfig()
let gpcGemPath = null

try {
  gpcGemPath = execSync('bundle show govuk_publishing_components')
    .toString()
    .replace(/\n$/, '')
} catch (err) {
  console.error("Expected govuk_publishing_components gem to be installed")
  process.exit()
}

config.resolve.modules.push(`${gpcGemPath}/app/assets/javascripts`)
module.exports = config
```

However we then later learnt that we have another problem with using govuk_publishing_components as they are all initialised by using javascript provided by govuk_frontend_toolkit which then requires jQuery which is provided by static. 

At this point is just seemed to be going too far to either try solve the problems of a deprecated govuk_frontend_toolkit now and instead I think the best path is raising an issue on govuk_publishing_components about what would be needed - hopefully @alex-ju and myself can get a chance to work on this over firebreak.


kevindew:
@benthorner the README should be ok to stay as is as we still have a package.json for standard linting and I guess since yarn is more Rails than NPM it's better to stick with that since it all works (I think we also might get a nag from Rails assets:precompile if we switched back to NPM since it thinks you've made a mistake)

alex-ju:
I think we've lost the `restrict` functionality with this one

kevindew:
@alex-ju Yeah I pulled it out of this. Sorry I should have made that clearer though.

I figured we didn't actually need it and could use a:

```
var $govspeak = document.querySelector('[data-module="markdown-editor"]')
if ($govspeak) {
  new MarkdownEditor($govspeak).init()
}
```

in the JS file for each component. 

It was also somewhat prompted by the linter hating it because it was just a file with a global function in, which felt a bit funky.

alex-ju:
Yeah, saw that. Will got his way then.

PR #218, opened 2018-09-05
Implement flash messages after 2i submit

tijmenb:
This is the flash message that occurs when the user has submitted the
content for review.

screen shot 2018-09-05 at 17 56 18

https://trello.com/c/9szY348f


app/views/documents/show/_submitted_for_review.html.erb
```diff
@@ -0,0 +1,10 @@
+<% copy_to_clipboard = render(
+  "govuk_publishing_components/components/copy_to_clipboard",
+    label: t("documents.show.flashes.submitted_for_review.label"),
+    copyable_content: "http://www.example.org",
```
kevindew: oopsy!

PR #226, opened 2018-09-05
Use the copy to clipboard component

tijmenb:
See commits for details.

https://trello.com/c/hAfZNnxt

### DIFF
--- a/app/views/components/_page_preview.html.erb
+++ b/app/views/components/_page_preview.html.erb
@@ -1,5 +1,8 @@
 <%= render "govuk_publishing_components/components/details", title: "Share" do %>
-  <p>Copy and send this link to someone and they’ll be able to preview your draft on GOV.UK.</p>
+  <%= render "govuk_publishing_components/components/copy_to_clipboard",
+    label: "Copy and send this link to someone and they’ll be able to preview your draft on GOV.UK.",
+    copyable_content: url,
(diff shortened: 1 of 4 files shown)


app/views/components/_page_preview.html.erb
```diff
@@ -1,5 +1,8 @@
 <%= render "govuk_publishing_components/components/details", title: "Share" do %>
-  <p>Copy and send this link to someone and they’ll be able to preview your draft on GOV.UK.</p>
+  <%= render "govuk_publishing_components/components/copy_to_clipboard",
+    label: "Copy and send this link to someone and they’ll be able to preview your draft on GOV.UK.",
+    copyable_content: url,
+    button_text: "Copy link" %>
 
   <%= link_to url, url %>
```
alex-ju: I think we don't need this line anymore.

config/locales/en/global.yml
```diff
@@ -0,0 +1,3 @@
+en:
```
benthorner: We decided not to use translations for any components, as they may be moved out of this repo in future into the main publishing components one, where the translations wouldn’t work.

tijmenb: This string isn't used inside the component, just passed into it.

benthorner: Since this is an interactive element, you don't need to use a translation for it.

PR #232, opened 2018-09-06
[Do not merge] Prototype for history

tijmenb:
This is a prototype for tracking history to aid in https://trello.com/c/MQfGA9yU. 

It's throwaway code and doesn't have any tests.

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2018_09_05_134625) do
+ActiveRecord::Schema.define(version: 2018_09_06_111857) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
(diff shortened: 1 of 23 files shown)


db/schema.rb
```diff
@@ -55,6 +55,46 @@
...
+    t.index ["document_id"], name: "index_event_document_submitteds_on_document_id"
+    t.index ["user_id"], name: "index_event_document_submitteds_on_user_id"
+  end
+
+  create_table "event_document_updateds", force: :cascade do |t|
+    t.bigint "document_id"
+    t.bigint "user_id"
+    t.json "before", default: {}
+    t.json "after", default: {}
+    t.json "changeset", default: {}
```
benthorner: What was this for?

db/schema.rb
```diff
@@ -55,6 +55,46 @@
...
+    t.datetime "created_at", null: false
+    t.datetime "updated_at", null: false
+    t.index ["document_id"], name: "index_event_document_submitteds_on_document_id"
+    t.index ["user_id"], name: "index_event_document_submitteds_on_user_id"
+  end
+
+  create_table "event_document_updateds", force: :cascade do |t|
+    t.bigint "document_id"
+    t.bigint "user_id"
+    t.json "before", default: {}
```
benthorner: Could we get this by looking at the previous event?

db/schema.rb
```diff
@@ -55,6 +55,46 @@
...
+    t.datetime "created_at", null: false
+    t.datetime "updated_at", null: false
+    t.index ["document_id"], name: "index_event_document_approveds_on_document_id"
+    t.index ["user_id"], name: "index_event_document_approveds_on_user_id"
+  end
+
+  create_table "event_document_publisheds", force: :cascade do |t|
+    t.bigint "document_id"
+    t.bigint "user_id"
+    t.string "review_state"
```
benthorner: I'm wondering if we could just store the whole document, so we don't have to worry about which attributes were 'in-play' at the time.

tijmenb: `review_state` is a confusing name here, it's supposed to be something that indicates whether or not this document was force-published. I think we might make that a separate event, or a boolean on the data record.

db/schema.rb
tijmenb: Yeah, or save just the changeset. The consideration for saving both is that we can then clean up old updated-events, and every event can be rendered without doing a weird query.

db/schema.rb
tijmenb: The changes in the document.

PR #236, opened 2018-09-07
Implement change note and update type feature

tijmenb:
This implements the change note and update type feature.

- Users can now add a change note and select an update type in the edit screen.
- When a document is first created, it has a "major" update type by default and "First published." as change note. This can't be changed.
- The update type and change note are sent to the publishing-api, so change notes for major updates will be displayed on the site. The update type is no longer part of the `publish` call.
- The change note is displayed on the summary page as well.
- After a document is published, we reset the update type and change note.
- I've added a test for the creation of a document, because there wasn't a feature test that tests all the things that need to be set up when the document is first created.

## Known issues

- The layout of the new form labels will be fixed by @alex-ju and https://github.com/alphagov/govuk_publishing_components/pull/508.
- If users don't select an update type, the publishing-api will reject the update. This needs a new validation rule, which will be part of https://trello.com/c/K52YqwPK. 

## Preview

sep-07-2018 12-49-16

https://trello.com/c/NEQWHpoU

### DIFF
--- a/app/services/document_publishing_service.rb
+++ b/app/services/document_publishing_service.rb
@@ -11,8 +11,8 @@ class DocumentPublishingService
 
   def publish(document, review_state)
     document.update!(publication_state: "sending_to_live", review_state: review_state)
-    publishing_api.publish(document.content_id, "major", locale: document.locale)
-    document.update!(publication_state: "sent_to_live")
+    publishing_api.publish(document.content_id, nil, locale: document.locale)
+    document.update!(publication_state: "sent_to_live", change_note: nil, update_type: "major", has_live_version_on_govuk: true)
   end
 
 private
--- a/db/migrate/20180906155513_add_change_notes_to_documents.rb
+++ b/db/migrate/20180906155513_add_change_notes_to_documents.rb
@@ -1,0 +1,10 @@
+# frozen_string_literal: true
+
+class AddChangeNotesToDocuments < ActiveRecord::Migration[5.2]
+  def change
+    change_table :documents, bulk: true do |t|
+      t.column :change_note, :text
+      t.column :update_type, :string
+    end
+  end
+end
--- a/spec/features/change_notes_spec.rb
+++ b/spec/features/change_notes_spec.rb
@@ -1,0 +1,69 @@
+# frozen_string_literal: true
+
+RSpec.feature "Change notes" do
+  scenario "User updates change notes and update type" do
+    given_there_is_a_previously_published_document
+    when_i_go_to_edit_the_document
+    and_i_fill_in_the_change_note_and_update_type
+
+    then_the_summary_page_displays_change_note_and_update_type
+    and_the_form_displays_the_change_note_and_update_type
+    and_the_publishing_api_should_know_the_update_type_and_change_note
+
+    when_i_publish_the_document
+    then_the_change_note_and_update_type_should_be_cleared_for_the_next_edition
+  end
+
+  def given_there_is_a_previously_published_document
+    @document = create(:document, has_live_version_on_govuk: true)
+  end
+
+  def when_i_go_to_edit_the_document
+    visit edit_document_path(@document)
+  end
+
+  def and_i_fill_in_the_change_note_and_update_type
+    @update_request = stub_any_publishing_api_put_content
+
+    fill_in "document[change_note]", with: "Updated banana pricing"
+    choose I18n.t("documents.edit.update_type.minor_name")
+
+    click_on "Save"
+  end
+
+  def then_the_summary_page_displays_change_note_and_update_type
+    expect(page).to have_content "Updated banana pricing"
+  end
+
+  def and_the_form_displays_the_change_note_and_update_type
+    visit edit_document_path(@document)
+    expect(find("textarea[name='document[change_note]']")).to have_text "Updated banana pricing"
+    expect(find("input[name='document[update_type]'][value='minor']")).to be_checked
+  end
+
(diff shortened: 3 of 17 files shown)


db/migrate/20180906155513_add_change_notes_to_documents.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+class AddChangeNotesToDocuments < ActiveRecord::Migration[5.2]
+  def change
+    change_table :documents, bulk: true do |t|
```
kevindew: ooh I've not seen this `bulk` argument before.

spec/features/change_notes_spec.rb
```diff
@@ -0,0 +1,69 @@
...
+  end
+
+  def when_i_go_to_edit_the_document
+    visit edit_document_path(@document)
+  end
+
+  def and_i_fill_in_the_change_note_and_update_type
+    @update_request = stub_any_publishing_api_put_content
+
+    fill_in "document[change_note]", with: "Updated banana pricing"
```
kevindew: 😁

spec/features/change_notes_spec.rb
```diff
@@ -0,0 +1,69 @@
...
+  end
+
+  def then_the_summary_page_displays_change_note_and_update_type
+    expect(page).to have_content "Updated banana pricing"
+  end
+
+  def and_the_form_displays_the_change_note_and_update_type
+    visit edit_document_path(@document)
+    expect(page).to have_selector("textarea[name='document[change_note]']", text: "Updated banana pricing")
+    expect(page).to have_selector("input[name='document[update_type]'][checked='checked'][value='minor']")
```
kevindew: Might be a bit cleaner for these two to be something like `expect(find("document[change_note]")). to have_text` and similar

db/migrate/20180906155513_add_change_notes_to_documents.rb
tijmenb: Me neither! Rubocop forced me to do this instead of 2 `add_columns`.

spec/features/change_notes_spec.rb
tijmenb: 🍌

spec/features/change_notes_spec.rb
tijmenb: How would you do that for a radio button? This was the only way I found to see if a radio button is checked.

tijmenb:
@kevindew updated with behaviour to not allow update type change for the first edition.

spec/features/change_notes_spec.rb
kevindew: I may be totally wrong here because all I've done is google but I think you could do `expect(find("document[update_type]")).to be_checked` since that's a boolean we can use on a field: https://www.rubydoc.info/github/jnicklas/capybara/Capybara%2FNode%2FElement:checked%3F

db/migrate/20180906155513_add_change_notes_to_documents.rb
kevindew: He'll make us all redundant.

kevindew:
> Yep. I've added this to https://trello.com/c/K52YqwPK to the acceptance criteria. Can't be a nil column though, because we'll still allow users to save the page without a note (just not publish).

Oh yes good point

> Mwoah, I think having a named state is preferable in this case because it better matches the UI and mental model of users. Do you feel strongly about this?

No not too strongly, it's consistent with our other states - yet another enum situation :( . Might be prudent to have a similar validation in: https://github.com/alphagov/content-publisher/blob/master/app/models/document.rb#L24-L25

> @kevindew updated with behaviour to not allow update type change for the first edition.

Awesome, do we need this to then have a default one that it does on first edition here: https://github.com/alphagov/content-publisher/pull/236/files#diff-9ebabcf3d75c8d23c983f107afd49258R24 ?

spec/features/change_notes_spec.rb
tijmenb: Ah! Nice. Much better now.

tijmenb:
@kevindew the default for the first edition is done here:

https://github.com/alphagov/content-publisher/blob/54abab93688968ee23ba6d623b624a07c4ca96bf/app/controllers/new_document_controller.rb#L41-L50

And tested here:

https://github.com/alphagov/content-publisher/blob/54abab93688968ee23ba6d623b624a07c4ca96bf/spec/features/create_a_document_spec.rb#L38-L40

kevindew:
@tijmenb Ah sorry missed that - should that have a change note of "First published." too?

tijmenb:
@kevindew updated, it now has the correct defaults. Squashed the commits.

app/services/document_publishing_service.rb
```diff
@@ -11,8 +11,8 @@ def publish_draft(document)
 
   def publish(document, review_state)
     document.update!(publication_state: "sending_to_live", review_state: review_state)
-    publishing_api.publish(document.content_id, "major", locale: document.locale)
-    document.update!(publication_state: "sent_to_live")
+    publishing_api.publish(document.content_id, nil, locale: document.locale)
+    document.update!(publication_state: "sent_to_live", change_note: nil, update_type: "major", has_live_version_on_govuk: true)
```
benthorner: Could we use the new event history to figure this (has live version) out instead?

tijmenb: Not sure what you mean by "all the historical event JSON would be wrong"?

Btw, I intended this flag to be one-way, and have it not affected by unpublishing. Maybe I'll rename it to reflect that.

PR #238, opened 2018-09-07
Remember if the document was ever live on GOV.UK

tijmenb:
This is important because:

- We'll be able to discard the entire document if it's never been published
- We'll be able to show a link to live GOV.UK if it's ever been published
- We need to make send the "major" update type when a document has never been published (https://github.com/alphagov/content-publisher/pull/236#pullrequestreview-153325299)

https://trello.com/c/NEQWHpoU


benthorner:
We just scan the new history for this, right?

PR #243, opened 2018-09-10
Set bold labels for edit pages

alex-ju:
This PR:
 - sets bold labels for textarea fileds on edit document page
 - sets bold labels for autocomplete fileds on edit tags page

### Before
screen shot 2018-09-10 at 10 13 20

### After
screen shot 2018-09-10 at 10 12 57

### DIFF
--- a/app/views/components/_markdown_editor.html.erb
+++ b/app/views/components/_markdown_editor.html.erb
@@ -7,7 +7,7 @@
   <%= render "govuk_publishing_components/components/label", {
       text: label[:text],
       html_for: textarea[:id]
-  } %>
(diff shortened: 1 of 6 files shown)


app/views/components/_markdown_editor.html.erb
```diff
@@ -7,7 +7,7 @@
   <%= render "govuk_publishing_components/components/label", {
       text: label[:text],
```
benthorner: Would it work the same if we removed this line?

alex-ju: yes, good point. removed them.

PR #244, opened 2018-09-10
Index page

benthorner:
https://trello.com/c/l8MAts3s/205-build-the-index-page-with-filtering

This adds basic filtering and sorting to the index page, as well as trying to match the design:

   - filter by document type and slug/URL
   - sort by updated_at
   - show document title, document type, creator (eventually editor) and (last) updated_at

We'll do future PRs to add more filters and sorts, but these were easiest to start with. Issues:

   - The document type dropdown should be grouped by supertype, using optgroups
   - The document type dropdown should allow blanks (currently manually inject a blank)
   - The table is significantly different from the designs (needs a custom component)

The designs also suggest a more elaborate description of the results (e.g. "1 retired and removed documents by ..."). As we don't have many filters right now, we're going to keep it as just a count.


app/controllers/documents_controller.rb
```diff
@@ -39,6 +48,10 @@ def generate_path
 
 private
 
+  def index_params
```
tijmenb: This doesn't seem to be used?

app/controllers/documents_controller.rb
```diff
@@ -7,7 +7,16 @@ class DocumentsController < ApplicationController
...
   def index
-    @documents = Document.page(params[:page]).per(50)
+    filter = DocumentFilter.new(
+      filters: params.permit(:title_or_url, :document_type).to_hash,
+      sort: params[:sort],
+      page: params[:page],
+      per_page: 50,
+    )
+
+    @documents = filter.documents
```
tijmenb: Perhaps consider [passing in a single instance variable](https://robots.thoughtbot.com/sandi-metz-rules-for-developers) here.

app/views/documents/index.html.erb
```diff
@@ -1,39 +1,89 @@
...
-      end
-    } %>
+    <p class="govuk-body">
+      <%= t("documents.index.search_results.summary_html", count: @documents.total_count) %>
+    </p>
+
+    <% if @documents.none? %>
+      <%= govspeak_to_html t("documents.index.search_results.guidance") %>
+    <% else %>
+      <table class="govuk-table">
```
tijmenb: Consider extracting this into a partial for readability (less indentation & easier to understand the `if` statement).

app/views/documents/index/_filters.html.erb
```diff
@@ -0,0 +1,43 @@
...
+  <%= render "govuk_publishing_components/components/input", {
+    label: {
+      text: t("documents.index.filter.title_or_url")
+    },
+    name: "title_or_url",
+    value: params[:title_or_url],
+    type: "search"
+  } %>
+
+  <% document_type_select = SupertypeSchema.all.inject([]) do |memo, supertype|
```
tijmenb: Perhaps this can be extracted to the `DocumentFilter` class?

app/controllers/documents_controller.rb
benthorner: Good spot. I've combined this with the second comment and now use this method to generate the single params variable to pass to DocumentFilter. I see what you mean that passing a single variable makes sense, especially as we're extracting each one in the initializer already anyway.

tijmenb: 👍

app/services/document_filter.rb
```diff
@@ -0,0 +1,54 @@
...
+    SORT_KEYS.flat_map { |item| [item, "-#{item}"] }.include?(sort)
+  end
+
+  def filtered_scope(scope)
+    filters.inject(scope) do |memo, (field, value)|
+      next memo if value.blank?
+
+      case field
+      when :title_or_url
+        memo.where("title ILIKE ? OR base_path ILIKE ?", "%#{value}%", "%#{value}%")
```
tijmenb: Does `"%#{value}%"` introduce an SQLi vulnerability? I think you have to use [sanitize_sql_like](https://api.rubyonrails.org/classes/ActiveRecord/Sanitization/ClassMethods.html#method-i-sanitize_sql_like).

kevindew: Good point, I don't think it introduces a vulnerability as it uses a prepared statement, but using sanitise_sql_like will mean we don't get unexpected searches if a user makes uses of a % character.

app/views/documents/index.html.erb
benthorner: Done :-).

kevindew:
@benthorner FYI the [WIP] is still on https://github.com/alphagov/content-publisher/pull/244/commits/8b8ac3f96eeb6fc0ebeab3abb306efae520d4b66

app/services/document_filter.rb
benthorner: Good point - I've done this and added a test for it. I thought the '?' technique automatically sanitised the strings, but obviously not for %!

app/views/documents/index/_filters.html.erb
benthorner: As it's very view-specific - generating an array of arrays - I thought it would be best here, a la choose_supertype.html.erb.

tijmenb: Ok, good point. Could you golf it down to this?

```rb
DocumentTypeSchema.all.reject(&:managed_elsewhere).map { |d| [d.label, d.id] }
```

benthorner: Done - 'twas an unnecessary relic.

PR #248, opened 2018-09-10
Add lead image workflow

emmabeynon:
For https://trello.com/c/kwumHz3Q/189-set-up-a-minimal-lead-image-workflow-upload-and-select-xl
Paired with @davbo 

This adds a minimal lead image workflow in order for users to be able to perform basic tasks relating to a document's lead image:
- Uploading an image
- Adding and modifying metadata relating to a lead image
- Selecting an image from a list of existing images as a lead image
- Removing an image as a lead image

We also display the chosen lead image on the summary page for a document and display an error if a user chooses an invalid image to upload.

We expect some new components to be built after this PR has been merged, for a example for the image list, so there is some temporary code here which will need to be removed after these components have been created.

### DIFF
--- a/app/controllers/document_lead_image_controller.rb
+++ b/app/controllers/document_lead_image_controller.rb
@@ -1,0 +1,13 @@
+# frozen_string_literal: true
+
+class DocumentLeadImageController < ApplicationController
+  def index
+    @document = Document.find_by_param(params[:document_id])
+  end
+
+  def create
+  end
+
+  def edit
+  end
+end
--- a/app/controllers/document_lead_image_controller.rb
+++ b/app/controllers/document_lead_image_controller.rb
@@ -6,8 +6,39 @@ class DocumentLeadImageController < ApplicationController
   end
 
   def create
+    document = Document.find_by_param(params[:document_id])
+    upload = UploadedImageService.new(params.require(:image)).process
+
+    if upload.valid?
+      image = create_image_from_upload(upload, document)
+      redirect_to edit_document_lead_image_path(params[:document_id], image.id)
+    end
   end
 
   def edit
+    @document = Document.find_by_param(params[:document_id])
+    @image = Image.find_by(id: params[:image_id])
+  end
+
+private
+
+  def create_image_from_upload(upload, document)
+    blob = ActiveStorage::Blob.create_after_upload!(
+      io: upload.file,
+      filename: upload.filename,
+      content_type: upload.mime_type,
+    )
+
+    Image.create!(
+      document: document,
+      blob: blob,
+      filename: blob.filename,
+      width: upload.dimensions[:width],
+      height: upload.dimensions[:height],
+      crop_x: upload.crop_dimensions[:x],
+      crop_y: upload.crop_dimensions[:y],
+      crop_width: upload.crop_dimensions[:width],
+      crop_height: upload.crop_dimensions[:height],
+    )
   end
 end
--- a/app/controllers/document_lead_image_controller.rb
+++ b/app/controllers/document_lead_image_controller.rb
@@ -7,6 +7,11 @@ class DocumentLeadImageController < ApplicationController
 
   def create
     document = Document.find_by_param(params[:document_id])
+    unless params[:image]
+      redirect_to document_lead_image_path, alert: t("document_lead_image.index.no_file_selected")
+      return
+    end
+
     upload = UploadedImageService.new(params.require(:image)).process
 
     if upload.valid?
--- a/app/controllers/document_lead_image_controller.rb
+++ b/app/controllers/document_lead_image_controller.rb
@@ -34,6 +34,12 @@ class DocumentLeadImageController < ApplicationController
     redirect_to document_path(document)
   end
 
+  def destroy
+    document = Document.find_by_param(params[:document_id])
+    document.update(lead_image_id: nil)
+    redirect_to document_path(document)
+  end
+
 private
 
   def update_params
--- a/app/controllers/document_lead_image_controller.rb
+++ b/app/controllers/document_lead_image_controller.rb
@@ -20,7 +20,19 @@ class DocumentLeadImageController < ApplicationController
     @image = Image.find_by(id: params[:image_id])
   end
 
+  def update
+    document = Document.find_by_param(params[:document_id])
+    image = Image.find_by(id: params[:image_id])
+    image.update(update_params)
+    document.update(lead_image_id: image.id)
+    redirect_to document_path(document)
+  end
+
 private
+
+  def update_params
+    params.permit(:caption, :alt_text, :credit)
+  end
 
   def create_image_from_upload(upload, document)
     blob = ActiveStorage::Blob.create_after_upload!(
--- a/app/controllers/document_lead_image_controller.rb
+++ b/app/controllers/document_lead_image_controller.rb
@@ -12,6 +12,8 @@ class DocumentLeadImageController < ApplicationController
     if upload.valid?
       image = create_image_from_upload(upload, document)
       redirect_to edit_document_lead_image_path(params[:document_id], image.id)
+    else
+      redirect_to document_lead_image_path, alert: upload.errors
     end
   end
 
--- a/app/controllers/document_lead_image_controller.rb
+++ b/app/controllers/document_lead_image_controller.rb
@@ -27,6 +27,10 @@ class DocumentLeadImageController < ApplicationController
     image = Image.find_by(id: params[:image_id])
     image.update(update_params)
     document.update(lead_image_id: image.id)
+    if params[:next] == "lead-image"
+      redirect_to document_lead_image_path(document)
+      return
+    end
     redirect_to document_path(document)
   end
 
--- a/app/controllers/document_lead_image_controller.rb
+++ b/app/controllers/document_lead_image_controller.rb
@@ -18,7 +18,10 @@ class DocumentLeadImageController < ApplicationController
       image = create_image_from_upload(upload, document)
       redirect_to edit_document_lead_image_path(params[:document_id], image.id)
     else
-      redirect_to document_lead_image_path, alert: upload.errors
+      redirect_to document_lead_image_path, alert: {
+        "alerts" => upload.errors,
+        "title" => t("document_lead_image.index.error_summary_title"),
+      }
     end
   end
 
--- a/app/controllers/document_lead_image_controller.rb
+++ b/app/controllers/document_lead_image_controller.rb
@@ -35,10 +35,16 @@ class DocumentLeadImageController < ApplicationController
     image = Image.find_by(id: params[:image_id])
     image.update(update_params)
     document.update(lead_image_id: image.id)
-    if params[:next] == "lead-image"
+    if params[:next_screen] == "lead-image"
       redirect_to document_lead_image_path(document)
       return
     end
+    redirect_to document_path(document)
+  end
+
+  def choose_image
+    document = Document.find_by_param(params[:document_id])
+    document.update(lead_image_id: params[:image_id])
     redirect_to document_path(document)
   end
 
--- a/app/views/document_lead_image/_image_list.html.erb
+++ b/app/views/document_lead_image/_image_list.html.erb
@@ -1,0 +1,38 @@
+<table>
+  <% @document.images.each_with_index do |image, i| %>
+    <tr>
+      <td class="image-list">
+        <%= image_tag(url_for(image.thumbnail), id: "image-#{i}") %>
+      </td>
+      <td class="image-list" id=<%= "image-#{i}-metadata" %>>
+        <div class="image-list__metadata">
+          <strong><%= t("document_lead_image.index.alt_text") %></strong>: <%= image.alt_text %> <br \>
+          <strong><%= t("document_lead_image.index.caption") %></strong>: <%= image.caption %> <br \>
+          <strong><%= t("document_lead_image.index.credit") %></strong>: <%= image.credit %> <br \>
+          <div class="image-list__lead-image-flag">
+            <% if image == @document.lead_image %>
+              <div class="image-list__lead-image-label">Lead image</div>
+              <div class="image-list__remove-button">
+                <%= form_tag remove_document_lead_image_path(@document), method: :delete do %>
+                <button>Remove</button>
+                <% end %>
+              </div>
+            <% end %>
+          </div>
+        </div>
+      </td>
+    </tr>
+    <tr>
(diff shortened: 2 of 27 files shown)


benthorner:
Ein bug...

screen shot 2018-09-10 at 14 48 59

benthorner:
It doesn't seem to cope if I don't select an image...

screen shot 2018-09-10 at 14 51 05

benthorner:
The summary page looks a bit long now with those extra image attributes - do we have a design for this?

benthorner:
Also, clicking the remove button doesn't quite give the feedback I was expecting...

screen shot 2018-09-10 at 14 55 34

app/views/document_lead_image/edit.html.erb
```diff
@@ -0,0 +1,34 @@
+<%= content_for :back_link, document_path(@document) %>
```
benthorner: This back link should probably go to the previous page.

emmabeynon:
> The summary page looks a bit long now with those extra image attributes - do we have a design for this?

No, the design focused in on the lead image section specifically.  Perhaps one to raise with the team?

benthorner:
Probably just a quick chat with @soniaturcotte would be enough. But yeah, something to raise.

app/views/document_lead_image/_image_list.html.erb
```diff
@@ -0,0 +1,38 @@
+<table>
```
benthorner: Perhaps we could give @alex-ju a hand here and put this in a component, so at least the boilerplate is there.

app/views/document_lead_image/_image_list.html.erb
```diff
@@ -0,0 +1,38 @@
...
+        <div class="image-list__metadata">
+          <strong><%= t("document_lead_image.index.alt_text") %></strong>: <%= image.alt_text %> <br \>
+          <strong><%= t("document_lead_image.index.caption") %></strong>: <%= image.caption %> <br \>
+          <strong><%= t("document_lead_image.index.credit") %></strong>: <%= image.credit %> <br \>
+          <div class="image-list__lead-image-flag">
+            <% if image == @document.lead_image %>
+              <div class="image-list__lead-image-label">Lead image</div>
+              <div class="image-list__remove-button">
+                <%= form_tag remove_document_lead_image_path(@document), method: :delete do %>
+                <button>Remove</button>
```
benthorner: We could use a component here...

https://govuk-publishing-components.herokuapp.com/component-guide/button

app/views/document_lead_image/_image_list.html.erb
```diff
@@ -0,0 +1,38 @@
+<table>
+  <% @document.images.each_with_index do |image, i| %>
+    <tr>
+      <td class="image-list">
+        <%= image_tag(url_for(image.thumbnail), id: "image-#{i}") %>
```
benthorner: Could we use a component here instead?

https://govuk-publishing-components.herokuapp.com/component-guide/image_card

benthorner: I'm thinking the whole component could be inside a form to select it as the lead image, with a special 'no lead image' one above the rest, to replace the remove button - it's a bit weird having that on all of the images.

app/controllers/document_lead_image_controller.rb
```diff
@@ -0,0 +1,67 @@
...
+    @document = Document.find_by_param(params[:document_id])
+    @image = Image.find_by(id: params[:image_id])
+  end
+
+  def update
+    document = Document.find_by_param(params[:document_id])
+    image = Image.find_by(id: params[:image_id])
+    image.update(update_params)
+    document.update(lead_image_id: image.id)
+    if params[:next] == "lead-image"
```
benthorner: This is a bit strange - why is the key called ':next'?

app/controllers/document_lead_image_controller.rb
```diff
@@ -0,0 +1,67 @@
...
+      return redirect_to document_lead_image_path(document)
```
benthorner: Little one: we're not actually returning the redirect, so it should really be on a separate line.

spec/features/choose_lead_image_spec.rb
```diff
@@ -0,0 +1,65 @@
+# frozen_string_literal: true
+
+RSpec.feature "Choose a lead image" do
+  scenario "User chooses an existing image as lead image" do
+    given_there_is_a_document_with_existing_images
+    when_i_visit_the_summary_page
+    when_i_visit_the_lead_images_page
```
benthorner: Another minor one: I think this could be an 'and'.

spec/features/choose_lead_image_spec.rb
```diff
@@ -0,0 +1,65 @@
...
+    then_i_should_be_able_to_see_a_list_of_existing_images
+    when_i_remove_the_lead_image
```
benthorner: This looks suspiciously like a separate scenario - can we make it one instead?

spec/features/choose_lead_image_spec.rb
```diff
@@ -0,0 +1,65 @@
...
+    when_i_visit_the_lead_images_page
+    and_i_select_an_image_to_be_the_lead_image
+    then_i_should_see_the_new_lead_image_on_the_summary_page
+  end
+
+  def given_there_is_a_document_with_existing_images
+    document_type_schema = build(:document_type_schema, lead_image: true)
+    document = create(:document, document_type: document_type_schema.id)
+    @image1 = create(:image, document: document, filename: "image-1.jpg",
+                              alt_text: "image 1 alt text", caption: "image 1 caption",
```
benthorner: Could the factory do this for us?

spec/features/choose_lead_image_spec.rb
```diff
@@ -0,0 +1,65 @@
...
+    visit document_path(Document.last)
+  end
+
+  def when_i_visit_the_lead_images_page
+    click_on "Change Lead image"
+  end
+
+  def then_i_should_be_able_to_see_a_list_of_existing_images
+    expect(find("#image-0")["src"]).to include("image-1.jpg")
+    within("td#image-0-metadata") do
```
benthorner: Do we need a within (and therefore the ids) here?

spec/features/choose_lead_image_spec.rb
```diff
@@ -0,0 +1,65 @@
...
+      expect(page).to have_content(@image1.alt_text)
+      expect(page).to have_content(@image1.caption)
+      expect(page).to have_content(@image1.credit)
+      expect(page).to have_content("Lead image")
```
benthorner: Ah, I see, it's so we can check it's the lead one. Perhaps this could be a class we could check on the image itself, so we can see it's selected.

spec/features/upload_invalid_lead_image_spec.rb
```diff
@@ -0,0 +1,33 @@
+# frozen_string_literal: true
+
+RSpec.feature "Edit a lead image" do
+  scenario "User edits a lead image" do
+    given_there_is_a_document
+    when_i_visit_the_summary_page
+    when_i_visit_the_lead_images_page
```
benthorner: Another small one: could be an 'and' here.

spec/features/edit_lead_image_spec.rb
```diff
@@ -0,0 +1,40 @@
+# frozen_string_literal: true
+
+RSpec.feature "Edit an existing lead image" do
+  scenario "User edits lead image" do
+    given_there_is_a_document_with_existing_images
+    when_i_visit_the_lead_images_page
+    then_i_click_edit_details
```
benthorner: This should be an 'and' - 'then' is for expectations.

spec/features/edit_lead_image_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+    then_i_edit_the_image_metadata
+    then_i_am_redirected_to_the_lead_images_page
```
benthorner: This step is doing a couple of things - perhaps it should be split in two.

spec/features/upload_lead_image_spec.rb
```diff
@@ -0,0 +1,55 @@
+# frozen_string_literal: true
+
+RSpec.feature "Upload a lead image" do
+  scenario "User uploads a lead image" do
+    given_there_is_a_document
+    when_i_visit_the_summary_page
+    then_i_see_there_is_no_lead_image
+    when_i_visit_the_lead_images_page
+    then_i_should_see_no_images_available
+    and_i_upload_a_new_image
```
benthorner: This should be a 'when'.

spec/features/upload_lead_image_spec.rb
```diff
@@ -0,0 +1,55 @@
...
+    and_i_fill_in_the_metadata
+    then_i_should_be_able_to_see_the_image
```
benthorner: We're implicitly checking it's the lead image - perhaps the step should express that too :-).

app/views/layouts/application.html.erb
```diff
@@ -40,9 +40,16 @@
     <% end %>
 
     <% if flash[:alert] %>
-      <%= render "components/error_alert", {
-        message: flash[:alert]
-      } %>
+      <% if flash[:alert].is_a? Enumerable %>
+        <%= render "govuk_publishing_components/components/error_summary", {
+          title: "TODO: Update gem to not require title",
```
tijmenb: According to https://trello.com/c/H09wIghW, we should [add the title "Something has gone wrong"](https://docs.google.com/document/d/15euIDtRWAClra8y4VX1ZMSnXqS3SN24fJof6EcGRdWQ/edit#heading=h.u0b4qt5inox0).

app/views/document_lead_image/_image_list.html.erb
davbo: Not sure I understand how the form is submitted without a button?

app/views/document_lead_image/_image_list.html.erb
davbo: We decided to not use a component until the "link-button" (link which behaves like a button) is created.

benthorner: Cool 🇹🇭 .

app/views/document_lead_image/_image_list.html.erb
benthorner: Oh yeah, that's true. Perhaps not then. We could still put it alongside and use the component - I was thinking it would get us to preview earlier. But you're right it's not part of this PR.

benthorner: That said, it seems silly to right a whole new component if we could just tweak the design a bit. Perhaps @soniaturcotte (hello again ;-) could comment on this?

spec/features/choose_lead_image_spec.rb
emmabeynon: Fair point 👌

spec/features/choose_lead_image_spec.rb
emmabeynon: Metadata isn't required for an image so I'd rather pass in the values when they're needed than set them as a default in the factory.

emmabeynon:
Your comments should be address now @benthorner

app/controllers/document_lead_image_controller.rb
benthorner: Ah, this is quite confusing/unusual. I think this should be split into two actions, especially as one of the ways this is called doesn't actually update the image attributes.

app/views/document_lead_image/edit.html.erb
```diff
@@ -0,0 +1,34 @@
...
+      } %>
+      <%= render "govuk_publishing_components/components/input", {
+        label: {
+          text: t("document_lead_image.edit.form_labels.credit")
+        },
+        name: "credit",
+        value: @image.credit,
+      } %>
+      <%= render "govuk_publishing_components/components/button", {
+        text: "Save image"
```
benthorner: Could this be 'Save and choose image', as it's actually doing two things?

spec/features/choose_lead_image_spec.rb
benthorner: Ah, yes, good point!

spec/features/choose_lead_image_spec.rb
benthorner: I still think this should be a class, as I doubt we're going to be showing the text 'Lead image' to the user.

spec/features/edit_lead_image_spec.rb
```diff
@@ -0,0 +1,44 @@
+# frozen_string_literal: true
+
+RSpec.feature "Edit an existing lead image" do
+  scenario "User edits lead image" do
+    given_there_is_a_document_with_existing_images
+    when_i_visit_the_lead_images_page
+    and_i_click_edit_details
+    then_i_edit_the_image_metadata
+    then_i_am_redirected_to_the_lead_images_page
+    then_i_should_be_able_to_see_the_edited_metadata
```
benthorner: This one should be an 'and'.

spec/features/upload_lead_image_spec.rb
```diff
@@ -0,0 +1,65 @@
+# frozen_string_literal: true
+
+RSpec.feature "Upload a lead image" do
+  scenario "User uploads a lead image" do
+    given_there_is_a_document
+    when_i_visit_the_summary_page
+    then_i_see_there_is_no_lead_image
+    when_i_visit_the_lead_images_page
+    then_i_should_see_no_images_available
+    when_i_upload_no_file
```
benthorner: This looks a bit like a separate scenario. Could we split it out in here?

spec/features/choose_lead_image_spec.rb
emmabeynon: We are showing "Lead image" text next to the image which has been selected as the lead image, so I feel this is a valid thing to test for.

app/views/document_lead_image/edit.html.erb
emmabeynon: This has made me realise that a user should be able to edit the metadata of an image without selecting it as the lead image, so I will do some work to solve this.

app/controllers/document_lead_image_controller.rb
davbo: it's a difference between if you're editing the metadata for an existing image (returns to lead-image page) or you're uploading a new lead image (returns to document summary).

spec/features/choose_lead_image_spec.rb
benthorner: Discussed. Should become clearer as we iterate the workflow/frontend.

PR #249, opened 2018-09-10
Add translations ADR to document our approach

benthorner:
https://trello.com/c/J55uTleS/242-adr-for-application-microcopy

One of the files I wanted to use as an example wasn't quite consistent
with the approach taken elsewhere, so I've corrected it in this commit.


docs/adr/0004-translations.md
```diff
@@ -0,0 +1,53 @@
...
+
+Pending
+
+The approach taken so far is primarily driven by the needs of developers, rather than the needs of the other roles in the team. We should avoid committing to this approach until other people in the team have made more use of it.
+
+## Consequences
+
+Using translations for the most of the strings has allowed members of the team who aren't developers to make changes without a developer or a development environment. Instead, we trained them to use the Github commit and PR interface.
+
+Some people found it hard to find extracted strings, due to them being in multiple, technology-specific locations. It's still unclear how we should structure the extracted strings in a way that makes it easy to find a given string. One way to ignore this problem is to use search (e.g. Github search) to locate the file where the string is stored; this doesn't always work because some strings are only shown under certain conditions (e.g. errors) and therefore may not be widely known.
```
kevindew: Sounds good - I think this is one of our trickier problems. I'm intrigued to learn if content designers get baffled by the controller/action layout but also short of better ideas. 

Only thing I've got is that content-designers might find it easier dealing with a long file with lots in rather than a short one so could jump from controller/action to just a controller.yml - but not yet a known problem or whether that would help.

docs/adr/0004-translations.md
```diff
@@ -0,0 +1,53 @@
...
+   * **Big guidance** is extracted into it's own Markdown files and stored alongside the corresponding HTML page that shows it. For example, the guidance for creating a new document is stored in `app/views/new_document/guidance.md`.
+   * **Domain data** that's static is extracted into custom YAML files. This application has two static models (DocumentTypeSchema and SupertypeSchema) that encapsulate domain concepts where the data is part of the application. We think domain data - whether it's a backend setting or a label - should be defined in one place so it can change together. In this application, we've chosen to store the backing files for the two static models in `app/formats`.
+   * **Global strings** (states and validation messages) are extracted using translations. As these strings aren't page-specific, we put them at the top-level of the translation hierarchy (in `states.yml` and `validations.yml`).
+   * **All other strings** are extracted using translations, in a hierarchy that follows the structure of the `app/views` directory. For example, the above example relates to `app/views/publish_document/published.html.erb`.
```
kevindew: Are we missing a section here on controller/action.yml setup for translations?

docs/adr/0004-translations.md
```diff
@@ -0,0 +1,53 @@
...
+
+```
+
+Rails translations have a few special behaviours, such as pluralization, raw HTML, and variables. The `%{title}` string in the above is an example of a variable, which a developer will set to the title of the document being published.
+
+## Decision
+
+Although we could use translations to extract all of the strings in the application, in some cases we felt this wasn't necessary, or that a different method should be used. The following is a summary of the rules we currently use.
+
+   * **Link and button labels** are not extracted. We think link and button labels are unlikely to change, and extracting them made the application tests harder to read by obfuscating some of the crucial steps in the test with translation keys.
```
kevindew: I think this implies we'd have to test against a translation which may not be unanimous. I'm making a related comment about a testing section below though where I'll expand on this.

docs/adr/0004-translations.md
```diff
@@ -0,0 +1,53 @@
...
+   * **Global strings** (states and validation messages) are extracted using translations. As these strings aren't page-specific, we put them at the top-level of the translation hierarchy (in `states.yml` and `validations.yml`).
+   * **All other strings** are extracted using translations, in a hierarchy that follows the structure of the `app/views` directory. For example, the above example relates to `app/views/publish_document/published.html.erb`.
+
+Every instance of a string in the tests has been replaced according to the above rules, such that the tests continue to pass when an extracted string is changed. **Link and button labels** are not replaced, as they are not extracted in the code.
+
+We also configured Rails to raise an exception when we hit a page where a translation is missing, as we don't test all of the translations; this is done by setting `config.action_view.raise_on_missing_translations = true` in `application.rb`.
+
```
kevindew: Might be an idea to add a section on testing and the various options that were considered. As that seems to be the controversial area and probably the bit most likely to trip up anyone new to the project. 

As I recall the options we had were:

- Register every phrase in i18n files and use only keys to those in our tests:
  - Pros: No doubts of what to put in locale config, changes to phrases can't change test outcome
  - Cons: Feature tests are very difficult to read
- Register every phrase in i18n files and use a mix of text and translation look ups in tests
  - Pros:  No doubts of what to put in locale config, Feature tests can be easy to read
  - Cons: Content designers can break tests with a content change, Unclear what text should be free text and what should be a translation look up in test.
- Register only phrases we expect a content designer to edit
  - Pros:  Feature tests can be easy to read, changes to phrases can't change test outcome
  - Cons: Inconsistency between what's in locale files and in views for a particular component

benthorner: Good point. I've added a section about this, trying to avoid my natural bias ;-).

docs/adr/0004-translations.md
```diff
@@ -0,0 +1,63 @@
...
+After trying the first option, we found that extracting links and button labels had the greatest impact on readability; we also think these kinds of strings are unlikely to change. As the original intention behind this work was to enable non-developers to seamlessly edit content, we adopted the third option, which appears to be working well for people making changes, but has caused some confusion among the developers as the subset of strings to extract wasn't specified as it is below.
+
+## Decision
+
+Although we could use translations to extract all of the strings in the application, in some cases we felt this wasn't necessary, or that a different method should be used. The following is a summary of the rules we currently use.
+
+   * **Link and button labels** are not extracted. We think link and button labels are unlikely to change, and extracting them made the application tests harder to read by obfuscating some of the crucial steps in the test with translation keys.
```
tijmenb: The approach of not translating clickable things still bugs me. For two reasons:

1. Having a bit translated and other bits untranslated looks to me like I forgot to put something in the locale files:
  
  https://github.com/alphagov/content-publisher/blob/52eae1b87cc7f258111ecacf83536d20c3dece5c/app/views/documents/show/_actions.html.erb#L4-L8

2. I definitely recognise the readability concern in the tests. But what I don't understand is that we do accept this readability drop when it comes to other translated strings. If we were to follow our logic, why are we still translating those strings?

https://github.com/alphagov/content-publisher/blob/52eae1b87cc7f258111ecacf83536d20c3dece5c/spec/features/choose_a_format_spec.rb#L12-L33 

Options in order of my preference:

1. Translate everything consistently
2. Translate the buttons too, but keep the raw strings in the tests
3. Somebody bribe me with cookies so I'll change my mind

tijmenb: I've put up a PR for an alternative solution here: https://github.com/alphagov/content-publisher/pull/254/files. Perhaps that could satisfy all parties involved?

benthorner: I'm actually happy with translating everything, and I think @kevindew didn't mind much either way. At the time @davbo was quite concerned about it, though - maybe we should revisit.

@tijmenb your PR reminds me of a concept I once used called [Page Objects](https://robots.thoughtbot.com/better-acceptance-tests-with-page-objects). I think we should revisit using translations directly in tests first, but if it's still a concern, perhaps we could look at doing something like this.

docs/adr/0004-translations.md
benthorner: I've added a testing section now, so this shouldn't be so surprising.

benthorner:
OK. Thanks for the comments everyone! A reminder that this is a 'pending' ADR, so it's definitely something we can continue discussing and tweaking.

PR #250, opened 2018-09-10
Deal consistently with errors sending to draft

tijmenb:
When a document is saved, we attempt to send it to the publishing-api as draft. This goes wrong sometimes, either because the publishing-api isn't available (in development), the publishing-api is slow to respond
(when it's busy) or when we're accidentally sending an invalid payload to the publishing-api (this should not be possible, but it's likely that we'll make mistakes in this).

This commit adds a "retry" button to the page when the draft publishing hasn't succeeded. This isn't a particularly elegant solution (as in, requiring fewer lines of code), but I can't think of anything better at the moment.

It's intent is to communicate to the user that something went wrong, and gives an option to fix the problem.

https://trello.com/c/VkVrVGeB

### DIFF
--- a/spec/features/create_a_document_api_down_spec.rb
+++ b/spec/features/create_a_document_api_down_spec.rb
@@ -7,6 +7,9 @@ RSpec.feature "Create a document when the API is down" do
     when_i_submit_the_form
     then_i_see_the_document_exists
     and_the_preview_creation_failed
+
+    when_the_api_is_up_again_and_i_click_the_retry_button
+    then_the_document_is_saved_again
   end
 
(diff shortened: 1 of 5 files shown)


spec/features/create_a_document_api_down_spec.rb
```diff
@@ -7,6 +7,9 @@
     when_i_submit_the_form
     then_i_see_the_document_exists
     and_the_preview_creation_failed
+
+    when_the_api_is_up_again_and_i_click_the_retry_button
```
benthorner: This must be our amazing 2ndline team :-).

PR #251, opened 2018-09-10
Use `update!` to save documents

tijmenb:
This will raise an error is the document is invalid, instead of silently swallowing it. `Model#update` is only to be used if we're using the return value to check if the update was successful or not.


benthorner:
Good spot, though 🇹🇭.

tijmenb:
@benthorner we're not doing validation of things that would be fixable for users here, so if this fails it's a developer error and we'll be notified by Sentry.

PR #256, opened 2018-09-11
Update to change base path label

benhazell:
Change Base Path label to 'Slug'

### DIFF
--- a/config/locales/en/documents/show.yml
+++ b/config/locales/en/documents/show.yml
@@ -14,7 +14,7 @@ en:
         title: Content
         items:
           title: Title
-          base_path: Base path
+          base_path: Slug
           summary: Summary
           update_type: Update type
           change_note: Change note


kevindew:
@benhazell I'm not sure "Slug" is the best terminology to use here as the most [common definition](https://en.wikipedia.org/wiki/Clean_URL#Slug) for it is just the page identifiable part of a URL.

So we'll be showing something like `/government/news/420-million-to-support-local-infrastructure` whereas traditionally a slug (and as Whitehall uses them) refers to just the `420-million-to-support-local-infrastructure`

I don't think we should go with base path though for users, but could consider options like GOV.UK Path or just URL ?

benhazell:
@kevindew The change from Base Path to Slug was part of a demo of GitHub YAML editing BenT and I just did. This isn't a real change - we can reject it.
I wrote slug more as a joke for the content community as everyone's always found it a silly word.

kevindew:
@benhazell d'oh! sorry

benthorner:
@kevindew we can close this. Although I notice that CI doesn't seem to have picked this up.

@benhazell can you try doing another change on a brand new branch name to see if it's something to do with the branch ('patch-2') you chose?

kevindew:
@benthorner CI won't pick this up as it's a fork of this repo. To make changes directly in this repo Ben needs to be part of the GOV.UK Content Designers Github group - there is a pending invite for him to join that.

PR #257, opened 2018-09-11
Refactor url preview

alex-ju:
This PR:
- extracts the URL preview section into its own component
- refactor the code to be consistent with the other components

Doesn't include:
- `Fetch` polyfill
- `URL`/`searchParams` polyfill

[Trello card](https://trello.com/c/9saZuUXt)


alex-ju:
@kevindew Given the current form of this code we'll need to bring in polyfills for both `Fetch` and `URL.searchParams` (updated this PR note to reflect that). I would create a separate card to go over all our JavaScript and ensure cross-browser support, if that's alright with you.

kevindew:
@alex-ju Yeah no problem 👍

PR #258, opened 2018-09-11
Handle unknown document creator on index page

kevindew:
This is to fix the index page currently throwing an exception when the
user who created the document does not exist.

I felt the key unknown_data was a bit abstract so went for a more
explicit unknown_creator

### DIFF
--- a/app/views/documents/index/_results.html.erb
+++ b/app/views/documents/index/_results.html.erb
@@ -31,7 +31,9 @@
           <br/>
           <%= document.document_type_schema.label %>
           <br/>
-          <%= t "documents.index.search_results.headings.creator", creator: document.creator.name %>
+          <%= t "documents.index.search_results.headings.creator",
+            creator: document.creator&.name || I18n.t("documents.index.search_results.unknown_creator")
+          %>
         </td>
 
         <td class="govuk-table__cell ">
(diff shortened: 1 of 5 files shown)


benthorner:
Although I think we ought to have a test for this somewhere...

benthorner:
@kevindew maybe one day I'll tell you ;-)

I think a quick feature test for this particular scenario would be fine. There's perhaps some development pain we're approaching where we have too many features!

tijmenb:
@kevindew perhaps remove this line from the factory:

https://github.com/alphagov/content-publisher/blob/ff2a6c0bfa3196d70abb33e68c4cb423555699ee/spec/factories/document_factory.rb#L12

Attributes that aren't required by database constraints/model validation are better not included in factories, since they can lead to relying on things being present like this.

benthorner:
That would be a quick way and consistent with our previous approach - we just have to be strict with ourselves to avoid it creeping in in future 🇭🇺 .

kevindew:
@benthorner yeah - I see we have a show a document without creator test which could include it as well. We may well have to ponder how to organise the feature tests.

@tijmenb that sounds a good idea re removing that (and saves a query). I also felt a bit uneasy about the duplicated translation. I saw we briefly flirted with a global.yml file in translations was there a discussion about whether or not that was a bad idea?

benthorner:
I think the global.yml discussion depends on the example. In this case, there's no reason why the strings on the index/show pages need to be the same, so it would be premature to make them refer to the same thing.

PR #259, opened 2018-09-11
Revert "Manually turn on brakeman"

tijmenb:
Reverts alphagov/content-publisher#247, which should not be necessary anymore if https://github.com/alphagov/govuk-jenkinslib/pull/29 works correctly.


tijmenb:
This works, as evidenced by `continuous-integration/jenkins/security` being present in the checks.

PR #260, opened 2018-09-11
Unknown user by default in tests

kevindew:
Follow up to https://github.com/alphagov/content-publisher/pull/258

This doesn't set a creator by default when we build document models via FactoryBot

It creates two new feature tests which test the expectations of a document with a creator and a document without one.


spec/features/viewing_a_document_with_no_creator_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+  scenario "User finds document without a creator and views it" do
+    given_there_is_a_document_with_no_creator
+    when_i_visit_the_index_page
+    then_i_see_we_dont_know_who_created_the_document
+    when_i_visit_the_document_page
+    then_i_again_see_we_dont_know_who_created_the_document
+  end
+
+  def given_there_is_a_document_with_no_creator
+    @document = create(:document, title: "Excellent document", creator: nil)
```
benthorner: Is this title important here?

benthorner:
This looks good, although I think there's another piece of potential pain that the feature tests are becoming harder to find as the names diverge.

spec/features/viewing_a_document_with_no_creator_spec.rb
kevindew: It is not, thanks - I've simplified in c82f825

kevindew:
> This looks good, although I think there's another piece of potential pain that the feature tests are becoming harder to find as the names diverge.

Yup I think working out a criteria to to group the feature tests is a looming challenge. Seems tricky as there's not a 1:1 mapping with them and an application concept.

PR #261, opened 2018-09-11
Deal consistently with errors publishing

tijmenb:
This is a continuation of https://github.com/alphagov/content-publisher/pull/250.

Something publishing to the publishing-api goes wrong, either because the publishing-api isn't available (in development), the publishing-api is slow to respond (when it's busy).

This commit adds a message and button to the summary page when something has gone wrong, so the user can fix the issue themselves.

screen shot 2018-09-11 at 14 49 46

https://trello.com/c/VkVrVGeB

### DIFF
--- a/app/services/user_facing_state.rb
+++ b/app/services/user_facing_state.rb
@@ -10,7 +10,7 @@ class UserFacingState
     case state.to_sym
     when :draft
       query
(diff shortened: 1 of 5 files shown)


benthorner:
Should we still show the other actions for this state here? It's a bit weird that you're 'not allowed' to delete your draft at this point.

tijmenb:
@benthorner yeah, perhaps. Left those out because they don't do anything, we'll think about that when implementing those features.

PR #262, opened 2018-09-11
Support gov responses and world news stories

benthorner:
https://trello.com/c/KhxwyMKK/243-fix-not-supporting-government-response-and-world-new-story

This commit adds two new 'edit only' document types, for which documents
can only be edited and not created - they are hidden in the selection.

Format specs for this kind of document are slightly different, as we
need to simulate importing them, rather than creating new ones.

### DIFF
--- a/app/formats/document_types.yml
+++ b/app/formats/document_types.yml
@@ -108,6 +108,110 @@
       type: multi_tag
       document_type: worldwide_organisation
 
+- id: government_response
+  label: Government response
+  edit_only: true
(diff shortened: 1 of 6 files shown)


benthorner:
Already ahead of you on that one...https://github.com/alphagov/content-publisher/pull/263.

@huminmiah was keen on this, though, and it only took half an hour.

benthorner:
Closing in favour of https://github.com/alphagov/content-publisher/pull/263.

PR #263, opened 2018-09-11
Filter out unwanted new document types

benthorner:
This is an alternative to https://trello.com/c/KhxwyMKK/243-fix-not-supporting-government-response-and-world-new-story.

We currently don't support 'government response' and 'world news story'
document types, so we should avoid importing them until the migration
path is clear.

### DIFF
--- a/lib/tasks/import.rake
+++ b/lib/tasks/import.rake
@@ -11,6 +11,9 @@ namespace :import do
     input = ENV["INPUT"] ? File.open(ENV["INPUT"]) : STDIN
 
     input.each_line do |line|
+      next if line.include?("government_response")
+      next if line.include?("world_news_story")
+
       importer.import(JSON.parse(line))
       imported += 1
     end


app/formats/document_types.yml
```diff
@@ -129,6 +129,13 @@
     hostname: whitehall-admin
     path: /government/admin/news/new
 
+- id: government_response
+  label: Government response
+  description: Government statements in response to media coverage, such as rebuttals and ‘myth busters’
```
tijmenb: Where does this come from? It's not in @benhazell's [document types document](https://docs.google.com/document/d/1ivZ34_SBeIRllwi31MVewyxudE7F0QnW3U3p5QoHZ6E/edit).

lib/tasks/whitehall_news_importer.rb
```diff
@@ -31,6 +31,8 @@ def create_or_update_document(translation, edition, document)
         review_state: "unreviewed",
         summary: translation["summary"],
         tags: tags(edition),
+        has_live_version_on_govuk: edition["state"] == "published",
+        update_type: "major",
```
kevindew: Does this do any filtering? I was expecting a next if doc_type == "government_response" or similar but not sure if I've missed something.

app/formats/document_types.yml
benthorner: This comes directly from Whitehall - if you create a new 'News' document, this is one of the options and therefore one of the kinds of document we're importing under 'NewsArticle'.

I can split this into a separate PR if you want - just trying to bung all the fixes in one.

lib/tasks/whitehall_news_importer.rb
benthorner: This was one of the other (smaller) fixes I found while investigating the problem. It'll make sure we're showing change notes consistently with how they appear in Whitehall. Again, happy to split into a separate PR if you want :-).

app/formats/document_types.yml
tijmenb: Ben's doc is the authoritative design of these pages, so if something is not on there you can assume it's for good reason. I wouldn't add this, but you could talk to Ben if you this there's been a mistake.

lib/tasks/whitehall_news_importer.rb
benthorner: But you're right, somehow the main filtering change got lost from the commit - it's in there now (import.rake).

lib/tasks/whitehall_news_importer.rb
```diff
@@ -31,6 +31,8 @@ def create_or_update_document(translation, edition, document)
         review_state: "unreviewed",
         summary: translation["summary"],
         tags: tags(edition),
+        has_live_version_on_govuk: edition["state"] == "published",
```
tijmenb: Is this correct? `has_live_version_on_govuk` should be true if there was _ever_ a live version of the document, not just if the edition was published.

app/formats/document_types.yml
benhazell: We're not supporting the creation of government response document type. 
I think @huminmiah wanted it cut based on usage data and @paola-roccuzzo  research.

lib/tasks/whitehall_news_importer.rb
benthorner: Yeah, I get that - I was trying to replicate the way we're setting it in the app at the moment. But I'm happy to write a separate card to backfill doing this on the import.

app/formats/document_types.yml
benthorner: No probs - cut it.

benthorner:
@tijmenb simplified to just the essentials.

PR #264, opened 2018-09-11
Rationalise page titles for document editing

tijmenb:
This makes the titles for the edit pages consistent.

They use the form "{thing} `{page title}`".


config/locales/en/globals.yml
```diff
@@ -0,0 +1,3 @@
+en:
```
benthorner: I'm keen to avoid a file called 'globals', as it could contain absolutely anything. We can take the approach for 'states' and 'validations' and call it 'documents'.

PR #265, opened 2018-09-11
Index filter by state

kevindew:
Trello: https://trello.com/c/rCESaDMh/240-filter-index-page-by-state

This adds in the state filter for the index page.

It does have a bit of a WTF going on for handling the scenario where there is a state passed in that is unknown to us: https://github.com/alphagov/content-publisher/blob/25c6b5d786600a50baf2e0096c67a1152357ded8/app/services/document_filter.rb#L49-L51

Alternative options if people dislike this approach are:
1 - moving the where("1 = 2") into [UserFacingState](https://github.com/alphagov/content-publisher/blob/master/app/services/user_facing_state.rb#L23) instead of exception
2 - be inconsistent with other query string filters and return all results

### DIFF
--- a/app/controllers/documents_controller.rb
+++ b/app/controllers/documents_controller.rb
@@ -50,7 +50,7 @@ private
 
   def filter_params
     {
-      filters: params.permit(:title_or_url, :document_type).to_hash,
+      filters: params.permit(:title_or_url, :document_type, :state).to_hash,
       sort: params[:sort],
       page: params[:page],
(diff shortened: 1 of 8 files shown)


benthorner:
Would `scope.none` be more obvious?

https://stackoverflow.com/questions/4877931/how-to-return-an-empty-activerecord-relation

kevindew:
@benthorner yes, had I known of that I'd have used it. Will update.

PR #269, opened 2018-09-12
Fix failing tests due to bad translation key

benthorner:


benthorner:
It did raise errors - the test failed. May have been caused by a branch not re-running tests after a non-conflicting change on master.

tijmenb:
@benthorner which PR introduced this change?

benthorner:
I think it was a combination of a5aef16b and cf8cc160.

kevindew:
I didn't know about `I18n.t!` raising errors that does seem a good option.

kevindew:
Looks like we could also do it through a config variable `config.action_view.raise_on_missing_translations = true` https://stackoverflow.com/questions/8066901/rails-how-to-raise-i18n-translation-is-missing-exceptions-in-the-testing-enviro

benthorner:
We already have this in `config/application.rb` - as now documented in the ADR ;-).

PR #270, opened 2018-09-12
Create MVP for history feature

tijmenb:
This is a first PR for the "history" feature. It's very rudimentary:

Whenever a relevant action takes place, we create a `TimelineEntry` record. I've called this a "timeline entry", because I couldn't come up with a better generic name. We expect that this page will also include comments, so "timeline entry" seems appropriate for that too. We display the timeline entry in the history tab:

screen shot 2018-09-12 at 12 47 49

## How to review

Note that this doesn't implement all of the things from the Trello ticket, there will be follow up PRs (see "To do" checklist).

- Does the naming of "timeline entry" make sense?
- Have we captured entries for all the relevant user-initiated actions at the moment?

https://trello.com/c/TVcThaKi

### DIFF
--- a/app/controllers/documents_controller.rb
+++ b/app/controllers/documents_controller.rb
@@ -24,6 +24,7 @@ class DocumentsController < ApplicationController
   def update
     document = Document.find_by_param(params[:id])
     document.update!(update_params(document))
+    TimelineEntry.create!(document: document, user: current_user, entry_type: "updated_content")
     DocumentPublishingService.new.publish_draft(document)
     redirect_to document, notice: t("documents.show.flashes.draft_success")
   rescue GdsApi::BaseError
--- a/app/views/documents/show/_history_tab.html.erb
+++ b/app/views/documents/show/_history_tab.html.erb
@@ -1,1 +1,11 @@
-Document history is not available right now.
+<% @document.timeline_entries.includes(:user).order(created_at: :desc).each do |entry| %>
+  <div class="timeline-entry govuk-!-margin-bottom-6">
+    <h4 class="govuk-heading-s govuk-!-margin-bottom-1">
+      <%= t "documents.history.entry_types.#{entry.entry_type}" %>
+    </h4>
+
+    <div class="govuk-body">
+      <%= entry.created_at.strftime("%l:%M%P on %d %B %Y") %> by <%= entry.username_or_unknown %>
+    </div>
+  </div>
+<% end %>
--- a/db/migrate/20180907121447_create_timeline_entries.rb
+++ b/db/migrate/20180907121447_create_timeline_entries.rb
@@ -1,0 +1,17 @@
+# frozen_string_literal: true
+
+class CreateTimelineEntries < ActiveRecord::Migration[5.2]
+  def change
+    create_table :timeline_entries do |t|
+      t.string :entry_type, null: false
+
+      # Delete all timeline_entries if a document is deleted
+      t.references :document, foreign_key: { on_delete: :cascade }, null: false
+
+      # Don't allow users to be deleted if they're related to a timeline entry
+      t.references :user, foreign_key: { on_delete: :restrict }
+
+      t.timestamps
+    end
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -79,6 +79,16 @@ ActiveRecord::Schema.define(version: 2018_09_07_121447) do
     t.index ["document_id"], name: "index_images_on_document_id"
   end
 
+  create_table "timeline_entries", force: :cascade do |t|
+    t.string "entry_type", null: false
+    t.bigint "document_id", null: false
+    t.bigint "user_id"
+    t.datetime "created_at", null: false
+    t.datetime "updated_at", null: false
+    t.index ["document_id"], name: "index_timeline_entries_on_document_id"
+    t.index ["user_id"], name: "index_timeline_entries_on_user_id"
+  end
+
   create_table "users", force: :cascade do |t|
     t.string "name"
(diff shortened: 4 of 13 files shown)


spec/models/timeline_entry_spec.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+RSpec.describe TimelineEntry do
```
benthorner: Do we need this if it's covered by a feature spec?

app/controllers/documents_controller.rb
```diff
@@ -24,6 +24,7 @@ def show
   def update
     document = Document.find_by_param(params[:id])
     document.update!(update_params(document))
+    TimelineEntry.create!(document: document, user: current_user, entry_type: "updated_content")
```
benthorner: I noticed that the 'create document' workflow means we end up with a 'created' and an 'updated' entry in the history. Would that confuse people? It did surprise me a bit.

benthorner: There's also a new entry after 'change content' -> 'save' when nothing has changed.

spec/models/timeline_entry_spec.rb
tijmenb: It's nice to have this checked, there might be things that haven't been feature-tested (yet).

app/controllers/documents_controller.rb
tijmenb: > I noticed that the 'create document' workflow means we end up with a 'created' and an 'updated' entry in the history. Would that confuse people? It did surprise me a bit.

I don't know, but I'd like to keep separate it so that we can attach different data to it.

> There's also a new entry after 'change content' -> 'save' when nothing has changed.

Why would you press save if nothing changed?

benthorner: I rarely press cancel in general. Reasoning: I don't care if I haven't changed anything.

spec/models/timeline_entry_spec.rb
```diff
@@ -0,0 +1,11 @@
...
+  describe "#entry_type" do
+    TimelineEntry::ENTRY_TYPES.each do |type|
+      it "`#{type}` has a translation" do
+        I18n.t!("documents.history.entry_types.#{type}")
```
kevindew: I think your meaning of this test might be clearer if you used the `exists?` method: https://github.com/svenfuchs/i18n/blob/master/lib/i18n.rb#L200

app/views/documents/show/_history_tab.html.erb
```diff
@@ -1 +1,11 @@
-Document history is not available right now.
+<% @document.timeline_entries.includes(:user).order(created_at: :desc).each do |entry| %>
+  <div class="timeline-entry govuk-!-margin-bottom-6">
+    <h4 class="govuk-heading-s govuk-!-margin-bottom-1">
+      <%= t "documents.history.entry_types.#{entry.entry_type}" %>
+    </h4>
+
+    <div class="govuk-body">
+      <%= entry.created_at.strftime("%l:%M%P on %d %B %Y") %> by <%= entry.user.name %>
```
kevindew: Be good to have a fallback for if a user is nil here.

db/migrate/20180907121447_create_timeline_entries.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+class CreateTimelineEntries < ActiveRecord::Migration[5.2]
+  def change
+    create_table :timeline_entries do |t|
+      t.string :entry_type, null: true
+      t.references :document, foreign_key: true, null: true
```
kevindew: Should we allow this to be null? I see we have `dependant: destroy` (although not sure if that's just the linter) but we could just have this as a foreign_key with `on_delete: cascade` to not worry about Ruby having to do anything if a document is deleted.

benthorner: Good point @kevindew - not sure why the 'entry_type' one should be null either.

tijmenb: 🤕 Ouch this should have been `null: false`, misremembered the syntax

app/views/documents/show/_history_tab.html.erb
tijmenb: Added a database constraint so the user can't be nil anymore.

db/migrate/20180907121447_create_timeline_entries.rb
tijmenb: Updated this to add `null: false` and constraints, including comments with the intention of the constraints. Rubocop still wants me to add `dependent: :destroy` though :|

emmabeynon:
Not a blocker, but given we're displaying "Document History", some alternative names could be "History Entry" or "History Event"?

db/schema.rb
```diff
@@ -79,6 +79,16 @@
     t.index ["document_id"], name: "index_images_on_document_id"
   end
 
+  create_table "timeline_entries", force: :cascade do |t|
+    t.string "entry_type", null: false
+    t.bigint "document_id", null: false
+    t.bigint "user_id", null: false
```
kevindew: Be good to keep this allowable to be nil so that it doesn't cause too many pain points with regards to importing data from Whitehall. Our other places where we have a user_id column are nullable.

tijmenb:
@kevindew removed the `null: true` from the user column.

app/views/documents/show/_history_tab.html.erb
kevindew: Sorry - think this now applies again

PR #271, opened 2018-09-12
Show correct flash messages according to design

tijmenb:
This PR:

- Removes the success flash messages for now, since aren't in the designs and Humin thinks there might not be a user need.
- Update the error flash messages [according to the content design in the error messages doc](https://docs.google.com/document/d/15euIDtRWAClra8y4VX1ZMSnXqS3SN24fJof6
EcGRdWQ/edit)

Last PR for https://trello.com/c/VkVrVGeB


app/controllers/application_controller.rb
```diff
@@ -4,4 +4,6 @@ class ApplicationController < ActionController::Base
   include GDS::SSO::ControllerMethods
 
   before_action :authenticate_user!
+
+  add_flash_types :alert_with_description
```
kevindew: I've not seen this before, is this so you don't need to put things in a `flash:` hash?

app/views/layouts/application.html.erb
```diff
@@ -39,6 +39,13 @@
       } %>
     <% end %>
 
+    <% if flash["alert_with_description"] %>
+      <%= render "govuk_publishing_components/components/error_alert", {
+        message: flash["alert_with_description"].fetch("title"),
+        description: sanitize(govspeak_to_html(flash["alert_with_description"].fetch("description")), tags: %w[a]),
```
kevindew: If I understand this right, we'll strip out all HTML that isn't links here? is there a particular reason we want to forbid any other elements? 

If so it'd probably useful to have this in a comment or commit message since it seems atypical to our usual handling of govspeak.

app/controllers/application_controller.rb
tijmenb: Yep! Some other apps [use this too](https://github.com/search?q=org%3Aalphagov+add_flash_types&type=Code).

app/views/layouts/application.html.erb
tijmenb: Caught me! It was a hack I shouldn't have committed to get around styling issues caused by nested `<p>` tags. I fixed it in the components with https://github.com/alphagov/govuk_publishing_components/pull/517.

PR #273, opened 2018-09-12
Add Jasmin for testing our JavaScript

benthorner:
https://trello.com/c/AgFY5UN1/214-setup-jasmine-testing-for-javascript-and-add-some-tests-m

This is based on the approach taken in govuk_publishing_components,
which uses the 'jasmine' gem locked to a 2.x version - there seems to be
a problem running 'rake jasmine:ci' for version 3.x.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -36,6 +36,7 @@ group :development, :test do
   gem "govuk-lint", "~> 3"
   gem "govuk_schemas", "~> 3.2"
   gem "govuk_test", "~> 0.2"
+  gem "jasmine", "~> 2.4"
   gem "rspec-rails", "~> 3"
   gem "webmock", "~> 3"
 end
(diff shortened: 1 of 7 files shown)


kevindew:
This looks good 👍 

I was wondering about the problem with jasmine > 3. I can see we use that on a few of our repos:

```
➜  govuk find . -name Gemfile -exec grep -n 'jasmine' /dev/null {} +
./signon/Gemfile:49:  gem 'jasmine', '3.2.0'
./maslow/Gemfile:34:  gem 'jasmine', '3.2.0'
./frontend/Gemfile:33:  gem 'jasmine', '~> 3.2'
...
```

and wondering if they have same problem you're facing? and/or is the problem already documented somewhere that we can link to?

benthorner:
It looks like the issues with jasmine 3.x are coming from our vendored assets, which we hope will fix itself over time - or if we move to webpacker. For now it makes sense to stick with 2.x.

benthorner:
@kevindew I've updated the commit message to explain the problem a little more.

kevindew:
@benthorner thanks 👍

PR #274, opened 2018-09-12
Index limits

benthorner:
https://trello.com/c/3OXJzoo5/206-limits-and-filtering-on-edit-content-page

This ensures we cleanup (no newlines) and truncate the title and summary
before we save them, if the user ignores the guidance for these fields.

For the base_path, we similarly cleanup and truncate the title before
passing it to the path generator service, thus limiting the slug length.

### DIFF
--- a/app/controllers/documents_controller.rb
+++ b/app/controllers/documents_controller.rb
@@ -58,7 +58,7 @@ private
   end
 
   def update_params(document)
-    DocumentAttributes.new(document).attributes(params)
+    DocumentUpdateParams.new(document).update_params(params)
       .merge(publication_state: "changes_not_sent_to_draft", review_state: "unreviewed")
   end
 end
--- a/app/controllers/documents_controller.rb
+++ b/app/controllers/documents_controller.rb
@@ -58,10 +58,7 @@ private
   end
 
   def update_params(document)
-    contents_params = document.document_type_schema.contents.map(&:id)
-    base_path = PathGeneratorService.new.path(document, params[:document][:title])
-
-    params.require(:document).permit(:title, :summary, :update_type, :change_note, contents: contents_params)
-      .merge(base_path: base_path, publication_state: "changes_not_sent_to_draft", review_state: "unreviewed")
+    DocumentAttributes.new(document).attributes(params)
+      .merge(publication_state: "changes_not_sent_to_draft", review_state: "unreviewed")
   end
 end
--- a/app/services/document_attributes.rb
+++ b/app/services/document_attributes.rb
@@ -1,0 +1,26 @@
+# frozen_string_literal: true
+
+class DocumentAttributes
+  TITLE_SLUG_MAX_LENGTH = 150
+  TITLE_MAX_LENGTH = 300
+  SUMMARY_MAX_LENGTH = 600
+
+  attr_reader :document
+
+  def initialize(document)
+    @document = document
+  end
+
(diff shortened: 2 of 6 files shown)


app/controllers/documents_controller.rb
```diff
@@ -58,10 +58,7 @@ def filter_params
   end
 
   def update_params(document)
-    contents_params = document.document_type_schema.contents.map(&:id)
-    base_path = PathGeneratorService.new.path(document, params[:document][:title])
-
-    params.require(:document).permit(:title, :summary, :update_type, :change_note, contents: contents_params)
-      .merge(base_path: base_path, publication_state: "changes_not_sent_to_draft", review_state: "unreviewed")
+    DocumentAttributes.new(document).attributes(params)
```
kevindew: Do you think this should be called something with a reference to update params?

app/services/document_attributes.rb
```diff
@@ -0,0 +1,26 @@
...
+
+  attr_reader :document
+
+  def initialize(document)
+    @document = document
+  end
+
+  def attributes(params)
+    path_title = params[:document][:title].chomp[0...TITLE_SLUG_MAX_LENGTH]
+    base_path = PathGeneratorService.new.path(document, path_title)
```
kevindew: What happens if title is nil?

app/services/document_attributes.rb
```diff
@@ -0,0 +1,26 @@
...
+  SUMMARY_MAX_LENGTH = 600
+
+  attr_reader :document
+
+  def initialize(document)
+    @document = document
+  end
+
+  def attributes(params)
+    path_title = params[:document][:title].chomp[0...TITLE_SLUG_MAX_LENGTH]
```
kevindew: We could probably do a `params.require(:document)` so we don't need to nest into params each time

app/services/document_attributes.rb
```diff
@@ -0,0 +1,26 @@
...
+    base_path = PathGeneratorService.new.path(document, path_title)
+
+    title = params[:document][:title].chomp[0...TITLE_MAX_LENGTH]
```
kevindew: I think chomp only strips new lines at the end of strings?

app/controllers/documents_controller.rb
benthorner: Renamed to DocumentUpdateParams.

app/services/document_attributes.rb
benthorner: OK - you win ;-).

app/services/document_attributes.rb
benthorner: Fixed - my misunderstanding.

app/services/document_attributes.rb
benthorner: Coping now - good spot :-).

app/services/document_update_params.rb
```diff
@@ -12,11 +12,11 @@ def initialize(document)
   end
 
   def update_params(params)
-    path_title = params[:document][:title].chomp[0...TITLE_SLUG_MAX_LENGTH]
+    path_title = params[:document][:title].squish[0...TITLE_SLUG_MAX_LENGTH]
```
emmabeynon: I've not seen `squish` before but I like it 😄

kevindew: Oh yes that's a new one for me too

PR #275, opened 2018-09-12
Keep track edition numbers for documents

tijmenb:
This PR adds an edition number to documents. This number will be shown in the document history, where it increases every time the document is published.

This doesn't have feature tests because it's not used in the UI yet.

https://trello.com/c/TVcThaKi

### DIFF
--- a/app/services/document_publishing_service.rb
+++ b/app/services/document_publishing_service.rb
@@ -16,7 +16,7 @@ class DocumentPublishingService
   def publish(document, review_state)
     document.update!(publication_state: "sending_to_live", review_state: review_state)
     publishing_api.publish(document.content_id, nil, locale: document.locale)
-    document.update!(publication_state: "sent_to_live", change_note: nil, update_type: "major", has_live_version_on_govuk: true)
+    document.update!(edition_number: document.edition_number + 1, publication_state: "sent_to_live", change_note: nil, update_type: "major", has_live_version_on_govuk: true)
   rescue GdsApi::BaseError => e
     GovukError.notify(e)
     document.update!(publication_state: "error_sending_to_live")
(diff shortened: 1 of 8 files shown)


app/services/document_publishing_service.rb
```diff
@@ -16,7 +16,7 @@ def publish_draft(document)
   def publish(document, review_state)
     document.update!(publication_state: "sending_to_live", review_state: review_state)
     publishing_api.publish(document.content_id, nil, locale: document.locale)
-    document.update!(publication_state: "sent_to_live", change_note: nil, update_type: "major", has_live_version_on_govuk: true)
+    document.update!(edition_number: document.edition_number + 1, publication_state: "sent_to_live", change_note: nil, update_type: "major", has_live_version_on_govuk: true)
```
emmabeynon: Just wondering if it's correct to increment the edition number before we know for sure that the publish has been successful?  I can image if there was an error publishing the document and the user tried to publish again we'd increment the edition number again which doesn't seem quite right?

tijmenb: Good point. This is below `publishing_api.publish` so if it reaches this line we're sure the publishing was successful, otherwise an error would have been raised.

kevindew: Are we incrementing this prematurely?

When we have a user_facing_state of "published" it will have an edition number of 2. Whereas that would still be the first edition until it goes to draft?

tijmenb:
I'm closing this for now, because we won't need the edition number in the design for a while. @kevindew's also right in that we need to increment the version number only when the user edits a previously published document. Otherwise we would technically withdraw, unpublish and approve documents with an incorrect version number.

PR #280, opened 2018-09-13
Update edit image page

alex-ju:
This PR:
- updates the image meta component to group meta attributes and only display meta list when presented
- updates the image meta component to allow no_border modifier
- updates the edit image page to use the updated image meta component
- updates tests to reflect changes

### Preview
screen shot 2018-09-13 at 18 17 07

[Trello card](https://trello.com/c/RCoMnQcK)

### DIFF
--- a/app/views/components/_image_meta.html.erb
+++ b/app/views/components/_image_meta.html.erb
@@ -1,12 +1,16 @@
 <%
   id ||= "image-meta-#{SecureRandom.hex(4)}"
   alt_text ||= nil
-  caption ||= nil
-  credit ||= nil
   actions ||= []
+  meta ||= {}
   lead_image ||= false
+  no_border ||= false
+
+  if meta.key?(:alt_text)
+    alt_text = meta[:alt_text]
+  end
 %>
-<div class="app-c-image-meta" id="<%= id %>">
+<div class="app-c-image-meta <%= 'app-c-image-meta--no-border' if no_border %>" id="<%= id %>">
   <div class="govuk-grid-row">
     <div class="govuk-grid-column-one-third">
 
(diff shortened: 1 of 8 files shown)


app/views/components/_image_meta.html.erb
```diff
@@ -1,12 +1,16 @@
...
   id ||= "image-meta-#{SecureRandom.hex(4)}"
   alt_text ||= nil
-  caption ||= nil
-  credit ||= nil
   actions ||= []
+  meta ||= []
   lead_image ||= false
+  no_border ||= false
+
+  if meta.any? && meta.key?(:alt_text)
```
alex-ju: If you know better ways of checking for a hash key to exists please let me know and I'll update.

kevindew: you can use [`key?`](https://ruby-doc.org/core-2.5.1/Hash.html#method-i-key-3F)

note that https://github.com/alphagov/content-publisher/pull/280/files#diff-3ffed0748d3f0d0a0349498ec9bc47d7R5 sets this object to an array rather than a hash

app/views/components/_image_meta.html.erb
```diff
@@ -18,23 +22,25 @@
...
+              value: meta[:alt_text]
+            },
+            {
+              field: t("documents.show.lead_image.fields.caption"),
+              value: meta[:caption]
+            },
+            {
+              field: t("documents.show.lead_image.fields.credit"),
+              value: meta[:credit]
+            }
```
kevindew: Are these ok to accept nil values?

app/views/components/_image_meta.html.erb
```diff
@@ -52,6 +58,6 @@
         <li class="app-c-image-meta__choose-image"><%= actions[:choose_button] %></li>
       <% end %>
       <li><%= link_to("Edit details", actions[:edit_link], class: "govuk-link") %></li>
-    <ul>
+    </ul>
```
kevindew: Good catch 👍

app/views/components/_image_meta.html.erb
```diff
@@ -1,12 +1,16 @@
 <%
   id ||= "image-meta-#{SecureRandom.hex(4)}"
   alt_text ||= nil
-  caption ||= nil
-  credit ||= nil
   actions ||= []
+  meta ||= []
```
kevindew: I think you want this to be a hash

app/views/components/_image_meta.html.erb
kevindew: Hopefully you won't need `.any?` once this is initialised as a hash.

app/views/components/_image_meta.html.erb
alex-ju: Updated

app/views/components/_image_meta.html.erb
alex-ju: I think so, the metadata component supports nil for these and if we have a nil for `value` we still want to display the `field` label.

PR #283, opened 2018-09-14
Add cropper component

alex-ju:
This PR:
- [x] adds an image cropper component which uses [Cropper.js](https://github.com/fengyuanchen/cropperjs)
- [x] creates a new view for the crop step
- [x] updates in controller/routes to get to that page

[Trello card](https://trello.com/c/3vS8DPkB)

- [x] integrate the page in the flow
- [x] read/send the crop coordinates from/to the backend
- [x] make it work with asset manager after rebase

[Trello card](https://trello.com/c/Yavdtg4i)

### Notes
We ended up attaching the cropper instance to `window.onload` as `image.load` in unreliable because of caching.

### Preview
When the uploaded image has an aspect ratio bigger than 3:2
crop example landscape

When the uploaded image has an aspect ratio smaller than 3:2
crop example portrait

### DIFF
--- a/app/assets/javascripts/components/image_cropper.js
+++ b/app/assets/javascripts/components/image_cropper.js
@@ -1,0 +1,27 @@
+//= require vendor/cropperjs/dist/cropper.js
+
+var $imageCroppers = document.querySelectorAll('[data-module="image-cropper"]')
+
+/* eslint-disable */
+if ($imageCroppers) {
+  $imageCroppers.forEach(function ($imageCropper) {
+    var $image = $imageCropper.querySelector('.app-c-image-cropper__image')
+    var width = $image.clientWidth;
+    var height = $image.clientHeight;
+    console.log(width + 'x' + height)
+    if ($image){
+      new Cropper($image,{
+        viewMode: 2,
+        aspectRatio: 3 / 2,
+        autoCrop: true,
+        autoCropArea: 1,
+        guides: false,
+        zoomable: false,
+        highlight: false,
+        minCropBoxWidth: 300,
+        minCropBoxHeight: 200
+      })
+    }
+  })
+}
+/* eslint-enable */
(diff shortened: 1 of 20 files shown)


app/assets/javascripts/components/image_cropper.js
```diff
@@ -0,0 +1,24 @@
...
+    if ($image){
+      new Cropper($image,{
+        viewMode: 2,
+        aspectRatio: 3 / 2,
+        autoCrop: true,
+        autoCropArea: .80,
+        guides: false,
+        zoomable: false,
+        highlight: false,
+        minCropBoxWidth: 300,
```
benthorner: Should we synchronise these settings with the minimum dimensions, which appear to be 960x640?

kevindew:
Are you thinking of handling the situation of constraining to a max height in this component? I wonder if it'd be good if in the component guide examples to have some of the more extreme cases for consideration?

alex-ju:
@kevindew this PR isn't dealing with the container constraining and recalculating the crop frame to match the aspect ratio that we discussed. I can change this to WIP and work to get that in too. I was trying to avoid raising huge PRs.

app/assets/javascripts/components/image_cropper.js
alex-ju: Setting the minimum to 960 means you currently won't be able to resize the cropping frame (as it's already constrained in width to 960 by the container), but that can be achieved if I do the extra work to allow the resize of the cropping frame based on the image size (as mentioned in my reply to Kevin)

benthorner:
@alex-ju +1 for small PRs :-). But I think @kevindew is right that coping with the resize is pretty fundamental to us using this component.

alex-ju:
@kevindew the component guide examples are quite useless right now because the javascript doesn't load for those pages (something to look into), this is one of the reasons I needed to create a page and test them there.

Thanks for helping me to get the original image. I updated the JavaScript code to deal with images constrained by the main container and added examples in this PR message to illustrate how it deals with images that are not respecting the 3:2 aspect ratio.

Let me know if you think we should bring anything else to this PR.

app/controllers/document_lead_image_controller.rb
```diff
@@ -33,14 +33,49 @@ def create
...
 
-  def edit
+  def crop
     @document = Document.find_by_param(params[:document_id])
     @image = @document.images.find(params[:image_id])
   end
 
+  def update_crop
+    document = Document.find_by_param(params[:document_id])
+    image = Image.find_by(id: params[:image_id])
```
tijmenb: For follow up: this should be `Image.find`, as `find_by` returns nil.

PR #286, opened 2018-09-17
Upload images to asset manager

emmabeynon:
For https://trello.com/c/2BSsLGFS/239-upload-images-to-asset-manager-and-set-lead-image-in-publishing-api
Paired with @davbo 

This PR allows images to be uploaded to Asset Manager as a draft, after which we save the returned Asset Manager file URL to the image.  Upon publish we change the image to no longer be a draft.
We also ensure that lead image data is sent to the Publishing API once a lead image is uploaded, removed or modified.


app/controllers/publish_document_controller.rb
```diff
@@ -8,6 +8,10 @@ def confirmation
   def publish
     document = Document.find_by_param(params[:id])
     review_state = params[:self_declared_review_state] == "has-been-reviewed" ? "reviewed" : "published_without_review"
+
+    images = [document.images, document.lead_image].flatten
```
tijmenb: Maybe this can go into `DocumentPublishingService`?

spec/features/upload_lead_image_spec.rb
```diff
@@ -12,6 +12,13 @@
     then_i_should_be_able_to_see_the_lead_image_on_the_summary_page
   end
 
+  scenario "User uploads a lead image and asset manager fails to upload it" do
```
tijmenb: Perhaps we can move this into its own file? Content tagger has multiple scenarios per file, and it's not improving readability (we should write this down).

spec/features/publish_a_document_spec.rb
```diff
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 RSpec.feature "Publishing a document" do
-  scenario "User publishes a document" do
+  scenario "User publishes a document with a lead image" do
```
tijmenb: It might be best to create a new feature for this, to keep the things that are being tested per scenario to a minimum. Alternatively, we could add the publish step to the lead image feature test? That way we're end-to-end testing the workflow.

app/controllers/publish_document_controller.rb
emmabeynon: Makes sense 👍

spec/features/upload_lead_image_spec.rb
emmabeynon: I think that's fair.  I'm a little concerned that we've got so many test files in the features directory that it's becoming difficult to navigate, but that's probably a conversation we can have outside of this PR 😸

app/services/document_publishing_service.rb
```diff
@@ -17,6 +17,7 @@ def publish(document, review_state)
     document.update!(publication_state: "sending_to_live", review_state: review_state)
     publishing_api.publish(document.content_id, nil, locale: document.locale)
     document.update!(publication_state: "sent_to_live", change_note: nil, update_type: "major", has_live_version_on_govuk: true)
+    publish_images(document.images)
```
tijmenb: Should image publishing happen before the sending to the publishing-api?

benthorner:
Also just noticed on https://trello.com/c/Yavdtg4i/254-create-crop-page-step-in-create-image-workflow that we're planning to lose the DocumentImagesController, so maybe it's not worth changing it in this PR?

app/models/image.rb
```diff
@@ -35,4 +35,33 @@ def crop_variant(resize = "#{WIDTH}x#{HEIGHT}")
...
+  # Used as a stand-in for a File / Rack::Multipart::UploadedFile object when
+  # passed to GdsApi::AssetManager#create_asset. The interface is required for
+  # uploading a file using the restclient we used in the backend.
+  #
+  # https://github.com/rest-client/rest-client/blob/master/lib/restclient/payload.rb#L181-L194
+  #
+  # This is done by delegating to 'io' which should implement the IO interface
+  # (e.g. StringIO) and adds methods to return filename, content_type and path.
+  class AssetManagerFile < SimpleDelegator
+    attr_reader :filename, :content_type, :url
```
kevindew: I think you want filename to be original_filename as per: https://www.rubydoc.info/gems/rack/Rack/Multipart/UploadedFile

app/controllers/document_lead_image_controller.rb
```diff
@@ -14,15 +14,26 @@ def create
...
-    if image_uploader.valid?
-      image = image_uploader.upload(document)
-      redirect_to edit_document_lead_image_path(params[:document_id], image.id)
-    else
+    unless image_uploader.valid?
       redirect_to document_lead_image_path, alert: {
         "alerts" => image_uploader.errors,
         "title" => t("document_lead_image.index.error_summary_title"),
       }
+      return
```
kevindew: Might make your intentions clearer if you did `return redirect_to ...`

app/services/publishing_api_payload.rb
```diff
@@ -43,13 +43,25 @@ def links
     document.tags.merge("organisations" => links.uniq)
   end
 
+  def image
+    {
+      "url" => document.lead_image.asset_manager_file_url,
+      "alt_text" => document.lead_image.alt_text || "",
```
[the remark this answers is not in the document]
emmabeynon: My understanding is that entering alt text will be encouraged in guidance, but I don't think we have any work planned to enforce it as a required field, so for now this just handles cases where it's not entered.

app/services/asset_manager_service.rb
```diff
@@ -0,0 +1,26 @@
...
+
+private
+
+  def asset_manager
+    @asset_manager ||= GdsApi::AssetManager.new(Plek.new.find("asset-manager"))
+  end
+
+  def asset_id(file_url)
+    url_array = file_url.split("/")
+    url_array[url_array.length - 2]
```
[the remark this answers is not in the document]
emmabeynon: It should return the 2nd to last element - perhaps `url_array[-2]` is clearer?

app/controllers/document_lead_image_controller.rb
benthorner: Hmm...I'd go the other way around, as the use of 'return <value>' to me implies we're actually returning something that's used by something else.

app/services/asset_manager_service.rb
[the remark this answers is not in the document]
benthorner: Yeah, I understand technically what it's doing, but I've no idea why. What does it achieve? Why is the number '2' so important here? (I did actually go and work out why, but it would be helpful to make it clearer here.)

app/services/document_publishing_service.rb
emmabeynon: I think you're right, it probably makes sense for the images to be published before the document is.

spec/features/publish_a_document_spec.rb
emmabeynon: Now added to `upload_lead_image_spec` 👍

davbo:
@benthorner we left the changes to the `DocumentImagesController` in there. You're right we might end up losing that controller all together but including the changes didn't seem to cause any issues.

PR #287, opened 2018-09-17
Handle no Publishing API for organisation filters

benthorner:
https://trello.com/c/GsK9GgOu/241-filter-index-page-by-organisation

### DIFF
--- a/app/views/documents/index/_filters.html.erb
+++ b/app/views/documents/index/_filters.html.erb
@@ -48,9 +48,15 @@
     }
   } %>
 
-  <% organisation_select = LinkablesService.new("organisation").select_options  %>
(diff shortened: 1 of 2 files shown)


kevindew:
Thanks for picking this up so promptly. Code looks good although I wonder if it makes more sense from a UX angle to not have a field at all rather than one that is empty?

benthorner:
@kevindew I tried that - unfortunately this causes a failure because it's required in order to select nothing!

benthorner:
@kevindew I see what you mean, although I think this makes it easier to see there's something wrong, rather than mysteriously removing the feature.

kevindew:
@benthorner ah what fails?

PR #290, opened 2018-09-18
Simplify `find_by` usage

tijmenb:
`find_by` will return nil if the record is not found, while `find` (or `find_by!`) will raise an exception. `find` takes an ID, so we can simplify these calls.

### DIFF
--- a/app/controllers/document_images_controller.rb
+++ b/app/controllers/document_images_controller.rb
@@ -23,7 +23,7 @@ class DocumentImagesController < ApplicationController
 
   def update
     document = Document.find_by_param(params[:document_id])
-    image = Image.find_by!(id: params[:id], document_id: document.id)
+    image = document.images.find(params[:id])
     image.assign_attributes(image_params)
 
     if image.valid?
--- a/app/controllers/document_lead_image_controller.rb
+++ b/app/controllers/document_lead_image_controller.rb
@@ -38,12 +38,12 @@ class DocumentLeadImageController < ApplicationController
 
   def edit
     @document = Document.find_by_param(params[:document_id])
-    @image = Image.find_by(id: params[:image_id])
+    @image = @document.images.find(params[:image_id])
   end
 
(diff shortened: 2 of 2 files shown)


app/controllers/document_images_controller.rb
```diff
@@ -22,8 +22,7 @@ def create
   end
 
   def update
-    document = Document.find_by_param(params[:document_id])
-    image = Image.find_by!(id: params[:id], document_id: document.id)
+    image = Image.find(params[:id])
```
kevindew: This originally had the document lookup to ensure we had an image for that document - since we have both document_id and image id in the url. 

To maintain this it might be best that this does `image = Document.find_by_param(params[:document_id]).images.find(params[:id])`

what do you think?

app/controllers/document_lead_image_controller.rb
```diff
@@ -38,12 +38,12 @@ def create
...
   def edit
     @document = Document.find_by_param(params[:document_id])
-    @image = Image.find_by(id: params[:image_id])
+    @image = Image.find(params[:image_id])
   end
 
   def update
     document = Document.find_by_param(params[:document_id])
-    image = Image.find_by(id: params[:image_id])
+    image = Image.find(params[:image_id])
```
kevindew: Similarly on these probably should be `document.images.find` if we want to strictly match the url

kevindew:
@benthorner say you have image 10 for document `123:en` and image 11 for document `456:en`. If you don't look up by document as well as image id then effectively you can edit images for different documents by say using a url of `/documents/456:en/lead-image/11`

benthorner:
Ah, OK. So it's more about the defensive programming side of things. That's fine with me - @tijmenb perhaps `find_by!` will be the best of both worlds.

tijmenb:
@kevindew @benthorner changed it to `document.images.find(x)`

PR #299, opened 2018-09-19
Implement edition numbering

tijmenb:
This PR adds an edition number to documents. This number will be shown in the document history, where it increases every time a new draft is created.

## What it looks like

screen shot 2018-09-19 at 15 14 42

https://trello.com/c/TVcThaKi

### DIFF
--- a/app/controllers/documents_controller.rb
+++ b/app/controllers/documents_controller.rb
@@ -23,6 +23,11 @@ class DocumentsController < ApplicationController
 
   def update
     document = Document.find_by_param(params[:id])
+
+    if document.publication_state == "sent_to_live"
+      document.current_edition_number += 1
+    end
+
     document.update!(update_params(document))
     TimelineEntry.create!(document: document, user: current_user, entry_type: "updated_content")
     DocumentPublishingService.new.publish_draft(document)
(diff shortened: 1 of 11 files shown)


app/views/documents/show/_history_tab.html.erb
```diff
@@ -1,5 +1,12 @@
+<% displayed_edition_number = nil %>
+
 <% @document.timeline_entries.includes(:user).order(created_at: :desc).each do |entry| %>
   <div class="timeline-entry govuk-!-margin-bottom-6">
+    <% if displayed_edition_number != entry.edition_number %>
+      <h3 class="govuk-heading-m"><%= entry.edition_number.ordinalize %> edition</h3>
```
kevindew: Should this go through the translations file?

app/controllers/documents_controller.rb
```diff
@@ -23,6 +23,11 @@ def show
 
   def update
     document = Document.find_by_param(params[:id])
+
+    if document.publication_state == "sent_to_live"
+      document.current_edition_number = document.current_edition_number + 1
```
emmabeynon: Can we do `document.current_edition_number += 1` here?

app/views/documents/show/_history_tab.html.erb
tijmenb: 👍 done

app/controllers/documents_controller.rb
tijmenb: 👍 done

tijmenb: Turns out `increment` exists as well! https://apidock.com/rails/ActiveRecord/Persistence/increment

PR #300, opened 2018-09-19
Add test for publishing when asset manager is down

emmabeynon:


spec/features/publish_a_document_asset_manager_down_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+  def given_there_is_a_document_with_a_lead_image
+    @image = create(:image, :in_asset_manager)
+    document = create(:document, publication_state: "sent_to_draft", lead_image: @image)
+    @image.update(document: document)
+  end
+
+  def and_asset_manager_is_down
+    #TODO: Add stub to gds-api-adapters test helpers
+    stub_request(:put, "https://asset-manager.test.gov.uk/assets/#{@image.asset_manager_id}").to_return(status: 500)
+    stub_publishing_api_publish(Document.last.content_id, update_type: nil, locale: Document.last.locale)
```
benthorner: You could use `stub_any_publishing_api_publish` to make the test less brittle here. Otherwise looks good :-)

emmabeynon: Done 👍

PR #302, opened 2018-09-19
Add audit log for debugging

tijmenb:
This adds the papertrail gem to the app and sets it up. Paperclip will keep track of all of the changes in the models (Document, Image, User).

The reason for this is to keep an audit log of all of the changes. While we're busy building this application, we're very likely to run into a situation where a document has entered a particular state, and we're not sure how this happened. Having a full log of changes will help us a lot in debugging. And with time being linear, we won't be able to add this feature retroactively if something goes wrong.

It's worth noting that this is a distinct & separate use case from the history timeline (https://github.com/alphagov/content-publisher/pull/270), which is a feature designed for end-users. The two shouldn't be coupled.

Downside of having this:

- It slow all of the writes down. When a document is edited, there will be multiple versions created because of the state changes.
- The versions table is likely to grow a lot. We'll need to make sure that we clean up if the table becomes too large. There's a ticket on the planning board for this: https://trello.com/c/XMrIPU88. This is also the reason that we need to keep the history feature and this audit trail separate, because we should be able to truncate the table without losing end-user functionality.

This PR also adds the entries to a document "debug" page.

screenshot_2018-09-19 internal metadata for hello - gov uk


kevindew:
Just to understand the context, Is this intended entirely for debugging purposes or are they any ideas about it's usage with regards to rolling back?

It seems a cool idea and I can see it being useful in the early days. I have a concern that if we're not looking at it often it could fill up a lot of data.

What sort of future do you see for this with the context of storing revisions?

tijmenb:
@kevindew argh, tried to communicate that in the PR description but it wasn't clear enough. Papertrail should _only_ be used as an audit log, and absolutely nothing else. This will make it easy for us to throw away the data and have a functioning application.

benthorner:
Can you add comment above the Papertrail lines to indicate what it's being used for. I reckon when we eventually have two kinds of history it'll confuse people unless we clearly signpost it for them.

PR #309, opened 2018-09-20
Use `Services` module for consistency

tijmenb:
Almost all other apps on GOV.UK use a `Services` module to instantiate the API clients.

### DIFF
--- a/app/services/linkables_service.rb
+++ b/app/services/linkables_service.rb
@@ -1,7 +1,5 @@
 # frozen_string_literal: true
 
-require "gds_api/publishing_api_v2"
-
 class LinkablesService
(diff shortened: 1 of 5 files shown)


app/services/linkables_service.rb
```diff
@@ -25,16 +23,7 @@ def linkables
...
-      publishing_api.get_linkables(document_type: document_type).to_hash
+      Services.publishing_api.get_linkables(document_type: document_type).to_hash
     end
   end
-
-  def publishing_api
-    GdsApi::PublishingApiV2.new(
-      Plek.new.find("publishing-api"),
-      disable_cache: true,
-      timeout: 1,
```
kevindew: Note this change loses this timeout

tijmenb:
Okido

PR #312, opened 2018-09-20
Reorganise feature tests

tijmenb:
We currently have around 40 feature tests, which makes it hard to find what you're looking for. This PR splits up the directory into 5 areas.

Since I did the organising, it makes sense to me personally. I'll do some pop-up user research to see if it makes sense to others as well.


kevindew:
Thanks for doing this 👍

From a quick first look it looks pretty good. I like that they're based around user behaviour rather than controllers/domain concepts. Some suggestions would be:

- index_page sticks out a bit that may make sense under `browsing` or `finding`
- maybe separate creating from editing
- where we've got specific content editing maybe group that under content for consistency with images and tagging

what do you think?

tijmenb:
Reoganised again into:

https://github.com/alphagov/content-publisher/tree/65079abe0c3716e7d591d3f7b277f3bbe15b8e93/spec/features

PR #313, opened 2018-09-20
Set state on imported documents

emmabeynon:
For https://trello.com/c/uBqWwIqu/234-set-the-correct-state-when-importing-from-whitehall-m

This sets `publication_state` and `review_state` on documents that are imported from Whitehall based on their Whitehall state.  Whitehall states have been mapped to Content Publisher publication states and review states in this document: https://docs.google.com/spreadsheets/d/1R3tfzhWDsO2zize6s-8EzjW7rvgKZQhu3NLYEVXouWk/edit#gid=0

We have chosen not to import documents with the following states:
- `imported` - There are only 4 imported documents which were last updated in 2013 it seems unncessary to import these.
- `scheduled` - We don't currently support scheduling in this app, so for now we won't import documents that are scheduled for publication.
- `superseded` - These are documents that have been superseded by other documents, so it seems unnecessary to import these for future editing.
- `withdrawn` - It seems unnecessary to import withdrawn documents for future editing.

In order to safeguard against any documents with rare/unidentified states from being imported we've created a whitelist of 'supported' states and will only import documents with these states.  As and when other states are deemed 'supported', we can add them to this whitelist.

### DIFF
--- a/lib/tasks/whitehall_news_importer.rb
+++ b/lib/tasks/whitehall_news_importer.rb
@@ -65,18 +65,14 @@ module Tasks
     end
 
     def publication_state(edition)
-      return "sent_to_draft" if edition["state"] == "draft"
-      return "sent_to_live" if edition["state"] == "published"
-      return "sent_to_draft" if edition["state"] == "rejected"
-      return "sent_to_draft" if edition["state"] == "submitted"
+      edition["state"] == "published" ? "sent_to_live" : "sent_to_draft"
     end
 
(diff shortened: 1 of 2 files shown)


emmabeynon:
@kevindew Will do 👍

emmabeynon:
I've run the import task and spot-checked a few Whitehall documents to check that the `publication_state` and `review_state` have been correctly set, so I'm going to merge this.

PR #314, opened 2018-09-20
Fix importing current edition number from Whitehall

emmabeynon:
We found a bug when running the "Whitehall Import" job where sometimes an edition imported from Whitehall may not have a `published_major_version`, which was causing a NilClass error when trying to
increment the version number.

### DIFF
--- a/lib/tasks/whitehall_news_importer.rb
+++ b/lib/tasks/whitehall_news_importer.rb
@@ -31,7 +31,7 @@ module Tasks
         review_state: "unreviewed",
         summary: translation["summary"],
         tags: tags(edition),
-        current_edition_number: edition["published_major_version"] + 1,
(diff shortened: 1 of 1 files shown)


kevindew:
Good catch, do we know why some have that number? I'm just wondering if every minor update misses this value or it's just a few anomaly documents?

emmabeynon:
@kevindew My guess is that these are documents that have never been published and so that value has never been set, but I'm not too familiar with how minor/major changes manifest themselves in Whitehall so I'll need to look into it.

kevindew:
@emmabeynon sure, I think it's worth working out why these are happening before having a fix in that may put the wrong number in. As I know this may be blocking your testing now I'd suggest we skip records without the number until we know?

emmabeynon:
Closed in favour of https://github.com/alphagov/content-publisher/pull/315

PR #319, opened 2018-09-21
Update web components polyfill

alex-ju:
This RP replaces the `webcomponents loader` with `webcomponents bundle`.

While the first is best for performance it's causing issues with our current minify/uglify assept pipeline.

Read more about the [web components bundle](https://github.com/webcomponents/webcomponentsjs#using-webcomponents-bundlejs).


app/assets/javascripts/components/markdown-editor.js
```diff
@@ -1,4 +1,4 @@
-//= require vendor/@webcomponents/webcomponentsjs/webcomponents-loader.js
+//= require vendor/@webcomponents/webcomponentsjs/webcomponents-bundle.js
```
davbo: I know it was like this before and seemed to be included correctly but it seems odd to have this included only in the markdown component ?

alex-ju: Makes sense to move it to `application.js` now

PR #325, opened 2018-09-21
Fix CORS errors caused by cropper

davbo:
Disables rotation and scaling in the cropper. This prevents cropper.js from
making an additional XHR request to download the image again. This XHR request
creates CORS errors and isn't necessary for us unless we want the rotation
feature to work.

Digging into why it makes the request it seems to be to download the image into
an arraybuffer to parse the exif data for things like handling image rotation.
The code which makes the XHR is here:

  https://github.com/fengyuanchen/cropperjs/blob/23b2e6c07c0c636082cc3bbf3509e3cb3d531271/src/js/cropper.js#L98-L159

Disabling these options stops the XHR from being sent.


kevindew:
Nice digging 👍

davbo:
🕵️

PR #327, opened 2018-09-24
Tagging to role appointments

kevindew:
Trello: https://trello.com/c/3GT17tkH/211-associate-content-with-ministers-l

This allows the functionality to tag news/press releases to ministers by allowing a tag to role appointments. It does this because there is currently no clear way to get a list of ministers from the Publishing API and there does not seem to be a particular reason why all roles/people can't be associated with a news story.

screen shot 2018-09-24 at 10 35 30

Things still to do:

- [x] Talk to a content designer about what to call the field (currently called government appointments) - after chat I've gone for Ministers and government appointments
- [x] Update the schemas to support adding roles and people as edition links (https://github.com/alphagov/govuk-content-schemas/pull/814)
- [x] Check/fix there is no where that relies on ministers that can rely people, and that we don't break anything with changes

### DIFF
--- a/spec/formats/news_story_spec.rb
+++ b/spec/formats/news_story_spec.rb
@@ -27,7 +27,7 @@ RSpec.feature "Create a news story", format: true do
 
   def and_i_add_some_tags
     stub_any_publishing_api_put_content
-    expect(Document.last.document_type_schema.tags.count).to eq(5)
+    expect(Document.last.document_type_schema.tags.count).to eq(6)
     publishing_api_has_linkables([linkable], document_type: "topical_event")
     publishing_api_has_linkables([linkable], document_type: "worldwide_organisation")
     publishing_api_has_linkables([linkable], document_type: "world_location")
--- a/spec/formats/news_story_spec.rb
+++ b/spec/formats/news_story_spec.rb
@@ -27,6 +27,9 @@ RSpec.feature "Create a news story", format: true do
 
   def and_i_add_some_tags
     stub_any_publishing_api_put_content
+
+    publishing_api_has_links(role_appointment_links)
+
     expect(Document.last.document_type_schema.tags.count).to eq(6)
     publishing_api_has_linkables([linkable], document_type: "topical_event")
     publishing_api_has_linkables([linkable], document_type: "worldwide_organisation")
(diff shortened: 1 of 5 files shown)


spec/services/publishing_api_payload_spec.rb
```diff
@@ -61,6 +61,33 @@
...
+                       })
+
+      person_id = SecureRandom.uuid
+      role_id = SecureRandom.uuid
+      links = {
+        "content_id" => role_appointment_id,
+        "links" => { "person" => [person_id], "role" => [role_id] },
+      }
+
+      # TODO: add this to gds-api-adapters test helpers
```
benthorner: Is this already there...

https://github.com/alphagov/gds-api-adapters/blob/master/lib/gds_api/test_helpers/publishing_api_v2.rb#L384

spec/formats/news_story_spec.rb
```diff
@@ -52,19 +58,39 @@ def then_i_can_publish_the_document
...
+        "roles" => [role_content_id],
+        "people" => [person_content_id],
+      ),
       "title" => "A great title",
       "document_type" => "news_story",
       "description" => "A great summary",
     }
   end
 
+  def role_appointment_links
```
benthorner: Could make this a little shorter by memoising here and merging `role_appointment_links["links"]` into `content_body["links"]`, which would reduce duplication in the two format specs.

spec/services/publishing_api_payload_spec.rb
kevindew: d'oh! thanks.

spec/formats/news_story_spec.rb
kevindew: I don't think we can (or I've not understood you correctly) as they have different link names. singular role/person for a role_appointment, and plural roles/people for news

benthorner: Oh yeah. Alternatively you could still memoize the role_appointment_links and extract out the content_ids in the content body - just thought you could save on a couple of methods.

spec/formats/news_story_spec.rb
```diff
@@ -52,19 +57,39 @@ def then_i_can_publish_the_document
 
   def content_body
     {
-      "links" => {
+      "links" => a_hash_including(
```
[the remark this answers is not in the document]
kevindew: That's to not couple it to our app ordering. If I remember right I got caught on a failing test because of hash key ordering and that felt a bit too coupled.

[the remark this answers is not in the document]
benthorner: So I think it would be better to leave it as a plain old hash, because the 'including' matcher implies there could be other stuff in there, event though we know there isn't. I don't think we're relying on keys here, so it should be OK?

kevindew:
@benthorner I've changed back to a hash

PR #333, opened 2018-09-24
Display user that last edited a document

emmabeynon:
For https://trello.com/c/OdqQe7J2/269-store-and-present-the-user-that-last-edited-a-document

This adds a method to retrieves the user that last edited a document by querying the most recent
Timeline Entry with type `updated_content` or `updated_tags`.  We have decided to perform a query on Timeline Entries so that we're using these entries as the source of truth for a document's history. We considered associating a `last_edited_by` user to a document and updating this every time
a document is edited, which could be a simpler approach. However since we are already storing the user in the Timeline Entry anyway it doesn't seem particularly useful to store this information again elsewhere. This approach also should also prevent any discrepancies that could occur between the user stored on the Document and the Timeline Entries.

Using `last_edited_by` we then display the user on the summary page:
screen shot 2018-09-24 at 15 24 31

And the index page:
screen shot 2018-09-24 at 15 24 48

### DIFF
--- a/app/views/documents/index/_results.html.erb
+++ b/app/views/documents/index/_results.html.erb
@@ -38,6 +38,10 @@
           <%= t "documents.index.search_results.state",
             state: t("user_facing_states.#{document.user_facing_state}.name")
           %>
+          <br/>
+          <%= t "documents.index.search_results.last_edited_by",
+            user: document.last_edited_by&.name || document.creator&.name || I18n.t("documents.index.search_results.unknown_creator")
+          %>
         </td>
 
         <td class="govuk-table__cell ">
--- a/app/models/document.rb
+++ b/app/models/document.rb
@@ -54,4 +54,12 @@ class Document < ApplicationRecord
   def title_or_fallback
     title.presence || I18n.t!("documents.untitled_document")
   end
+
+  def last_edited_by
+    entries = self.timeline_entries
+      .where(entry_type: %w(updated_content updated_tags))
+      .order(:created_at)
+
+    entries.last.user if entries.any?
+  end
 end
--- a/app/views/documents/show/_document_metadata.html.erb
+++ b/app/views/documents/show/_document_metadata.html.erb
@@ -17,6 +17,10 @@
         field: t("documents.show.metadata.created_by"),
         value: @document.creator&.name || I18n.t("documents.show.metadata.unknown_creator")
       },
+      {
+        field: t("documents.show.metadata.last_edited_by"),
+        value: @document.last_edited_by&.name || @document.creator&.name || I18n.t("documents.show.metadata.unknown_creator")
+      }
     ]
   } %>
 </div>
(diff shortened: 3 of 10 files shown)


app/views/documents/index/_results.html.erb
```diff
@@ -38,6 +38,10 @@
           <%= t "documents.index.search_results.state",
             state: t("user_facing_states.#{document.user_facing_state}.name")
           %>
+          <br/>
+          <%= t "documents.index.search_results.last_edited_by",
+            user: document.last_edited_by&.name || document.creator&.name || I18n.t("documents.index.search_results.unknown_creator")
```
tijmenb: Do we have a N+1 query here? I think that for every document we'll end up doing a query to fetch the latest timeline entry, and then a query to fetch the user. For 50 results that would mean 100 extra database calls.

emmabeynon: That's a good point.  I've been swinging between this approach and just writing the user who last edited to the document.  This is making me think the latter is going to be more performant.  What do you think?

tijmenb: One way would be to do `self.timeline_entries.includes(:user)`, which will prevent the 2nd query, but that still leaves the 50 for the entries.

Another approach depends on what the original user need for this feature is. I could imagine that it would be useful for users to see all of the people involved, rather than just the last, so you'd have something like this:

```
Publishers: Tijmen, Emma and Kevin
```

All of the participating people is easier to fetch in a single query with something like:

```rb
documents = # whatever we search for
User.where(id: TimelineEntry.where(document: documents).select(:user_id))
```

@kevindew what do you think?

tijmenb: Update: above doesn't at all solve the problem, I don't know why I thought it would.

tijmenb: Ok, this would work. The trick is to do the queries but do the filtering in Ruby:

```rb
Document
  .all
  .includes(timeline_entries: [:user])
  .each { |d| 
    puts d.timeline_entries
      .sort_by(&:created_at)
      .select { |t| t.entry_type.in?(%w(updated_content updated_tags)) }
      .first.user.name 
  }
```

kevindew: I'd imagine it's good to go down the route of explicitly storing who last edited the document on the document. Then you can get around the N+1 problem. 

I think if you're dealing with a list of documents you're not going to want to deal a list of x length of publishers.

I think we'll handle the full grouping through the history.

It'd probably useful to have in the filterable options selecting anyone who has ever edited a document but I think that could be achieved in a relatively performant way of doing an exists sub query on timeline entry.

benthorner: Given that we're already storing the creator on the document itself, it's consistent if we store the editor on the document too, although it's a shame as we could potentially use the history for both of these things. Are the user needs different in some way?

emmabeynon:
Closed in favour of https://github.com/alphagov/content-publisher/pull/336

PR #336, opened 2018-09-25
Associate document with user who last edited it

emmabeynon:
For https://trello.com/c/OdqQe7J2/269-store-and-present-the-user-that-last-edited-a-document

This adds a `last_editor` to Document which is updated every time an edit is made to a document.  We have defined an edit to mean a change to the document itself rather than a state change such as sending for review or publishing.

Make `last_editor` available allows us to display that last person who edited a document:

**Summary page**
screen shot 2018-09-25 at 09 51 07

**Index page results**
screen shot 2018-09-25 at 12 10 05

### DIFF
--- a/app/views/documents/index/_results.html.erb
+++ b/app/views/documents/index/_results.html.erb
@@ -40,7 +40,7 @@
           %>
           <br/>
           <%= t "documents.index.search_results.last_edited_by",
-            user: document.last_editor&.name || document.creator&.name || I18n.t("documents.index.search_results.unknown_creator")
+            user: document.last_editor&.name || I18n.t("documents.index.search_results.unknown_creator")
           %>
         </td>
 
--- a/app/views/documents/index/_results.html.erb
+++ b/app/views/documents/index/_results.html.erb
@@ -38,6 +38,10 @@
           <%= t "documents.index.search_results.state",
             state: t("user_facing_states.#{document.user_facing_state}.name")
           %>
+          <br/>
+          <%= t "documents.index.search_results.last_edited_by",
+            user: document.last_editor&.name || document.creator&.name || I18n.t("documents.index.search_results.unknown_creator")
+          %>
         </td>
 
         <td class="govuk-table__cell ">
--- a/app/views/documents/index/_results.html.erb
+++ b/app/views/documents/index/_results.html.erb
@@ -32,7 +32,7 @@
           <%= document.document_type_schema.label %>
           <br/>
           <%= t "documents.index.search_results.creator",
-            creator: document.creator&.name || I18n.t("documents.index.search_results.unknown_creator")
+            creator: document.creator&.name || I18n.t("documents.index.search_results.unknown_user")
           %>
           <br/>
           <%= t "documents.index.search_results.state",
@@ -40,7 +40,7 @@
           %>
           <br/>
           <%= t "documents.index.search_results.last_edited_by",
-            user: document.last_editor&.name || I18n.t("documents.index.search_results.unknown_creator")
+            user: document.last_editor&.name || I18n.t("documents.index.search_results.unknown_user")
           %>
         </td>
 
--- a/spec/features/editing_content/edit_a_document_spec.rb
+++ b/spec/features/editing_content/edit_a_document_spec.rb
@@ -10,6 +10,7 @@ RSpec.feature "Edit a document" do
     and_the_content_is_available_for_preview
     and_the_content_needs_review
     and_i_see_the_content_is_in_draft_state
+    and_i_see_i_was_the_last_user_to_edit_the_document
   end
 
   def given_there_is_a_document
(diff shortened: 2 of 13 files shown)


app/views/documents/index/_results.html.erb
```diff
@@ -38,6 +38,10 @@
           <%= t "documents.index.search_results.state",
             state: t("user_facing_states.#{document.user_facing_state}.name")
           %>
+          <br/>
+          <%= t "documents.index.search_results.last_edited_by",
+            user: document.last_editor&.name || document.creator&.name || I18n.t("documents.index.search_results.unknown_creator")
+          %>
```
kevindew: Should we change this key to be more generic such as unknown user?

spec/features/workflow/viewing_a_document_with_no_creator_spec.rb
```diff
@@ -1,12 +1,14 @@
 # frozen_string_literal: true
 
 RSpec.feature "Viewing a document with no creator" do
-  scenario "User finds document without a creator and views it" do
+  scenario "User finds an unedited document without a creator and views it" do
```
benthorner: I think this spec is more 'User views a document without a creator/editor'. We should avoid implying that this is possible within the app, as it only happens for imported stuff, which may have been edited by someone, we just don't know who.

spec/features/workflow/viewing_a_document_with_no_creator_spec.rb
```diff
@@ -24,6 +26,13 @@ def then_i_see_we_dont_know_who_created_the_document
     )
   end
 
+  def and_i_see_the_document_has_no_last_editor
+    expect(page).to have_content(
+      I18n.t("documents.index.search_results.last_edited_by",
+        user: I18n.t("documents.index.search_results.unknown_creator")),
```
benthorner: Could we create a separate translation for this, to avoid confusion between creator/editor?

app/views/documents/index/_results.html.erb
```diff
@@ -38,6 +38,10 @@
           <%= t "documents.index.search_results.state",
             state: t("user_facing_states.#{document.user_facing_state}.name")
           %>
+          <br/>
+          <%= t "documents.index.search_results.last_edited_by",
+            user: document.last_editor&.name || document.creator&.name || I18n.t("documents.index.search_results.unknown_creator")
```
kevindew: Can we set last edited when we mark who created the document so that we don't need to do this or here? 

It feels like this should follow same pattern as db created_at and updated_at records where updated_at is set to same time as created_at when an item is created.

spec/features/editing_content/edit_a_document_spec.rb
```diff
@@ -10,6 +10,8 @@
     and_the_content_is_available_for_preview
     and_the_content_needs_review
     and_i_see_the_content_is_in_draft_state
+    and_i_see_i_was_the_last_user_to_edit_the_document
+    and_on_the_index_page_i_see_the_document_was_last_edited_by_me
```
benthorner: This last step is doing a lot - could this check be done in the index page spec instead?

spec/features/workflow/viewing_a_document_with_no_creator_spec.rb
emmabeynon: I think I'm going to go for Kevin's suggestion of an `unknown_user` here.

app/views/documents/index/_results.html.erb
emmabeynon: Makes sense - added here: https://github.com/alphagov/content-publisher/pull/336/commits/6820219e0cc9ea8d06f07bb0d5627f132d360fac

app/views/documents/index/_results.html.erb
emmabeynon: I think this is reasonable given we're using it in a couple of places now - changed here: https://github.com/alphagov/content-publisher/pull/336/commits/b21fd9edfe5dbfd6cf18e0ba87ad4d97a601d63d

spec/features/workflow/viewing_a_document_with_no_creator_spec.rb
emmabeynon: Agreed, this makes the test intention clearer - now updated 👍

spec/features/editing_content/edit_a_document_spec.rb
emmabeynon: The index page specs seem to be focused on filtering, which is why I chose not to add this test to one of those files.  This seemed to be the best place to put it given that it's related to editing a document, though I agree it's perhaps a little shoehorned in.  Would it help if I split it into visiting the index page and seeing the last editor as separate steps?

benthorner: I think it would be better to create a quick 'index_page' spec to check content that appears on the index page - we're not checking most of it at the moment. Line 13 already checks the edit feature works, so line 14 is only checking that we're showing this information on the index page and is therefore more related to a spec about the index page specifically.

emmabeynon: Added here: https://github.com/alphagov/content-publisher/pull/336/commits/e3792f23ed6238a0be1ec103c7e136c34687945e

PR #338, opened 2018-09-25
Ensure cross browser support

alex-ju:
This PR brings in polyfills needed for the existing code in order to work with [supported browsers](https://www.gov.uk/service-manual/technology/designing-for-different-browsers-and-devices).

We vendor the following polyfills:
- `fetch` and `abortcontroller` for the request to preview markdown using govspeak in the *markdown-editor* and for the *url-preview*
- `custom-elements` and `reflect` for *markdown-toolbar* injectors

For the rest of polyfills (e.g. `Promise`, `URL`, `.closest`) instead of shipping code to all browsers we switch to using polyfill.io which will only return the polyfills that are suitable for the requesting browser.

Testing progress:
- [x] Google Chrome (latest versions) + Android and iOS
- [x] Mozilla Firefox (latest versions)
- [x] Edge (latest versions)
- [x] Internet Explorer 11
- [x] Safari (latest version) + iOS

[Trello card](https://trello.com/c/Me0hBC5a)


kevindew:
Great to improve cross browser support 🎉 

Would you be able to add some info on the polyfill.io commit about what we get from that, why we're using that rather than polyfills ourselves? 

(also probably good to squash the fix abort controller commit)

alex-ju:
@kevindew Yes, I'll surely do that so we can have a view of what component needs what. Deployed this branch as WIP as I need to test it in on a Windows computer.

benthorner:
I had a quick look through the fetch polyfill - does it support a timeout? I don't think we care what it is so long as it's reasonably short.

alex-ju:
@benthorner the `fetch` polyfill doesn't provide any timeout, it brings the polyfill to the browsers that won't support it; it can be used however with `abortcontroller`. The question is what are we bringing to the users by adding this experimental feature? If we were to make requests while typing the title, the aborting would save server requests, otherwise I don't find it mandatory in our scenarios. We can bring it back but dropping 52b4790408c5fc96a067667046b8eb087c9c28a5, but I'm questioning the need for it.

benthorner:
@alex-ju the only technical requirement was to ensure a timeout is implemented, so that we handle errors properly - especially for the markdown preview where it affects the actions a user can perform. I suspect it's a product-level decision if we want to compromise on timeout support - would it be hard to keep the existing abort behaviour?

alex-ju:
@benthorner ok, then I'll drop the commit which removes the `abortcontroller` and will leave it for a separate discussion; does that sound ok?

benthorner:
@alex-ju when I put it in it looked like we could use a polyfill for the abort stuff...https://www.npmjs.com/package/abortcontroller-polyfill. 

Is that an option here? It's probably a call for Humin, but I reckon fulfilling this PR is better than a bad timeout behaviour, if we can't make the abort stuff work for some reason.

alex-ju:
@benthorner that polyfill is already part of this PR, but vendored, as we do with the rest of them.

PR #339, opened 2018-09-25
Add placeholder screens for retire/remove

benthorner:
https://trello.com/c/QAEhhJdf/244-build-workflow-for-delete-draft-remove-and-retire-document-m

This adds two separate placeholder screens for retiring/removing a
document, so that a user can see the feature isn't supported and know
what to do. I wanted to keep the new actions out of the already bulging
DocumentsController, so I've created two completely separate placeholder
routes, which will allow these features to be implemented separately.
In order to render the textual content on the page, I've used the same
pattern as for new_document/guidance.html.erb.

### DIFF
--- a/app/views/remove_document/remove.govspeak.md
+++ b/app/views/remove_document/remove.govspeak.md
@@ -1,0 +1,5 @@
+If you need to remove this content please [raise a support request and we’ll do it for you](https://support.publishing.service.gov.uk/).
+
+You must include any public explanation or new web address required.
+
+[Full guidance on removing content](https://www.gov.uk/guidance/content-design/gov-uk-content-retention-and-withdrawal-archiving-policy)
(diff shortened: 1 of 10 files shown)


app/views/remove_document/remove.govspeak.md
```diff
@@ -0,0 +1,5 @@
+If you need to remove this content please [raise a support request and we’ll do it for you](https://support.publishing.service.gov.uk/).
+
+You must include any public explanation or new web address required.
+
+[Full guidance on removing content]( https://www.gov.uk/guidance/content-design/gov-uk-content-retention-and-withdrawal-archiving-policy)
```
kevindew: Looks like you have an extra space character here (and in the retire one)

benthorner: Thanks for spotting :-)

PR #344, opened 2018-09-26
Delete images

benthorner:

This adds the ability to delete a (lead) image for a document. Related changes:

   - Making the red destructive class work for buttons (553330d)
   - Making the actions for each image more flexible (853ac49)
   - Making the API errors for images more generic (d19ed2a)

Looking through the code has highlighted a few issues to address separately:

   - The lead images controller is very long and complicated
   - The update_crop action doesn't cope with Asset Manager down
   - Some of the image specs still don't validate the preview payload

### DIFF
--- a/app/assets/stylesheets/core/_link.scss
+++ b/app/assets/stylesheets/core/_link.scss
@@ -3,8 +3,9 @@
 @import "../../../node_modules/govuk-frontend/tools/all";
 @import "../../../node_modules/govuk-frontend/core/links";
 
-.app-link--destructive {
-  @include govuk-link-common;
+
+// TODO: support this modifier in govuk-frontend
+.govuk-link.app-link--destructive {
   color: $govuk-error-colour;
 
   &:hover {
--- a/app/assets/stylesheets/utilities/_display.scss
+++ b/app/assets/stylesheets/utilities/_display.scss
@@ -1,3 +1,7 @@
 .app-hidden {
   display: none;
 }
+
+.app-inline-block {
+  display: inline-block;
+}
--- a/app/controllers/document_lead_image_controller.rb
+++ b/app/controllers/document_lead_image_controller.rb
@@ -7,6 +7,7 @@ class DocumentLeadImageController < ApplicationController
 
   def create
     document = Document.find_by_param(params[:document_id])
+
     unless params[:image]
       redirect_to document_lead_image_path, alert: t("document_lead_image.index.no_file_selected")
       return
@@ -26,12 +27,10 @@ class DocumentLeadImageController < ApplicationController
       image = image_uploader.upload(document)
       image.asset_manager_file_url = upload_image_to_asset_manager(image)
     rescue GdsApi::BaseError
-      redirect_to document_lead_image_path, alert_with_description: {
-        "title" => t("document_lead_image.index.flashes.asset_manager_error.title"),
-        "description" => t("document_lead_image.index.flashes.asset_manager_error.description"),
-      }
+      redirect_to document_lead_image_path, alert_with_description: t("document_lead_image.index.flashes.api_error")
       return
     end
+
     image.save!
     redirect_to crop_document_lead_image_path(params[:document_id], image.id)
   end
@@ -44,6 +43,7 @@ class DocumentLeadImageController < ApplicationController
   def update_crop
     document = Document.find_by_param(params[:document_id])
     image = Image.find(params[:image_id])
+
     Image.transaction do
       image.update!(update_crop_params)
       asset_manager_file_url = upload_image_to_asset_manager(image)
@@ -51,6 +51,7 @@ class DocumentLeadImageController < ApplicationController
       image.asset_manager_file_url = asset_manager_file_url
       image.save!
     end
+
     begin
       if image.id == document.lead_image_id
         DocumentUpdateService.update!(
@@ -61,13 +62,15 @@ class DocumentLeadImageController < ApplicationController
         )
       end
     rescue GdsApi::BaseError
-      redirect_to document_lead_image_path(document), alert_with_description: t("document_lead_image.index.flashes.publishing_api_error")
+      redirect_to document_lead_image_path(document), alert_with_description: t("document_lead_image.index.flashes.api_error")
       return
     end
+
     if params[:next_screen] == "lead-image"
       redirect_to document_lead_image_path(document)
       return
     end
+
     redirect_to edit_document_lead_image_path(document, image)
   end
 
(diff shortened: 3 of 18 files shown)


kevindew:
There seems to have been a ghost Ben Thorner who worked on this as well.

app/controllers/document_lead_image_controller.rb
```diff
@@ -117,6 +117,30 @@ def choose_image
     redirect_to document_path(document)
   end
 
+  def delete_image
+    document = Document.find_by_param(params[:document_id])
+    image = document.images.find(params[:image_id])
+
+    begin
+      if image.id == document.lead_image_id
```
kevindew: I notice you've checked by id here and by object equivalence in https://github.com/alphagov/content-publisher/pull/344/commits/4a9c904b24b7622a94db6b17cb6aa0ac3c085d0e#diff-da8c1e6384debf22b83216459a49c33fR21 any reason for discrepancy? do we save a query at all checking by id

app/controllers/document_lead_image_controller.rb
```diff
@@ -117,6 +117,30 @@ def choose_image
...
+        DocumentUpdateService.update!(
+          document: document,
+          user: current_user,
+          type: "lead_image_removed",
```
kevindew: Is a timeline entry type? It doesn't look to be in https://github.com/alphagov/content-publisher/blob/fb0d834d356a980f0e8ff12fb0a4e68895d01d58/app/models/timeline_entry.rb#L7 but I can see it was already used in this controller before so not sure how/if it works properly

app/controllers/document_lead_image_controller.rb
```diff
@@ -117,6 +117,30 @@ def choose_image
...
+          attributes_to_update: { lead_image_id: nil },
+        )
+      end
+
+      AssetManagerService.new.delete(image)
+      image.destroy
```
kevindew: I guess we have a somewhat annoying situation here where if the request to asset manager fails you remove the item as a lead image but don't remove the image from your document. Could consider putting this in a transaction so it's an all or nothing? (though I know it's not great practice to do a http request while in a transaction)

spec/features/editing_images/delete_image_spec.rb
```diff
@@ -0,0 +1,70 @@
...
+  def then_i_see_the_document_has_no_lead_image
+    expect(page).to have_content(I18n.t("documents.show.lead_image.no_lead_image"))
+    expect(page).to have_content(I18n.t("documents.history.entry_types.lead_image_removed"))
+  end
+
+  def and_the_preview_creation_succeeded
+    expect(@request).to have_been_requested
+    expect(page).to have_link("Preview")
+
+    expect(a_request(:put, /content/).with { |req|
```
kevindew: Does this test again the request on line 63? I wonder if you need both - as it took me quite a while to work out where @request was set

spec/features/editing_images/delete_image_spec.rb
```diff
@@ -0,0 +1,70 @@
...
+  end
+
+  def then_i_see_the_document_has_no_lead_image
+    expect(page).to have_content(I18n.t("documents.show.lead_image.no_lead_image"))
+    expect(page).to have_content(I18n.t("documents.history.entry_types.lead_image_removed"))
+  end
+
+  def and_the_preview_creation_succeeded
+    expect(@request).to have_been_requested
+    expect(page).to have_link("Preview")
```
kevindew: Is this to check that we don't have this element hidden by the error message? 

If so I wonder if that's a level of knowledge in the UI that this test doesn't really need. I'd imagine checking the page doesn't have error copy is a clearer intention. What do you think?

app/assets/stylesheets/core/_link.scss
```diff
@@ -3,8 +3,7 @@
 @import "../../../node_modules/govuk-frontend/tools/all";
 @import "../../../node_modules/govuk-frontend/core/links";
 
-.app-link--destructive {
-  @include govuk-link-common;
+.govuk-link.app-link--destructive {
```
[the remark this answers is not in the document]
benthorner: I went through this with @alex-ju already and we think this is the best we can do for now. It's only necessary because our CSS is spread over multiple repos, but I've added a TODO to highlight that it should be fixed :-).

app/assets/stylesheets/utilities/_display.scss
```diff
@@ -1,3 +1,7 @@
 .app-hidden {
   display: none;
 }
+
+.app-inline {
```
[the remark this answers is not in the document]
benthorner: Fair enough - changed to 'inline-block'. We were wondering where this should go, and came up with "it's the responsibility of the thing (in this case, the form in disguise) to make itself look like a link". Perhaps "the thing" could be some sort of component, but arguably that's what the Rails tag helpers are already doing - that's one layer of magic enough for me ;-).

app/controllers/document_lead_image_controller.rb
benthorner: Ah, good spot. I was being consistent with what was in the controller. In fact, the call in the view is worse because it does an unnecessary DB query, so I changed that instead :-).

app/controllers/document_lead_image_controller.rb
benthorner: You're right - it should be `validates_inclusion_of` - will fix in a separate PR.

app/controllers/document_lead_image_controller.rb
benthorner: So this behaviour is intended (the Asset Manager down spec covers it), but I get it might not be what the user wants. As long as it leaves everything in a valid state and the user understands that their action may not have completely worked, I think it's reasonable to leave it as it is for now to avoid the additional complexity - we'd have to do another Pub API request to get the state consistent again.

spec/features/editing_images/delete_image_spec.rb
benthorner: So this is a 'pattern', although not a very good one. Unfortunately WebMock doesn't allow you to inspect the request body unless you use an `a_request` matcher, but then we miss checking the request was made to the correct content_id, so that's what 63 is doing. This is really a TODO for gds-api-adapters to support a block in `stub_publishing_api_put_content` - does that make sense?

kevindew: I think we could check this was to the correct content_id if the /content/ regex included the content_id ?

makes sense about the block but I guess for any thing complex with webmock you end up having to break out (or there'd be a lot of changes to GdsApi)

app/assets/stylesheets/utilities/_display.scss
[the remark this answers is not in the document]
alex-ju: @kevindew ideally this sort of attributes should be all set at a component level, but in this case we need it for a form that wraps a button. We don't have any components for forms yet and I doubt we'll have one as it's a wrapper element so I'm fine with using an application-level class for this. `app-inline-block` is more appropriate indeed.

PR #352, opened 2018-09-28
Iterate image handling

benthorner:
https://trello.com/c/8D48qG5D/280-minor-iterations-for-lead-image-workflow

This makes several changes to the way we deal with images:

   - Adding more specs to cover API interactions (cfee9ae, b431d4e, eeac1be, b594223, ddf2655, ec015e8, 1f7f806)
   - Invert the contextual redirect behaviour to use a more explicit 'wizard' param (66291c4)
   - Change the edit metadata button to "Save details" (in edit context) (66291c4)
   - List the lead image first in the list of images (81e980c)

Part of the aim here is to separate the handling of generic images and lead images, which made the individual endpoints simpler and also helped me to understand where we had coverage issues. It also supports a future refactoring to separate the code for generic image handling.


spec/features/editing_images/edit_image_crop_publishing_api_down_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+    and_the_publishing_api_is_down
+    and_i_crop_the_image
+    then_i_see_the_image_is_updated
+    and_the_preview_creation_failed
+  end
+
+  def given_there_is_a_document_with_images
+    document_type_schema = build(:document_type_schema, lead_image: true)
+    document = create(:document, document_type: document_type_schema.id)
+    create(:image, :in_asset_manager, document: document)
```
kevindew: Do you think this would be clearer if you explicitly specified the file used? I was wondering if the drag actually tested anything and so had to dig into the factories to work out we were defaulting to a 1000x1000 image.

app/views/document_lead_image/crop.html.erb
```diff
@@ -5,7 +5,9 @@
   <%= t("document_lead_image.crop.description") %>
 </p>
 
-<%= form_tag(crop_document_lead_image_path(@document, @image, next_screen: params[:next_screen]), method: :patch) do %>
+<%= form_tag(crop_document_lead_image_path(@document, @image, wizard: params[:wizard]), method: :patch) do %>
+  <%= hidden_field_tag :wizard, params[:wizard] %>
```
kevindew: Do we need this hidden field if it's in the query string?

app/views/document_lead_image/edit.html.erb
```diff
@@ -8,7 +8,12 @@
   no_border: true
 } %>
 
-<%= form_tag(update_document_lead_image_path(@document, @image, next_screen: params[:next_screen]), method: :patch) do %>
+<% submit_path = case params[:wizard]
+                 when "lead_image" then update_and_choose_document_lead_image_path(@document, @image)
+                 else update_document_lead_image_path(@document, @image)
+                 end %>
```
kevindew: Feels like this case should just be an if?

app/views/document_lead_image/edit.html.erb
```diff
@@ -91,7 +96,10 @@
   </div>
 
   <%= render "govuk_publishing_components/components/button", {
-    text: "Save and choose",
-    margin_bottom: true
+    margin_bottom: true,
+    text: case params[:wizard]
+          when "lead_image" then "Save and choose"
+          else "Save details"
+          end
```
kevindew: Same comment again regarding case

app/views/document_lead_image/crop.html.erb
benthorner: Ah, good spot - I was aiming for the hidden field to keep the method call short.

app/views/document_lead_image/edit.html.erb
benthorner: Sure - updated to a ternary :-) I was thinking we may have other wizards in future...

app/views/document_lead_image/edit.html.erb
benthorner: Also fixed.

spec/features/editing_images/edit_image_crop_publishing_api_down_spec.rb
benthorner: Sure - I've enhanced it a bit more to make it clearer what the crop was and how it's changed.

benthorner:
@kevindew I think I've addressed your specific comments. I'm not sure what to do about the API down specs - each has a custom behaviour and therefore it's own unique pattern of setup and UI checks. Could you clarify what you mean by 'granular', as I think that would imply having even more of them? For me the API down cases are just another route through the workflow, so I feel OK about testing them as features. Perhaps we could add a pain card to sort this out - one suggestion could be to merge groups of specs as multiple scenarios, so they can all change together.

kevindew:
@benthorner cool thanks.

Re granular, I was thinking on tests like this: https://github.com/alphagov/content-publisher/blob/9d6c187869e3aa7f79ff916d3ceb8c84de003d84/spec/features/editing_images/edit_image_crop_publishing_api_down_spec.rb we could start later in the flow (like directly at the crop point) to maybe be testing more specifically just the api down flow. But thinking this through it wouldn't actually remove much from the test. I'm probably just caught up on the lines of code each one is.

Multiple scenarios could be good for error cases when there's a high percentage of steps duplicated

PR #353, opened 2018-09-28
Disable image deletion for published documents

benthorner:

### DIFF
--- a/app/controllers/document_lead_image_controller.rb
+++ b/app/controllers/document_lead_image_controller.rb
@@ -162,6 +162,7 @@ class DocumentLeadImageController < ApplicationController
 
   def destroy
     document = Document.find_by_param(params[:document_id])
+    raise "Trying to delete image for a live document" if document.has_live_version_on_govuk
 
     begin
       DocumentUpdateService.update!(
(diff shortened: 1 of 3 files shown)


benthorner:
@kevindew I was thinking this was more of a future feature than something we need to fix - in my head, nothing appears to be broken. The user needs around deletion are a bit confusing to me and probably depend on the overall needs for editions - should an edition be a totally isolated snapshot (assets and all), or are they incremental in some ways (e.g. assets are shared between editions)? You could argue that "we're keeping everything forever", but maybe that's what we want to do.

kevindew:
@benthorner Sure, I think we think about it the same way just with different wording. From our perspective it's a feature needed, from a user perspective it might be that it needs a fix. But either why I was just keen to have a note on it as I think it's one of the few areas where we do something a little weird to get past our lack of history.

Definitely an interesting problem ahead about asset isolation/storage.

benthorner:
@kevindew I've added a card about it on the planning board, which should be a good prompt for a discussion about this with the new people in the team, when we come to review the board.

https://trello.com/c/DBtIaRyi/348-we-dont-know-how-to-handle-deletion-of-images-assets

PR #357, opened 2018-10-01
Split the document images controller in two

benthorner:

The document (lead) images controller was getting large and complicated,
with a mixture of image- and lead-image-specific operations. Combing the
endpoints has lead to a few problems:

  * the routes don't make sense e.g. delete_document_lead_image_path has
nothing to do with the lead image
  * the redirect for for deleting lead images is inconsistent with the remove
workflow

Splitting the endpoints into two controllers - one for image operations
and another for lead image operations - should make it clearer to reason
about what each endpoint should do.

This split also makes the image pages more generic, moving towards an
'image management' workflow rather than something specific to lead
images.

### DIFF
--- a/app/controllers/document_images_controller.rb
+++ b/app/controllers/document_images_controller.rb
@@ -45,7 +45,6 @@ class DocumentImagesController < ApplicationController
     document = Document.find_by_param(params[:document_id])
     image = Image.find(params[:image_id])
 
-
     begin
       Image.transaction do
         image.update!(update_crop_params)
--- a/app/controllers/document_images_controller.rb
+++ b/app/controllers/document_images_controller.rb
@@ -1,0 +1,145 @@
+# frozen_string_literal: true
+
+class DocumentImagesController < ApplicationController
+  def index
+    @document = Document.find_by_param(params[:document_id])
+  end
+
+  def create
+    document = Document.find_by_param(params[:document_id])
+
+    unless params[:image]
+      redirect_to document_images_path, alert: t("document_images.index.no_file_selected")
+      return
+    end
+
+    image_uploader = ImageUploader.new(params.require(:image))
+
+    unless image_uploader.valid?
+      redirect_to document_images_path, alert: {
+        "alerts" => image_uploader.errors,
+        "title" => t("document_images.index.error_summary_title"),
+      }
+      return
+    end
+
+    begin
+      image = image_uploader.upload(document)
+      image.asset_manager_file_url = upload_image_to_asset_manager(image)
+    rescue GdsApi::BaseError => e
+      Rails.logger.error(e)
+      redirect_to document_images_path, alert_with_description: t("document_images.index.flashes.api_error")
+      return
+    end
+
+    image.save!
+    redirect_to crop_document_image_path(params[:document_id], image.id, wizard: params[:wizard])
+  end
+
+  def crop
+    @document = Document.find_by_param(params[:document_id])
+    @image = @document.images.find(params[:image_id])
+  end
+
+  def update_crop
+    document = Document.find_by_param(params[:document_id])
+    image = Image.find(params[:image_id])
+
+
+    begin
+      Image.transaction do
+        image.update!(update_crop_params)
+        asset_manager_file_url = upload_image_to_asset_manager(image)
+        delete_image_from_asset_manager(image)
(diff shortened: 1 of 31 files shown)


app/controllers/document_images_controller.rb
```diff
@@ -0,0 +1,140 @@
...
+  def crop
+    @document = Document.find_by_param(params[:document_id])
+    @image = @document.images.find(params[:image_id])
+  end
+
+  def update_crop
+    document = Document.find_by_param(params[:document_id])
+    image = Image.find(params[:image_id])
+
+
```
kevindew: Super minor, but we seem to have an extra new line here

app/controllers/document_lead_image_controller.rb
```diff
@@ -5,79 +5,6 @@ def index
...
-
-    redirect_to document_lead_image_path(document)
-  end
-
-  def edit
-    @document = Document.find_by_param(params[:document_id])
-    @image = Image.find_by(id: params[:image_id])
-  end
-
   def update
```
kevindew: update and destroy seem very similar to those in document_images controller? Is there a way we can have document_images_controller handle them to avoid the duplication? 

Otherwise feels like we may we want to consider a service class.

app/views/document_images/_image_list.html.erb
```diff
@@ -15,9 +15,17 @@ end %>
     <% end %>
   <% end %>
 
-  <% delete_image_button = capture do %>
-    <%= form_tag delete_document_lead_image_path(@document, image), method: :delete, class: "app-inline-block" do %>
-      <button class="govuk-link app-link--destructive">Delete image</button>
+  <% if image.id == @document.lead_image_id %>
```
kevindew: The way this is presented is a little bit confusing with the capture assigned to same variable name in each condition. 

I wonder if it'd be easier to follow if it was `<% delete_image_button = capture do %>` and then the conditional inside that?

app/controllers/document_lead_image_controller.rb
benthorner: Indeed they are, but there are two differences that make them difficult to combine - the redirect and the lead_image_id. We needed conditionals and in a couple of instances implemented the workflow for one case but not the other (e.g. for lead images but not other images).

But I agree there's some duplication here, and I'd be up for centralising update/destroy into an ImageService or similar. If that sounds good to you, I can do another commit for it? The ImageUploader service could also be a good thing to refactor into this new service.

benthorner:
@kevindew Sorry about the commit - it was easy-ish to do but I realise I could have done it action-by-action, which would have been easier for you to read. I'll try to remember that :-).

benthorner:
@kevindew I get your comment about the exceptions - it would make it clearer and support customisable handling depending on what's down - but perhaps we should be motivated by a user need for this?

app/controllers/document_lead_image_controller.rb
kevindew: Sure, could be worth spiking on it to see if that's a nicer way to handle the conditional aspect of it.

On first thought it seemed pretty ripe for a service class since we have about 20 lines of code: https://github.com/alphagov/content-publisher/blob/10890b6cbbc047cbce6951f8fc5fbc8ceb7d2208/app/controllers/document_images_controller.rb#L9-L35 but then actually looking at the code about 3/4 is handling different responses so it might well be that there isn't a lot to gain? What's your thoughts?

kevindew:
@benthorner yeah not something that's necessary now. I wasn't thinking of it so much as a user need though more to just slim down the code

e.g. we have this block in similar ways 4 times with same error behaviour

https://github.com/alphagov/content-publisher/blob/10890b6cbbc047cbce6951f8fc5fbc8ceb7d2208/app/controllers/document_images_controller.rb#L26-L32

whereas we could have a single `rescue_from PublishingApiFailedError, with: -> { redirect_to document_images_path, alert_with_description: t("document_images.index.flashes.api_error") }` if the error was more specific.

benthorner:
@kevindew As we discussed, I've refactored the controller exceptions using `rescue_from`. For the update/destroy actions, on closer inspection there's still enough difference between the code for images and lead images that combining them into a service would introduce more complexity. I'd like to leave that for now - they may even diverge in future when we look to handle deletion properly ;-).

benthorner:
Yay :-)

PR #360, opened 2018-10-02
Fix locale for unknown user

alex-ju:
Before
screen shot 2018-10-02 at 10 40 30

After
screen shot 2018-10-02 at 10 40 53

### DIFF
--- a/app/views/documents/index/_results.html.erb
+++ b/app/views/documents/index/_results.html.erb
@@ -28,7 +28,7 @@
               },
               metadata: {
                 document_type: document.document_type_schema.label,
-                document_create: t("documents.index.search_results.creator", creator: document.creator&.name || I18n.t("documents.index.search_results.unknown_creator")),
+                document_create: t("documents.index.search_results.creator", creator: document.creator&.name || I18n.t("documents.index.search_results.unknown_user")),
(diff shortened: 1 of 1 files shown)


app/views/documents/index/_results.html.erb
```diff
@@ -28,7 +28,7 @@
               },
               metadata: {
                 document_type: document.document_type_schema.label,
-                document_create: t("documents.index.search_results.creator", creator: document.creator&.name || I18n.t("documents.index.search_results.unknown_creator")),
+                document_create: t("documents.index.search_results.creator", creator: document.creator&.name || I18n.t("documents.index.search_results.unknown_user")),
```
tijmenb: Instead of `I18n.t()` we should use `t()`, here. `t()` raises an exception when a translation isn't found, but `I18n.t()` doesn't (`I18n.t!()` will also do that).

alex-ju: @tijmenb good to know. thanks!

kevindew: Ahh! that's why https://github.com/alphagov/content-publisher/blob/master/config/application.rb#L48 wasn't catching it.

PR #365, opened 2018-10-08
Remove last image border and fix bottom padding

benthorner:
https://trello.com/c/8D48qG5D/280-minor-iterations-for-lead-image-workflow

### DIFF
--- a/app/assets/stylesheets/components/_image_meta.scss
+++ b/app/assets/stylesheets/components/_image_meta.scss
@@ -6,12 +6,16 @@
   @include govuk-font(19);
 
   margin-bottom: govuk-spacing(4);
-  padding-bottom: govuk-spacing(3);
+  padding-bottom: govuk-spacing(4);
   border-bottom: 1px solid $govuk-border-colour;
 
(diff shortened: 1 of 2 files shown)


benthorner:
Good spot @alex-ju. Through some crystal ball you seem to have, I discovered you already added a 'no_border' attribute to the 'image_meta' component, which provides a way to achieve the design with the unintended side-affect, which @soniaturcotte wants to avoid.

alex-ju:
@benthorner I'm a bit confused; I was more happy with the CSS approach though.

benthorner:
@alex-ju oh, I thought you'd prefer using the attribute :-| We still want a bottom border for the lead image on the show page, but there's no reliable difference in the DOM between the image_meta on the show page and the image_meta for the images list, so I think we need to tell the last item that it shouldn't have a bottom border, which I was surprised to find you already had an attribute for :-).

alex-ju:
It's totally fine, I added the attribute for this type of cases indeed. As long as it's following the design I'm happy with it. 👍

PR #370, opened 2018-10-08
Log exceptions so we can fix them ;-)

benthorner:
This ensures we call 'Rails.logger.error(e)' whenever we rescue an
exception, so that we have sufficient information to fix the error. Most
of the errors are 'Intermittent retryable' and not right for Sentry.

https://github.com/alphagov/govuk-rfcs/blob/master/rfc-087-dealing-with-errors.md#intermittent-retryable-errors

This also removes an untested rescue for when we cannot generate a
unique path for a document, as this behaviour was not consistent for the
no-javascript path and it feels like we should solve this problem first.

This also removes Sentry reporting for exceptions originating from the
views, as we anticipate these are 'Intermittent errors with user
impact', such as timeouts, which don't require immediate attention.

### DIFF
--- a/app/controllers/document_lead_image_controller.rb
+++ b/app/controllers/document_lead_image_controller.rb
@@ -26,7 +26,8 @@ class DocumentLeadImageController < ApplicationController
     begin
       image = image_uploader.upload(document)
       image.asset_manager_file_url = upload_image_to_asset_manager(image)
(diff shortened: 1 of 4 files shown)


benthorner:
@kevindew I was thinking the same thing ;-).

PR #372, opened 2018-10-09
Calculate valid crop height instead of validating

benthorner:
https://trello.com/c/WQYgUVRO/274-look-into-sentry-errors-related-to-image-uploads

This fixes https://sentry.io/govuk/app-content-publisher/issues/718579676/?query=is:unresolved.

Previosuly we tried to validate whether the given crop_height and
crop_width conformed to the ideal aspect ratio determined by the
Image::WIDTH and Image::Height. This had a few unexpected edge cases, so
rather than make the validator more complicated we instead compute the
crop_height from the crop_width using the ideal aspect ratio.

We also decided the other image validations were unnecessary and removed
these, as the worst case scenario of them being invalid should only
affect individual users, and we don't anticipate a developer wanting to
manually override image attributes using the console.


app/models/image.rb
```diff
@@ -13,21 +13,6 @@ class Image < ApplicationRecord
...
-            numericality: { only_integer: true, greater_than_or_equal_to: 0 }
-  validates :crop_y,
-            numericality: { only_integer: true, greater_than_or_equal_to: 0 }
-  validates :crop_width,
-            numericality: { only_integer: true, greater_than_or_equal_to: WIDTH }
-  validates :crop_height,
-            numericality: { only_integer: true, greater_than_or_equal_to: HEIGHT }
-
-  validates_with ImageAspectRatioValidator
-
```
DilwoarH: Should it still do the validation?

benthorner: Good question - I think I addressed this in the second paragraph of the commit. Because we're confident the UI will enforce the image constraints, we don't feel it adds value to have these validations here. The other possibility is a programmer might try to set these values when debugging something, but we think this is unlikely for images ;-).

DilwoarH: I think the best rule to follow is to never trust the browser, however, as you are working out the height based on the width - I think it might be fine.

app/controllers/document_images_controller.rb
```diff
@@ -119,6 +119,8 @@ def delete_image_from_asset_manager(image)
   end
 
   def update_crop_params
-    params.permit(:crop_x, :crop_y, :crop_width, :crop_height)
+    image_aspect_ratio = Image::HEIGHT.to_f / Image::WIDTH
+    crop_height = params[:crop_width].to_i * image_aspect_ratio
```
DilwoarH: Are the images always horizontal? would they ever be verticle?

app/models/image.rb
benthorner: I think you've got @kevindew 's thoughts on this ;-). You're right that we should ideally have a validator on the critical path here; we do this for documents, because we anticipate the user doing something wrong, but as you say, the use shouldn't be able to specify an invalid crop.

app/controllers/document_images_controller.rb
benthorner: I'm not sure what you mean here - could you elaborate? The image constants imply a horizontal image orientation, if that helps.

PR #374, opened 2018-10-10
Remove no image text when no images

1pretz1:
Remove the text of "No images uploaded" when we have no images

This applies to documents which have no images on the upload images
page.

## Before
screen shot 2018-10-09 at 16 36 53

## After 
screen shot 2018-10-09 at 16 37 26

Trello: 
https://trello.com/c/IMdJcbEs/303-remove-the-text-of-no-images-uploaded-when-we-have-no-images

### DIFF
--- a/spec/features/editing_images/upload_lead_image_asset_manager_down_spec.rb
+++ b/spec/features/editing_images/upload_lead_image_asset_manager_down_spec.rb
@@ -33,6 +33,6 @@ RSpec.feature "Upload a lead image when Asset Manager is down" do
   end
 
   def and_the_image_does_not_exist
-    expect(page).to have_content(I18n.t("document_images.index.no_existing_image"))
+    expect(page).to_not have_content(I18n.t("document_images.index.existing_image"))
   end
 end
--- a/app/views/document_images/index.html.erb
+++ b/app/views/document_images/index.html.erb
@@ -34,10 +34,6 @@
         <%= t("document_images.index.existing_image") %>
       </h2>
       <%= render "image_list" %>
-    <% else %>
-      <h2 class="govuk-heading-m">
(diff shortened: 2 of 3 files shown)


1pretz1:
@benthorner I'm not sure i know what you mean? Are you referring to using something like the stub method `asset_manager_does_not_have_an_asset` to assert an asset has not been created?

spec/features/editing_images/upload_lead_image_asset_manager_down_spec.rb
```diff
@@ -33,6 +33,6 @@ def then_i_should_see_an_error
   end
 
   def and_the_image_does_not_exist
-    expect(page).to have_content(I18n.t("document_images.index.no_existing_image"))
+    expect(page).to_not have_content(I18n.t("document_images.index.existing_image"))
```
benthorner: Sorry for being vague. It’s this line here. It’s very easy for a page to not have content - maybe we change the workflow so it’s never there anyway. Do you think there’s a positive way to check there’s no image?

1pretz1: Ah okay, understood!

1pretz1: Hmmm well you could check that there is an image tag with the css class but this may be flakey due to changes in the styling.

1pretz1: Ah ah, just found that content-publisher has a table `images` which keeps a record of each image uploaded. So could check there to make sure nothing has been created!

1pretz1: I'll merge this then make the tests better in another pr

PR #376, opened 2018-10-11
Handle no-javascript/non-croppable experience for lead images

alex-ju:
## Common experience
screen shot 2018-10-11 at 11 20 27

## No-javascript experience
The image is automatically resized and cropped center-based

screen shot 2018-10-11 at 11 20 04

## Fixed dimensions experience
The image is displayed as is without the cropping functionality

screen shot 2018-10-11 at 11 25 19

[Trello card](https://trello.com/c/Mx0dBCHb)

### DIFF
--- a/app/views/document_images/crop.html.erb
+++ b/app/views/document_images/crop.html.erb
@@ -16,8 +16,18 @@
     crop_height: @image.crop_height,
   } %>
 
-  <%= render "govuk_publishing_components/components/button", {
-    text: (@image.at_exact_dimensions?) ? "Continue" : "Crop image",
-    margin_bottom: true
-  } %>
+  <div class="app-js-only">
+    <%= render "govuk_publishing_components/components/button", {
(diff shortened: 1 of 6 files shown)


app/views/document_images/crop.html.erb
```diff
@@ -1,25 +1,26 @@
...
     crop_x: @image.crop_x,
     crop_y: @image.crop_y,
     crop_width: @image.crop_width,
-    crop_height: @image.crop_height
+    crop_height: @image.crop_height,
   } %>
 
   <%= render "govuk_publishing_components/components/button", {
-    text: "Crop image",
+    text: (@image.at_exact_dimensions?) ? "Continue" : "Crop image",
```
benthorner: I think we also have a weirdness here: if JavaScript is disabled, we still say 'Crop image', even though the description above says it has already been 'automatically resized to fit'. Perhaps one for @benhazell :-).

alex-ju:
@benthorner updated; we now have the copy inside the component as per ADR.

PR #378, opened 2018-10-11
Make upload_lead_image_asset_manager_down_spec more robust

1pretz1:
This test now checks that an image is not displayed on the page via the
css and that an image has not been created in the db. Previously, this
test only looked for some text that may or may not change in the future.


1pretz1:
Cool 👍

PR #379, opened 2018-10-12
Add drafting requirements

1pretz1:
Adds drafting requirements to a document, currently meaning that a document has to have the presence of a title which is no more than 150 characters long and spans only a single line. If these validations fail the user won't proceed to the next page and will be shown the relevant error messages as a flash and on the input boxes.

Also, the trimming behaviour has been removed as this was deemed not necessary and not a great user experience.

## Preview 
screen shot 2018-10-19 at 11 22 58

Trello:
https://trello.com/c/sAXUVPxo/275-make-title-a-compulsory-field-to-avoid-preview-errors


benthorner:
@1pretz1 there's a useful concept on the card ('drafting requirements'), which I think we should put into the codebase here. This is also a separate feature from the existing 'create a document' happy path, so let's make a separate spec for it as well :-). Another thing is that there are some comments on the card, which talk about what the design is meant to be - you've got the flash, but we're missing the other error indicator that Sonia pointed out. Let me know if you need some more clarification ;-).

1pretz1:
@1pretz1 yep i've just spotted the comments, changes incoming 💨

spec/features/workflow/drafting_requirements_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+
+RSpec.feature "Drafting requirements" do
+  scenario do
+    given_there_is_a_document_with_no_content
+    when_i_visit_the_edit_document_page
+    and_save_without_inputting_anything
+    then_i_should_see_an_error_message_regarding_a_missing_title
+
+    when_i_visit_the_edit_document_page
+    and_save_a_title_which_doesnt_meet_the_validations
```
benthorner: Can we make these more specific and split them up?

spec/features/workflow/drafting_requirements_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+    then_i_should_see_error_messages_regarding_validations
+
+    when_i_visit_the_edit_document_page
```
benthorner: We don't need this one as it's covered by other features :-)

config/locales/en/drafting_requirements.yml
```diff
@@ -0,0 +1,7 @@
+en:
```
benthorner: We can push this into the edit.yml for documents as this copy will only appear on this page.

1pretz1: sure

spec/features/workflow/drafting_requirements_spec.rb
1pretz1: Good spot

spec/features/workflow/drafting_requirements_spec.rb
1pretz1: Yep, previously they were split up and specific but due to adding more and more validations, there is a risk of having a realllllllly long test file. But, as there are only a few, your approach makes more sense.

app/helpers/drafting_requirements_helper.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
```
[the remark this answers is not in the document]
1pretz1: 👍

PR #385, opened 2018-10-17
Adds markdown toolbar to the editor

DilwoarH:
## User story
As a publisher
I need to add and format content with markdown
So that it appears on GOV.UK

## Background
We've iterated our markdown approach in Q2 and have prioritised the tool bar approach for private beta. We would like to test this approach in our first round  of research for Q2 next week.

The aim here is to make it easier for publishers to add markdown. We know from data and the survey that publishers spend time on finding the best markdown and editing the markdown. Also, for longer documents they spend lots of the time finding the errors in markdown.

Ticket: https://trello.com/c/TvmLqnxi/332-add-toolbar-to-markdown-editor

### Normal
screen shot 2018-10-18 at 12 02 46

### Hover state
screen shot 2018-10-18 at 12 02 53

### Focus state
screen shot 2018-10-18 at 14 26 05


huminmiah:
This looks great! A few comments: 

- Can we add hover text for people who don't know what the icons mean 
- The colour of the tool bar should be consistent to the shade of the edit tab. Currently, it looks like the tool bar is part of the preview section

@benhazell can you provide the content for the right hand side.

app/assets/stylesheets/components/_markdown-editor.scss
```diff
@@ -79,5 +79,36 @@
 
   .govuk-textarea {
     overflow: scroll;
+    padding-top: govuk-spacing(8);
   }
 }
+
+.app-c-markdown-editor-toolbar__icon {
+  display: inline-block;
```
alex-ju: This should be `display:block;` and `.app-c-markdown-editor-toolbar__button` should be `inline-block;`. This way you won't have alignment issues and the focus will be rectangular.

alex-ju: The buttons should follow the focus pattern we have with other components. You can achieve that with `@include govuk-focusable;` in `.app-c-markdown-editor-toolbar__button`

benthorner:
I notice we now have a toolbar in the editor, and a separate component called 'markdown toolbar', which is confusing. Not really part of this PR, but just making sure you're aware and consider renaming the toolbar ('sidebar') one :-)

app/assets/stylesheets/components/_markdown-editor.scss
DilwoarH: Hi alex, this has all been done.

PR #386, opened 2018-10-17
Replace autocompletes with selects for filters

alex-ju:
This PR replaces the autocomplete components with selects for filters in the index page.

The reason we're not using the select component is that the data doesn't come in a form that's usable for the component. It currently comes as two arrays one with for the options values and another with the selected items. The component takes a single array with a boolean flag (`selected: true`) for the selected options.

[Trello card](https://trello.com/c/DjDrH3lY)

### DIFF
--- a/app/views/documents/index/_filters.html.erb
+++ b/app/views/documents/index/_filters.html.erb
@@ -9,67 +9,66 @@
     type: "search"
   } %>
 
-  <% document_type_select = DocumentTypeSchema.all
-    .reject(&:managed_elsewhere)
-    .map { |d| [d.label, d.id] }
-  %>
+  <div class="govuk-form-group">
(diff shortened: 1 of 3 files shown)


app/views/documents/index/_filters.html.erb
```diff
@@ -9,67 +9,63 @@
...
-    selected_options: [params[:organisation]],
-    size: 10,
-    data: {
-      module: "autocomplete"
-    }
-  } %>
+    } %>
+
+    <%= select_tag "organisation",
+      options_for_select([["", ""]] + organisation_select, [params[:organisation]]),
```
benthorner: I think we can replace the empty tuple with 'include_blank: true'.

alex-ju: Updated to use `include_blank` 👍

PR #387, opened 2018-10-17
Fix component guide

alex-ju:
Now 5 out of 13 components can not be previewed in the component guide.

This PR:
- Ensure components preview
- Ensure SCSS dependencies using imports
- Add examples for components (where not previously defined)
- Uses consistent file name for SCSS files

This PR doesn't:
- Deal with `Markdown raw`, which doesn't have a template file (only yaml and scss)
- Enable JavaScript support for components in the component guide (needs more investigation)


alex-ju:
@benthorner added examples to cover the mentioned attributes for image_meta component

PR #392, opened 2018-10-19
Show topics

benthorner:
https://trello.com/c/WdQGc5jT/330-allow-a-user-to-see-a-complete-hierarchy-of-govuk-taxons

This adds a new section to the document summary that shows a complete bulleted hierarchy of the entire GOV.UK taxonomy, with support for the Publishing API being down.

screen shot 2018-10-19 at 12 11 48

 This adds a new TopicsService to iteratively fetch and cache the topic
 taxonomy from the Publishing API, which is then shown using nested
 bullets on a new topics index page. All the naming uses the term topic
 instead of taxon to be consistent with what we show to the user.
 
 The GOV.UK taxonomy is a tree with many levels, most of which are accessed
 through child_taxons links, with the exception of level_one_taxons.
 Because level_one_taxons are not fully expanded for the GOV.UK homepage,
 building the entire topic tree is an involved, multi-request process.
 
 The load time without caching is similar to the tags page, which should
 be acceptable for now - we can take the Whitehall approach and adopt
 Redis with async refresh if we need to in future.


app/services/topics_service.rb
```diff
@@ -0,0 +1,47 @@
...
+    publishing_api.get_expanded_links(GOVUK_HOMEPAGE_CONTENT_ID)
+      .dig("expanded_links", "level_one_taxons")
+      .to_a
+  end
+
+  def publishing_api
+    GdsApi::PublishingApiV2.new(
+      Plek.new.find("publishing-api"),
+      disable_cache: true,
+      timeout: 60,
```
kevindew: this seems a heck of a timeout for something running at web request, may as well set it way lower since the request will be terminated long before it reaches.

spec/features/editing_topics/edit_topics_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+                {
+                  "title" => "Level Three Topic",
+                  "links" => {},
+                },
+              ],
+            },
+          },
+        ],
+      },
+    )
```
kevindew: This is pretty beastie. I think if we need to do this again for a different test we should create a helper.

app/views/document_topics/edit.html.erb
```diff
@@ -0,0 +1,22 @@
+<% content_for :back_link, document_path(@document) %>
+<% content_for :title, t("document_topics.edit.title", title: @document.title_or_fallback) %>
+
+<% def unroll(topic) %>
+  <% capture do %>
+    <li>
+      <%= topic[:title] %>
+
+      <ul>
+        <% topic[:children].each do |child_topic| %>
```
kevindew: Can this sometimes create an empty list on the last item when `topic[:children]` is empty?

app/services/topics_service.rb
benthorner: The timeout is what's in Whitehall, but thinking back I recall it was arbitrary. Makes sense to remove unless it's needed 👍 .

spec/features/editing_topics/edit_topics_spec.rb
benthorner: We think alike here :-). I thought I'd wait for further specs so we have more working examples.

app/services/topics_service.rb
kevindew: Sure, makes sense for Whitehall if it's running asynchronously. Probably want a really low one rather than none.

app/views/document_topics/edit.html.erb
benthorner: It seems like the 'child_taxons' link is either present or not, so this shouldn't happen. But I agree it's logical to see the code there anyway ;-).

app/services/topics_service.rb
benthorner: Well it seemed to work fine without - even to a remote machine. The [default timeout is 4 seconds](https://github.com/alphagov/gds-api-adapters/blob/a0c99d4895057be3cd2590a486a5e24cea49548f/lib/gds_api/json_client.rb#L150), which seems OK to me.

benthorner:
@kevindew updated - how does it look now?

benthorner:
Will wait until we've done https://trello.com/c/pt9c438U/361-feature-flag-for-upcoming-user-research until this is merged.

benthorner:
@kevindew we've now changed this to use a lookup table approach, which is similar in length/complexity but should make doing https://trello.com/c/5tAUflyY/331-show-current-taxons-for-a-document a lot easier. Could you take another look?

PR #393, opened 2018-10-19
Add sidebar to markdown text editor

DilwoarH:
As a publisher
I need to add and format content with markdown
So that it appears on GOV.UK

This is only to add the sidebar to the markdown editor found in the create and edit screen in content publisher.

Ticket: https://trello.com/c/EHO8aGQs/345-add-sidebar-to-markdown-text-editor

### Screenshots

screenshot-content-publisher dev gov uk-2018 10 19-13-01-24


app/formats/markdown_guidance.yml
```diff
@@ -0,0 +1,83 @@
+---
```
benthorner: Is this used anywhere?

benthorner:
@soniaturcotte I think @DilwoarH opened them all manually to take the screenshot ;-).

soniaturcotte:
> @soniaturcotte I think @DilwoarH opened them all manually to take the screenshot ;-).

Yes I figured that, but we should only be opening one at a time. So if you have acronyms open, and you select tables, it automatically closes acronyms. I think this might come with the accordion component. Is that right @alex-ju

soniaturcotte:
I would say option one is fine if we are going to get the component in soon. 

Final comment, the copy at the beginning should be 

**Formatting**
[Full markdown guidance](https://docs.google.com/document/d/1RHvAf1MEjwOuR4POpcLAkhBQWLejYhe4WOyVCAdCb0I/edit)

As per the content document: https://docs.google.com/document/d/1RHvAf1MEjwOuR4POpcLAkhBQWLejYhe4WOyVCAdCb0I/edit

DilwoarH:
@soniaturcotte - updated

screen shot 2018-10-19 at 14 27 01

benthorner:
@alex-ju we had a think about using a locale file, but without going all the way and creating a custom DSL. Most of the guidance could be specified using markdown and injected into the component after rendering it, but we thought it wasn't worth the effort for now - let's see how often it changes :-).

DilwoarH:
The decision to not include the guidance was discussed in length with @benthorner. The main reason why this was not done was that of the following:
- Markdown guidance is not content that will change very frequently and there work/effort/time required would not be justified
- The markdown guidance is a part of a component which means logic would need to be implemented on a component level and the current structure of components do not support an additional file for content.

PR #396, opened 2018-10-19
Remove redundant test

1pretz1:
This has been removed because all the test coverage in this file has
been covered by the publishing requirements workflow feature.


benthorner:
Although technically this only becomes redundant when we've removed the min length validations, but it's a step in the right direction :-).

PR #401, opened 2018-10-22
Add organisation as a custom dimension via dataLayer

alex-ju:
This PR adds current user's organisation to `dimension1` (which is configured to store organisations in GTM) via dataLayer.

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -6,6 +6,9 @@
   <% end %>
 
   <% if ENV["GOOGLE_TAG_MANAGER_ID"] %>
+    <script>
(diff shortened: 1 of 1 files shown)


app/views/layouts/application.html.erb
```diff
@@ -6,6 +6,9 @@
   <% end %>
 
   <% if ENV["GOOGLE_TAG_MANAGER_ID"] %>
+    <script>
+      dataLayer = [{ 'dimension1': <%= current_user.organisation_slug %> }];
```
benthorner: The developer guide suggests we use meaningful keys for the data layer keys. Is there a particular reason for calling it 'dimension1'?

https://developers.google.com/tag-manager/devguide#adding-data-layer-variables-to-a-page

PR #394, opened 2018-10-19
Remove preview back button

benthorner:

PR #408, opened 2018-10-23
Shorten and standardise spec names a bit

benthorner:
This removes some of the redundant words from the spec file names and
standardises others:

   - Remove occassional uses of 'a' e.g. 'show_a_document'
   - Prefix 'api_down' with 'publishing_' or 'asset_'
   - Remove usages of 'a_document' for all workflow specs
   - Change occassional uses of 'viewing/showing' to 'show'

Using fewer words should make the file tree easier to scan, and make the
file naming a bit clearer for new specs, which is what I was attempting
to do before starting this commit :-).


benthorner:
Looks like this accidentally got merged as part of https://github.com/alphagov/content-publisher/pull/394.

PR #411, opened 2018-10-23
Debug permission flag

kevindew:
Trello: https://trello.com/c/pt9c438U/361-feature-flag-for-upcoming-user-research

This adds a "debug" permission which can be used by signon for features that provide information that shouldn't be available to normal users of the app and just to provide functionality for someone learning more about the app and how it works.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,6 +33,22 @@ yarn install
 bundle exec rake
 ```
 
+### Permissions
+
+Functionality of this application is enabled with permissions. There is a
+`pre_release_features` permission, for using functionality not yet available to
+users, and there is a `debug` permission, to access debug and documentation in
+app.
+
+To enable these permissions in a development environment run:
+
+```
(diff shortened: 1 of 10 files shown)


README.md
```diff
@@ -22,17 +22,45 @@ This is a Ruby on Rails application.
...
 ### Running the test suite
 
 ```
 yarn install
 bundle exec rake
 ```
 
+### Permissions
+
+Functionality of this application is enabled with permissions. There is a
```
benthorner: Could we just tell them to run `rake db:seed` and put the permissions in there?

kevindew: The problem we've got is that if you clone data then you won't have run (or plan to run) db:seed.

benthorner: Yup. I think I forgot the seed deliberately doesn't have the 'debug' permission, so we need a separate task for it.

PR #418, opened 2018-10-25
Restrict access to managing topics to pre-release features

kevindew:
Trello: https://trello.com/c/pt9c438U/361-feature-flag-for-upcoming-user-research

This PR is to merge into https://github.com/alphagov/content-publisher/pull/392 so that can move from DO NOT MERGE. I intend to then update https://github.com/alphagov/content-publisher/pull/411 to be consistent with approaches in this.

See commits for details.

### DIFF
--- a/app/controllers/document_topics_controller.rb
+++ b/app/controllers/document_topics_controller.rb
@@ -1,6 +1,9 @@
 # frozen_string_literal: true
 
 class DocumentTopicsController < ApplicationController
+  include GDS::SSO::ControllerMethods
+  before_action { authorise_user!(User::PRE_RELEASE_FEATURES_PERMISSION) }
+
   rescue_from GdsApi::BaseError do |e|
     Rails.logger.error(e)
     render "#{action_name}_api_down", status: :service_unavailable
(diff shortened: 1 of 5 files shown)


kevindew:
The testing aspect is a bit of a pain I agree and does feel a bit overkill in the feature tests. Does feel though like we should have something that tests that it can't be accessed without the permission.

I do expect this to be a pattern we use multiple times though particularly as we go into beta rather than limited to taxonomy.

How'd you feel about if we add this as a default permission and then only modify the test to check that you don't have the permission?

benthorner:
@kevindew that sounds like an easier approach to me. I can also see that, while it's not that important right now, we'll want to be more careful about testing what users can see when we go live ;-).

I did have another thought, which is about those additional user methods: technical the permissions aren't mutable, as they're assigned by signon. But I think having those methods implies they are, but actually those methods shouldn't be called by the code itself. I'm going to go with 'objective' on this one, and ask: would it be OK to remove them?

kevindew:
@benthorner re: the user methods, the values are inherently mutable. It's just a shortcut to the means to do it. Though you are right that the code shouldn't be used outside of testing. The other option was a test helper for this - Are you happier with that?

All ways feel a bit clunky as I don't think anyone has thought of a nice way of integrating permissions as part of feature tests for GDS-SSO

benthorner:
@kevindew that sounds better to me. I guess we don't need the 'add' method anymore anyway, so even shorter :-).

kevindew:
@benthorner I've changed this up now. Didn't go with a helper after all since it was just one usage, will add one if/when it expands.

PR #420, opened 2018-10-26
Fix markdown buttons break line of content after first word

alex-ju:
We found in user research that publishers don't select the item they want to style, but rather position the cursor at the beginning of a line. With this approach, after pressing a markdown toolbar button (i.e. list) it currently makes the content break with a new line after the first word.

This PR fixes the issue by removing the new inserted line and applies to all the current items (h2, h3, blockquote and lists) except for for link which is an inline wrapper injector.

The updated code comes from https://github.com/alphagov/markdown-toolbar-element. Plan to do releases and changelog soon, so I can point to those for changes instead.

[Trello card](https://trello.com/c/oVHHFHWJ)

### Before
screen shot 2018-10-26 at 10 03 04

### After
screen shot 2018-10-26 at 10 03 42

### DIFF
--- a/app/assets/javascripts/vendor/@github/markdown-toolbar-element/dist/index.umd.js
+++ b/app/assets/javascripts/vendor/@github/markdown-toolbar-element/dist/index.umd.js
@@ -158,17 +158,77 @@
     window.customElements.define('md-header', MarkdownHeaderButtonElement);
   }
 
-  var MarkdownBoldButtonElement = function (_MarkdownButtonElemen2) {
-    _inherits(MarkdownBoldButtonElement, _MarkdownButtonElemen2);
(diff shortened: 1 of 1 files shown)


DilwoarH:
Hey Alex,

Had a look, this is the result I get:

oct-30-2018 10-49-49

DilwoarH:
I think to solve the problem, we need to modify the following function https://github.com/alphagov/markdown-toolbar-element/blob/master/dist/index.esm.js#L579

alex-ju:
@DilwoarH I deployed a [standalone version of the markdown toolbar element](https://alphagov.github.io/markdown-toolbar-element/examples/) and works as expected. I'm unable to deploy my branch not due to issues with vagrant and Heroku deploy isn't working so let's review it together tomorrow.

PR #421, opened 2018-10-29
Enhance ability to detect live publications with imported whitehall d…

kgarwood:
Trello: https://trello.com/c/vAsVUGzV/245-import-hasliveversionongovuk-from-whitehall

These changes are about enhancing the importing task for whitehall records to they are better able to determine if a document has any edition that is already live.

A document will have a live edition of some kind if it meets one of two
criteria:
(1) the publication state of the imported document record is 'published'
(2) it has at least two editions.

If it has a live edition, these flags must be set properly:
- publication_state must be set to 'sent-to-live'
- has_live_version_on_govuk must be set to true
- review state must be set to 'reviewed'

Previously the code was only checking for criterion (1).

Main changes:
- add document as parameter to methods that determine flag states and
  use the edition count method to support criterion (2)
- modify tests to accommodate criterion (2)
- modify most tests so they also check for has_live_version_on_govuk state
- change wording of test case descriptions because now we have three
  states to check which seem linked

### DIFF
--- a/spec/tasks/whitehall_news_importer_spec.rb
+++ b/spec/tasks/whitehall_news_importer_spec.rb
@@ -123,7 +123,7 @@ RSpec.describe Tasks::WhitehallNewsImporter do
     expect { importer.import(parsed_json) }.not_to(change { Document.count })
   end
 
-  context "when an imported document has more than on edition" do
+  context "when an imported document has more than one edition" do
     let(:import_published_then_drafted_data) do
       {
         content_id: SecureRandom.uuid,
@@ -178,7 +178,7 @@ RSpec.describe Tasks::WhitehallNewsImporter do
       Tasks::WhitehallNewsImporter.new.import(parsed_json)
 
       expect(Document.last.publication_state).to eq("sent_to_live")
-      expect(Document.last.review_state).to eq("reviewed")
+      expect(Document.last.review_state).to eq("unreviewed")
       expect(Document.last.has_live_version_on_govuk).to eq(true)
     end
   end
--- a/spec/tasks/whitehall_news_importer_spec.rb
+++ b/spec/tasks/whitehall_news_importer_spec.rb
@@ -25,56 +25,10 @@ RSpec.describe Tasks::WhitehallNewsImporter do
           topical_events: [SecureRandom.uuid, SecureRandom.uuid],
           world_locations: [SecureRandom.uuid, SecureRandom.uuid],
           state: "draft",
+          minor_change: false,
           force_published: false,
         },
       ],
-    }
-  end
-  let(:import_published_then_drafted_data) do
-    {
-        content_id: SecureRandom.uuid,
-        editions: [
-            {
-                created_at: Time.zone.now,
-                news_article_type: { key: "news_story" },
-                translations: [
-                    {
-                        locale: "en",
-                        title: "Title",
-                        summary: "Summary",
-                        body: "Body",
-                        base_path: "/government/news/title",
-                    },
-                ],
-                lead_organisations: [SecureRandom.uuid, SecureRandom.uuid],
-                supporting_organisations: [SecureRandom.uuid, SecureRandom.uuid],
-                worldwide_organisations: [SecureRandom.uuid, SecureRandom.uuid],
-                topical_events: [SecureRandom.uuid, SecureRandom.uuid],
-                world_locations: [SecureRandom.uuid, SecureRandom.uuid],
-                state: "draft",
-                force_published: false,
-            },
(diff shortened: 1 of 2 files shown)


spec/tasks/whitehall_news_importer_spec.rb
```diff
@@ -65,41 +113,54 @@
     expect(Document.last.review_state).to eq("unreviewed")
   end
 
-  it "sets the correct publication state and review state when Whitehall document state is 'published'" do
+  it "sets the correct publication, review and has live states when Whitehall document state is 'published'" do
```
benthorner: Not a blocker, but could we just say 'sets the correct states'? The checks within the tests make it pretty clear which states we're checking :-).

kevindew:
Nice 👍

spec/tasks/whitehall_news_importer_spec.rb
```diff
@@ -60,46 +62,58 @@
...
+    expect(Document.last.has_live_version_on_govuk).to eq(true)
   end
 
-  it "sets the correct publication state and review state when Whitehall document is force published" do
+  it "can set minor update type" do
+    import_data[:editions][0][:minor_change] = true
+    parsed_json = JSON.parse(import_data.to_json)
+    Tasks::WhitehallNewsImporter.new.import(parsed_json)
+    expect(Document.last.update_type).to eq("minor")
+  end
```
kevindew: This looks good 👍

spec/tasks/whitehall_news_importer_spec.rb
```diff
@@ -108,4 +122,64 @@
...
+      }
+    end
+
+
+    it "sets the correct states when Whitehall document has more than one edition" do
+      parsed_json = JSON.parse(import_published_then_drafted_data.to_json)
+      Tasks::WhitehallNewsImporter.new.import(parsed_json)
+
+      expect(Document.last.publication_state).to eq("sent_to_live")
+      expect(Document.last.review_state).to eq("reviewed")
```
kevindew: Hmm I'm not sure this test is doing what we'd expect it to do?

We'd expect the document to be a draft (and thus sent_to_draft) with a not been reviewed state?

spec/tasks/whitehall_news_importer_spec.rb
```diff
@@ -108,4 +122,64 @@
...
+              },
+            ],
+            lead_organisations: [SecureRandom.uuid, SecureRandom.uuid],
+            supporting_organisations: [SecureRandom.uuid, SecureRandom.uuid],
+            worldwide_organisations: [SecureRandom.uuid, SecureRandom.uuid],
+            topical_events: [SecureRandom.uuid, SecureRandom.uuid],
+            world_locations: [SecureRandom.uuid, SecureRandom.uuid],
+            state: "published",
+            force_published: false,
+            },
```
kevindew: This and the line below needs to be indented in a couple of spaces. Again no idea why our linter isn't getting it :-/

spec/tasks/whitehall_news_importer_spec.rb
```diff
@@ -108,4 +122,64 @@
...
+          ],
+      }
+    end
+
+
+    it "sets the correct states when Whitehall document has more than one edition" do
+      parsed_json = JSON.parse(import_published_then_drafted_data.to_json)
+      Tasks::WhitehallNewsImporter.new.import(parsed_json)
+
+      expect(Document.last.publication_state).to eq("sent_to_live")
```
kevindew: It doesn't seem right that this is sent_to_live. I'd have thought it was a sent_to_draft matching https://github.com/alphagov/content-publisher/pull/421/files#diff-fc6876b3da5d8f577dec523eb02e8beeR148

spec/tasks/whitehall_news_importer_spec.rb
```diff
@@ -169,4 +122,64 @@
...
+                  lead_organisations: [SecureRandom.uuid, SecureRandom.uuid],
+                  supporting_organisations: [SecureRandom.uuid, SecureRandom.uuid],
+                  worldwide_organisations: [SecureRandom.uuid, SecureRandom.uuid],
+                  topical_events: [SecureRandom.uuid, SecureRandom.uuid],
+                  world_locations: [SecureRandom.uuid, SecureRandom.uuid],
+                  state: "published",
+                  force_published: false,
+              },
+          ],
+      }
```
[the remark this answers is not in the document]
kgarwood: I'm trying to still understand why it's doing that.  I've corrected the changes and I'm going back to make sure the space settings in my editor are at 2

PR #422, opened 2018-10-29
Improve error handling

benthorner:
https://trello.com/c/WdQGc5jT/330-allow-a-user-to-see-a-complete-hierarchy-of-govuk-topics

Please see the commit messages for details...


app/services/topics_service.rb
```diff
@@ -33,11 +34,15 @@ def unroll(all_topics, topics)
   end
 
   def level_one_topics
+    start_time = Time.zone.now
+
     @level_one_topics ||= publishing_api.get_expanded_links(GOVUK_HOMEPAGE_CONTENT_ID)
       .dig("expanded_links", "level_one_taxons")
 
     @level_one_topics.each do |level_one_topic|
+      raise GdsApi::TimedOutException.new if Time.zone.now - start_time > TOPICS_TIMEOUT
```
kevindew: It seems a bit fishy for us to be raising a GdsApi exception as you'd expect them to come from the Gem. Should we instead have our own exception defined in this class and roll GdsApi errors into that?

benthorner: Admittedly, yes. I wanted to avoid having another exception, which we then have to catch in addition to GdsApi::BaseError. I'm not sure how much value it adds to raise our own custom exception here, other than to bloat the class - the outward behaviour should be the same as if it were a single API call, to my thinking was that the 'surprise factor' is pretty limited.

benthorner: I think we have a general pain-point of using GdsApi::BaseError throughout the app, instead of our own exception stack. That said, I can't recall it causing any harm, so far ;-).

kevindew: Yeah I'd agree the pain point is the catching of GdsApi errors, and us throwing the gems exceptions manually is a clear manifestation of that.

I'd not be advocating we catching both exceptions just would think it'd be better if we wrapped all the errors we expect consumers to handle in a single one.

I'm happy to approve this though if doing that is either not clearly a better way forward or requires a bunch of work that doesn't add value now, but would be good to consider addressing at some point.

benthorner: I think I'm struggling to see it as a better way forward right now - it hasn't caused me any pain and I'm struggling to imagine a scenario where it would cause problems for others. At some point I expect I'll see the wisdom of application-specific exceptions - still waiting ;-).

PR #423, opened 2018-10-29
Show topic on summary

benthorner:
https://trello.com/c/5tAUflyY/331-show-current-taxons-for-a-document

This adds the ability to see the current topics for a document on the summary page. We extended the TopicsService to allow us to get the current topics, which meant doing some renaming to keep the purpose of the methods clear. We also cope with the Publishing API being down or slow.

We have some duplication of stubs in the tests, but I don't think it's enough to get a clear refactoring out of it. I'm expecting the final piece of work - to enable editing of topics - will involve revisiting these specs, rather than creating new ones, so it seems unlikely the duplication will spread.

### DIFF
--- a/app/services/topics_service.rb
+++ b/app/services/topics_service.rb
@@ -5,31 +5,46 @@ require "gds_api/publishing_api_v2"
 class TopicsService
   GOVUK_HOMEPAGE_CONTENT_ID = "f3bbdec2-0e62-4520-a7fd-6ffd5d36e03a"
   CACHE_OPTIONS = { expires_in: 5.minutes, race_condition_ttl: 10.seconds }.freeze
-  TOPICS_TIMEOUT = 4.seconds
+  TOPIC_INDEX_TIMEOUT = 4.seconds
 
-  def topics
+  def topic_index
     Rails.cache.fetch("topics", CACHE_OPTIONS) do
-      all_topics = {}
+      index = {}
 
-      all_topics[GOVUK_HOMEPAGE_CONTENT_ID] = {
+      index[GOVUK_HOMEPAGE_CONTENT_ID] = {
         title: "GOV.UK Homepage",
         children: level_one_topics.map { |level_one_topic| level_one_topic["content_id"] },
       }
 
-      unroll(all_topics, level_one_topics)
-      all_topics
+      unroll(index, level_one_topics, GOVUK_HOMEPAGE_CONTENT_ID)
+      index
     end
+  end
+
+  def topics_for_document(content_id)
+    publishing_api.get_links(content_id)
+      .dig("links", "taxons").to_a
+      .map { |topic_content_id| topic_index[topic_content_id] }
+  end
+
+  def topic_breadcrumb(topic)
+    parent = topic_index[topic[:parent]]
+    parent ? [topic] + topic_breadcrumb(parent) : [topic]
   end
 
 private
 
-  def unroll(all_topics, topics)
+  def unroll(index, topics, parent_topic)
     topics.each do |topic|
       child_topics = topic.dig("links", "child_taxons").to_a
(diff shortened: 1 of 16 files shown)


spec/features/editing_topics/show_topics_spec.rb
```diff
@@ -0,0 +1,84 @@
...
+                  "content_id" => "level_three_topic",
+                  "title" => "Level Three Topic",
+                  "links" => {},
+                },
+              ],
+            },
+          },
+        ],
+      },
+    )
```
kevindew: Is this the dreaded second instance of this or is this the same test?

app/services/topics_service.rb
```diff
@@ -5,31 +5,46 @@
 class TopicsService
   GOVUK_HOMEPAGE_CONTENT_ID = "f3bbdec2-0e62-4520-a7fd-6ffd5d36e03a"
   CACHE_OPTIONS = { expires_in: 5.minutes, race_condition_ttl: 10.seconds }.freeze
-  TOPICS_TIMEOUT = 4.seconds
+  TOPIC_INDEX_TIMEOUT = 4.seconds
 
-  def topics
+  def topic_index
```
kevindew: I'm a little confused about the name change from all_topics to index. Does it mean something different?

spec/features/editing_images/upload_lead_image_spec.rb
```diff
@@ -32,7 +32,7 @@ def and_i_crop_the_image
     asset_manager_delete_asset(@asset_id)
     stub_publishing_api_put_content(Document.last.content_id, {})
     click_on "Crop image"
-    WebMock.reset!
+    WebMock::RequestRegistry.instance.reset!
```
kevindew: This looks a lot safer. Is it a private API? Should we add a todo comment to update this?

app/services/topics_service.rb
benthorner: So using the word `topics` or `all_topics` kind of implied an array, whereas the return value is actually a lookup table or `index` and the rest of the code uses it in this way. We thought `index` was a little easier to comprehend than `lookup_table`.

spec/features/editing_images/upload_lead_image_spec.rb
benthorner: Can do - an upcoming release of WebMock should expose it as a proper method :-).

spec/features/editing_topics/show_topics_spec.rb
benthorner: This is the dreaded second instance ;-). I think it will also be the last, though, and I'd like to avoid refactoring it without more evidence - the two specs are very isolated, after all.

kevindew: Ok - I'll trust you to do the responsible thing on the third time 😉

benthorner:
@kevindew I think I've responded to everything here. This is a branch off #422, but I can't see anything that's leaked onto this one. Was there something in particular that looked anomalous?

spec/features/editing_topics/show_topics_spec.rb
benthorner: Thanks. I suppose it may not be me doing it, but I expect one or other of us'll be reviewing it!

app/services/topics_service.rb
kevindew: Sure, lookup_table == 👎 - do be cautious of slipping into hungarian notation.

As an outsider index doesn't really tell me more than topics, but I know I'm only bikeshedding so don't worry about it.

benthorner: Ah, I see what you mean. Really we struggled to come up with a name for it, so probably the best thing about it is that it suggests it's not an array!

PR #430, opened 2018-11-01
Fix markdown buttons break content after firts word

alex-ju:
We found in user research that publishers don't select the item they want to style, but rather position the cursor at the beginning of a line. With this approach, after pressing a markdown toolbar button (i.e. list) it currently makes the content break with a new line after the first word.

This PR fixes the issue by selecting the current line by default if not selection is made by the user.

The updated code comes from https://github.com/alphagov/markdown-toolbar-element. Plan to do releases and changelog soon, so I can point to those for changes instead.

[Trello card](https://trello.com/c/oVHHFHWJ)

👉[Test me in the preview deploy](https://content-publisher-revie-pr-430.herokuapp.com/)

### Before
screen shot 2018-10-26 at 10 03 04

### After
screen shot 2018-10-26 at 10 03 42

### DIFF
--- a/app/assets/javascripts/components/markdown-toolbar.js
+++ b/app/assets/javascripts/components/markdown-toolbar.js
@@ -557,15 +557,15 @@
     return Array(n + 1).join(string);
   }
 
-  function wordSelectionStart(text, index) {
-    while (text[index] && text[index - 1] != null && !text[index - 1].match(/\s/)) {
+  function lineSelectionStart(text, index) {
+    while (text[index] && text[index - 1] != null && !text[index - 1].match(/\n/)) {
       index--;
     }
     return index;
   }
 
-  function wordSelectionEnd(text, index) {
-    while (text[index] && !text[index].match(/\s/)) {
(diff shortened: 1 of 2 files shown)


kevindew:
Should we be editing the umd file? 

I feel like we shouldn't really still have this in a vendor directory now we're modifying this ourselves. It seems to give the impression that this is a github release but it's now a hybrid. Maybe we should have forked their project and be bringing one of our own releases in?

alex-ju:
@kevindew we're not editing the file, it's the latest version from the repo, as mentioned in the PR description and can be found here https://github.com/alphagov/markdown-toolbar-element/blob/master/dist/index.umd.js. We would be able to point to it via package.json if we're not vendoring.

kevindew:
@alex-ju D'oh sorry for not reading the description. My bad 😞 

Hmm, we really should have a review about if we definitely need to vendor and can't use package.json

DilwoarH:
Works 👍

alex-ju:
Thanks @DilwoarH!

PR #434, opened 2018-11-02
Edit topics

benthorner:
https://trello.com/c/82J6eESL/356-allow-the-user-to-modify-the-topics-for-a-document

This enables a user to edit the topics for a document:

   * The user can see which topics are currently selected and change the selections (https://github.com/alphagov/content-publisher/commit/b0dadb13e92a9a8bb7273fb9d6e8a3c983da29ee)
   * We cope when the Publishing API is down (https://github.com/alphagov/content-publisher/commit/3eb253272cbe78c2f0b8cdd61b115ad3ca1a7c61)
   * We abort when the topics have been changed externally (https://github.com/alphagov/content-publisher/commit/ee9d022078bd7a1e5390d620b847dd4ecd8d22e8)

Because we don't store the topic taxonomy locally, we need to repeatedly stub parts of the tree, up to three levels deep. To avoid duplicate code, we've refactored this into our first helper, which is included explicitly in the necessary specs. (https://github.com/alphagov/content-publisher/commit/917f165e89506efcdb298b4465042f199ac2d7a5)

### DIFF
--- a/app/controllers/document_topics_controller.rb
+++ b/app/controllers/document_topics_controller.rb
@@ -12,13 +12,16 @@ class DocumentTopicsController < ApplicationController
   def edit
     @document = Document.find_by_param(params[:document_id])
     @topic_index = TopicsService.new.topic_index
-    @topic_content_ids = TopicsService.new.topics_for_document(@document)
+    @topic_content_ids, @version = TopicsService.new.topics_for_document(@document)
   end
 
   def update
     document = Document.find_by_param(params[:document_id])
-    TopicsService.new.patch_topics(document, params.fetch(:topics, {}))
+    TopicsService.new.patch_topics(document, params.fetch(:topics, []), params[:version])
     redirect_to document_path(document), notice: t("documents.show.flashes.topics_updated")
+  rescue GdsApi::HTTPConflict => e
+    Rails.logger.error(e)
+    redirect_to document, alert_with_description: t("documents.show.flashes.topic_update_conflict")
   rescue GdsApi::BaseError => e
     Rails.logger.error(e)
     redirect_to document, alert_with_description: t("documents.show.flashes.topic_update_error")
--- a/app/controllers/document_topics_controller.rb
+++ b/app/controllers/document_topics_controller.rb
@@ -21,7 +21,7 @@ class DocumentTopicsController < ApplicationController
     redirect_to document_path(document), notice: t("documents.show.flashes.topics_updated")
   rescue GdsApi::HTTPConflict => e
     Rails.logger.error(e)
-    redirect_to document, alert_with_description: t("documents.show.flashes.topic_update_conflict")
+    redirect_to document_topics_path, alert_with_description: t("document_topics.edit.flashes.topic_update_conflict")
   rescue GdsApi::BaseError => e
     Rails.logger.error(e)
     redirect_to document, alert_with_description: t("documents.show.flashes.topic_update_error")
--- a/app/controllers/document_topics_controller.rb
+++ b/app/controllers/document_topics_controller.rb
@@ -17,7 +17,10 @@ class DocumentTopicsController < ApplicationController
 
   def update
     document = Document.find_by_param(params[:document_id])
-    TopicsService.new.patch_topics(document, params.require(:topics))
+    TopicsService.new.patch_topics(document, params.fetch(:topics, {}))
     redirect_to document_path(document), notice: t("documents.show.flashes.topics_updated")
+  rescue GdsApi::BaseError => e
+    Rails.logger.error(e)
+    redirect_to document, alert_with_description: t("documents.show.flashes.topic_update_error")
   end
 end
--- a/app/controllers/document_topics_controller.rb
+++ b/app/controllers/document_topics_controller.rb
@@ -12,5 +12,12 @@ class DocumentTopicsController < ApplicationController
   def edit
     @document = Document.find_by_param(params[:document_id])
     @topic_index = TopicsService.new.topic_index
+    @topic_content_ids = TopicsService.new.topics_for_document(@document)
+  end
+
+  def update
+    document = Document.find_by_param(params[:document_id])
+    TopicsService.new.patch_topics(document, params.require(:topics))
+    redirect_to document_path(document), notice: t("documents.show.flashes.topics_updated")
   end
 end
--- a/app/controllers/document_topics_controller.rb
+++ b/app/controllers/document_topics_controller.rb
@@ -17,7 +17,7 @@ class DocumentTopicsController < ApplicationController
 
   def update
     document = Document.find_by_param(params[:document_id])
-    TopicsService.new.patch_topics(document, params.fetch(:topics, []), params[:version])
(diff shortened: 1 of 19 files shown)


config/locales/en/documents/show.yml
```diff
@@ -69,3 +69,10 @@ en:
           chosen: "Image “%{file}” has been selected as the lead image"
           removed: "Image “%{file}” has been deselected. Document will now use the default lead image"
           deleted: "Image “%{file}” has been deleted. Document wil now use the default lead image"
+        topics_updated: Topics updated on all editions
```
kevindew: Is this content designer approved? Feels like it could be a point of confusion.

app/controllers/document_topics_controller.rb
```diff
@@ -12,5 +12,18 @@ class DocumentTopicsController < ApplicationController
...
+    @topic_content_ids, @version = TopicsService.new.topics_for_document(@document)
+  end
+
+  def update
+    document = Document.find_by_param(params[:document_id])
+    TopicsService.new.patch_topics(document, params.fetch(:topics, []), params[:version])
+    redirect_to document_path(document), notice: t("documents.show.flashes.topics_updated")
+  rescue GdsApi::HTTPConflict => e
+    Rails.logger.error(e)
+    redirect_to document, alert_with_description: t("documents.show.flashes.topic_update_conflict")
```
kevindew: Should we be redirecting them to the topic edit page here? seems a bit of a bummer to ask them to go back to the form when we could take them there.

app/controllers/document_topics_controller.rb
```diff
@@ -12,5 +12,18 @@ class DocumentTopicsController < ApplicationController
   def edit
     @document = Document.find_by_param(params[:document_id])
     @topic_index = TopicsService.new.topic_index
+    @topic_content_ids, @version = TopicsService.new.topics_for_document(@document)
+  end
+
+  def update
+    document = Document.find_by_param(params[:document_id])
+    TopicsService.new.patch_topics(document, params.fetch(:topics, []), params[:version])
+    redirect_to document_path(document), notice: t("documents.show.flashes.topics_updated")
```
kevindew: I'm not sure of the answer but the naming structure we have for the locale key feels a bit weird with it being documents.show but that's entirely controlled by the documents topics controller.

app/views/document_topics/edit.html.erb
```diff
@@ -1,24 +1,39 @@
 <% content_for :back_link, document_path(@document) %>
 <% content_for :title, t("document_topics.edit.title", title: @document.title_or_fallback) %>
 
-<% def unroll(topic) %>
+<% def unroll(topic_content_id) %>
   <% capture do %>
     <li>
-      <%= @topic_index[topic][:title] %>
+      <%= tag.input type: "checkbox",
+        id: topic_content_id,
```
kevindew: Not that it's likely to cause a problem but it'd be more conventional if this was prefixed by a string e.g. "topic-#{content_id}"

app/views/document_topics/edit.html.erb
```diff
@@ -1,24 +1,39 @@
...
+        name: "topics[]",
+        value: topic_content_id,
+        checked: @topic_content_ids.include?(topic_content_id) %>
 
-      <% if @topic_index[topic][:children].any? %>
+      <%= @topic_index[topic_content_id][:title] %>
```
kevindew: This may be moot if this all replaced by the miller component but I'd guess we'd want this to be a label of the checkbox.

app/views/document_topics/edit.html.erb
```diff
@@ -1,24 +1,39 @@
 <% content_for :back_link, document_path(@document) %>
 <% content_for :title, t("document_topics.edit.title", title: @document.title_or_fallback) %>
 
-<% def unroll(topic) %>
+<% def unroll(topic_content_id) %>
   <% capture do %>
```
kevindew: Maybe I'm missing something but this need to do a capture?

config/locales/en/documents/show.yml
```diff
@@ -69,3 +69,10 @@ en:
...
+        topic_update_error:
+          title: Something has gone wrong
+          description: Something went wrong when saving your changes. Please try again or [raise a support request](https://support.publishing.service.gov.uk/technical_fault_report/new).
```
kevindew: Do we repeat this translation in a few places?

app/views/documents/show/_topics.html.erb
```diff
@@ -1,13 +1,13 @@
...
 
   <% breadcrumbs = capture do %>
-    <% if topics.any? %>
+    <% if topic_content_ids.any? %>
       <div class="topic-breadcrumb">
-        <% topics.each do |topic| %>
+        <% topic_content_ids.each do |topic_content_id| %>
           <ol>
-            <% service.topic_breadcrumb(topic).reverse.each do |crumb| %>
+            <% service.topic_breadcrumb(topic_content_id).reverse.each do |crumb| %>
```
kevindew: Do we reverse this because this method returns results in a [child, parent, grandparent, great-grandparent] fashion? If so I wonder if we'd want to reverse the method as I'd suggest breadcrumbs suggests the inverse.

config/locales/en/documents/show.yml
benthorner: Yup - @benhazell has commented on the card and we've updated 👍 .

app/controllers/document_topics_controller.rb
benthorner: Can do - my thought was to be consistent with the 'API down' case. But actually Whitehall does this, so we may as well behave the same - @MuriloDalRi would you like to change this?

benthorner: Done this :-)

app/views/document_topics/edit.html.erb
benthorner: Done - this was interesting!

app/controllers/document_topics_controller.rb
benthorner: So the convention is that copy should be put in the locale file for the page it is shown on. Otherwise it's difficult for a content designer to work out which file a particular piece of copy might be in e.g. if I see an alert on the document summary page, the copy should be in documents/show.yml.

config/locales/en/documents/show.yml
benthorner: We have a lot of translations that are similar, but this one is still unique. I think we should keep them separate to allow a content designer to evolve them separately.

app/views/documents/show/_topics.html.erb
benthorner: Good spot - I think we did it like this originally because the method was called something else. Fixed :-).

app/views/document_topics/edit.html.erb
benthorner: Done :-).

app/views/document_topics/edit.html.erb
benthorner: Also done - thanks for pointing out :-). We tried to use the checkbox component, but it's not really designed for single item usage. Will leave this to @alex-ju.

PR #435, opened 2018-11-02
Auto populate legacy topics (specialist sectors)

benthorner:
https://trello.com/c/lE8z7rmq/360-automatically-select-specialist-sectors-when-a-user-selects-a-topic

The new topic taxonomy has replaced a range of legacy alternatives,
except for 'topics' a.k.a specialist sectors. This makes sure we set
the specialist sectors for a document when the topics change, using the
legacy mappings for the selected topics and their parent lineage.

If the document already has specialist sectors, we will overwrite them.
This is a departure from the Whitehall behaviour, which we will validate
as with user testing. Hopefully we will retire specialist sectors before
the need arises to better support their existence ;-).


benthorner:
Discussed offline and we think this should be fine, as the recursive call relies on one common, cached API call to populate the topic_index; and the maximum recursion depth is limited to around 5 stack frames.

PR #442, opened 2018-11-05
Separate editions from publishing

benthorner:
https://trello.com/c/IjE24eFh/372-update-publishing-requirements-for-creating-an-edition-of-a-document

This modifies the workflow to better support users creating a new edition of a document:

   * Hide the 'Change' links when the content isn't editable, except for topics (https://github.com/alphagov/content-publisher/commit/984be24c15233ab7186257306e5d1087d5aa0762)
   * Swap the order of the update type and change note fields when editing content (https://github.com/alphagov/content-publisher/commit/8f3c7ad6bb6b221b367c67aa9bd0ce25f019a07a)
   * Centralising new edition 'creation' into the document drafting service (https://github.com/alphagov/content-publisher/commit/650eb43eb988e2afc6f98fbd6ad72c4e92f2d7b1, https://github.com/alphagov/content-publisher/commit/d2f7af71db9acd49341d9110ca79f61e1f9384ca, https://github.com/alphagov/content-publisher/commit/95d28ae287f069a10c9f02022f4442f789107cf0)

I also noticed we weren't sending the update type on publish...now we are (https://github.com/alphagov/content-publisher/commit/895c312902580b6f0a09c76053e125cbd4bc61ee).


app/services/document_publishing_service.rb
```diff
@@ -14,7 +14,7 @@ def publish_draft(document)
   def publish(document, review_state)
     document.update!(publication_state: "sending_to_live", review_state: review_state)
     publish_assets(document.images)
-    GdsApi.publishing_api_v2.publish(document.content_id, nil, locale: document.locale)
+    GdsApi.publishing_api_v2.publish(document.content_id, document.update_type, locale: document.locale)
```
kevindew: We shouldn't actually do this. Sending update type here is deprecated. 

The method arguments on GDS API Adapters are somewhat unfortunate for this.

It has some effect on change note that I can't quite remember: https://github.com/alphagov/publishing-api/blob/45b3b427c8f4330fae45367cd79ade3961ff75ee/app/commands/v2/publish.rb#L36

benthorner: Thanks for spotting - all gone :-).

PR #447, opened 2018-11-06
Embed contacts via markdown

kevindew:
Trello: https://trello.com/c/gEAMmD0g/386-ability-to-embed-a-contact-via-markdown

This is marked as DNM as https://github.com/alphagov/govspeak/pull/130 needs to be released first

nov-07-2018 09-42-07

This PR introduces the ability to convert markdown such as `[Contact:e3dd86ba-508c-4e4c-9cba-1b98208d4de0]` and then insert the HTML of a contact. 

It does this by looking up all contact content_ids in the govspeak and then for each of them [looking them up](https://github.com/alphagov/publishing-api/blob/master/doc/api.md#get-v2contentcontent_id) in the Publishing API.

The means to look up in the Publishing API isn't perfect, it's as many calls as there are contacts embedded and may indadvertedly embed draft contact information. The expectation though is that the UI work to embed contacts may offer an opportunity to review those issues.

### DIFF
--- a/app/services/contacts_service.rb
+++ b/app/services/contacts_service.rb
@@ -1,0 +1,12 @@
+# frozen_string_literal: true
+
+class ContactsService
+  def by_content_id(content_id)
+    # FIXME: This might return a draft as this Publishing API method doesn't
+    # distinguish between these
+    document = GdsApi.publishing_api_v2.get_content(content_id)
+    document["schema_name"] == "contact" ? document.to_h : nil
+  rescue GdsApi::HTTPNotFound
+    nil
+  end
+end
--- a/app/services/publishing_api_payload.rb
+++ b/app/services/publishing_api_payload.rb
@@ -97,7 +97,7 @@ private
   # or class system.
   def perform_input_type_specific_transformations(field)
     if field.type == "govspeak"
-      GovspeakService.new.to_html(document.contents[field.id])
+      GovspeakService.new(document.contents[field.id]).to_html
     else
       document.contents[field.id]
     end
--- a/app/services/publishing_api_payload.rb
+++ b/app/services/publishing_api_payload.rb
@@ -97,7 +97,7 @@ private
   # or class system.
   def perform_input_type_specific_transformations(field)
     if field.type == "govspeak"
-      GovspeakService.new(document.contents[field.id]).to_html
+      GovspeakDocument.new(document.contents[field.id]).to_html
     else
       document.contents[field.id]
     end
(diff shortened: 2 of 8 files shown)


spec/services/contacts_service_spec.rb
```diff
@@ -0,0 +1,52 @@
+# frozen_string_literal: true
+
+RSpec.describe ContactsService do
+  describe "#by_content_id" do
+    context "when a contact is found" do
+      it "returns the contact" do
+        content_id = SecureRandom.uuid
```
benthorner: It's nice to see the full attributes of a contact, but it took me a while to see that most of it isn't necessary for the test, but it's presence kind of suggested it was. Could we keep the test data to a minimum so it's easier to see the data under test?

spec/services/govspeak_service_spec.rb
```diff
@@ -0,0 +1,44 @@
+# frozen_string_literal: true
+
+RSpec.describe GovspeakService do
+  describe "#to_html" do
+    it "converts the provided text to HTML" do
+      expect(GovspeakService.new("## Hullo").to_html)
+        .to match(%{<h2 id="hullo">Hullo</h2>})
+    end
+
+    context "when a contact that isn't known is provided in the govspeak" do
```
benthorner: This took me a couple of readings - something about the 'is provided' bit. Perhaps just 'when we reference an unknown contact'?

app/services/contacts_service.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+class ContactsService
+  def by_content_id(content_id)
+    # FIXME: This might return a draft as this Publishing API method doesn't
+    # distinguish between these
+    document = GdsApi.publishing_api_v2.get_content(content_id)
+    document["schema_name"] == "contact" ? document.to_h : nil
```
benthorner: You probably thought of this already, but it looks like the response says which content store it's coming from. Could we add that filter in here as well?

app/services/publishing_api_payload.rb
```diff
@@ -97,7 +97,7 @@ def roles_and_people(role_appointments)
   # or class system.
   def perform_input_type_specific_transformations(field)
     if field.type == "govspeak"
-      GovspeakService.new.to_html(document.contents[field.id])
+      GovspeakService.new(document.contents[field.id]).to_html
```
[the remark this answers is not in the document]
kevindew: I've switched to GovspeakDocument like we discussed.

spec/services/govspeak_service_spec.rb
kevindew: Thanks - I like your suggestion

app/services/contacts_service.rb
kevindew: Unfortunately that'd remove things that are live and have a draft

kevindew:
@benthorner all updated now

PR #451, opened 2018-11-08
Update lead image

alex-ju:
This PR updates the lead image with the changes resulted from user research analysis.

## Before
lead image before

## After
lead image after

[Trello card](https://trello.com/c/Yu5DjMk3)

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -4,6 +4,7 @@
 
 @import "objects/grid";
 @import "objects/history";
+@import "objects/pane";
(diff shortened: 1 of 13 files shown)


benthorner:
This looks better :-). We seem to have lost the functionality to unselect but not delete ('remove') a lead image. Was this intentional? If it is, we should do an additional PR to remove the backend support for this, so we don't leave it to rot away.

alex-ju:
@benthorner I realised the same after seeing the tests failing. I will put them back for now (remove lead image and delete lead image) until we make an informed decision about it.

PR #458, opened 2018-11-09
Added guidance links

benhazell:
Added guidance links to markdown guidance section on right hand side of the Edit screen.
Links should open in a new tab or window.


kevindew:
Had a quick look at this with Ben.

The YAML can be corrected by using the pipe character for the multiline selections. Though once this is done the markdown is not rendered as these sections do not render govspeak.

benthorner:
@benhazell the code needed a bit of tweaking here to render the guidance body as govspeak. This should become more obvious when we do https://trello.com/c/AiMKvMd2/424-update-naming-of-html-and-govspeak-fields-in-locale-and-config-files.

PR #459, opened 2018-11-09
Index page design updated

DilwoarH:
- Adds additional column for status
- make index page layout changes (filters now 1/4, results now 3/4)
- changes font size to 16 for table row text

https://trello.com/c/jok0QIR6/286-index-page-design

## Screenshot

screen shot 2018-11-13 at 13 44 13

https://trello.com/c/cJmGlDh9/374-update-design-for-index-page-in-content-publisher

### DIFF
--- a/app/views/documents/index.html.erb
+++ b/app/views/documents/index.html.erb
@@ -1,11 +1,11 @@
 <% content_for :title, t("documents.index.title") %>
 
 <% content_for :title_side, render("govuk_publishing_components/components/button", {
-  text: "New document", href: new_document_path, margin_bottom: true
+  text: "Create new document", href: new_document_path, margin_bottom: true
 }) %>
 
 <div class="govuk-grid-row">
-  <div class="govuk-grid-column-one-third">
+  <div class="govuk-grid-column-one-quarter">
     <%= render "documents/index/filters" %>
 
     <hr class="govuk-section-break govuk-section-break--visible govuk-section-break--l">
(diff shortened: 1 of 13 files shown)


app/assets/stylesheets/objects/_grid.scss
```diff
@@ -1,3 +1,7 @@
 .app-grid-column--end {
   text-align: right;
 }
+
+.govuk-width-container {
+  max-width: 1200px;
```
alex-ju: As discussed, if we're going to go with a new breaking point we should do it by defining it in `media-queries`. The rest of the code is fine, but we should discuss this with the rest of the team and understand implications.

app/assets/stylesheets/views/index.scss
```diff
@@ -0,0 +1,12 @@
+.results-table-wrapper {
+  .govuk-table__body {
```
alex-ju: I would move these to `app/assets/stylesheets/utilities/_overwrites.scss` as it's an actual overwrite.

app/views/documents/index.html.erb
```diff
@@ -24,7 +24,9 @@
         <%= govspeak_to_html t("documents.index.search_results.guidance") %>
       </div>
     <% else %>
-      <%= render "documents/index/results" %>
+      <div class="results-table-wrapper">
```
alex-ju: We won't need the additional wrapper as the overwrite should be applied at the project level.

app/views/documents/index/_results.html.erb
```diff
@@ -28,13 +29,12 @@
               },
               metadata: {
                 document_type: document.document_type_schema.label,
-                document_create: t("documents.index.search_results.creator", creator: document.creator&.name || I18n.t!("documents.index.search_results.unknown_user")),
```
alex-ju: We've lost the `document_create` on the way or we dropped it intentionally?

DilwoarH: Hey alex - we dropped it intentionally. @soniaturcotte said we don't need both creator and last editor here.

alex-ju: OK

PR #463, opened 2018-11-12
Copy changes to show.yml

paola-roccuzzo:
Amend 'base path' to 'page address path'

### DIFF
--- a/config/locales/en/documents/show.yml
+++ b/config/locales/en/documents/show.yml
@@ -20,7 +20,7 @@ en:
         title: Content
         items:
           title: Title
-          base_path: Base path
+          base_path: Page address path
           summary: Summary
           update_type: Update type
           change_note: Change note


kevindew:
Are we confident that users will understand this anymore than base path?

It ends up a bit confusing as base path has quite a nuanced meaning and we lose that with "Page address path" - I guess also page address path relies on user still understanding what a path is so I wonder if the first two words are redundant?

I wondered if some of the suggestions to just remove showing this field would produce an easier experience for users?

benthorner:
Superseded by https://github.com/alphagov/content-publisher/pull/471.

paola-roccuzzo:
@kevindew agreed - best solution is to remove "Base path" from the summary screen.
I believe we will need to add a caption/hint to "Page address" in the edit screen - and a different one when users create a new edition (to alert them of what happens with the existing one).

PR #464, opened 2018-11-12
Update navigation header to include phase banner

DilwoarH:
screen shot 2018-11-12 at 13 31 20

Ticket: https://trello.com/c/6hvQZTVK/259-update-navigation-header

### DIFF
--- a/app/views/layouts/_phase_banner_elements.html.erb
+++ b/app/views/layouts/_phase_banner_elements.html.erb
@@ -1,0 +1,6 @@
+<span class="gem-c-phase-banner__banner-item">
+  <a class="govuk-link govuk-link--no-visited-state" href="<%= Plek.new.external_url_for("support") + "/general_request/new" %>"><%= t("application.phase_banner.send_us_feedback") %></a>
+</span>
+<span class="gem-c-phase-banner__banner-item">
+  <a class="govuk-link govuk-link--no-visited-state" href="<%= publisher_updates_path %>"><%= t("application.phase_banner.whats_new") %></a>
+</span>
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -25,12 +25,21 @@
   <%= render "govuk_publishing_components/components/layout_header", {
     environment: GovukPublishingComponents::AppHelpers::Environment.current_acceptance_environment,
     navigation_items: [
+      { text: "Switch app", href: Plek.new.external_url_for("signon") },
       { text: current_user.name, href: Plek.new.external_url_for("signon") },
       { text: "Log out", href: gds_sign_out_path }
     ]
   }%>
 
   <div class="govuk-width-container">
+    <%= render "govuk_publishing_components/components/phase_banner", {
+      app_name: "Content publisher",
+      phase: "beta",
+      message: capture do
+        render "layouts/phase_banner_elements"
+      end,
+    } %>
+
     <% if yield(:back_link).present? %>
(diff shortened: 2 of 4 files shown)


app/views/layouts/application.html.erb
```diff
@@ -25,12 +25,21 @@
   <%= render "govuk_publishing_components/components/layout_header", {
     environment: GovukPublishingComponents::AppHelpers::Environment.current_acceptance_environment,
     navigation_items: [
+      { text: "Switch app", href: Plek.new.external_url_for("signon") },
```
DilwoarH: Need a external url for this

app/views/layouts/_phase_banner_elements.html.erb
```diff
@@ -0,0 +1,7 @@
+<span class="gem-c-phase-banner__banner-item">
+    <a class="govuk-link govuk-link--no-visited-state" href="https://support.publishing.service.gov.uk/">Send us feedback</a>
```
DilwoarH: Need to set appropriate URL

app/views/layouts/_phase_banner_elements.html.erb
```diff
@@ -0,0 +1,7 @@
...
+</span>
+<span class="gem-c-phase-banner__banner-item">
+    <a class="govuk-link govuk-link--no-visited-state" href="#">What's new</a>
```
DilwoarH: Need to set appropriate URL

app/assets/stylesheets/components/_phase-banner.scss
```diff
@@ -0,0 +1,10 @@
+.gem-c-phase-banner {
```
alex-ju: As in the other pull request, this is an overwrite should stay in `app/assets/stylesheets/utilities/_overwrites.scss`

app/views/layouts/application.html.erb
```diff
@@ -25,12 +25,21 @@
...
       { text: "Log out", href: gds_sign_out_path }
     ]
   }%>
 
   <div class="govuk-width-container">
+    <%= render "govuk_publishing_components/components/phase_banner", {
+      app_name: "Content publisher",
+      phase: "beta",
+      message: capture do
+        render "layouts/phase_banner_elements"
```
alex-ju: I'm not sure we need a partial for this - we could capture it above - but if we do `layouts` isn't probably the best place. I would check with @benthorner to see where it fits in the project structure.

app/views/layouts/_phase_banner_elements.html.erb
```diff
@@ -0,0 +1,6 @@
+<span class="gem-c-phase-banner__banner-item">
+  <a class="govuk-link govuk-link--no-visited-state" href="<%= Plek.new.external_url_for("support") + "/general_request/new" %>">Send us feedback</a>
```
alex-ju: We should use translation files for these two pieces of copy.

app/views/layouts/application.html.erb
DilwoarH: @benthorner - where would you put something like this, I put it in here as it is a generic element used across the whole app.

DilwoarH:
@alex-ju - all changes have been done except for moving the template.

app/views/layouts/application.html.erb
benthorner: I think I'm more concerned about the footer part, which has made this file long. I think it's fine to make a few partials here, as this file is becoming unreadable for someone trying to get an idea of its structure. And layouts/_partial.html.erb is the place I'd look first to find them - you might need to specify the path explicitly when you render them, though.

PR #469, opened 2018-11-12
Topic publishing requirements

benthorner:
https://trello.com/c/7AGwIoeJ/395-ensure-a-user-selects-at-least-one-topic-before-publishing

This ensures we warn the user if they do not select a topic before publishing; the warning turns into an error if they try to publish (or submit for 2i review), or a fatal error if the Publishing API is down.

Doing this work was made easier following a couple of refactorings to the [code for publishing requirements](https://github.com/alphagov/content-publisher/commit/1f8a5b2fb32011f373ff33f5c5351046614389de) and the [code for topics](https://github.com/alphagov/content-publisher/commit/73bf844bfab119b1d6fefa8f45a33f80744edb4a); please set the linked commits.


app/models/document.rb
```diff
@@ -59,4 +59,8 @@ def user_facing_state
   def title_or_fallback
     title.presence || I18n.t!("documents.untitled_document")
   end
+
+  def topics
+    DocumentTopics.new(self)
```
kevindew: Since this isn't memo'd doesn't this mean we do a HTTP request each time you call this method?

app/models/topic.rb
```diff
@@ -0,0 +1,37 @@
...
+    find(TopicIndex::GOVUK_HOMEPAGE_CONTENT_ID)
+  end
+
+  def self.find(topic_content_id)
+    TopicIndex.new.index[topic_content_id]
+  end
+
+  attr_reader :title, :child_content_ids, :legacy_topic_content_ids, :parent_content_id, :content_id
+
+  def initialize(**params)
```
kevindew: Do we need a double splat here?

app/services/publishing_requirements.rb
```diff
@@ -1,41 +1,46 @@
...
     end
 
     messages
   end
+
+private
+
+  def check_contents(field)
+    return [] if document.contents[field.id].present?
+    [{ text: I18n.t!("publishing_requirements.no_content_#{field.id}"), href: "#content" }]
```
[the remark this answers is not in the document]
benthorner: I believe we actually want to keep them on summary; if we sent them to another screen, it would be unclear how they get there manually.

app/models/document.rb
benthorner: I don't think so - the memoisation is on the specific methods of the DocumentTopics class, which I now notice I didn't add yet (was postponing it to the final commit). Thanks for spotting :-).

app/models/topic.rb
benthorner: Oops again - changed tack as I wrote the method. Fixed.

benthorner:
@kevindew I know what you mean about the models thing. I wanted to do something better than shove them in services, but I don't mind adding to that bit of creeping debt if that sits better with you. Alternatively, the author of the article seems to be keen on `app/models/something/topics_stuff`, but that doesn't feel very intuitive to me. Definitely not a fan of lib as it's hard to use code in there.

app/services/publishing_requirements.rb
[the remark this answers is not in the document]
kevindew: We'll handle this outside of this PR.

app/models/document.rb
kevindew: You'd still get a new instance of DocumentTopics though so that wouldn't have access to the memoisation of the previous one?

kevindew:
@benthorner Lets have a chat about the class model pattern and consequences. Feels like something easier to discuss than through github comments.

app/models/document.rb
benthorner: Oops again - fixed!

benthorner:
@kevindew I did a bit more thinking and came up with a new commit when you've got a moment...

app/models/document_topics.rb
```diff
@@ -0,0 +1,45 @@
...
+
+  def self.find_by_document(document)
+    publishing_api = GdsApi.publishing_api_v2
+    links = publishing_api.get_links(document.content_id)
+    topic_content_ids = links.dig("links", "taxons").to_a
+
+    new(
+      document: document,
+      version: links["version"],
+      topics: topic_content_ids.map(&Topic.method(:find)),
```
kevindew: Sorry to be a pain with late comments.

One thing that doesn't sit happy with me and this implementation is that we call this class method on `Topic` and then that creates a new instance of `TopicIndexService` which itself goes off and makes HTTP requests (or loads data from a cache). This seems inefficient when an instance of TopicIndexService could hold this data in memory. 

I'd be suggesting that when we create a DocumentTopics we pass around a TopicIndexService (or something along those lines) so this data can be stored. What do you think?

benthorner: I'm not really sure what the problem is - the TopicIndexService deals with caching, so there's no need to bother other classes with the same issue?

kevindew: Maybe I didn't explain well.

So if we had one instance of TopicIndexService that stored the tree in memory and was passed around, we'd only ever have to do one one look up of the taxonomy tree. 

As this implementation stands each time we create a Topic we need to retrieve the data from cache all pull the request. Which means:

1) In dev - since you have no cache HTTP requests each time
2) with a file system cache - you have to read from file system each time this is used rather than memory
3) with a redis or memcache server - a HTTP request each cache hit

Normally this wouldn't be the sort of thing that'd be a concern as maybe you hit a cache once or twice in a request duration but here we have loops of things doing this.

I'm probably on the verge of premature optimisation but it just sits weird with me and feels more like a candidate for dependency resolution. Does that make sense?

benthorner: Ah, I understand what you mean now, and can see we were operating under different assumptions. In particular, the Rails FileStore does [have an in-memory cache](https://github.com/rails/rails/blob/master/activesupport/lib/active_support/cache/file_store.rb#L15).

But I see what you mean that a Topic could be more conscious of the service it's relying on, but at the expense of passing it around everywhere, which I think is what lead you on to dependency resolution.

My original thought was that the TopicIndexService is acting like a 'bare metal' adapter for a topic, but this has made me think that really I'm treating it more like a TopicIndex model. It's a similar compromise between doing stuff in Ruby vs. SQL.

Without making the code more complicated, I think the best solution to this would be to download the topics index so that it can truly be a model, which makes the overhead of requesting all the data negligible - kind of the same as the original sync/async discussion we had about taxonomy. While the caching is sufficient, I think we're OK for now.

Does that address your concern (at least in principle)?

benthorner:
@kevindew I think I've fixed the efficiency problem in development now, while trying to keep the interface simple - the caching of the TopicIndexService at the document level avoids the injection from ballooning across the app.

PR #479, opened 2018-11-14
Design tweeks to make publishing information screens better visually

DilwoarH:
Ticket: https://trello.com/c/7ARz6trT/432-design-tweaks-to-publisher-information-pages

Preview: https://content-publisher-revie-pr-479.herokuapp.com/publisher-updates


app/assets/stylesheets/utilities/_overwrites.scss
```diff
@@ -78,3 +78,19 @@
     font-size: inherit;
   }
 }
+
+.govuk-link {
```
alex-ju: We shouldn't need to nest these - a modifier class should not depend on the non-modified class to be present.

app/assets/stylesheets/utilities/_overwrites.scss
```diff
@@ -78,3 +78,19 @@
...
+  &.govuk-link--no-decoration {
+    text-decoration: none;
+  }
+}
+
+.govuk-list {
+  .govuk-list-item--with-border-bottom {
```
alex-ju: Similar here, we shouldn't nest them.

These two in particular look quite ugly, especially the --font-size-16. I would push this work to be part of the side navigation component instead as per https://trello.com/c/ZhPzVCUr

DilwoarH:
@alex-ju - it's now been moved to a separate component

PR #480, opened 2018-11-15
MVP add contacts feature

kevindew:
Trello: https://trello.com/c/a80pDKyV/412-browse-and-insert-a-contact-into-markdown

nov-15-2018 13-40-06

This card will be followed up with some design tweaks. More info in commits, main commit info below: 

This implements the design to insert contacts into a document that
involved loading a new page - with the intention of avoiding a modal
window and having an identical JS and No JS experience.

It involves allowing a user to submit the edit content with a different
button and because of that button changing the behaviour of the update
action to redirect the user to a select contact view. When this form is
submitted it then redirects the user back to the edit content form.

This approach is hopefully not something we will keep long term or use
again as it suffers from a number of issues:

- User can be prevented from getting to the insert contact screen by a
  validation error
- A new entry in the document history will be created each time the user
  submits either form
- We can't insert at a particular point in the textarea field easily and
  thus append to the end of the string
- We have to closely embed which field will be amended with the
  DocumentContactsController which doesn't make it particularly friendly
  to the format system

There are a number of approaches that could solve one or more of these
issues but will likely cause the complications and exceptions for this
to increase - whereas it probably makes sense to be entirely in the
concern of the govspeak editor.

### DIFF
--- a/app/controllers/document_contacts_controller.rb
+++ b/app/controllers/document_contacts_controller.rb
@@ -4,6 +4,9 @@ class DocumentContactsController < ApplicationController
   def search
     @document = Document.find_by_param(params[:id])
     @contacts_by_organisation = ContactsService.new.all_by_organisation
+  rescue GdsApi::BaseError => e
+    Rails.logger.error(e)
+    render "search_api_down", status: :service_unavailable
   end
 
   def insert
--- a/app/controllers/document_contacts_controller.rb
+++ b/app/controllers/document_contacts_controller.rb
@@ -1,0 +1,28 @@
+# frozen_string_literal: true
+
+class DocumentContactsController < ApplicationController
+  def search
+    @document = Document.find_by_param(params[:id])
+    @contacts_by_organisation = ContactsService.new.all_by_organisation
+  end
+
+  def insert
+    document = Document.find_by_param(params[:id])
+    contact_markdown = "[Contact:#{params.require(:contact_id)}]\n"
+
+    body = document.contents.fetch("body", "").chomp
+    document.contents["body"] = if body.present?
+                                  "#{body}\n\n#{contact_markdown}"
+                                else
+                                  contact_markdown
+                                end
+
+    DocumentDraftingService.update!(
+      document: document,
+      user: current_user,
+      type: "updated_content",
+    )
+
+    redirect_to edit_document_path(document) + "#body"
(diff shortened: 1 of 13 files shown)


app/controllers/document_contacts_controller.rb
```diff
@@ -0,0 +1,32 @@
+# frozen_string_literal: true
+
+class DocumentContactsController < ApplicationController
+  def search
+    @document = Document.find_by_param(params[:id])
+    @options = ContactsService.new.all_by_organisation.flat_map do |org|
+      org["contacts"].map do |contact|
+        ["#{contact['title']} - #{org['name']}", contact["content_id"]]
```
benthorner: This feels a little bit weird in here as it's specific to the view code that wants to use it. Could we leave it to the view code, like we do elsewhere?

app/controllers/document_contacts_controller.rb
```diff
@@ -0,0 +1,32 @@
...
+      end
+    end
+  end
+
+  def insert
+    document = Document.find_by_param(params[:id])
+    contact_markdown = "[Contact:#{params.require(:contact_id)}]\n"
+
+    body = document.contents.fetch("body", "").chomp
+    document.contents["body"] = if body.present?
```
benthorner: Could do `body.to_s + ...`.

benthorner:
Oh, and I noticed we're not explicitly checking that we bypass the requirements stuff without a title, as the factory puts one in by default. I'm kind of tempted to just remove the check and let the debt be more obvious; or you could unset it in the test so we check it works implicitly :-).

spec/features/editing_content/insert_contact_spec.rb
```diff
@@ -0,0 +1,52 @@
...
+    and_i_go_to_add_a_contact
+    when_i_select_a_contact
+    then_i_can_see_the_contact_in_the_body
+  end
+
+  let(:organisation) { { "content_id" => SecureRandom.uuid, "internal_name" => "Organisation" } }
+  let(:contact) do
+    {
+      "content_id" => SecureRandom.uuid,
+      "locale" => "en",
```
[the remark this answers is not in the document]
kevindew: Good point, was used in an earlier iteration before I just went full on only get english - now removed.

PR #485, opened 2018-11-15
Centralise requirements

benthorner:
https://trello.com/c/4iKDzhBY/417-separate-saving-a-change-from-the-creation-of-a-preview

This is my attempt to centralise the way we check draft and publish requirements for documents and images in the application, so that it's more consistent and everything is easier to find.

All the centralising stuff is done in the first commit, which could be split up for images and documents, but I think seeing it all together helps identify the change more clearly.

The checks haven't changed, and we still show any pre-draft issues on the document and image edit pages when the user tries to save. But we also now show them on the summary screen, which means the user has a single point of reference for issues with their document; this has the side-effect of resolving a bug where the user aborts uploading an image, which then has no alt text.


app/controllers/publish_document_controller.rb
```diff
@@ -4,7 +4,7 @@ class PublishDocumentController < ApplicationController
   def confirmation
     @document = Document.find_by_param(params[:id])
 
-    if PublishingRequirements.new(@document).errors?(tried_to_publish: true)
+    if Requirements::Checker.new(@document).pre_publish_issues(raise_exceptions: true).any?
```
kevindew: I found the raise_exceptions argument a bit confusing as I read it as I expected that to mean we wanted to raise exceptions for issues rather than return an array (I was initially going to comment asking if `any?` did anything)

If I understand right this raise_exceptions is for when we fail to check if they added topics or not - working on that assumption I don't think we actually want to raise an exception and instead just have an issue that they need to require topics? so argument could be like `require_topics_check` or similar - what do you think?

app/services/requirements/checker.rb
```diff
@@ -0,0 +1,28 @@
...
+        issues += ImageChecker.new(image).pre_draft_issues.to_a
+      end
+
+      issues += DocumentChecker.new(document).pre_draft_issues.to_a
+      CheckerIssues.new(issues)
+    end
+
+    def pre_publish_issues(params)
+      issues = []
+      issues += DocumentChecker.new(document).pre_publish_issues(params).to_a
```
kevindew: Super minor - but these two lines could be just `issues = DocumentChecker.new(document).pre_publish_issues(params).to_a`

app/services/requirements/image_checker.rb
```diff
@@ -0,0 +1,36 @@
...
+      if @image.caption.to_s.length > CAPTION_MAX_LENGTH
+        issues << Issue.new(:caption, :too_long, max_length: CAPTION_MAX_LENGTH, filename: image.filename)
+      end
+
+      CheckerIssues.new(issues)
+    end
+
+    def pre_draft_issues
+      pre_draft_metadata_issues
+    end
```
kevindew: I think this would be better if we just had a single `pre_draft_issues` method here as I found the metadata aspect to be a bit confusing. My thought in https://github.com/alphagov/content-publisher/pull/485/files#diff-3fc221a1b655a04cb0f87aa766b94161R74 was "why do we only care about metadata issues" ? 

I'm assuming you've done this with consistency with DocumentChecker but I think these classes are somewhat different as that class has a wider responsibility of the document - whereas this is focussed on images

app/services/requirements/document_checker.rb
```diff
@@ -0,0 +1,64 @@
+# frozen_string_literal: true
+
+module Requirements
+  class DocumentChecker
```
kevindew: Should this class have a concern about required tags? like organisations for instance?

app/services/requirements/checker.rb
```diff
@@ -0,0 +1,28 @@
+# frozen_string_literal: true
+
+module Requirements
+  class Checker
```
kevindew: I'm curious about the relationship between this class, DocumentChecker and their responsibilities. Is this class intended to cover the entire document from a user perspective and the DocumentChecker and ImageChecker are mapped to the models of `Document` and `Image` respectively or is it different?

app/services/requirements/document_checker.rb
```diff
@@ -0,0 +1,64 @@
...
+        issues << Issue.new(:change_note, :blank)
+      end
+
+      begin
+        if document.document_type_schema.topics && document.topics.none?
+          issues << Issue.new(:topics, :none)
+        end
+      rescue GdsApi::BaseError => e
+        Rails.logger.error(e)
+        raise if raise_exceptions
```
kevindew: Another minor one, but if we're intending to re-raise the exception it's probably reasonable for the higher point in the stack to decide on whether to log it or not (to save it being logged multiple times)

config/locales/en/requirements.yml
```diff
@@ -0,0 +1,39 @@
+en:
+  requirements:
+    title:
+      blank:
+        short_message: Enter a title
+        long_message: Enter a title
+      too_long:
+        short_message: "Enter a title that is fewer than %{max_length} characters long"
+        long_message: Enter a shorter title
```
kevindew: When you explained the long and short message they made a lot of sense but seeing them they're actually pretty confusing as often the short is longer than the long message. I assume you're also a little uncomfortable with this ;)

Of course the big challenge is what is a more appropriate name. Suggestions I have (with no presumptions they are good) are:

in_context_message / out_context_message
context_message / summary_message

with my brain trying to formulate ideas about how some messages are instant, transient, in context, in response to an initial form submission and other ones are more permanent, out of context, shown in a wider summaries. 

Don't know if that's helped at all?

app/services/requirements/checker.rb
benthorner: Done :-).

app/services/requirements/document_checker.rb
benthorner: Good point - removed.

app/services/requirements/document_checker.rb
benthorner: My intention here was just to centralise the existing requirements, rather than add new ones. We don't currently check for tag stuff, but maybe this should be a card. I think this would be in a different PR, anyway.

app/controllers/publish_document_controller.rb
benthorner: It is actually intended that we raise an exception and fail - this is documented on the card. Without the Publishing API we can't know if the document meets the requirements or not, but we also don't want to show an incorrect message 'Add a topic' to the user. Telling a little lie and saying their publish operation 'failed' was decided to be the best behaviour. This is why the flag is generically 'raise_exceptions', as it applies for anything where the check is undecidable.

app/services/requirements/image_checker.rb
benthorner: Resolved as discussed - the update action should check specifically for metadata issues, as this is what it's updating.

app/controllers/publish_document_controller.rb
benthorner: Resolved as discussed ^^^.

app/services/requirements/checker.rb
benthorner: We discussed this and came up with a plan to split the DocumentChecker into separate classes, freeing up the name 'DocumentChecker' to be the thing that runs all the checks.

config/locales/en/requirements.yml
benthorner: Good point - I've gone with 'form_message' and 'summary_message', which I think captures it quite well.

benthorner:
@kevindew this should be ready for another look now :-).

app/services/requirements/topic_checker.rb
```diff
@@ -0,0 +1,28 @@
+# frozen_string_literal: true
+
+module Requirements
+  class TopicChecker
+    TITLE_MAX_LENGTH = 150
```
kevindew: I don't think you meant for this to be here.

app/views/document_images/edit.html.erb
```diff
@@ -25,7 +27,7 @@
         },
         name: "alt_text",
         value: @image.alt_text,
-        error_message: @errors&.fetch("alt_text", nil)&.join("<br/>"),
+        error_message: alt_text_errors&.any? ? alt_text_errors.join("<br/>") : nil,
```
kevindew: Super minor and hopefully I didn't say this before but can't this just be `alt_text_errors&.join("<br>")` as that'll return nil?

spec/features/workflow/drafting_requirements_spec.rb
```diff
@@ -1,56 +1,30 @@
...
+  scenario do
+    given_there_is_a_document
+    when_the_document_has_issues_to_fix
+    then_i_see_a_warning_to_fix_the_issues
 
-  def and_save_a_title_which_has_multiple_lines
-    fill_in "document[title]", with: "MULTIPLE LINES ARGHHH \r\n 2nd line"
-    click_on "Save"
+    #when_i_try_to_generate_a_preview
+    #then_i_see_an_error_to_fix_the_issues
```
kevindew: Are these meant to be commented out?

app/services/requirements/topic_checker.rb
benthorner: Thanks :-).

app/views/document_images/edit.html.erb
benthorner: Not quite - if there are no errors, we need to return nil instead of an empty string.

spec/features/workflow/drafting_requirements_spec.rb
benthorner: I thought I'd write them here, as otherwise it didn't feel like a 'drafting requirements' spec. Would you prefer them left out?

app/views/document_images/edit.html.erb
kevindew: ooh good edge case catch 👍

spec/features/workflow/drafting_requirements_spec.rb
kevindew: Yeah it looks a bit like a WIP to me with these commented out. I think the commit message would be the appropriate place to explain further info.

The presence of them confuses me so I'd suggest removing.

benthorner: OK. Done.

app/views/document_images/edit.html.erb
benthorner: And now it's redundant with the update to publishing components :-).

PR #487, opened 2018-11-16
Make image messages consistent

kevindew:
This drops our mention of SVG as we don't support these yet
(unfortunately the guidance link may contradict that but then there's
various things in that document that currently don't accurate represent
Content Publisher.

It also drops the oxford comma and our use of jpg and jpeg.

### DIFF
--- a/config/locales/en/document_images/index.yml
+++ b/config/locales/en/document_images/index.yml
@@ -2,7 +2,7 @@ en:
   document_images:
     index:
       title: "Images for ‘%{title}’"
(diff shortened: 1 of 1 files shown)


config/locales/en/document_images/index.yml
```diff
@@ -13,7 +13,7 @@ en:
         upload_requirements:
           title: You need to
           no_file_selected: Select a file to upload
-          invalid_format: Select a a jpg, jpeg, png or gif image file
+          invalid_format: Select a a jpg, png or gif image file
```
leenagupte: @kevindew Typo police: "a a"?

kevindew: gold ⭐️to you @leenagupte

PR #494, opened 2018-11-19
Add link to raise support request

soniaturcotte:
In research, users thought they should use the feedback form for technical support. In private beta, we should be super clear about where they should go when they need help.

https://trello.com/c/WOEm2VQ2/452-review-the-entire-app-for-design-fixes


kevindew:
@soniaturcotte I've amended your commit here I fixed an extra comma in `<%= Plek.new.external_url_for("support") + "/technical_fault_report/new", %>` and added the key to the translation file.

It looks like this:

screen shot 2018-11-20 at 18 11 08

PR #496, opened 2018-11-19
Publisher information changes

soniaturcotte:
Remove dummy content from What's new page and replace with first release note

https://trello.com/c/JN04xUyt/458-update-publisher-information-in-whats-new-section

### DIFF
--- a/app/views/publisher_information/publisher_updates.govspeak.md
+++ b/app/views/publisher_information/publisher_updates.govspeak.md
@@ -1,3 +1,2 @@
-
-
-
+### Content Publisher is available in beta <span class="govuk-caption-m">29 November 2018</span>
+Beta partners can now access Content Publisher to publish news stories and press releases. Most features are available, [some will be released later](https://content-publisher.integration.publishing.service.gov.uk/beta-capabilities).
--- a/app/views/publisher_information/publisher_updates.govspeak.md
+++ b/app/views/publisher_information/publisher_updates.govspeak.md
@@ -1,9 +1,3 @@
-### Scheduling <span class="govuk-caption-m">22 October 2018</span>
-You can now schedule documents in to publish in the future
 
-
-
-### Inline images <span class="govuk-caption-m">16 September 2018</span>
-Inline images are now available for News stories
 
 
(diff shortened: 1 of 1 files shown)


kevindew:
@soniaturcotte We can't do these back links as we're not in a document context. @kgarwood can probably give you more information (I think he was intending to speak to you about them) but the problem we have is that you can get to these pages from anywhere so it's unclear what a back link should take you to and what role it should have compared to your browser back button.

soniaturcotte:
> @soniaturcotte I've dropped the commits on this to do with the back button since they won't work and we need a change to govuk_publishing_components to make the change you've suggested (which is going on in a different stream of work). Lets grab a chat about the back buttons though when you next have a chance see if we can work out how they should work ideally.
Ok 

> Do we have any idea what the initial content for the publisher updates page should be? as with this change it comes out looking like this:
Does it make sense to put something about our first release ”You can now publish News articles and Press releases“ ? (I’ll speak to Paola/Sam.)

paola-roccuzzo:
@kevindew 
This is the copy change I'm proposing - I would have changed it myself, but I'm just not sure how to create links to an internal page in markdown:

screenshot-content-publisher integration publishing service gov uk-2018 11 21-11-30-02

Copy is (using Integration link): 

> **Content Publisher is available in beta**
> 29 November 2018
> Beta partners can now access Content Publisher to publish news stories and press releases. Most features are available, [some will be released later](https://content-publisher.integration.publishing.service.gov.uk/beta-capabilities).

app/views/publisher_information/publisher_updates.govspeak.md
```diff
@@ -1,9 +1,2 @@
...
-You can now schedule documents in to publish in the future
-
-
-
-### Inline images <span class="govuk-caption-m">16 September 2018</span>
-Inline images are now available for News stories
-
-
+### Content Publisher is available in beta <span class="govuk-caption-m">29 November 2018</span>
+Beta partners can now access Content Publisher to publish news stories and press releases. Most features are available, [some will be released later](https://content-publisher.integration.publishing.service.gov.uk/beta-capabilities).
```
kevindew: This link currently points to our integration environment, if you change it to be `[some will be released later](/beta-capabilities)` then it should work for all.

soniaturcotte: Done

PR #497, opened 2018-11-19
Remove default image content

soniaturcotte:
We don’t have this yet. 

Let’s remove now in case we don’t by the time we launch. We can add the line back once the feature is built.

https://trello.com/c/WOEm2VQ2/452-review-the-entire-app-for-design-fixes


kevindew:
I tried to fix this one as it was failing based on the back link branch but I've ended up with those commits coming through by me changing the base branch.

PR #498, opened 2018-11-20
[DO NOT MERGE] Send auth_bypass_ids to asset manager

kevindew:
Trello: https://trello.com/c/dawXVwdW/439-asset-manager-files-are-not-available-when-using-access-tokens-or-if-someone-doesnt-have-asset-manager-permission

DNM until https://github.com/alphagov/authenticating-proxy/pull/108 is out

This is to allow these files to be available when a document is
previewed with a auth bypass token as a means of authentication.


spec/services/asset_manager_service_spec.rb
```diff
@@ -5,15 +5,21 @@
 RSpec.describe AssetManagerService do
   describe "#upload_bytes" do
     it "uploads a byte stream to Asset Manager and returns the asset URL" do
-      asset = double(:asset, content_type: "type", filename: "foo/bar.jpg")
+      asset = double(:asset,
+                     content_type: "type",
+                     filename: "foo/bar.jpg",
+                     document: FactoryBot.build(:document))
```
leenagupte: I think you can do `create(:document)` here and in the test below.

kevindew: Oh cool - I've updated. Note I went for build rather than create to avoid the db write.

leenagupte: Ah. Ok.

kevindew: Yeah you often have to have it written to the db if you're going to do any finds or similar, but using build is way faster if you can do so.

PR #499, opened 2018-11-20
Allow documents to be "removed" or "retired"

leenagupte:
Trello: https://trello.com/c/H3jeLoZW

## Motivation

We don't currently support unpublishing or withdrawing a document (which in our app has the terminology of remove and retire respectively).

Our intention is to do this for users if they ask us to do so while in private beta.

This implements the minimal functionality required to allow users to request a document be retired/removed via zendesk. Only the backend call to publishing-api has been implemented, and is executed via a rake task.

The frontend and changes to the "state" of the document will be implemented later.

### DIFF
--- a/app/services/document_unpublishing_service.rb
+++ b/app/services/document_unpublishing_service.rb
@@ -5,9 +5,15 @@ class DocumentUnpublishingService
     GdsApi.publishing_api_v2.unpublish(document.content_id, type: "withdrawal", explanation: explanatory_note, locale: locale)
   end
 
-  def remove(document, explanatory_note: nil, alternative_path: nil)
+  def remove(document, explanatory_note: nil, alternative_path: nil, locale: "en")
     delete_assets(document.images)
-    GdsApi.publishing_api_v2.unpublish(document.content_id, type: "gone", explanation: explanatory_note, alternative_path: alternative_path)
+    GdsApi.publishing_api_v2.unpublish(
+      document.content_id,
+      type: "gone",
+      explanation: explanatory_note,
+      alternative_path: alternative_path,
+      locale: locale,
+    )
   end
 
   def remove_and_redirect(document, redirect_path)
--- a/app/services/document_unpublishing_service.rb
+++ b/app/services/document_unpublishing_service.rb
@@ -5,9 +5,9 @@ class DocumentUnpublishingService
     GdsApi.publishing_api_v2.unpublish(document.content_id, type: "withdrawal", explanation: explanatory_note, locale: locale)
   end
 
-  def remove(document)
+  def remove(document, explanatory_note: nil)
     delete_assets(document.images)
-    GdsApi.publishing_api_v2.unpublish(document.content_id, type: "gone")
+    GdsApi.publishing_api_v2.unpublish(document.content_id, type: "gone", explanation: explanatory_note)
   end
 
   def remove_and_redirect(document, redirect_path)
--- a/app/services/document_unpublishing_service.rb
+++ b/app/services/document_unpublishing_service.rb
@@ -4,4 +4,8 @@ class DocumentUnpublishingService
   def retire(document, explanatory_note)
     GdsApi.publishing_api_v2.unpublish(document.content_id, type: "withdrawal", explanation: explanatory_note)
   end
+
+  def remove(document)
+    GdsApi.publishing_api_v2.unpublish(document.content_id, type: "gone")
+  end
 end
--- a/app/services/document_unpublishing_service.rb
+++ b/app/services/document_unpublishing_service.rb
@@ -1,8 +1,8 @@
 # frozen_string_literal: true
 
 class DocumentUnpublishingService
-  def retire(document, explanatory_note)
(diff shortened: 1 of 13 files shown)


spec/features/workflow/unpublishing_spec.rb
```diff
@@ -0,0 +1,61 @@
+# frozen_string_literal: true
+
+require "spec_helper"
```
benthorner: We've setup the app so you don't need to require the spec_helper :-).

spec/services/document_unpublishing_service_spec.rb
```diff
@@ -0,0 +1,72 @@
+# frozen_string_literal: true
+
+require "spec_helper"
+
+RSpec.describe DocumentUnpublishingService do
+  describe "#retire" do
+    let(:document) { create(:document) }
+
+    before do
+      @explanatory_note = "The document is out of date"
```
benthorner: Could this be a let?

spec/features/workflow/unpublishing_spec.rb
```diff
@@ -0,0 +1,61 @@
...
+require "rake"
+
+Rails.application.load_tasks
```
benthorner: I think this will have an effect for all the specs - would be good to move this somewhere central so it's clear it applies globally.

leenagupte: I'll move it to `spec_helper`

spec/features/workflow/unpublishing_spec.rb
leenagupte: Fixed up. The other test files have been cleaned up too.

spec/features/workflow/unpublishing_spec.rb
leenagupte: Fixed up.

spec/services/document_unpublishing_service_spec.rb
benthorner: That's interesting - you mean like `let(:explanatory_note) { "blah" }`?

leenagupte: Exactly like that.

leenagupte: Just tried it again, and it's working now. I must have had a syntax error earlier.

leenagupte: Fixed up.

benthorner: It works for me (removing the @ everywhere).

benthorner: Just missed your comment :-).

leenagupte:
@kevindew @benthorner If you're both happy, I'll merge.

benthorner:
👍

app/services/document_unpublishing_service.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+
+class DocumentUnpublishingService
+  def retire(document, explanatory_note)
+    GdsApi.publishing_api_v2.unpublish(document.content_id, type: "withdrawal", explanation: explanatory_note)
```
kevindew: For these api calls I think it'd make sense to include the locale in this. I know for now we'll only being dealing with "en" but be good to not have surprises later when this becomes more pertinent.

lib/tasks/unpublish.rake
```diff
@@ -0,0 +1,28 @@
+# frozen_string_literal: true
+
+namespace :unpublish do
+  desc "Retire a document"
+  task :retire_document, %I(base_path explanatory_note) => :environment do |_, args|
+    raise "Missing base_path parameter" unless args.base_path
+    raise "Missing explanatory_note parameter" unless args.explanatory_note
+
+    document = Document.find_by(base_path: args.base_path)
```
kevindew: I'm not sure we can safely rely on base path here to remove documents as it could have changed in a later draft of their document. 

I'd probably suggest using the `fe7c00c1-a34f-419c-88eb-c844ba70b1c5:en` hybrid since that's the identifier we use for all our URLs?

app/services/document_unpublishing_service.rb
leenagupte: I'll update and fix up.

lib/tasks/unpublish.rake
leenagupte: I'll add locale as a optional parameter to the rake task, with "en" as the default and make "content_id" the required param rather than base_path.

leenagupte:
@kevindew 

> These tasks will fail if these are drafts and not published on the Publishing API - should we handle that here?

It probably doesn't need to be on the service in the same way that `publish_draft` doesn't check the state of the document before trying to publish. But something could be added to the rake task to fail early if the document is in an invalid state.

> Similarly if we have a draft and are published we may want to be able to discard the published item but keep the draft

the default action is not to discard drafts. See: https://github.com/alphagov/gds-api-adapters/blob/master/lib/gds_api/publishing_api_v2.rb#L147

leenagupte:
@kevindew @benthorner I'm closing this PR. It's been replaced with #510

PR #501, opened 2018-11-20
Iterate design of 'Content has been published' confirmation screen

MuriloDalRi:
https://trello.com/c/JsrpqR3h/419-iterate-design-of-content-has-been-published-confirmation-screen

- Removes empty `<div class="govuk-grid-row">` above the confirmation
  panel, which was causing extra spacing.
- Removes 30px margin below title in confirmation panel (https://github.com/alphagov/govuk_publishing_components/pull/628)
- Replace copy-to-clipboard section to just show link in normal govuk-body styling

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -79,6 +79,7 @@
     <% end %>
 
     <main class="govuk-main-wrapper<%= " govuk-main-wrapper--l" if yield(:back_link).blank?%>" id="main-content" role="main">
+      <% if yield(:title).present? %>
       <div class="govuk-grid-row">
         <div class="govuk-grid-column-two-thirds">
           <span class="govuk-caption-l"><%= yield(:context) %></span>
@@ -88,6 +89,7 @@
           <%= yield(:title_side) %>
         </div>
       </div>
+      <% end %>
       <%= yield %>
     </main>
   </div>
--- a/app/views/publish_document/published.html.erb
+++ b/app/views/publish_document/published.html.erb
@@ -12,10 +12,7 @@
       <%= render "govuk_publishing_components/components/govspeak" do %>
         <%= govspeak_to_html t("publish_document.published.reviewed.body", title: @document.title) %>
 
-        <%= render "govuk_publishing_components/components/copy_to_clipboard",
-          label: nil,
-          copyable_content: DocumentUrl.new(@document).public_url,
-          button_text: "Copy link" %>
+        <%= link_to(nil, DocumentUrl.new(@document).public_url, class: "govuk-link") %>
       <% end %>
     <% else %>
       <%= render "govuk_publishing_components/components/panel", {
@@ -25,10 +22,9 @@
       <%= render "govuk_publishing_components/components/govspeak" do %>
         <%= govspeak_to_html t("publish_document.published.published_without_review.body", title: @document.title) %>
 
-        <%= render "govuk_publishing_components/components/copy_to_clipboard",
-          label: t("publish_document.published.published_without_review.send_label"),
-          copyable_content: document_url(@document, utm_content: "2i-link"),
-          button_text: "Copy link" %>
+        <p> <%= t("publish_document.published.published_without_review.send_label") %> </p>
+
+        <%= link_to(nil, document_url(@document, utm_content: "2i-link"), class: "govuk-link") %>
       <% end %>
     <% end %>
   </div>


app/views/layouts/application.html.erb
```diff
@@ -79,15 +79,17 @@
...
+          </div>
+          <div class="govuk-grid-column-one-third app-grid-column--end">
+            <%= yield(:title_side) %>
+          </div>
         </div>
-        <div class="govuk-grid-column-one-third app-grid-column--end">
-          <%= yield(:title_side) %>
-        </div>
-      </div>
+      <% end %>
```
kevindew: 👍

app/views/publish_document/published.html.erb
```diff
@@ -1,4 +1,9 @@
-<% content_for :back_link, document_path(@document) %>
+<% if @document.review_state == "reviewed" %>
+  <% content_for :back_link, t("application.back_link.back"), documents_path %>
```
kevindew: I don't think this will work

config/locales/en/application.yml
```diff
@@ -7,3 +7,5 @@ en:
       support_and_feedback: Support and feedback
       documentation: Documentation
       how_to_write_publish_improve: How to write, publish, and improve content
+    back_link:
+      back: Back to document list
```
kevindew: This looks to be a global change - and we only want to change this for a specific view.

app/views/publish_document/published.html.erb
MuriloDalRi: Do you mean the implementation or because my change to the component hasn't been merged?

config/locales/en/application.yml
MuriloDalRi: I wasn't sure whether to hardcode this in that view or put it in a locale file and thought it might go here. Should I just hardcode it in the view?

app/views/publish_document/published.html.erb
kevindew: it looks like we're passing three arguments to `content_for` when it can just accept two: the key and the value. The value of this is then passed to the component as the href argument: https://github.com/alphagov/content-publisher/blob/6ec7337fea74534cce1ea515633a72981cd936c3/app/views/layouts/application.html.erb#L44

config/locales/en/application.yml
kevindew: Yeah I'd hardcode it as that seems consistent with: https://github.com/alphagov/content-publisher/blame/6ec7337fea74534cce1ea515633a72981cd936c3/docs/adr/0004-editing-microcopy.md#L43

app/views/layouts/application.html.erb
```diff
@@ -41,7 +41,14 @@
     } %>
 
     <% if yield(:back_link).present? %>
-      <%= render "govuk_publishing_components/components/back_link", href: yield(:back_link) %>
+      <% if yield(:back_link).include?('text') %>
```
benthorner: I think we could have a problem here if a link happens to include the word 'text' - e.g. a title. Perhaps it's safer to just have a ':custom_back_link' content block, so it doesn't interfere with the current one?

MuriloDalRi: 👍

app/views/publish_document/published.html.erb
```diff
@@ -1,4 +1,9 @@
-<% content_for :back_link, document_path(@document) %>
+<% if @document.review_state == "reviewed" %>
+  <% content_for :custom_back_link, {text: "Back to document list", href: documents_path}.to_json.html_safe %>
```
benthorner: Another idea to make this simpler: you could render the custom back link as a block right here instead of passing JSON hashes around.

MuriloDalRi: But wouldn't that render the component inside the `<main>` block?

benthorner: I don't think so. You can just yield the raw HTML in the `custom_back_link` section.

kevindew:
@MuriloDalRi it might be an idea to separate the back button part of this PR into a separate PR as the other two parts of this seem good to be merged but that seems to be ongoing.

Although it seems clever to use JSON as means to pass an object around I think that's something quite unexpected for views so I think that's an approach we should try avoid.

Based on the things Ben was suggesting previously the simplest approach could be that we have:

In the global layout:

```
yield(:back_link)
```

and then in each view that sets a back link have either:

```
<% content_for :back_link,
  render "govuk_publishing_components/components/back_link", {
     href: document_path(@document),
  } %>
```

and for a more complex one:

```
<% content_for :back_link,  
  render "govuk_publishing_components/components/back_link", {
     text: "Back to document list",
     href: documents_path,
  } %>
%>

```

MuriloDalRi:
@kevindew I'll open a new PR for the back link 👍

PR #510, opened 2018-11-21
Allow documents to be "removed" or "retired"

leenagupte:
Trello: https://trello.com/c/H3jeLoZW

Supersedes PR #499 

## Motivation

We don't currently support unpublishing or withdrawing a document (which in our app has the terminology of remove and retire respectively).

Our intention is to do this for users if they ask us to do so while in private beta.

This implements the minimal functionality required to allow users to request a document be retired/removed via zendesk. Only the backend call to publishing-api has been implemented, and is executed via a rake task.

The frontend and changes to the "state" of the document will be implemented later.

### DIFF
--- a/app/services/document_unpublishing_service.rb
+++ b/app/services/document_unpublishing_service.rb
@@ -22,6 +22,10 @@ class DocumentUnpublishingService
     delete_assets(document.images)
   end
 
+  def remove_and_redirect(document, redirect_path)
+    GdsApi.publishing_api_v2.unpublish(document.content_id, type: "redirect", alternative_path: redirect_path)
+  end
+
 private
 
   def delete_assets(assets)
--- a/app/services/document_unpublishing_service.rb
+++ b/app/services/document_unpublishing_service.rb
@@ -22,8 +22,13 @@ class DocumentUnpublishingService
     delete_assets(document.images)
   end
 
-  def remove_and_redirect(document, redirect_path)
-    GdsApi.publishing_api_v2.unpublish(document.content_id, type: "redirect", alternative_path: redirect_path)
+  def remove_and_redirect(document, redirect_path, explanatory_note: nil)
+    GdsApi.publishing_api_v2.unpublish(
+      document.content_id,
+      type: "redirect",
+      explanation: explanatory_note,
+      alternative_path: redirect_path,
+    )
 
     delete_assets(document.images)
   end
--- a/app/services/document_unpublishing_service.rb
+++ b/app/services/document_unpublishing_service.rb
@@ -10,11 +10,12 @@ class DocumentUnpublishingService
     )
   end
 
-  def remove(document, explanatory_note: nil)
+  def remove(document, explanatory_note: nil, alternative_path: nil)
     GdsApi.publishing_api_v2.unpublish(
       document.content_id,
       type: "gone",
       explanation: explanatory_note,
+      alternative_path: alternative_path,
     )
 
     delete_assets(document.images)
--- a/app/services/document_unpublishing_service.rb
+++ b/app/services/document_unpublishing_service.rb
@@ -28,6 +28,7 @@ class DocumentUnpublishingService
       type: "redirect",
       explanation: explanatory_note,
       alternative_path: redirect_path,
+      locale: document.locale,
     )
 
     delete_assets(document.images)
--- a/app/services/document_unpublishing_service.rb
+++ b/app/services/document_unpublishing_service.rb
@@ -2,6 +2,11 @@
 
 class DocumentUnpublishingService
   def retire(document, explanatory_note)
-    GdsApi.publishing_api_v2.unpublish(document.content_id, type: "withdrawal", explanation: explanatory_note)
+    GdsApi.publishing_api_v2.unpublish(
+      document.content_id,
+      type: "withdrawal",
+      explanation: explanatory_note,
+      locale: document.locale,
+    )
   end
 end
--- a/app/services/document_unpublishing_service.rb
+++ b/app/services/document_unpublishing_service.rb
@@ -1,0 +1,7 @@
+# frozen_string_literal: true
+
+class DocumentUnpublishingService
+  def retire(document, explanatory_note)
+    GdsApi.publishing_api_v2.unpublish(document.content_id, type: "withdrawal", explanation: explanatory_note)
+  end
+end
--- a/app/services/document_unpublishing_service.rb
+++ b/app/services/document_unpublishing_service.rb
@@ -10,8 +10,13 @@ class DocumentUnpublishingService
     )
   end
 
-  def remove(document)
-    GdsApi.publishing_api_v2.unpublish(document.content_id, type: "gone")
+  def remove(document, explanatory_note: nil)
+    GdsApi.publishing_api_v2.unpublish(
+      document.content_id,
+      type: "gone",
+      explanation: explanatory_note,
+    )
+
     delete_assets(document.images)
   end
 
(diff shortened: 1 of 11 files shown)


app/services/document_unpublishing_service.rb
```diff
@@ -0,0 +1,44 @@
+# frozen_string_literal: true
+
+class DocumentUnpublishingService
+  def retire(document, explanatory_note, locale: "en")
```
kevindew: If we're passing in document we can use that as the source of the locale rather than an extra argument

docs/retiring-and-removing-documents.md
```diff
@@ -0,0 +1,61 @@
...
+
+- CONTENT_ID
+- NOTE
+
+Optional parameters:
+
+- LOCALE (set to "en" by default)
+
+```
+rake unpublish:retire_document CONTENT_ID='a-content-id' NOTE='A note'
```
kevindew: Just a suggestion and not a blocker but since all these commands have content_id as an argument shall we pass that in through the normal rake way e.g. `rake unpublish:retire_document[a-content-id] NOTE='A note'` ? Just feels a bit more succinct and it would save you needing lines like https://github.com/alphagov/content-publisher/pull/510/files#diff-b0f8989c20523cc86786623565a5f99cR21 and associated tests

docs/retiring-and-removing-documents.md
```diff
@@ -0,0 +1,61 @@
...
+- CONTENT_ID
+- NEW_PATH
+
+Optional parameters:
+
+- LOCALE (set to "en" by default)
+- NOTE
+
+```
+rake unpublish:remove_document CONTENT_ID='a-content-id' NEW_PATH='/redirect-to-here'
```
kevindew: heads up this should be remove_and_redirect_document

lib/tasks/unpublish.rake
```diff
@@ -0,0 +1,63 @@
+# frozen_string_literal: true
+
+namespace :unpublish do
```
kevindew: Another non blocker - do you think we could avoid having the task as unpublish? just wonder if we can try avoid that terminology and we've mostly managed it here so far.

Since everything applies to document what do you think of a namespace of document then we can have `document:retire` which seems to mean the same as `unpublish:retire_document` but is shorter and avoids that dreaded word 😉

lib/tasks/unpublish.rake
```diff
@@ -0,0 +1,63 @@
...
+    raise "Missing NOTE value" if ENV["NOTE"].nil?
+
+    content_id = ENV["CONTENT_ID"]
+    explanatory_note = ENV["NOTE"]
+    locale = ENV["LOCALE"] || "en"
+
+    document = Document.find_by(content_id: content_id, locale: locale)
+    raise "Document must have a published version before it can be retired" unless document.has_live_version_on_govuk
+
+    DocumentUnpublishingService.new.retire(document, explanatory_note, locale: locale) if document.present?
```
kevindew: I guess you don't need this conditional here as it would exception on `document.has_live_version_on_govuk`. You could kill it off by using `find_by!` in the initial call

spec/features/workflow/unpublishing_spec.rb
```diff
@@ -0,0 +1,208 @@
...
+      expect { Rake::Task["unpublish:retire_document"].invoke }.to raise_error("Missing CONTENT_ID value")
+    end
+
+    it "raises an error if a NOTE is not present" do
+      ClimateControl.modify CONTENT_ID: "a-content-id" do
+        expect { Rake::Task["unpublish:retire_document"].invoke }.to raise_error("Missing NOTE value")
+      end
+    end
+
+    it "copes with commas in the explanatory note" do
```
kevindew: I'm super pleased you checked for this but I don't think we actually need a test in the code for it.

Mainly as it's not realistic as it's down to bash's string handling rather than in the ruby.

Plus since we're using an env var and not multiple rake arguments we're free from that concern

spec/services/document_unpublishing_service_spec.rb
```diff
@@ -0,0 +1,116 @@
+# frozen_string_literal: true
+
+RSpec.describe DocumentUnpublishingService do
+  describe "#retire" do
+    let(:document) { create(:document) }
+    let(:explanatory_note) { "The document is out of date" }
+
+    it "withdraws a document in publishing-api with an explanatory note" do
+      stub_publishing_api_unpublish(document.content_id, body: { type: "withdrawal", explanation: explanatory_note, locale: "en" })
```
kevindew: On these stubs here it'd be good to use the locale from the document too (I appreciate it may be news to you that we even store it) save the tests breaking if we set different locales in the factory

lib/tasks/unpublish.rake
```diff
@@ -0,0 +1,63 @@
...
+        alternative_path: alternative_path,
+        locale: locale,
+      )
+    end
+  end
+
+  desc "Remove and redirect a document on GOV.UK e.g. unpublish:remove_and_redirect_document CONTENT_ID='a-content-id' NEW_PATH='/redirect-to-here'"
+  task remove_and_redirect_document: :environment do
+    raise "Missing CONTENT_ID value" if ENV["CONTENT_ID"].nil?
+    raise "Missing NEW_PATH value" if ENV["NEW_PATH"].nil?
```
kevindew: might want to use unless ENV["NEW_PATH"].present? just to rule out empty strings too

app/services/document_unpublishing_service.rb
leenagupte: Good point!

docs/retiring-and-removing-documents.md
leenagupte: Yeah, that's a good point. I could change them all to only use ENV VARS for the optional arguments.

kevindew: I think you'd still probably want an env var for a note even when that's required. But that's mostly because of rake's yucky syntax and the comma pains associated.

lib/tasks/unpublish.rake
leenagupte: Well, they do all call the DocumentUnpublishingService, so we aren't really avoiding that words. I couldn't think of another word that groups the tasks.

kevindew: You could have DocumentRemover and DocumentRetirer and not group them?

docs/retiring-and-removing-documents.md
leenagupte: True, and to keep it consistent with the other tasks that have note as an optional parameter.

lib/tasks/unpublish.rake
leenagupte: I think `blank?` might be the better option:

```
irb(main):001:0> a = ""
=> ""
irb(main):002:0> a.present?
=> false
irb(main):003:0> a.nil?
=> false
irb(main):004:0> a.blank?
=> true
irb(main):005:0> b = nil
=> nil
irb(main):006:0> b.present?
=> false
irb(main):007:0> b.nil?
=> true
irb(main):008:0> b.blank?
=> true
```

kevindew: Ah yes I agree!

leenagupte: It's odd. For some reason the tests broke when I changed it to `present?`.

lib/tasks/unpublish.rake
kevindew: sorry! not joking but your call on it - definitely don't want to suggest something that'll take a load of time and not add much value

PR #513, opened 2018-11-22
Manual preview

benthorner:
https://trello.com/c/4iKDzhBY/417-separate-saving-a-change-from-the-creation-of-a-preview

This builds on our new, centralised requirements to introduce a manual 'create preview' action that allows to fail silently in other cases where we no longer need to rely on automatic preview creation.

The commits should give a good overview of the changes. I thought of a couple of issues that I didn't think worth addressing as part of this PR:

   * We don't show requirements issues if the document is in an error state (seems unlikely to happen?)
   * We don't protect the final publish action of the PublishDocumentController when our requirements aren't met


benthorner:
Thanks @kevindew. And the answer to your question is 'yes' :-). I'm not quite sure what you mean in your last paragraph. Could you elaborate a bit more?

I'm going to merge this one now, but happy to continue the discussion :-).

PR #519, opened 2018-11-23
Render topics on summary page using bread crumb

kgarwood:
Changing appearance of each topic so that rather than being rendered as an enumerated list
of progressively more specific terms, it would render it like "A > B > C" on one line,
in much the way a breadcrumb for reaching a specific web page appears elsewhere on GOV.UK
forms.

GOV.UK's govuk-breadcrumbs__list helps render a single topic.  But this class looks like
it was not designed to render well within a list of breadcrumb lists.  In order to correct
this, I made a new component that specifies that for any new govuk-breadcrumb__list.
The component adds vertical space between successive topics so it looks more readable.

See Trello card: https://trello.com/c/QHbq9VOt/455-show-topics-as-breadcrumbs-in-the-summary-page

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -21,6 +21,7 @@
 @import "components/image-meta";
 @import "components/image-cropper";
 @import "components/side-nav";
+@import "components/topic-breadcrumb-trail";
 
 @import "utilities/display";
 @import "utilities/overwrites";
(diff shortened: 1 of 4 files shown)


kevindew:
```
  <% breadcrumbs = capture do %>
    <% if @document.topics.any? %>
      <ul class="govuk-list">
        <% @document.topics.each do |topic| %>
          <li>
            <%= render "govuk_publishing_components/components/breadcrumbs", {
              breadcrumbs: topic.breadcrumb.map { |crumb_topic| { title: crumb_topic.title } }
            } %>
          </li>
        <% end %>
      </ul>
    <% else %>
      <%= t("documents.show.topics.no_topics") %>
    <% end %>
  <% end %>
```

Here's suggestion for using the breadcrumbs component

PR #522, opened 2018-11-23
Corrections, added links, and various bits

paolaroccuzzo:
Loads of edits

### DIFF
--- a/app/views/publisher_information/how_to_use_publisher.govspeak.md
+++ b/app/views/publisher_information/how_to_use_publisher.govspeak.md
@@ -1,19 +1,17 @@
-Content publisher is a new way of publishing content on GOV.UK. It is designed to make it 
+Content Publisher is a new way of publishing content on GOV.UK. It is designed to make it 
 easier to publish useful content and to help you use data to manage that content.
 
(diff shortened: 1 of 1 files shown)


app/views/publisher_information/how_to_use_publisher.govspeak.md
```diff
@@ -1,19 +1,17 @@
...
 
 ##How to create content
 
-Content publisher can be used to create news stories and press releases. Other document types still need to be created in Whitehall publisher.
+Content Publisher can be used to create news stories and press releases. Other document types still need to be created in Whitehall publisher.
 
 You should always consider if content really needs to be published on GOV.UK. Unnecessary content makes it harder for users to find what the need.
 
-Read guidance on what to publish on GOV.UK.
+Read [guidance on what to publish on GOV.UK](documents/publishing-guidance).
```
benthorner: I think we have a syntax to make this open in a separate tab - should we do that for the links on this page?

paola-roccuzzo: My understanding is that we use that syntax for pages where we could lose input data in case users click a link - as this is on a guidance page we shouldn't open the link in a new tab.

PR #525, opened 2018-11-26
Render topics using GOV.UK's breadcrumb component

kgarwood:
Changing the way of rendering a topic from an enumerated list of more
specific categories to rendering it like "A > B > C", in the same way
breadcrumb trails are used to show web page path ways to users in other
parts of GOV.UK.

### DIFF
--- a/spec/features/editing_topics/show_topics_spec.rb
+++ b/spec/features/editing_topics/show_topics_spec.rb
@@ -46,7 +46,7 @@ RSpec.feature "Show the topics for a document" do
   end
 
   def then_i_see_the_topic_breadcrumb
-    within("#topics .topic-breadcrumb") do
+    within("#topics .govuk-list") do
       expect(page).to have_content("Level One Topic")
       expect(page).to have_content("Level Two Topic")
       expect(page).to have_content("Level Three Topic")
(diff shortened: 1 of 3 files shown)


kgarwood:
Ah... I hadn't noticed that.  I've fixed it so that the breadcrumbs will use font size 19 instead of the default 16 it would normally use.

spec/features/editing_topics/show_topics_spec.rb
```diff
@@ -46,7 +46,7 @@ def when_the_document_has_a_topic
   end
 
   def then_i_see_the_topic_breadcrumb
-    within("#topics .topic-breadcrumb") do
+    within("#topics .govuk-list") do
```
kevindew: I think you can drop the .govuk-list specificity here - since that's just used for generic list styling.

kgarwood: Done. Within removed

kevindew: the #topics one is still good

kevindew: Sorry @kgarwood I deleted your comment by mistake when I meant to kill one of mine

kgarwood: And good idea with an image, the work in this pull request controls rendering of the topics:
content_publisher_pull_525

kgarwood: Fixed, within #topic is retained in the test.

PR #526, opened 2018-11-26
Remove contenteditable

alex-ju:
as it currently blocks users from inserting text after using the injectors

### Note
This change was done in the main repository: https://github.com/alphagov/markdown-toolbar-element/pull/2. This PR updates the compiled JavaScript file.

https://trello.com/c/fLfbI7Vi

### DIFF
--- a/app/assets/javascripts/components/markdown-toolbar.js
+++ b/app/assets/javascripts/components/markdown-toolbar.js
@@ -583,13 +583,11 @@
     var after = textarea.value.slice(textarea.selectionEnd);
 
     if (canInsertText === null || canInsertText === true) {
-      textarea.contentEditable = 'true';
       try {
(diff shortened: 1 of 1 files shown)


benthorner:
I should probably ask if this attribute is set to true by default? Can't seem to find a clear spec anywhere :-|.

alex-ju:
@benthorner `contenteditable` enables the browser to perform a set of action (including character insertion, spelling and grammar) on the editing host. When the editing host is a textarea (or input text or other editable elements) this set of actions is enabled by default. On other elements, it can be enabled using this attribute. I'm not sure this helps, but W3C doesn't specify these elements anywhere so it's a 'de facto' standard.

benthorner:
Thanks @alex-ju - sounds like we're safe here :-).

PR #527, opened 2018-11-26
Fix ordered list injector

alex-ju:
Run the `orderedList` style args only when we have a selection, otherwise we default on a simple prefixing of the current line.

### Note
This change was done in the main repository: https://github.com/alphagov/markdown-toolbar-element/pull/3. This PR updates the compiled JavaScript file.

https://trello.com/c/37P9Kykk

### DIFF
--- a/app/assets/javascripts/components/markdown-toolbar.js
+++ b/app/assets/javascripts/components/markdown-toolbar.js
@@ -397,7 +397,7 @@
 
       var _this14 = _possibleConstructorReturn(this, (MarkdownOrderedListButtonElement.__proto__ || Object.getPrototypeOf(MarkdownOrderedListButtonElement)).call(this));
 
-      styles.set(_this14, { prefix: '1. ', multiline: true, orderedList: true });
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/components/markdown-toolbar.js
```diff
@@ -620,7 +620,7 @@
     var text = textarea.value.slice(textarea.selectionStart, textarea.selectionEnd);
 
     var result = void 0;
-    if (styleArgs.orderedList) {
+    if (styleArgs.orderedList && text) {
```
benthorner: There's a slight difference in behaviour between this one and bullet points: clicking the button multiple times for bullets (or quotes) inserts on a new line; clicking for a numbered list just keeps adding to the same line.

alex-ju: Updated the code so they behave the same (adding a new line after the item - as it's needed so the markdown is rendered correctly)

PR #528, opened 2018-11-26
Change back link

MuriloDalRi:
Ticket: https://trello.com/c/JsrpqR3h/419-iterate-design-of-content-has-been-published-confirmation-screen

We need to display a back link button with custom text in a specific view.


benthorner:
I think we can avoid modifying every page by having a 'custom_back_link' block, like we had before ;-).

PR #529, opened 2018-11-26
When content is changed on the current page we should alert users bef…

DilwoarH:
…ore leaving page

## User need
As a publisher
I want to be alerted when I'm about to leave the page
So that I decide if I want to drop the current edits or go back and save

## What
We need to provide users with an alert message to make them aware of the fact that leaving for another page means they're loosing changes on the current page.

## Background
___How to replicate:___

1. go to content publisher
2. create a new document (news > press release)
3. at the create and edit screen add text to body field
4. press a link in the footer or header or try to reload the page

Ticket: https://trello.com/c/uM6zb4sM/470-when-content-is-changed-on-the-current-page-we-should-alert-users-before-leaving-page


benthorner:
Could we apply this as a (not visible) component on the page to avoiding executing it everywhere? This seems to be the pattern we've got for contextual guidance.

alex-ju:
Looks good overall 👍 . I agree with @benthorner, would be great if we use a similar pattern with the contextual guidance.

app/assets/javascripts/views/edit.js
```diff
@@ -0,0 +1,18 @@
+var editDocument = document.querySelectorAll('.edit_document')
+if (editDocument) {
+  document.querySelectorAll('[data-warn-before-unload=true]').forEach(function (element) {
```
kevindew: Rather than having this set on a particular form input could we have this set on a form itself.

So we can have `<form data-warn-before-unload=true>`  and just listen for change events that have bubbled up from it? 

Then we can just listen to the submit event on the form too rather than having to put an event specifically on buttons within the form.

app/assets/javascripts/views/edit.js
```diff
@@ -0,0 +1,18 @@
...
+    element.addEventListener('change', function () {
+      if (!window.onbeforeunload) {
+        window.onbeforeunload = function () { return true }
```
kevindew: Do we have to do more than this for full browser support? 

I don't know if there are many browsers that still support custom text (maybe IE?) but I wonder if those ones may not show a message if this just returns true?

Also, any reason why we use `window.onbeforeunload` rather than `window.addEventListener` like we use for other events?

app/assets/javascripts/views/edit.js
DilwoarH: @kevindew - thanks for this, this was a good idea. This has been done.

kevindew: Great thanks 👍

app/assets/javascripts/views/edit.js
```diff
@@ -0,0 +1,19 @@
...
+    e.returnValue = true
+  }
+
+  document.querySelectorAll('[data-warn-before-unload=true]').forEach(function (element) {
+    element.addEventListener('change', function () {
+      window.addEventListener('beforeunload', beforeUnloadEvent)
+    })
+  })
+
+  document.querySelectorAll('button[type=submit]').forEach(function (element) {
```
kevindew: I think rather than listening for a click event on this button we can add an eventListener on the form submit event

app/assets/javascripts/components/warn-before-unload.js
```diff
@@ -0,0 +1,27 @@
+function WarnBeforeUnload ($module, $scope) {
+  if (!$module) $module = window
+  if (!$scope) $scope = document
+  this.$module = $module
+  this.$forms = $scope.querySelectorAll('[data-warn-before-unload=true]')
+}
+
+WarnBeforeUnload.prototype.init = function () {
+  var _this = this
+  this.$forms.forEach(function ($form) {
```
alex-ju: This may result in a `Cannot read property 'forEach' of undefined`. Can we check for this.$forms to exist and return from this function if not?

DilwoarH: updated

PR #531, opened 2018-11-26
Update autocomplete setup for tags selection

alex-ju:
This PR:
- updates the autocomplete component to use the latest version
- updates the autocomplete setup to narrow down suggestions after typing 3 characters

[Trello card](https://trello.com/c/2aQtJkCZ)


benthorner:
@alex-ju I think you need to turn off 'showAllValues', otherwise it'll ignore the minLength option.

alex-ju:
@benthorner that's right. I did that by removing the `showAllValues` options, which is false by default.

alex-ju:
@benthorner thanks! I'll wait to discuss the approach with our PM before merging it.

PR #540, opened 2018-11-27
Add GOV.UK live URL to force publish confirmation page

MuriloDalRi:
https://trello.com/c/MrK9ulAL/474-add-govuk-live-url-to-force-publish-confirmation-page

screen shot 2018-11-27 at 16 03 25

### DIFF
--- a/app/views/publish_document/published.html.erb
+++ b/app/views/publish_document/published.html.erb
@@ -29,8 +29,10 @@
 
         <p> <%= t("publish_document.published.published_without_review.send_label") %> </p>
 
-        <%= link_to(nil, document_url(@document, utm_content: "2i-link"), class: "govuk-link") %>
-      <% end %>
+      <%= render "govuk_publishing_components/components/copy_to_clipboard",
+        label: t("publish_document.published.published_without_review.send_label"),
+        copyable_content: document_url(@document, utm_content: "2i-link"),
+        button_text: "Copy link" %>
     <% end %>
   </div>
 </div>
--- a/app/views/publish_document/published.html.erb
+++ b/app/views/publish_document/published.html.erb
@@ -29,7 +29,7 @@
         <% stripped_url = DocumentUrl.new(@document).public_url.sub(/^https?\:\/\//,'') %>
         <%= link_to(nil, stripped_url, class: "govuk-link govuk-link--no-visited-state") %>
 
-        <h2 id="what-to-do-next">What to do next</h2>
+        <%= govspeak_to_html t("publish_document.published.published_without_review.what_to_do_govspeak") %>
       <% end %>
 
       <%= render "govuk_publishing_components/components/copy_to_clipboard",
--- a/app/views/publish_document/published.html.erb
+++ b/app/views/publish_document/published.html.erb
@@ -26,8 +26,11 @@
 
       <%= render "govuk_publishing_components/components/govspeak" do %>
         <%= govspeak_to_html t("publish_document.published.published_without_review.body_govspeak", title: @document.title) %>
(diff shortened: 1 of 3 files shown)


app/views/publish_document/published.html.erb
```diff
@@ -17,19 +17,20 @@
...
       } %>
 
       <%= render "govuk_publishing_components/components/govspeak" do %>
-        <%= govspeak_to_html t("publish_document.published.published_without_review.body_govspeak", title: @document.title) %>
+        <%= govspeak_to_html t("publish_document.published.published_without_review.body_govspeak", title: @document.title, link_to_live: DocumentUrl.new(@document).public_url) %>
 
-        <p> <%= t("publish_document.published.published_without_review.send_label") %> </p>
-
-        <%= link_to(nil, document_url(@document, utm_content: "2i-link"), class: "govuk-link") %>
+        <%= render "govuk_publishing_components/components/copy_to_clipboard",
```
benthorner: I think this should be outside of the govspeak component.

MuriloDalRi: 👍

soniaturcotte:
Yaas!

MuriloDalRi:
@kevindew the requirements changed and I was asked to show the URL and also to strip the http(s) so I had to try a different approach as govspeak (and markdown) doesn't support that.

Could you have another look?

app/views/publish_document/published.html.erb
```diff
@@ -26,11 +26,16 @@
 
       <%= render "govuk_publishing_components/components/govspeak" do %>
         <%= govspeak_to_html t("publish_document.published.published_without_review.body_govspeak", title: @document.title) %>
+        <% stripped_url = DocumentUrl.new(@document).public_url.sub(/^https?\:\/\//,'') %>
+        <%= link_to(nil, stripped_url, class: "govuk-link govuk-link--no-visited-state") %>
 
-        <p> <%= t("publish_document.published.published_without_review.send_label") %> </p>
-
-        <%= link_to(nil, document_url(@document, utm_content: "2i-link"), class: "govuk-link") %>
+        <h2 id="what-to-do-next">What to do next</h2>
```
kevindew: Probably want to put this in a translation file, also not sure we need the id - I don't think we're anticipating people to deep link to ti.

app/views/publish_document/published.html.erb
```diff
@@ -26,11 +26,16 @@
 
       <%= render "govuk_publishing_components/components/govspeak" do %>
         <%= govspeak_to_html t("publish_document.published.published_without_review.body_govspeak", title: @document.title) %>
+        <% stripped_url = DocumentUrl.new(@document).public_url.sub(/^https?\:\/\//,'') %>
```
kevindew: Nice 👍 - Do we need to do the same thing with the URL above too for when it doesn't need a review? If so might want to have a helper method for this

app/views/publish_document/published.html.erb
MuriloDalRi: Cool, I'll do that

app/views/publish_document/published.html.erb
MuriloDalRi: I think you're right, we can do the same to both URls. Would you put that method in the DocumentUrl service or the view? Or maybe in a helper?

kevindew: Seems sufficiently generic that it can live in the application helper

MuriloDalRi: Done

MuriloDalRi:
@kevindew This should be ready for a final review now hopefully 🙏

PR #542, opened 2018-11-27
ADR on Javascript initialisation flickering

kevindew:
Trello: https://trello.com/c/H1TqpkiC/491-strategy-for-avoiding-flickering-of-content-as-js-initialises

Rendered View 👉 https://github.com/alphagov/content-publisher/blob/js-enabled-adr/docs/adr/0005-avoiding-initialisation-flickers.md

### DIFF
--- a/docs/adr/0005-avoiding-initialisation-flickers.md
+++ b/docs/adr/0005-avoiding-initialisation-flickers.md
@@ -1,0 +1,85 @@
+# 5. Avoiding initialisation flickering in HTML
+
+Date: 2018-11-27
+
+## Context
+
+A common problem with website initialisation is that there can be elements of the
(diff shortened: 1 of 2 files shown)


docs/adr/0005-avoiding-initialisation-flickers.md
```diff
@@ -0,0 +1,85 @@
...
+```
+document.body.className = ((document.body.className) ? document.body.className + ' js-enabled' : 'js-enabled');
+setTimeout(function() {
+  if (!window.jsInitialised) {
+    document.body.className += ' js-failed';
+  }
+}, 2000);
+```
+
+Which would allow a 2 second grace period before changing the body class so
```
DilwoarH: is 2 seconds long enough?

PR #544, opened 2018-11-27
Double quotes to single quotes

paola-roccuzzo:
Updated double quotes to single quotes in the flashes

### DIFF
--- a/config/locales/en/document_images/index.yml
+++ b/config/locales/en/document_images/index.yml
@@ -2,7 +2,7 @@ en:
   document_images:
     index:
       title: "Images for ‘%{title}’"
-      description_govspeak: Images can be jpg, png or gif files. [Full guidance on images](https://www.gov.uk/guidance/how-to-publish-on-gov-uk/images-and-videos)
+      description_govspeak: Images can be jpg, png or gif files. Full [guidance on images](https://www.gov.uk/guidance/how-to-publish-on-gov-uk/images-and-videos).
       upload_image: Upload an image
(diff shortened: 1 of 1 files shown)


config/locales/en/document_images/index.yml
```diff
@@ -16,9 +16,9 @@ en:
           invalid_format: Select a jpg, png or gif image file
           max_size: "Select an image file under %{max_size} in size"
           min_dimensions: "Select an image at least %{width} pixels wide by %{height} pixels high"
-        cropped: "Image “%{file}” has been cropped"
-        details_edited: "Image details for “%{file}” have been updated"
-        deleted: "Image “%{file}” has been deleted"
+        cropped: "Image ‘%{file}‘ has been cropped"
+        details_edited: "Image details for ‘%{file}‘ have been updated"
```
kevindew:
```suggestion
        details_edited: "Image details for ‘%{file}’ have been updated"
```

config/locales/en/document_images/index.yml
```diff
@@ -16,9 +16,9 @@ en:
...
+        deleted: "Image ‘%{file}‘ has been deleted"
```
kevindew:
```suggestion
        deleted: "Image ‘%{file}’ has been deleted"
```

paolaroccuzzo:
Ooops, should have thought about that - thanks for spotting it! The web interface wouldn't find them, so I had to do it by hand and I might have missed some as well...

PR #548, opened 2018-11-27
Update the state of retired and removed documents

leenagupte:
Trello: https://trello.com/c/2VIaBEOi

## Motivation

In PR #510 a task was added to allow us manually remove or retire documents in publishing-api. Now we need to reflect the state of those documents in Content Publisher.

## Terminology

An explanation of what retired and removed documents are can be found [here](https://github.com/alphagov/content-publisher/blob/master/docs/retiring-and-removing-documents.md#retiring-and-removing-documents). 

## What's changed
1. Documents can be filtered by removed or retired from the "States" drop-down
2. State of document is displayed as "retired" or "removed" on index page
3. Document history has an entry for retired documents that displays a note about why the document was retired. This is the same as the explanation that appears in the banner on the live site.
4. Document history has an entry for removed documents that displays a note about why the document was retired and an alternative path.
5. Internally sets a new `live_state` to reflect the status of a document once it has been published.

### Index page
screen shot 2018-11-30 at 11 27 44

screen shot 2018-11-30 at 11 28 47

### Document history
screen shot 2018-11-29 at 14 34 17 2

### DIFF
--- a/app/models/document.rb
+++ b/app/models/document.rb
@@ -27,8 +27,15 @@ class Document < ApplicationRecord
     reviewed
   ].freeze
 
+  LIVE_STATES = %w[
+    published
+    retired
+    removed
+  ].freeze
+
   validates_inclusion_of :publication_state, in: PUBLICATION_STATES
   validates_inclusion_of :review_state, in: REVIEW_STATES
+  validates_inclusion_of :live_state, in: LIVE_STATES, allow_nil: true, unless: :has_live_version_on_govuk
   validates_inclusion_of :update_type, in: %w[major minor], allow_nil: true
 
   def document_type_schema
--- a/app/models/remove.rb
+++ b/app/models/remove.rb
@@ -1,0 +1,5 @@
+# frozen_string_literal: true
+
+class Remove < ApplicationRecord
+  belongs_to :timeline_entry, class_name: "TimelineEntry", foreign_key: :timeline_entries_id, inverse_of: :remove
+end
--- a/app/models/remove.rb
+++ b/app/models/remove.rb
@@ -2,4 +2,9 @@
 
 class Remove < ApplicationRecord
   belongs_to :timeline_entry, class_name: "TimelineEntry", foreign_key: :timeline_entries_id, inverse_of: :remove
+  validates_inclusion_of :entry_type, in: %w[removed]
+
+  def entry_type
+    self.timeline_entry.entry_type
+  end
 end
--- a/app/services/document_unpublishing_service.rb
+++ b/app/services/document_unpublishing_service.rb
@@ -35,6 +35,7 @@ class DocumentUnpublishingService
     )
 
     delete_assets(document.images)
+    TimelineEntry.create!(document: document, entry_type: "removed")
   end
 
 private
--- a/app/services/document_unpublishing_service.rb
+++ b/app/services/document_unpublishing_service.rb
@@ -9,6 +9,7 @@ class DocumentUnpublishingService
       locale: document.locale,
     )
 
+    document.update!(live_state: "retired")
     TimelineEntry.create!(document: document, entry_type: "retired")
   end
 
--- a/app/services/document_unpublishing_service.rb
+++ b/app/services/document_unpublishing_service.rb
@@ -23,6 +23,7 @@ class DocumentUnpublishingService
     )
 
     delete_assets(document.images)
+    document.update!(live_state: "removed")
     TimelineEntry.create!(document: document, entry_type: "removed")
   end
 
--- a/app/services/document_unpublishing_service.rb
+++ b/app/services/document_unpublishing_service.rb
@@ -19,7 +19,12 @@ class DocumentUnpublishingService
   def remove(document, explanatory_note: nil, alternative_path: nil)
     Document.transaction do
       document.update!(live_state: "removed")
-      TimelineEntry.create!(document: document, entry_type: "removed")
+      timeline_entry = TimelineEntry.create!(document: document, entry_type: "removed")
+      Remove.create!(
+        timeline_entry: timeline_entry,
+        explanatory_note: explanatory_note,
+        alternative_path: alternative_path,
+      )
 
       GdsApi.publishing_api_v2.unpublish(
         document.content_id,
--- a/app/services/document_unpublishing_service.rb
+++ b/app/services/document_unpublishing_service.rb
@@ -33,17 +33,20 @@ class DocumentUnpublishingService
   end
 
   def remove_and_redirect(document, redirect_path, explanatory_note: nil)
-    GdsApi.publishing_api_v2.unpublish(
-      document.content_id,
-      type: "redirect",
-      explanation: explanatory_note,
-      alternative_path: redirect_path,
-      locale: document.locale,
-    )
+    Document.transaction do
+      document.update!(live_state: "removed")
+      TimelineEntry.create!(document: document, entry_type: "removed")
+
+      GdsApi.publishing_api_v2.unpublish(
+        document.content_id,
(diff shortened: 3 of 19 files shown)


app/models/document.rb
```diff
@@ -20,6 +20,12 @@ class Document < ApplicationRecord
     sent_to_live
     error_sending_to_live
     error_deleting_draft
+    retiring
+    retired
+    error_retiring
+    removing
+    removed
+    error_removing
```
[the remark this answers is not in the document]
leenagupte: @kevindew will kill off the intermediary states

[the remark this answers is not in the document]
leenagupte: > The retire and removing states probably belong in a different grouping to publication_state (I know it's super confusing). As whether we publish, remove or retire we're always sending to live and can always have error_sending_to_live state.

I'd suggest maybe we have a live_state which has values of published, retired and removed. We can work out whether it had an error updating based on error_sending_to_live.

So the publication_state wouldn't change? There would just be extra `live_state` that in most cases would be set to `live`?

app/services/document_unpublishing_service.rb
```diff
@@ -1,16 +1,24 @@
 # frozen_string_literal: true
 
 class DocumentUnpublishingService
-  def retire(document, explanatory_note)
+  def retire(document, explanatory_note, user)
```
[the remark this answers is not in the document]
leenagupte: TimelineEntry does have the ability to set an "Unknown user" but I don't know if this every gets set at the moment.

docs/retiring-and-removing-documents.md
```diff
@@ -41,7 +43,7 @@ Optional parameters:
 - NEW_PATH
 
 ```
-rake unpublish:remove_document['a-content-id']
+rake unpublish:remove_document['a-content-id'] USERNAME='firstname lastname'
```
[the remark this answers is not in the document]
leenagupte: I'll see if I can remove the need for a user at all.

leenagupte:
@kevindew Thanks for your comments. We probably should talk earlier, but on the bright-side, I have a better understanding of how the app sets state and history than I did a day ago, so it wasn't a waste.

app/services/document_unpublishing_service.rb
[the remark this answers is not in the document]
leenagupte: I've made the association with `user` optional. Will fix-up and push.

docs/retiring-and-removing-documents.md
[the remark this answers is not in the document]
leenagupte: I've removed the need for a user to passed in, so these changes have been removed. Will fix-up and push.

app/models/document.rb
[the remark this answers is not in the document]
leenagupte: @kevindew I've updated the state of the document to include a `live_state`, but I'm not sure how `error_sending_to_live` fits into that.

app/models/document.rb
```diff
@@ -27,8 +27,15 @@ class Document < ApplicationRecord
...
 
+  LIVE_STATES = %w[
+    published
+    retired
+    removed
+  ].freeze
+
   validates_inclusion_of :publication_state, in: PUBLICATION_STATES
   validates_inclusion_of :review_state, in: REVIEW_STATES
+  validates_inclusion_of :live_state, in: LIVE_STATES, allow_nil: true
```
kevindew: Might want to only allow_nil when it's not sent_to_live

leenagupte: Updated and fixed-up

app/services/user_facing_state.rb
```diff
@@ -37,6 +41,10 @@ def to_s
       "draft"
     elsif review_state == "published_without_review"
       "published_but_needs_2i"
+    elsif publication_state == "sent_to_live" && live_state == "retired"
```
benthorner: This might want to go above the previous clause, as I think if you retire/remove an unreviewed document it'll still come out as 'published_but_needs_2i'.

app/views/documents/show/_history_tab.html.erb
```diff
@@ -11,6 +11,17 @@
       <%= t "documents.history.entry_types.#{entry.entry_type}" %>
     </h4>
 
+    <% if entry.retire.present? %>
+      <%= entry.retire.explanatory_note %>
+    <% elsif entry.remove.present? %>
```
benthorner: Small thing, but is the else needed here? I'm reading it as "either we'll have a remove, or we'll have a retire", but the .present? check then makes that redundant.

spec/factories/remove_factory.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+
+FactoryBot.define do
+  factory :remove do
+    explanatory_note { "The document is out of date" }
```
benthorner: I think we try to use random stuff where possible, just so the test data isn't static - we use SecureRandom a lot around the factories. Do you think it would be OK with that?

spec/services/document_unpublishing_service_spec.rb
```diff
@@ -29,11 +33,30 @@
...
+
+      expect(document.timeline_entries.first.entry_type).to eq("retired")
+      expect(document.timeline_entries.first.retire.explanatory_note).to eq(explanatory_note)
+    end
+
+    it "keeps track of the live state" do
+      stub_publishing_api_unpublish(document.content_id, body: { type: "withdrawal", explanation: explanatory_note, locale: document.locale })
+      DocumentUnpublishingService.new.retire(document, explanatory_note)
+
+      expect(document).to have_received(:update!).with(live_state: "retired")
```
benthorner: Could we check the document directly here? I think we stubbed update! elsewhere only when we needed to inspect the state halfway through the method - should be OK here :-).

benthorner: (You can always use .reload to be sure we actually saved something.)

app/models/remove.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+class Remove < ApplicationRecord
```
benthorner: I think it'd be clearer to have a more noun-y name here, as that's how we do most of the models. Had a chat with @kevindew and we came up with 'Retirement' and 'Removal'. What do you think?

app/views/documents/show/_history_tab.html.erb
leenagupte: You could have neither.

spec/factories/remove_factory.rb
leenagupte: Yep. The note can be anything.

spec/services/document_unpublishing_service_spec.rb
leenagupte: I think I was following the same pattern in another test file for consistency.  But, yeah, I can change to actually check the value of `live_state` on the document.

app/services/user_facing_state.rb
leenagupte: Done. Will fix-up.

spec/factories/remove_factory.rb
leenagupte: Have updated. Will fix-up.

app/models/remove.rb
leenagupte: I've renamed the models in a new branch: https://github.com/alphagov/content-publisher/pull/575

leenagupte:
Replaced with: https://github.com/alphagov/content-publisher/pull/575

leenagupte:
@kevindew @benthorner Renaming the models and rebasing got a bit hairy so I moved to a new PR: #575 

Can you re-review when you get a chance? Cheers.

PR #551, opened 2018-11-28
Disables grammarly plugin for markdown editor

DilwoarH:
Grammarly interferes with markdown editor making it impossible to insert and preview

How to replicate:

- Install grammarly
- Go to edit markdown field
- Experience grim position absolute issues

https://trello.com/c/1pgYWuWB/447-enable-markdown-editor-to-work-with-grammarly


alex-ju:
While this is a quick fix that we may want to introduce, the AC on this card was to make it work with Grammarly.

PR #571, opened 2018-11-30
Update autocomplete component to support custom template

alex-ju:
This PR:
- adds a new option for the autocomplete to support custom template
- add the custom template version of autocomplete to contacts

👉 [Preview component](https://content-publisher-revie-pr-571.herokuapp.com/component-guide/autocomplete/autocomplete_with_custom_template)

[Trello card](https://trello.com/c/05U8l7DF)

### DIFF
--- a/app/assets/javascripts/components/autocomplete.js
+++ b/app/assets/javascripts/components/autocomplete.js
@@ -15,7 +15,7 @@ if ($autocompletes) {
     new accessibleAutocomplete.enhanceSelectElement({
       selectElement: $el,
       minLength: 3,
-      showNoOptionsFound: false,
+      defaultValue: '',
       customAttributes: customAttributes
     })
   })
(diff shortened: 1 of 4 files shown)


app/assets/javascripts/components/autocomplete.js
```diff
@@ -40,4 +40,37 @@ if ($multiselects) {
...
+    defaultValue: '',
+    templates: {
+      inputValue: function (result) {
+        if (result) {
+          return result.split(' - ')[0]
+        }
+      },
+      suggestion: function (result) {
+        if (result) {
+          var resultItems = result.split(' - ')
```
benthorner: Are we sure this character sequence won't occur in what should be the first entry?

app/assets/javascripts/components/autocomplete.js
```diff
@@ -15,7 +15,7 @@ if ($autocompletes) {
     new accessibleAutocomplete.enhanceSelectElement({
       selectElement: $el,
       minLength: 3,
-      showNoOptionsFound: false,
+      defaultValue: '',
```
benthorner: Not sure why we're changing this?

app/views/components/docs/autocomplete.yml
```diff
@@ -80,3 +94,24 @@ examples:
       selected_options:
           - fr
           - de
+  autocomplete_with_custom_template:
```
benthorner: This one doesn't seem to work in preview :-(

app/assets/javascripts/components/autocomplete.js
alex-ju: Yes, we tested and for now it's safe to use it. It's only present in one organisation which doesn't have any contacts.

app/views/components/docs/autocomplete.yml
alex-ju: Works here https://content-publisher-revie-pr-571.herokuapp.com/component-guide/autocomplete/autocomplete_with_custom_template if you type `Uni` for example

app/assets/javascripts/components/autocomplete.js
alex-ju: We want to show if there're no results found and we want to set the default for this field to an empty option. These changed make the preview of the components work as expected - which didn't until now.

alex-ju:
@benthorner I'm happy to close this in favour of other approaches. Did the best I could given the way the data is presented in the HTML.

benthorner:
@alex-ju sorry - it's a tricky problem. That empty organisation could potentially have contacts in future, so this doesn't seem like a net improvement as it would introduce a waiting bug. We do currently have the same presentation style for the ministerial roles, which seemed to be OK in user testing. Perhaps, as you say, we could leave it as it is for the time being.

app/views/components/docs/autocomplete.yml
benthorner: Oops - me being dim.

PR #573, opened 2018-12-03
Remove support for animated gif image uploads

1pretz1:
Currently animated gifs don't upload / resize correctly, this change
removes the support for these uploads and shows a invalid_format error
if a user tries to upload this image type.

This is worked out by checking how many different frames there
are at the image path, animated gifs have multiple frames per image.
Static images have one frame is my understanding.

Trello:
https://trello.com/c/82HzbseV/514-handle-animated-gif-uploads


1pretz1:
Doh, is it bad that i thought all gifs were animated! 😆I blame my youth

app/services/image_upload_requirements.rb
```diff
@@ -19,7 +21,7 @@ def errors
       return [I18n.t!("document_images.index.flashes.upload_requirements.no_file_selected")]
     end
 
-    if SUPPORTED_FORMATS.exclude?(Marcel::MimeType.for(file))
+    if SUPPORTED_FORMATS.exclude?(Marcel::MimeType.for(file)) || MiniMagick::Image.new(file.tempfile.path).frames.count > 1
```
benthorner: Maybe make this a separate if block, as it's quite specific/long?

1pretz1: Yep it is a bit long, i just thought that was better than repeating code with another if when both blocks would be concerning invalid formats?

1pretz1: I could move it into a method with a descriptive name which would make the line shorter and clearer

PR #574, opened 2018-12-03
Use npm packages

benthorner:
https://trello.com/c/Ti3wYWAa/506-can-we-use-npm-rather-than-vendoring
https://trello.com/c/wcpxNtrO/507-set-up-a-self-hosted-polyfill-processs

This is an attempt to remove our use of vendored and externally loaded JavaScripts.


benthorner:
@alex-ju I've already tried it with IE11 and Edge and I'm going to try on Firefox. Any other supported browsers I should know about ;-).

alex-ju:
@benthorner the ones you mentioned and Safari will match the service manual

benthorner:
@alex-ju checked that now too (the edit page and the topics page). I think this is ready to merge, but I'll just have a chat with @kevindew first.

package.json
```diff
@@ -16,5 +16,15 @@
   },
   "devDependencies": {
     "standard": "^11.0.1"
+  },
+  "dependencies": {
+    "@webcomponents/custom-elements": "^1.2.1",
+    "abortcontroller-polyfill": "^1.2.1",
+    "core-js-bundle": "^3.0.0-alpha.1",
+    "cropperjs": "^1.4.3",
+    "es5-polyfill": "0.0.6",
```
kevindew: this version is missing a ^, is that intentional?

PR #580, opened 2018-12-04
Disable mobile preview on mobile and use the default one instead

alex-ju:
This PR updates the page preview component to drop the mobile frame and only present the default preview and the Google search snippet.

### Before and After
screen shot 2018-12-04 at 16 17 44

### Before
screen shot 2018-12-04 at 16 21 55

### After
screen shot 2018-12-04 at 16 19 04

[Trello card](https://trello.com/c/qMVliSBq)

### DIFF
--- a/app/assets/stylesheets/components/_page-preview.scss
+++ b/app/assets/stylesheets/components/_page-preview.scss
@@ -48,8 +48,13 @@ $mobile-border-bottom-width: 68px;
 }
 
 .app-c-preview__google {
-  max-width: 600px;
-  -webkit-font-smoothing: auto;
+  display: none;
+
+  @include govuk-media-query($from: tablet) {
+    display: block;
+    max-width: 600px;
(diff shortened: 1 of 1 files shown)


benthorner:
That looks much nicer :-). It's a bit weird having all three previews with no separator between them - especially now we've lost the contents list at the top. Could we just have the mobile preview on mobile? Or maybe show the search engine snippet at the top?

alex-ju:
@benthorner removed the google snippet as per card

benthorner:
@alex-ju thanks :-). When I was taking a look, I notice we still show the tab sections even though the inner content is hidden. They do seem to take up some space on the page - would it be better to hide the entire section instead?

alex-ju:
@benthorner the sections are still present as they're part of markup generated by the tabs component, but they're not taking space on the page though as they're empty containers and only have a margin property.

PR #581, opened 2018-12-04
Remove redundant markdown-toolbar component

benthorner:
https://trello.com/c/Ti3wYWAa/506-use-npm-rather-than-vendoring

This was made redundant as this component actually referred to a
vendored JavaScript file, which is now in vendor/@alphagov.


benthorner:
Superseded by https://github.com/alphagov/content-publisher/pull/582#pullrequestreview-181380467.

PR #589, opened 2018-12-05
Fix javascript tests

benthorner:
https://trello.com/c/Xil1z0No/523-jasmine-tests-dont-run-in-ci

Awww yeah!


Gemfile
```diff
@@ -39,7 +39,8 @@ group :development, :test do
   gem "govuk-lint", "~> 3"
   gem "govuk_schemas", "~> 3.2"
   gem "govuk_test", "~> 0.3"
-  gem "jasmine", "~> 2.4"
+  gem "jasmine", "~> 3.3"
```
kevindew: woo - no longer in the 2.99 world

spec/javascripts/support/jasmine_helper.rb
```diff
@@ -1,16 +1,13 @@
...
-# end
-#
-# Example: prevent PhantomJS auto install, uses PhantomJS already on your path.
+require "jasmine/runners/selenium"
+
 Jasmine.configure do |config|
-  config.prevent_phantom_js_auto_install = true
+  config.runner = lambda { |formatter, jasmine_server_url|
+    options = Selenium::WebDriver::Chrome::Options.new
+    options.headless!
```
kevindew: oh neat that we can use headless

benthorner:
I'm still sceptical with Jasmine - the technique I've used is undocumented and involved reading a lot of code; the phrase 'shoestring' still applies ;-). That said, it would be difficult to avoid a Ruby approach because of the necessary integration with Rails' asset pipeline. At least this avoids a long discussion about switching to something else, and I'll post this technique so that others can benefit :-).

kevindew:
@benthorner oh yeah darn I'd forgot about asset pipeline :( that would make anything else a total nightmare to set up. Guess we're a bit caught in a legacy trap with that.

PR #595, opened 2018-12-06
Only show edition metadata for subsequent editions

benthorner:
https://trello.com/c/hj7RXU4S/532-content-summary-change-note-meta-data-and-fix-mobile-styles

This also fixes the metadata on the document summary page so that
we don't show the change note for minor changes, so that it's consistent
with the way we handle this on the edit page.


benthorner:
Closio...

PR #597, opened 2018-12-06
Fix error alert

benthorner:
https://sentry.io/govuk/app-content-publisher/issues/761072360/?environment=integration-blue-aws


benthorner:
Looks like Seal won't accept this one's closed :-(.

PR #598, opened 2018-12-07
Push submited radio inputs into GA's dataLayer

erino:
If a form is given the data-attribute gtm then `GTMFormListener` will listen for the form to be submitted and push an event to the `dataLayer` for each checked radio input. Events are in the form

```json
  { "gtm-form-attribute": { "radio-name": "radio-value" } }
```

This tracking is added to the new document forms and also the confirm that a document has been reviewed form.


app/assets/javascripts/application.js
```diff
@@ -18,6 +18,7 @@
 //= require components/autocomplete.js
 //= require components/image-cropper.js
 //= require components/input-length-suggester.js
+//= require components/gtm-form-listener.js
```
kevindew: I'd suggest this should be in the modules directory of JS since it isn't actually a component

app/assets/javascripts/components/gtm-form-listener.js
```diff
@@ -0,0 +1,35 @@
+function GTMFormListener ($form, $dataLayer) {
```
kevindew: The dollar convention is used for HTML elements so shouldn't be on dataLayer as far as I'm aware.

(it's not my favourite convention)

app/assets/javascripts/components/gtm-form-listener.js
```diff
@@ -0,0 +1,35 @@
...
+  if (!$form) {
+    return
+  }
+
+  $form.addEventListener('submit', function (e) {
+    this.handleSubmit(e)
+  }.bind(this), false)
+}
+
+var $gtmFormListener = document.querySelector('form[data-gtm]')
```
kevindew: should this be a querySelectorAll so that we can support multiple forms on a page?

app/assets/javascripts/components/gtm-form-listener.js
```diff
@@ -0,0 +1,35 @@
+function GTMFormListener ($form, $dataLayer) {
+  this.$form = $form
+  this.dataLayer = $dataLayer || window.dataLayer
+}
+
+GTMFormListener.prototype.handleSubmit = function (event) {
+  var eventName = this.$form.dataset.gtm
+  var message = {}
```
kevindew: Does this need to be set to an empty object if we do that inside the element loop?

app/assets/javascripts/components/gtm-form-listener.js
```diff
@@ -0,0 +1,35 @@
...
+
+  var inputElements = this.$form.querySelectorAll('input:checked')
+  inputElements.forEach(function (element) {
+    if (element.value.length > 0) {
+      message = {}
+      message[eventName] = {}
+      message[eventName][element.name] = element.value
```
kevindew: Is this code ok for a set of checkboxes with the same name too? e.g. "field[]=foo&field[]=bar"

app/assets/javascripts/components/gtm-form-listener.js
```diff
@@ -0,0 +1,35 @@
...
+      this.dataLayer.push(message)
+    }
+  }, this)
+}
+
+GTMFormListener.prototype.init = function () {
+  var $form = this.$form
```
kevindew: I think we may as well just use this.$form in the two places this is used in this method?

app/views/new_document/choose_document_type.html.erb
```diff
@@ -3,7 +3,7 @@
 
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
-    <%= form_tag create_document_path do %>
+    <%= form_tag create_document_path, data: { gtm: "new_document" } do %>
```
kevindew: minor nitpick, but we've gone with kebab case on all the previous gtm attributes

spec/javascripts/components/gtm-form-listener-spec.js
```diff
@@ -0,0 +1,33 @@
+var $ = window.jQuery
+
+describe('GTM dataLayer messages for radio button submissions', function () {
```
kevindew: Woo for adding a test!

spec/javascripts/components/gtm-form-listener-spec.js
```diff
@@ -0,0 +1,33 @@
...
+  'use strict'
+
+  var form
+  var dataLayer
+  var gtmFormListener
+
+  beforeEach(function () {
+    form = $(`
```
kevindew: I think @benthorner did some work stripping out jQuery from all our tests so if we could use DOM methods that'd be consistent

app/assets/javascripts/application.js
erino: Fixed in b955ad0fc27b661e8da6b97adbf4a1e88e58a07c

app/assets/javascripts/components/gtm-form-listener.js
erino: Fixed in bb39bebd9eb99562cc209abf2441f0513e3c18ea

app/assets/javascripts/components/gtm-form-listener.js
erino: Fixed in 80e2f7107c683c971a0a6273d97f089659ed63c1

app/views/new_document/choose_document_type.html.erb
erino: Fixed in 63c57810ee3b5f7e31cafb73611c5e5ea256067f

spec/javascripts/components/gtm-form-listener-spec.js
erino: 👍 I mostly copied another spec...

spec/javascripts/components/gtm-form-listener-spec.js
erino: Fixed in 8cfc33a092bf9321f17a35c744e9120106ec3965

app/assets/javascripts/components/gtm-form-listener.js
erino: No, fixed in 900499c48755f2913ce8bf958f3ae6e520c9b905

app/assets/javascripts/components/gtm-form-listener.js
erino: ah, this PR is mostly cargo culting 😟

erino: Fixed in 2b3f06b78e86bdb01b2282793ac191ab6803ed9a

app/assets/javascripts/components/gtm-form-listener.js
erino: Yes and added a spec in 569a19254b15d163e49cc45c2b0ad2e0087c89ec

app/views/publish_document/confirmation.html.erb
```diff
@@ -3,7 +3,7 @@
 
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
-    <%= form_tag publish_document_path(@document) do %>
+    <%= form_tag publish_document_path(@document), data: { gtm: "publish_document" } do %>
```
kevindew: We've got another snake case here

app/views/new_document/choose_supertype.html.erb
```diff
@@ -3,7 +3,7 @@
 
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
-    <%= form_tag(choose_document_type_path, method: :get, enforce_utf8: false) do %>
+    <%= form_tag(choose_document_type_path, method: :get, enforce_utf8: false, data: { gtm: "new-document" } ) do %>
```
kevindew: Is this meant to be the same tag as choosing document type? and should they both reflect document type / supertype rather than new-document?

app/views/publish_document/confirmation.html.erb
erino: Fixed in 91c1a216

app/views/new_document/choose_supertype.html.erb
erino: Yes, I went through it with Andy

PR #601, opened 2018-12-07
Add more Google Tag Manager data attributes

kevindew:
https://trello.com/c/yRnJfrDN/487-deployment-of-remaining-data-attributes-for-consistent-event-tracking-in-gtm

I paired with Andy to work out these bits of tracking so he doesn't have to rely on CSS selectors.

This does need https://github.com/alphagov/govuk_publishing_components/pull/662 to fully work as the error alert and error summary fields don't support data attributes yet, but to be honest it's probably not much of a concern if we merge this before.

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -57,6 +57,7 @@
     <% if flash["alert_with_description"] %>
       <%= render "govuk_publishing_components/components/error_alert", {
         message: flash["alert_with_description"].fetch("title"),
+        data_attributes: { gtm: "global-alert" },
         description: capture do
           render "govuk_publishing_components/components/govspeak" do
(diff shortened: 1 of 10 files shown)


app/views/layouts/application.html.erb
```diff
@@ -4,11 +4,12 @@
     <meta name="sentry-current-env" content="<%= ENV["SENTRY_CURRENT_ENV"] %>">
   <% end %>
 
+  <script>
```
benthorner: How come this isn't in the conditional anymore?

kevindew: Sorry that should have been made clear in a commit. I'll sort that.

But it's because other pages can add data to the dataLayer and if it's not always there then you can't see whether it's doing what you expect until you are integration.

PR #603, opened 2018-12-10
[WIP] Replace attr_accessors by inerhiting ReadonlyModel

benthorner:
We have a number of classes, such as DocumentType and Topic, that
act as interfaces to static (readonly) data. Previously, some of these
classes used attr_accessors so they could include ActiveModel::Model,
which helps to DRY-up the way their instances are constructed.

This replaces the use of attr_accessor with attr_reader by changing each
model to extend a slightly modified version of ActiveModel::Model that
also makes the update_attributes and attributes= methods private.

### DIFF
--- a/app/models/concerns/readonly_model.rb
+++ b/app/models/concerns/readonly_model.rb
@@ -1,0 +1,17 @@
+# frozen_string_literal: true
+
+class ReadonlyModel
+  include ActiveModel::Model
+
+private
+
+  def assign_attributes(new_attributes)
+    new_attributes.each do |key, value|
+      instance_variable_set(:"@#{key}", value)
+    end
+  end
+
+  def attributes=(args)
+    assign_attributes(args)
+  end
+end
--- a/app/models/document_topics.rb
+++ b/app/models/document_topics.rb
@@ -1,9 +1,7 @@
 # frozen_string_literal: true
 
-class DocumentTopics
-  include ActiveModel::Model
-
-  attr_accessor :document, :version, :topics, :index
+class DocumentTopics < ReadonlyModel
+  attr_reader :document, :version, :topics, :index
 
   def self.find_by_document(document, index)
     publishing_api = GdsApi.publishing_api_v2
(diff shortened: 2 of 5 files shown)


benthorner:
I think you're right - I rationalised it myself, but it's probably not for everyone. There are two things I'd like to like to achieve: a DRY constructor and some kind of minimal model convention. But perhaps the only real convention here is that we have a DRY constructor. Perhaps something like...

```
module InitializeWithHash
  def initialize(hash = {})                                                                                                                                                                                                                                                   
    hash.each do |key, value|
      instance_variable_set(:"@#{key}", value)
    end 
  end 
end
```

I'd really like to avoid doing "@a = a, @b = b, @c = c, etc.' if we can.

kevindew:
> I'd really like to avoid doing "@A = a, @b = b, @c = c, etc.' if we can.

Yeah that can be annoyingly verbose in Ruby - I remember being surprised that Ruby didn't do something like Scala where their set directly from the constructor e.g.

```
class Point(val x: Double = 0, val y: Double = 0)

val point1 = new Point(y = 1)
```

Which would make a load of the short examples here so much simpler 😞

People have tried to do similar things to Ruby e.g. https://github.com/mbj/anima but I think that'd be rather overkill for us to consider.

> There are two things I'd like to like to achieve: a DRY constructor 

Sure, I could see why that'd benefit some of the short classes e.g: https://github.com/alphagov/content-publisher/blob/4b373fdf28f952aa14c0a39b552b9997f2727a65/app/services/document_type.rb#L42-L44

Some of them that have no methods could just be frozen OpenStructs too.

I think your suggestion of InitializeWithHash seems good for those sort of examples, that can keep them super concise. I don't think it should be used everywhere though just when it's a simple case that can reduce verbosity. 

> and some kind of minimal model convention

I don't think you necessarily have to strive for that - and if you have a nice simple module that is mixed in for the constructor then I don't think that needs to be coupled to the concept of a model. You do risk potentially indirection - where I've found the change to document type to now be confusing to fit the ReadOnlyModel: https://github.com/alphagov/content-publisher/pull/603/files#diff-0bf513ab215322410adda90965b8591c

benthorner:
Superseded by https://github.com/alphagov/content-publisher/pull/622.

PR #604, opened 2018-12-10
Fix `TypeError`s on autocomplete

alex-ju:
Update autocomplete to the latest code which (hopefully 🤞- unable to test locally and the component guide doesn't load the js for the autocomplete) fixes the `TypeError` on `this.state.query` when `undefined` as well as other few states where TypeError can occur.

It seems to be fixing things when testing the autocomplete in isolation but **if someone is able to help testing the component in `content-publisher` would be great**.

Change in the [accessible-autocomplete repo](https://github.com/alphagov/accessible-autocomplete/commit/2b15dcbbf34281bfbf3648c82582e2a768bc074e#diff-42909ce1eee782a4254962bd5bfe2a04R234).

[Trello card](https://trello.com/c/BTJwV1S1)


benthorner:
@alex-ju do we have a known scenario where this manifests itself?

alex-ju:
@benthorner: nope. Did the fix based on the Sentry report and was able to replicate one of them in the component's repo examples. I was hoping to replicate the other in content-publisher.

kevindew:
I did some work trying to work out how it happens and didn't really crack it. 

I think it's one or all of: https://github.com/alphagov/accessible-autocomplete/blob/c4b65244702928b71606dad0009487630f1f1494/src/autocomplete.js#L206, https://github.com/alphagov/accessible-autocomplete/blob/c4b65244702928b71606dad0009487630f1f1494/src/autocomplete.js#L195, and https://github.com/alphagov/accessible-autocomplete/blob/c4b65244702928b71606dad0009487630f1f1494/src/autocomplete.js#L280

where there are calls to `options[selected]` as presumably there's are cases where selected does not equal the value of an option (maybe initially?) but I wasn't able to confirm it.

alex-ju:
Thanks a lot @kevindew. I'll add more examples on the component repo with some edge cases for the multiselect to cover these and the start covering them by writing appropriate tests. It's probably a better approach than trying to fix specifics and will make us be more confident on the component itself.

PR #607, opened 2018-12-11
Add download link to images

1pretz1:
Clicking the download link will crop the image to 960x640 and download
it to the users computer. This is useful for whitehall users as images
uploaded here can easily be cropped and used in whitehall publisher.

Due to a discussion with Sonia, the link order has also been amended
which moves the `delete image` link to the end of the list.

Trello:
https://trello.com/c/w4aI961C/529-provide-a-960x640-download-link-for-images

## Before
screen shot 2018-12-11 at 12 02 14

## After
screen shot 2018-12-11 at 11 47 35

### DIFF
--- a/app/controllers/document_images_controller.rb
+++ b/app/controllers/document_images_controller.rb
@@ -132,6 +132,17 @@ class DocumentImagesController < ApplicationController
     end
   end
 
+  def download
+    image = Image.find(params[:image_id])
+    variant = image.crop_variant("960x640!").processed
+
+    send_data(
+      image.blob.service.download(variant.key),
+      filename: image.filename,
+      type: image.content_type,
+    )
+  end
+
 private
 
   def update_params
--- a/app/views/document_images/_image_list.html.erb
+++ b/app/views/document_images/_image_list.html.erb
@@ -9,19 +9,20 @@
     <% actions = [
       link_to("Edit crop", crop_document_image_path(@document, image), class: "govuk-link"),
       link_to("Edit details", edit_document_image_path(@document, image), class: "govuk-link"),
+      link_to('Download 960x640 image', download_image_path(@document, image), class: "govuk-link")
     ] %>
+
+    <% actions << capture do %>
+      <%= form_tag choose_document_lead_image_path(@document, image), class: "app-inline-block" do %>
+        <button class="govuk-link">Select as lead image</button>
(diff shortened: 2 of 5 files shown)


app/controllers/document_images_controller.rb
```diff
@@ -132,6 +132,17 @@ def destroy
     end
   end
 
+  def download
+    image = Image.find(params[:image_id])
+    variant = image.crop_variant("960x640!").processed
```
benthorner: I think you can just use `image.crop_variant` here, as the width/height are defaulted in the model :-).

1pretz1: There is an added explanation mark `960x640!` here which the default doesn't have. This forces the image to crop at the exact dimensions specified, without this the image can be a few pixels out due to ImageMagik trying to not distort the image on resizing.

I'm now curious as to whether we want to set the default to a force crop e.g add a explanation mark?

benthorner: Ooh, nice spot! Maybe it's worth adding a specific method on the model to explain this?

1pretz1: Hmmm yep, when i uploaded this image and then downloaded it without the bang it was 1 pixel out at 960x639 which obviously isn't suitable as whitehall errors that the image isn't the correct size. 
tree_png3498

I've had a brief look to find the discussion i found about this on the web but can't seem to find it -.-

kevindew: Ah yeah sure, that makes sense that occasionally it could be a pixel off. Hmm I wonder whether we should drop a pixel normally or just distort ever so slightly :-/ but this step seems necessary for Whitehall so this is good.

1pretz1: I don't think distorting the image by a pixel is going to affect the visual appearance of the image to the naked eye, so if it's required that the images are the exact dimensions they are cropped with then i think setting it to the default is okay?

However, if this needs further discussion i'll just use the bang with the download crop for now

kevindew: Yeah I'm not sure beyond Whitehall if these need to be the exact dimensions or not. I think if we roll with it being exact for this Whitehall upload but imagemagick defaults elsewhere, and then we'll see whether that causes any pain.

PR #608, opened 2018-12-11
[DO NOT MERGE] Add document withdrawal workflow

emmabeynon:
I've marked this as DNM as it's being changed for versioning work - Kevin

For https://trello.com/c/GGxgDv3W/533-add-form-to-let-users-retire-a-document

This adds a frontend interface so that users with managing editor permissions can withdraw a document.

When I started on this story this functionality was called "retire"; we've now decided to use "withdraw" instead.  While the public-facing elements of the application with only refer to "withdrawing", a lot of the internals still refer to "retiring".  A separate PR will deal with updating internal references to make it clear and consistent.

1) All users see a button to withdraw a document:
screen shot 2018-12-21 at 11 21 41

2) Users without managing editor permissions will see guidance about asking their managing editor to withdraw a document if they click the 'Withdraw' button:
screen shot 2018-12-21 at 11 28 42

3) Only users with managing editor permissions will see a form to withdraw a document, and will see guidance for the public explanation:
public_explanation

4) Users will see an error if they do not provide a public explanation:
screen shot 2018-12-21 at 11 23 24

5) Users will see an error if they provide a public explanation that is too long:
screen shot 2018-12-21 at 11 24 30

6) Users will see an error if the Publishing API is down when they attempt to withdraw a document:
screen shot 2018-12-21 at 11 26 30

7) All users will see a withdrawal notice with the date the document was withdrawn, the public explanation and a link to the change the public explanation.  They will also see a withdrawn status and the name of the person that withdrew the document.  
withdrawn_summary_screen

8) When users click "Undo withdraw" they will see a page explaining that the feature hasn't been built yet.
screen shot 2018-12-21 at 11 27 14

Outstanding items:
- [ ] @paola-roccuzzo to provide a character limit for the explanatory note

### DIFF
--- a/app/models/document.rb
+++ b/app/models/document.rb
@@ -81,4 +81,8 @@ class Document < ApplicationRecord
   def withdrawn_by
     @withdrawn_by ||= self.timeline_entries.where(entry_type: "retired").order(:created_at).last.user
   end
+
+  def withdrawn?
+    user_facing_state == "retired"
+  end
 end
--- a/app/models/document.rb
+++ b/app/models/document.rb
@@ -73,4 +73,8 @@ class Document < ApplicationRecord
     @document_topics_index ||= TopicIndexService.new
     DocumentTopics.find_by_document(self, @document_topics_index)
   end
+
+  def retirement
+    self.timeline_entries.where(entry_type: "retired").order(:created_at).last&.retirement
+  end
 end
--- a/app/models/document.rb
+++ b/app/models/document.rb
@@ -77,4 +77,8 @@ class Document < ApplicationRecord
   def retirement
     self.timeline_entries.where(entry_type: "retired").order(:created_at).last&.retirement
   end
+
+  def withdrawn_by
+    @withdrawn_by ||= self.timeline_entries.where(entry_type: "retired").order(:created_at).last.user
+  end
 end
--- a/app/models/document.rb
+++ b/app/models/document.rb
@@ -6,6 +6,7 @@ class Document < ApplicationRecord
   has_many :images, dependent: :destroy
   has_many :timeline_entries, dependent: :destroy
   has_many :internal_notes, dependent: :destroy
+  has_many :retirements, dependent: :destroy
 
   belongs_to :lead_image, class_name: "Image", optional: true, foreign_key: :lead_image_id, inverse_of: :document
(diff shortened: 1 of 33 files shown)


paola-roccuzzo:
@emmabeynon I think that this part of the screen is getting really crowded and hard to use - and for a use case that's not very frequent. I'd like to review this with @soniaturcotte and check if it's worth to have one CTA (something like "Change publishing status") on the Summary screen, and then move the other CTAs to the next screen, so we can also support users with guidance that we can't really provide at this stage. I'll sketch something out and I will share it with you, Sonia and @benhazell

app/models/document.rb
```diff
@@ -71,4 +71,8 @@ def document_topics
     @document_topics_index ||= TopicIndexService.new
     DocumentTopics.find_by_document(self, @document_topics_index)
   end
+
+  def retirement
+    self.timeline_entries.where(entry_type: "retired").order(:created_at).last&.retirement
```
kevindew: I think we should find a way to associate with a retirement without using timeline entry.

Timeline entry's role should be just around the timeline and the document should be able to know it's own state without consulting it.

app/views/retire_document/create.html.erb
```diff
@@ -0,0 +1,35 @@
+<% content_for :back_link, render_back_link(href: document_path(@document)) %>
+<% content_for :title, t("retire_document.retire.title", title: @document.title) %>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <%= form_tag(new_retirement_path(@document)) do %>
+      <%= render "govuk_publishing_components/components/textarea", {
```
kevindew: Not sure if it was clear from the card but we support (it probably wasn't) but we support govspeak as an input so we may need to ponder some sort of cutdown markdown editor for this (could totally be a future card but something to bear in mind)

kevindew:
Oh and I guess we also need to work out with Paola or Sonia about whether we give people the link to retire or not if they don't have access.

Probably need a card to style up the no permissions page a bit.

app/models/document.rb
emmabeynon: This should be addressed in this commit https://github.com/alphagov/content-publisher/pull/608/commits/82bf51135c41cec2d7ef58880cbfb2dac56fca21

emmabeynon:
Closed in favour of https://github.com/alphagov/content-publisher/pull/714

PR #612, opened 2018-12-13
Add internal notes backend

1pretz1:
Initial creation of the internal notes feature. Currently you can only create a internal note from the console which then displays in the document history section of the document page.

Trello:
https://trello.com/c/jilfIus0/547-mvp-of-reading-internal-notes

## Example:
screen shot 2018-12-17 at 16 16 48

### DIFF
--- a/app/models/internal_note.rb
+++ b/app/models/internal_note.rb
@@ -1,0 +1,7 @@
+# frozen_string_literal: true
+
+class InternalNote < ApplicationRecord
+  belongs_to :document
+  belongs_to :user, optional: true
+  belongs_to :timeline_entry, class_name: "TimelineEntry", foreign_key: :timeline_entries_id, inverse_of: :internal_note
+end
--- a/app/models/timeline_entry.rb
+++ b/app/models/timeline_entry.rb
@@ -24,6 +24,7 @@ class TimelineEntry < ApplicationRecord
     create_preview
     retired
     removed
+    internal_note
   ].freeze
 
   validates_inclusion_of :entry_type, in: ENTRY_TYPES
--- a/app/models/timeline_entry.rb
+++ b/app/models/timeline_entry.rb
@@ -3,8 +3,10 @@
 class TimelineEntry < ApplicationRecord
   belongs_to :document
   belongs_to :user, optional: true
+
   has_one :retirement, foreign_key: :timeline_entries_id, dependent: :destroy, inverse_of: :timeline_entry
   has_one :removal, foreign_key: :timeline_entries_id, dependent: :destroy, inverse_of: :timeline_entry
+  has_one :internal_note, foreign_key: :timeline_entries_id, dependent: :destroy, inverse_of: :timeline_entry
 
   ENTRY_TYPES = %w[
     created
--- a/app/views/documents/show/_history_tab.html.erb
+++ b/app/views/documents/show/_history_tab.html.erb
@@ -1,31 +1,39 @@
 <% displayed_edition_number = nil %>
 
-<% @document.timeline_entries.includes(:user).order(created_at: :desc).includes(:retirement).includes(:removal).each do |entry| %>
-  <div class="app-timeline-entry govuk-!-margin-bottom-6">
-    <% if displayed_edition_number != entry.edition_number %>
-      <h3 class="govuk-heading-m"><%= t "documents.history.edition_header", number: entry.edition_number.ordinalize %></h3>
-      <% displayed_edition_number = entry.edition_number %>
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <% @document.timeline_entries.includes(:user).order(created_at: :desc).includes(:retirement, :removal, :internal_note).each do |entry| %>
+      <div class="app-timeline-entry govuk-!-margin-bottom-6">
+        <% if displayed_edition_number != entry.edition_number %>
+          <h3 class="govuk-heading-m"><%= t "documents.history.edition_header", number: entry.edition_number.ordinalize %></h3>
+          <% displayed_edition_number = entry.edition_number %>
+        <% end %>
+
+        <h4 class="govuk-heading-s govuk-!-margin-bottom-1">
+          <%= t "documents.history.entry_types.#{entry.entry_type}" %>
+        </h4>
+
+        <div class="app-timeline-entry__dateline">
+          <%= entry.created_at.strftime("%d %B %Y at %l:%M%P") %> by <%= entry.username_or_unknown %>
+        </div>
+
+        <% if entry.internal_note.present? %>
+          <%= simple_format(escape_and_link(entry.internal_note.body)) %>
+        <% end %>
+
+        <% if entry.retirement.present? %>
+          <%= entry.retirement.explanatory_note %>
+        <% elsif entry.removal.present? %>
+          <% if entry.removal.explanatory_note.present? %>
+            <p><%= entry.removal.explanatory_note %></p>
+          <% end %>
+          <% if entry.removal.redirect %>
+            <p>Redirected to <%= link_to(Plek.new.website_root + entry.removal.alternative_path) %></p>
+          <% elsif entry.removal.alternative_path.present? %>
+            <p><%= link_to(Plek.new.website_root + entry.removal.alternative_path) %></p>
+          <% end %>
+        <% end %>
+      </div>
     <% end %>
-
-    <h4 class="govuk-heading-s govuk-!-margin-bottom-1">
-      <%= t "documents.history.entry_types.#{entry.entry_type}" %>
-    </h4>
-
-    <% if entry.retirement.present? %>
-      <%= entry.retirement.explanatory_note %>
-    <% elsif entry.removal.present? %>
-      <% if entry.removal.explanatory_note.present? %>
-        <p><%= entry.removal.explanatory_note %></p>
-      <% end %>
-      <% if entry.removal.redirect %>
-        <p>Redirected to <%= link_to(Plek.new.website_root + entry.removal.alternative_path) %></p>
-      <% elsif entry.removal.alternative_path.present? %>
-        <p><%= link_to(Plek.new.website_root + entry.removal.alternative_path) %></p>
-      <% end %>
-    <% end %>
-
-    <div class="app-timeline-entry__dateline">
-      <%= entry.created_at.strftime("%d %B %Y at %l:%M%P") %> by <%= entry.username_or_unknown %>
-    </div>
   </div>
-<% end %>
+</div>
--- a/db/migrate/20181212165840_create_internal_notes.rb
+++ b/db/migrate/20181212165840_create_internal_notes.rb
@@ -1,0 +1,14 @@
+# frozen_string_literal: true
+
+class CreateInternalNotes < ActiveRecord::Migration[5.2]
+  def change
+    create_table :internal_notes do |t|
+      t.text :body, null: false
+      t.references :document, foreign_key: { on_delete: :cascade }, null: false
+      t.references :user, foreign_key: { on_delete: :nullify }
+      t.references :timeline_entries, foreign_key: true, null: false
+
+      t.timestamps
+    end
+  end
+end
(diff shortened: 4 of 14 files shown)


app/models/internal_note.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+class InternalNote < ApplicationRecord
+  belongs_to :document
+  belongs_to :user, optional: true
+  belongs_to :timeline_entry, class_name: "TimelineEntry", foreign_key: :timeline_entries_id, inverse_of: :internal_note
```
kevindew: I think the class_name and foreign_key options are likely superfluous here.

app/models/internal_note.rb
```diff
@@ -0,0 +1,9 @@
...
+
+  validates :body, presence: true
```
kevindew: I think we now do all of validation through Requirements in services directory so I think this would be inconsistent here - be fine to just drop it.

app/models/timeline_entry.rb
```diff
@@ -3,8 +3,10 @@
 class TimelineEntry < ApplicationRecord
   belongs_to :document
   belongs_to :user, optional: true
+
   has_one :retirement, foreign_key: :timeline_entries_id, dependent: :destroy, inverse_of: :timeline_entry
   has_one :removal, foreign_key: :timeline_entries_id, dependent: :destroy, inverse_of: :timeline_entry
+  has_one :internal_note, foreign_key: :timeline_entries_id, dependent: :destroy, inverse_of: :timeline_entry
```
MuriloDalRi: Should this be has_many?

app/views/documents/show/_history_tab.html.erb
```diff
@@ -11,6 +11,14 @@
       <%= t "documents.history.entry_types.#{entry.entry_type}" %>
     </h4>
 
+    <div class="app-timeline-entry__dateline">
+      <%= entry.created_at.strftime("%d %B %Y at %l:%M%P") %> by <%= entry.username_or_unknown %>
+    </div>
+
+    <% if entry.internal_note.present? %>
+      <p><%= entry.internal_note.body %></p>
```
kevindew: Sorry this is probably only mentioned on the [epic](https://trello.com/c/d2ukJ9kE/592-internal-notes_) rather than the story - my bad

but this should support preserving line breaks and convert urls automatically into links

app/models/timeline_entry.rb
1pretz1: Nope i don't think so, as one timeline_entry is associated with one internal comment. E.g, you don't have one timeline_entry with multiple internal notes attached.

db/migrate/20181212165840_create_internal_notes.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+class CreateInternalNotes < ActiveRecord::Migration[5.2]
+  def change
+    create_table :internal_notes do |t|
+      t.text :body, null: false
+      t.references :document, foreign_key: { on_delete: :cascade }, null: false
+      t.references :user, foreign_key: { on_delete: :restrict }
```
kevindew: I think this should probably be a nullify rather than a restrict - I'm guessing though you've done it in a way that's consistent with retire and remove.

So if this is consistent I'd suggest it's good and we'll try resolve it and them at same time, but if it's different it probably should be nullify as I guess we want to allow users to be removed from the system if they so chose.

app/models/timeline_entry.rb
MuriloDalRi: Oh yeah right, I was thinking about something else. Nevermind, this looks good 👍

app/views/documents/show/_history_tab.html.erb
```diff
@@ -11,6 +11,14 @@
       <%= t "documents.history.entry_types.#{entry.entry_type}" %>
     </h4>
 
+    <div class="app-timeline-entry__dateline">
+      <%= entry.created_at.strftime("%d %B %Y at %l:%M%P") %> by <%= entry.username_or_unknown %>
+    </div>
+
+    <% if entry.internal_note.present? %>
```
kevindew: I imagine by accessing these this way we have an [n+1 issue](https://semaphoreci.com/blog/2017/08/09/faster-rails-eliminating-n-plus-one-queries.html) - which I guess we probably already had with removal and retire, it'd be good if we could try resolve that.

app/models/internal_note.rb
1pretz1: Without these rails doesn't recognise the foreign key for timeline_entry when creating a InternalNote as i believe rails is expecting timeline_entry_id instead of timeline_entries_id

kevindew: D'oh that's a pain - makes sense to have them then

app/models/internal_note.rb
1pretz1: Yep good point

app/views/documents/show/_history_tab.html.erb
1pretz1: 👍

benthorner:
Looks good - nice one @1pretz1 :-).

db/migrate/20181212165840_create_internal_notes.rb
1pretz1: Yep that makes sense, i followed the pattern as done with retire and remove.

app/views/documents/show/_history_tab.html.erb
1pretz1: Yep sure, i'll look into it

app/views/documents/show/_history_tab.html.erb
```diff
@@ -1,31 +1,39 @@
...
-    <% end %>
+        <div class="app-timeline-entry__dateline">
+          <%= entry.created_at.strftime("%d %B %Y at %l:%M%P") %> by <%= entry.username_or_unknown %>
+        </div>
 
-    <div class="app-timeline-entry__dateline">
-      <%= entry.created_at.strftime("%d %B %Y at %l:%M%P") %> by <%= entry.username_or_unknown %>
-    </div>
+        <% if entry.internal_note.present? %>
+          <p><%= sanitize(simple_format(entry.internal_note.body).gsub(URI.regexp, '<a href="\0">\0</a>')) %></p>
```
kevindew: I think there's a couple of problems with this and I'm also a bit confused.

So reading the docs [simple_format](https://apidock.com/rails/ActionView/Helpers/TextHelper/simple_format) will convert into a <p> element so I think this means that this doesn't need wrapping inside a p element? this also sanitizes right?

Then we do the gsub on the resulting HTML - which is a nice approach I didn't consider you could do that and thought you might need https://github.com/vmg/rinku or https://github.com/tenderlove/rails_autolink - though I guess that means we lose mailto links which would be nice to have. It'd also be good to just restrict the URI links to http and https which I think you can do with `URI.regexp(%w[http https])`

Then where I get confused is the sanitize call as I'd have thought that would have stripped out the elements we just added? but I assume that was fine with you trying it out?

Then finally I'd have thought (unless rails added it along the way somewhere) that this needs a html_safe call on it.

It might be easier to create a view helper method that we can test `simple_format_with_links` that takes simple_format arguments and any needed for links. What do you think?

1pretz1: You're correct in that `simple_format` wraps the text in a `<p>` tag, so the `<p>` i've added isn't needed which i will remove and yep we could restrict URI links as you've stated to http and https . 

The sanitize method works in a similar way to `html_safe` in terms of rendering the actual html, however it will strip the text of any potentially unsafe tags leaving the href links intact. As just calling `html_safe` on the text would create a vulnerability where publishers (highly unlikely i know) could submit a script which would be executed when the `internal_note` was rendered. This is a snippet from the docs about the sanitize method.

```Sanitizes HTML input, stripping all but known-safe tags and attributes.

It also strips href/src attributes with unsafe protocols like javascript:, while also protecting against attempts to use Unicode, ASCII, and hex character references to work around these protocol filters. All special characters will be escaped.

The default sanitizer is Rails::Html::WhiteListSanitizer. See Rails HTML Sanitizers for more information.```

I tested it locally by submitting links and hyperlinks were displayed on the internal note with the correct line breaks. However, if we were to keep this approach i'd agree that it'd probably be best to move it to its own method so its testable.

I could write another regexp that swaps any email address with it's corresponding `mailto` link, however this maybe feels like a case of recreating the wheel?

So, in summary i'd say if we want the mailto links we could use one of the gems you suggested instead of my approach but i'll leave the decision up to you :)

Comment (kevindew) [reply app/views/documents/show/_history_tab.html.erb]: Yeah if gem seems the best way to include mailto then give that a whirl.

I'm a bit confused how the santize method knows to allow `<a>`, `<p>` and `<br>` - are they considered the "known-safe tags" 

I thought that after we ran simple_format we'd then have a string that's safe to run html safe on as html entities would have been encoded accordingly. 

Probably good to have this as a tested method since it's generated quite a lot of conversation with us that could all be represented in the test. Sound good to you?

File: app/views/documents/show/_history_tab.html.erb [hunk 8]
```diff
@@ -1,31 +1,39 @@
...
-    <% end %>
+        <div class="app-timeline-entry__dateline">
+          <%= entry.created_at.strftime("%d %B %Y at %l:%M%P") %> by <%= entry.username_or_unknown %>
+        </div>
 
-    <div class="app-timeline-entry__dateline">
-      <%= entry.created_at.strftime("%d %B %Y at %l:%M%P") %> by <%= entry.username_or_unknown %>
-    </div>
+        <% if entry.internal_note.present? %>
+          <%= simple_format(Rinku.auto_link(entry.internal_note.body)) %>
```
kevindew: So what I think this should do is:

- take the note body and convert HTML into HTML entities
- then run auto link
- then run simple format

That would mean if a user entered any HTML it would be preserved.

So in terms of test cases:

input: "here is a note\nwith new lines" should output "<p>here is a note\n<br />with new lines</p>"(which it does currently)
input: "here is a note\nwith new lines\nand an email@example.com address" should output: "<p>here is a note\n<br />with new lines\n<br />and an <a href=\"mailto:email@example.com\">email@example.com</a> address</p>" (which it does currently)
input: "here is a note with <b>html</b> inside it" should output: "<p>here is a note with &lt;b&gt;html&lt;/b&gt; inside it</p>" (currently renders the bold)
input: "here is a note with a <script>script</script> inside it" should output: "<p>here is a note with &lt;script&gt;script&lt;/script&gt; inside it</p>" (currently strips the tag)

spec/features/workflow/internal_note_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+RSpec.feature "Displaying an internal note" do
+  scenario do
+    given_there_is_a_document_with_a_internal_note
+    when_i_visit_the_document_page
+    and_i_click_on_the_tab_document_history
+    then_i_should_see_the_internal_note
+  end
+
+  def given_there_is_a_document_with_a_internal_note
+    @document = create(:document, :publishable, base_path: "/news/belvita-soft-bakes")
```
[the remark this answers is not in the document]
1pretz1: We don't! Nor do we need the `publishable trait` thinking about it

spec/features/workflow/internal_note_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+      timeline_entry: @timeline_entry,
+      document: @document,
+    )
+  end
+
+  def when_i_visit_the_document_page
+    visit document_path(@document)
+  end
+
+  def and_i_click_on_the_tab_document_history
```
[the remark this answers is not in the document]
1pretz1: Yep you're correct!

spec/factories/internal_notes_factory.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
```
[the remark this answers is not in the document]
1pretz1: Good spot 👁

app/views/documents/show/_history_tab.html.erb
1pretz1: Thanks for the detailed explanation! I've made the changes as you've requested so the html is escaped and links / line breaks are added and have tested accordingly.

PR #615, opened 2018-12-14
Nullify users foreign key on delete for timeline_entries

1pretz1:
This change alters what happens when a user is deleted. Previously, a
user could not be deleted if they had timeline_entries. Now, users_id
gets set to null on deletion of a user.

This enables us to remove users from the system without having errors
related to database relations.


1pretz1:
Yep it got me! & I've now made the migration reversible.

PR #616, opened 2018-12-14
Refactor image upload requirements as a checker

benthorner:
https://trello.com/c/4iKDzhBY/379-separate-saving-a-change-from-the-creation-of-a-preview

This refactors the code to compute issues for image uploads so that it
matches the checker approach we use everywhere else. Part of this
approach is that we only check for one or two issues in the feature
specs, and use a unit spec to cover all of them in one place.


app/services/requirements/image_upload_checker.rb
```diff
@@ -0,0 +1,57 @@
...
+
+    def issues
+      issues = []
+
+      unless file
+        issues << Issue.new(:image_upload, :no_file)
+        return CheckerIssues.new(issues)
+      end
+
+      if allowed_format? || animated_image?
```
1pretz1: I think this should be `disallowed_format?` or equivalent

app/services/requirements/image_upload_checker.rb
```diff
@@ -0,0 +1,57 @@
...
+    def issues
+      issues = []
+
+      unless file
+        issues << Issue.new(:image_upload, :no_file)
+        return CheckerIssues.new(issues)
+      end
+
+      if unsupported_type? || animated_image?
+        issues << Issue.new(:image_upload, :format_not_allowed)
```
kevindew: One of the things I'd suggest we do (but not a priority) is to split up type and animated into separate checks. I'd suggest the static gif terminology of the message is a bit confusing and it'd be easier to just have a separate message that was "You must... Upload an animate that is not animated"

benthorner: Yeah, I was wondering about that - wanted it to be a pure refactor, though ;-). I've made a note and might get on to it later.

kevindew: Yup, no problem - I missed that first time round so thought I'd drop it in.

PR #617, opened 2018-12-14
Fix publish twice

benthorner:
https://sentry.io/govuk/app-content-publisher/issues/682788212/


app/controllers/publish_document_controller.rb
```diff
@@ -15,15 +15,17 @@ def confirmation
...
-    else
-      TimelineEntry.create!(document: document, user: current_user, entry_type: "published_without_review")
+    if document.publication_state == "sent_to_live"
+      redirect_to document_published_path(document)
+      return
     end
 
+    PublishService.new(document).publish(
+      user: current_user,
+      review_state: params[:review_state] || "published_without_review",
```
kevindew: We don't seem to be checking this part of user input

benthorner:
Strangely your comment isn't appearing, so this reply might look a bit strange! I noticed we weren't exactly checking the 'self_declared_review_state' param before either; perhaps we shouldn't be defaulting it either. My thought is that it'll be handled implicitly by the model validation, but we can do an explicit check instead if you prefer? 

Perhaps now is also a good time to remove the inconsistent default and show an array if the user doesn't select something, unless it's intentional - but then we should preselect the radio button like we do with the change note.

kevindew:
I'd suggest the ternary option is the best for the scope of what you're trying to achieve here, as the latter feels a substantial increase in scope. Can you write a card for the latter so we can check it with product too and have a ponder if there are other parts of the app that are similar?

benthorner:
Got a card for the defaulting https://trello.com/c/LhoqJDhC/623-revisit-the-radio-buttons-for-the-publish-confirmation-screen.

benthorner:
Should be OK now - can you take another look?

PR #620, opened 2018-12-17
Remove redundant uses of 'document' in our code

MuriloDalRi:
https://trello.com/c/YpJhGyqy/548-remove-redundant-uses-of-document-in-our-code

Some of our controllers and services unnecessarily use the word 'document' to describe themselves e.g. 'DocumentImagesController' and 'DocumentUnpublishingService'. We should remove these uses to make our naming more consistent.


app/services/unpublishing_service.rb
```diff
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-class DocumentUnpublishingService
+class UnpublishingService
```
benthorner: This one should probably be 'UnpublishService' for consistency.

MuriloDalRi: Oops missed a few letters here, well spotted!

benthorner:
Hmm...I think you missed renaming the unpublishing specs. Could you do another PR for that?

MuriloDalRi:
It's just the filename, right? I'll open a PR

PR #622, opened 2018-12-17
Revamp models

benthorner:
https://trello.com/c/qYHMcyDZ/550-revisit-our-approach-to-models

Another attempt to improve our format-related models and the way we use them in tests. I'm pretty happy with the overall result here: the models are more DRY and the factories consistent.


benthorner:
Yay 🎉

PR #624, opened 2018-12-17
Rename test file

MuriloDalRi:


benthorner:
There's also the service spec, I think :-).

MuriloDalRi:
Found another spec that needed renaming. Is that the one you meant?

PR #626, opened 2018-12-17
Making feedback and support request links spawn new windows

kgarwood:
https://trello.com/c/TiZgaPxw/537-open-support-links-in-a-new-tab

Changing links for 'Send us feedback' and 'Raise a support request' in the banner and footer so that when users click them the pages appear in a new window.


benthorner:
Do we want the rest of the links in the footer to open new tabs? It seems weirdly inconsistent as the links do different things without any visual indicator.

kevindew:
| Do we want the rest of the links in the footer to open new tabs? It seems weirdly inconsistent as the links do different things without any visual indicator.

For now we just want the support and feedback ones. You're right about it being an inconsistency and it pains me. But continuing further down the road of opening new tabs seems to deviate further and further from GOV.UK norm so feels a bit rock and hard place. Support and Feedback are the ones that give us the immediate concern anyway.

I guess the ones that make sense to be new tab are all the offsite ones (these, status and "how to write, improve...") whereas it feels correct that the ones in the right list open in target self since they're the same app. I guess that is consistent with our guidance links - as they are target blank.

But yeah I'm not sure whether it's best to just remove the target blanks when in production. 

That was all a bit rambley, so the tl;dr is the inconsistency is annoying but these seem to be the only links that can get into a nasty bind with.

app/views/layouts/application.html.erb
```diff
@@ -122,11 +122,14 @@
         items: [
           {
             href: "https://support.publishing.service.gov.uk/technical_fault_report/new",
-            text: "Raise a support request"
+            text: "Raise a support request",
+            attributes: {target: "_blank"}
```
kevindew: The linter doesn't seem to cover erb, but if it did it'd be asking you to put spaces inside the hash

benthorner:
@kevindew I thought we'd be wanting to avoid them, but if we must then we must.

app/views/layouts/application.html.erb
kgarwood: Done. Spaces put in

PR #628, opened 2018-12-17
Use separate messages for invalid image formats

benthorner:
The combined message might be confusing when a user uploads an animated
image, as the message doesn't higlight this problem specifically.

### DIFF
--- a/app/services/requirements/image_upload_checker.rb
+++ b/app/services/requirements/image_upload_checker.rb
@@ -23,8 +23,13 @@ module Requirements
         return CheckerIssues.new(issues)
       end
 
-      if unsupported_type? || animated_image?
-        issues << Issue.new(:image_upload, :format_not_allowed)
+      if unsupported_type?
+        issues << Issue.new(:image_upload, :unsupported_type)
+        return CheckerIssues.new(issues)
+      end
+
+      if animated_image?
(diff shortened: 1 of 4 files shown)


app/services/requirements/image_upload_checker.rb
```diff
@@ -23,8 +23,13 @@ def issues
...
       end
 
-      if unsupported_type? || animated_image?
-        issues << Issue.new(:image_upload, :format_not_allowed)
+      if unsupported_type?
+        issues << Issue.new(:image_upload, :unsupported_type)
+        return CheckerIssues.new(issues)
+      end
+
+      if animated_image?
```
kevindew: Do you know if we were relying on it being a supported type before checking if it was animated? I wonder if MiniMagick might raise an exception if it was say given a text file

benthorner: I can't see the issue here. We're checking the type first, so we know it's a supported type by the time we check animation. We've got a test that checking for unsupported types in the form of a text file, so that should cover any impact of switching these checks around - is that what you're concerned about?

kevindew: Oh yes sorry - I missed that it returns before proceeding.

PR #632, opened 2018-12-17
Add copy and paste tracking using GTM data layer

alex-ju:
This PR enables tracking of `copy` and `paste` events using GTM on specific components (initially markdown-editor and copy-link).

Author of PR: @erino; I did some clean up and updated tests.

[Trello card](https://trello.com/c/QHmf1x47)

### DIFF
--- a/app/assets/javascripts/modules/gtm-track-copy.js
+++ b/app/assets/javascripts/modules/gtm-track-copy.js
@@ -3,33 +3,33 @@ function GTMCopyListener ($module, dataLayer) {
   this.dataLayer = dataLayer || window.dataLayer
 }
 
-GTMCopyListener.prototype.handleCopyEvent = function (component) {
+GTMCopyListener.prototype.handleCopyEvent = function (value) {
   this.dataLayer.push({
     'event': 'text-copied',
-    'component': component
+    'element': value
   })
 }
 
-GTMCopyListener.prototype.handlePasteEvent = function (component) {
+GTMCopyListener.prototype.handlePasteEvent = function (value) {
   this.dataLayer.push({
     'event': 'text-pasted',
-    'component': component
+    'element': value
   })
 }
 
 GTMCopyListener.prototype.init = function () {
   var $module = this.$module
-  var moduleName = this.$module.dataset.module
+  var datasetValue = this.$module.dataset.gtmCopyPasteTracking
   $module.addEventListener('copy', function (e) {
-    this.handleCopyEvent(moduleName)
+    this.handleCopyEvent(datasetValue)
   }.bind(this), false)
 
   $module.addEventListener('paste', function (e) {
-    this.handlePasteEvent(moduleName)
+    this.handlePasteEvent(datasetValue)
   }.bind(this), false)
 }
 
-var $trackedComponents = document.querySelectorAll('[data-module="markdown-editor"], [data-module="copy-to-clipboard"]')
+var $trackedComponents = document.querySelectorAll('[data-gtm-copy-paste-tracking]')
 $trackedComponents.forEach(function ($component) {
   new GTMCopyListener($component).init()
 })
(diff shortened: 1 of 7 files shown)


app/assets/javascripts/modules/gtm-track-copy.js
```diff
@@ -0,0 +1,25 @@
+function GTMCopyListener () { }
+
+GTMCopyListener.prototype.handleCopyEvent = function () {
+  window.dataLayer.push({
+    'event': 'text-copied'
```
kevindew: From my conversation with Andy he was suggesting that you could only append data to the dataLayer variable prior to GTM initialisation - which would suggest these method calls wouldn't have any effect aside from modifying our global variable.

alex-ju: Events can be pushed at any time if a listener is setup in GTM to record them.

alex-ju:
> If I understand correctly any time a user copy and pastes text in Content Publisher we'd be trying to track an event of it - and I'm not sure that's a data source we really want - say a user copies and pastes 20 times in the edit of a doc.

> I seem to remember originally us trying to track the manual usage of copying links in the copy link component - but I'm not sure how this would fit into that.

We should definitely narrow down/scope this tracking to specific component. As I got from our PA, that should be the markdown editor and the copy link component. It's worth having clarifying these before moving on, indeed.

app/assets/javascripts/modules/gtm-track-copy.js
erino: I don't think that's the case, as we're pushing the value of the checked radio buttons on form submit and that's being tracked.

kevindew: Ah great - that's better than setting them in the header like we were doing.

app/assets/javascripts/modules/gtm-track-copy.js
```diff
@@ -0,0 +1,35 @@
...
+  $module.addEventListener('copy', function (e) {
+    this.handleCopyEvent(moduleName)
+  }.bind(this), false)
+
+  $module.addEventListener('paste', function (e) {
+    this.handlePasteEvent(moduleName)
+  }.bind(this), false)
+}
+
+var $trackedComponents = document.querySelectorAll('[data-module="markdown-editor"], [data-module="copy-to-clipboard"]')
```
kevindew: Rather than adding this to an explicit selector could we add this as a data attribute that could be selected on - in a similar vein to: https://github.com/alphagov/content-publisher/blob/b161ecb623c642e6fc3199e619de320b9adca5c4/app/assets/javascripts/modules/warn-before-unload.js#L26

app/assets/javascripts/modules/gtm-track-copy.js
```diff
@@ -0,0 +1,35 @@
+function GTMCopyListener ($module, dataLayer) {
+  this.$module = $module
+  this.dataLayer = dataLayer || window.dataLayer
+}
+
+GTMCopyListener.prototype.handleCopyEvent = function (component) {
+  this.dataLayer.push({
+    'event': 'text-copied',
+    'component': component
```
kevindew: If I remember right Andy was trying to track which copy-to-clipboard component was used so just the component name might not suffice as you'd not be able to differentiate between them.

We have a data attribute set on the input: https://github.com/alphagov/content-publisher/blob/a656889ebe093551a5a2d848cbc937bad2bee886/app/views/documents/show/_submitted_for_review.html.erb#L7

I wonder if we could send that rather than the component name?

alex-ju: Sure, I thought we want to track the component overall and then narrow differentiate using the page they're being used on. Can update to use explicit gtm values for each usage.

app/views/documents/fields/_govspeak_input.html.erb
```diff
@@ -9,7 +9,8 @@
         data: {
           "contextual-guidance": "document-contents-guidance",
           gramm: "false", # Disables grammerly plugin for markdown editor
-          gtm: "#{field.id}-input"
+          gtm: "#{field.id}-input",
+          "gtm-copy-paste-tracking": "copy-url-for-2i-approval-cta",
```
alex-ju: This should be `copy-paste-markdown`. Will raise a PR with a fix for it.

PR #638, opened 2018-12-18
Adds no results message

DilwoarH:
Ticket: 
- ~https://trello.com/c/pYDNDc14/565-issue-search-contacts-does-not-start-showing-choices-until-the-third-character-typed~
- https://trello.com/c/qXFrMELH/624-autocomplete-does-not-show-users-any-message-in-an-event-of-no-result

Screenshots:
screen shot 2018-12-19 at 10 41 00

### DIFF
--- a/app/assets/javascripts/components/autocomplete.js
+++ b/app/assets/javascripts/components/autocomplete.js
@@ -14,7 +14,7 @@ if ($autocompletes) {
     new window.accessibleAutocomplete.enhanceSelectElement({ // eslint-disable-line no-new, new-cap
       selectElement: $el,
       minLength: 3,
-      showNoOptionsFound: false,
+      showNoOptionsFound: true,
       customAttributes: customAttributes
     })
   })


kevindew:
Hey @DilwoarH. I thought I'd wrote notes on all of the issues Kevin raised but I missed this one 😞

Unfortunately we introduced the minimum of 3 characters because IE 11 freezes up for some of the accessible autocompletes (the actual problem is when 4,000 LI elements are inserted into the DOM as can happen with "Ministers and government appointments" so I think we'd find that if we tried this change out on IE 11 it'd be back to being very slow.

Using the `showNoOptionsFound` seems good to me though. @soniaturcotte do you have concerns about that? Only thing I wonder is can we change the text, I'm not sure if "No results found" feels a bit wrong for our select box replacements?

soniaturcotte:
Yes agree we should have the option for no results. That text is from the component so I'm ok with it: https://alphagov.github.io/accessible-autocomplete/examples/

app/assets/javascripts/components/autocomplete.js
```diff
@@ -14,7 +14,7 @@ if ($autocompletes) {
     new window.accessibleAutocomplete.enhanceSelectElement({ // eslint-disable-line no-new, new-cap
       selectElement: $el,
       minLength: 3,
-      showNoOptionsFound: false,
+      showNoOptionsFound: true,
```
alex-ju: 👍

app/views/components/_autocomplete.html.erb
```diff
@@ -18,10 +18,6 @@
       <span class="govuk-hint app-c-autocomplete__multiselect-instructions">To select multiple items in a list, hold down Ctrl (PC) or Command (Mac) key.</span>
     <% end %>
 
-    <% if hint %>
```
alex-ju: I think we must keep the hint before the select element, as we do with the other form components and it seems like we're using it like this in the tags page.

DilwoarH:
@soniaturcotte @kevindew @alex-ju - I have removed the hint text. This PR only add the no results message now.

PR #640, opened 2018-12-18
Image filename snippet

benthorner:
This normalises the filename associated with an image so that we can show it as a markdown snippet on the images index page. Making the changes to the views was a bit painful, so I've included a few commits to refactor the partials and image_meta component, which made the change easier; the refactoring means it's probably easier to review this PR one commit at a time, though.


app/services/image_filename_service.rb
```diff
@@ -0,0 +1,47 @@
...
+    "#{filename.base}-#{suffix}.#{filename.extension}"
+  end
+
+  def unused_suffix(suggested_base, bases)
+    suffixes = bases.map do |base|
+      match = base.match(/^#{suggested_base}-([0-9]+)$/)
+      match.to_a[1].to_i
+    end
+
+    ((1..(bases.count + 1)).to_a - suffixes).min
```
1pretz1: Took me a mo to wrap my head round this method, clever!

benthorner: I'm happy to tweak this if you can think of something that makes it easier to understand :-).

1pretz1: Hmm apart from generating a SecureRandom uuid for the suffix which isn't necessarily unique and makes horrendous file names, i can't think of a better approach then what you've implemented!

leenagupte: I don't normally advocate comments in code, but in this case I think it is needed. Or, can you split this method up so that the child methods are named in a way that explains exactly what they are returning?

benthorner: As the method is only 4 lines long, there's not much scope for splitting it up. I do appreciate it's a little tricky to understand, though. Could you try and describe what's most confusing?

spec/services/image_filename_service_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+      document = build :document
+      name = ImageFilenameService.new(document).call("file.jpg", "image/png")
+      expect(name).to eq "file.png"
+    end
+
+    it "ensures the filename is unique" do
+      document = create(:document)
+      create(:image, document: document, filename: "name.jpg")
+      create(:image, document: document, filename: "name-1.jpg")
+      create(:image, filename: "name-2.jpg")
```
leenagupte: I think you can remove this line. Unless you change the test name to be "ensures filenames of images associated with a document are unique"?

benthorner:
@kevindew the concurrent request issue seems to be something we live with, as it's also a problem for the base_path generator. I'm not too worried about it - we may as well wait and see if it's an issue.

I'll wait for @leenagupte to comment on the code. I'm happy to spend some time making it easier to understand, as I'd also like to re-use it for the base_path generator - this algorithm only requires one DB request and doesn't throw exceptions.

app/services/image_filename_service.rb
leenagupte: Love the while loop! 

But @benthorner, to answer your question, the bits that I found confusing where:

`match.to_a[1].to_i`

and

`((1..(bases.count + 1)).to_a - suffixes).min`

benthorner:
@kevindew @leenagupte @1pretz1  thanks for the suggestions :-). I've updated now to also truncate the name as specified on the card. What do you think?

app/services/image_filename_service.rb
```diff
@@ -1,15 +1,22 @@
...
   def call(suggested_name, mime_type)
-    normalised_name = normalise(suggested_name, mime_type)
-    ensure_unique(normalised_name)
+    filename = ActiveStorage::Filename.new(suggested_name)
+
+    base = filename.base.parameterize.slice 0...MAX_LENGTH
+    base = ensure_unique(base)
+
+    extension = Rack::Mime::MIME_TYPES.invert[mime_type]
+    "#{base}#{extension}"
```
leenagupte: Is there a missing `.` here, between base and extension?

app/services/image_filename_service.rb
```diff
@@ -18,30 +25,23 @@ def images
...
   end
 
-  def unused_suffix(suggested_base, bases)
-    suffixes = bases.map do |base|
-      match = base.match(/^#{suggested_base}-([0-9]+)$/)
-      match.to_a[1].to_i
+  def unused_suffix(suggested_base, potential_conflicts)
+    suffix = 1
+
+    while potential_conflicts.include?("#{suggested_base}-#{suffix}")
```
leenagupte: Nice!

app/services/image_filename_service.rb
benthorner: Weirdly the dot's included - I thought it would be more confusing to manually take it out and then manually put it in again :-|.

app/views/images/_image.html.erb
```diff
@@ -0,0 +1,55 @@
+<%
+  actions = []
```
leenagupte: There is a lot of code in these views. Have you considered moving some of it out into a Helper class?
govuk_publishing_components does something similar for the step-by-step component: https://github.com/alphagov/govuk_publishing_components/blob/8b575b247e1eeaa0b3bf8544492fe4303de050d2/app/views/govuk_publishing_components/components/_step_by_step_nav.html.erb#L17

app/services/image_filename_service.rb
```diff
@@ -1,15 +1,22 @@
...
   def initialize(document)
     @document = document
   end
 
   def call(suggested_name, mime_type)
-    normalised_name = normalise(suggested_name, mime_type)
-    ensure_unique(normalised_name)
+    filename = ActiveStorage::Filename.new(suggested_name)
+
+    base = filename.base.parameterize.slice 0...MAX_LENGTH
```
leenagupte: I would split out the code to truncate the filename into separate commit, to make it clear what was changed because of truncation, and what has just been refactored, for example, method renames.

kevindew:
Looks good to me, do we know what happens or if it's possible to have an empty file name?

app/services/image_filename_service.rb
benthorner: As the service is new in this PR, I've squashed the changes into the original commit instead, to make it easier for people in future.

benthorner:
@kevindew we've assumed there will always be a filename. Probably worth waiting to see if it's actually a problem :-).

app/services/image_upload_service.rb
```diff
@@ -31,8 +31,8 @@ def upload_to_asset_manager(image)
     AssetManagerService.new.upload_bytes(image, image.cropped_bytes)
   end
 
-  def filename
-    file.respond_to?(:original_filename) ? file.original_filename : File.basename(file)
+  def filename(document)
+    @filename ||= ImageFilenameService.new(document).call(file.original_filename, mime_type)
```
leenagupte: Should the changes to this service be in a separate commit like: "Use new ImageFilenameService to generate the filename"?

spec/features/editing_images/upload_lead_image_spec.rb
```diff
@@ -13,11 +13,11 @@
 
   def given_there_is_a_document
     document_type = build(:document_type, lead_image: true)
-    create(:document, document_type_id: document_type.id)
+    @document = create(:document, document_type_id: document_type.id)
```
leenagupte: Perhaps split the changes to this file out too as it's not related to the commit message?

config/locales/en/documents/show.yml
```diff
@@ -37,11 +37,9 @@ en:
...
         no_lead_image: No image selected. The default image for your department will be used.
         upload_image: Upload an image
-        fields:
-          alt_text: Alt text
-          caption: Image caption
-          credit: Image credit
-          image: Image
+        alt_text: Alt text
+        caption: Caption
+        credit: Credit
```
leenagupte: Were these changes, and the ones in images/index.yml, requested by a content designer/designer/user researcher?

app/views/images/_image_list.html.erb
```diff
@@ -28,25 +28,36 @@
...
-        {
-          field: t("images.index.meta.caption"),
-          value: image.caption,
-        },
-        {
-          field: t("images.index.meta.credit"),
-          value: image.credit,
-        }
-      ]
+      metadata_items: metadata_items,
```
leenagupte: Nitpick - This refactor could have been squashed into the previous commit.

spec/features/editing_images/upload_inline_image_spec.rb
```diff
@@ -5,7 +5,7 @@
     given_there_is_a_document
     when_i_visit_the_images_page
     and_i_upload_a_new_image
-    then_it_has_a_normalised_filename
+    then_i_see_it_has_a_snippet
```
leenagupte: Is there a missing step that checks that the user has the correct permissions to see the snippet?

PR #646, opened 2018-12-19
Lint JavaScript tests

alex-ju:
Quick clean in the JavaScript specs as a few StandardJS linting errors made their way in.
Also replaced the template literals for consistency, and, if I remember well they're not supported but our testing framework - also helps running the tests cross browser.


alex-ju:
Nope, but some of them could be.

PR #650, opened 2018-12-20
Add document history title for mobile view

DilwoarH:
Ticket: https://trello.com/c/hj7RXU4S/532-content-summary-change-note-meta-data-and-fix-mobile-styles

Screenshots:
screen shot 2018-12-20 at 13 50 54
screen shot 2018-12-20 at 13 52 43


app/assets/stylesheets/objects/_history.scss
```diff
@@ -1,3 +1,11 @@
+.app-timeline-heading {
+  display: none;
```
alex-ju: This would be a great improvement if we do it mobile first and make it available for screen readers too by keeping it in the accessibility tree.
```
.app-timeline-heading {
    @include govuk-media-query($from: tablet) {
        @include govuk-visually-hidden;
    }
}
```

PR #653, opened 2018-12-20
[WIP] ADR 8 - Reserving paths in Publishing API

kevindew:
I've flagged this as a work in progress as we've currently got 3 ADR's on the go so this one will slip onto on hold for time being.

👉 [View rendered version](https://github.com/alphagov/content-publisher/blob/adr-8/docs/adr/0008-reserving-paths-in-publishing-api.md) 👈 

This ADR is an attempt to try formulate a plan around a path reservation strategy. This ADR attempts to lay out both the goals we want to aim for and a proposed solution. In reviewing it'd be good to focus first on the goals rather than the solution as if we change the goals the solution proposal would need re-thinking.

Reasons to think about doing this:

- We currently have the non-optimal requirement that asks a user to change a title if they try use a title that already exists
- We've had a number of conflicts from Publishing API: https://sentry.io/govuk/app-content-publisher/issues/682788212/

Related stories/issues on trello: https://trello.com/c/snc4ihBE/575-delegate-unique-basepath-generation-to-the-publishing-api, https://trello.com/c/NPBZppvg/604-prevent-changes-to-the-basepath-after-its-been-published

### DIFF
--- a/docs/adr/0008-reserving-paths-in-publishing-api.md
+++ b/docs/adr/0008-reserving-paths-in-publishing-api.md
@@ -1,0 +1,176 @@
+# 8. Reserving paths in Publishing API
+
+Date: 2018-12-20
+
+## Context
+
+Content Publisher and Whitehall can both publish the same document types to
+GOV.UK that use the same URL formats. This can lead to conflict scenarios where
+a publishing in one application is blocked by a publishing in the other.
+
+For example, if Whitehall has a document with a base path of
+`/government/news/breaking-story` and Content Publisher also has a document
+with a base path of `/government/news/breaking-story` only the first app can
+publish to GOV.UK.
+
+Both applications prevent a user from entering a base path themselves and
+instead generate one based on a title. They currently both append numeric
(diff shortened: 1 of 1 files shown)


docs/adr/0008-reserving-paths-in-publishing-api.md
```diff
@@ -0,0 +1,176 @@
...
+
+When a user makes changes to the title input on the edit content form a HTTP
+request is sent to Content Publisher to generate a title. This should operate
+as follows:
+
+- The title and path prefix should be used to check whether there is
+  an available `ReservedPath` available in Content Publisher database;
+- If one is found it should be verified that the `ReservedPath` belongs to the
+  corresponding document (or session);
+- If there is not one found the Publishing API should be called to generate
```
erino: What happens to the existing `ReservedPath`? do we need to be explicit here that it is kept by Content Publisher?

kevindew:
This clearly isn't a work in progress anymore so I'm going to close it.

PR #659, opened 2018-12-27
Re-approach the markdown editor styling

alex-ju:
Update component so is doesn't rely on absolute positioning while preserving the visual styles for focus state.

👉 [Preview markdown editor component](https://content-publisher-revie-pr-659.herokuapp.com/component-guide/markdown_editor)

Problems solved with the new approach:
 - scrolling the textarea overlaps the editor header (https://trello.com/c/uiCA2sAo)
 - grammarly interferes with the component positioning hiding the editor header (https://trello.com/c/1pgYWuWB)

[Trello card](https://trello.com/c/C7RtuUhV)

### DIFF
--- a/app/assets/javascripts/components/markdown-editor.js
+++ b/app/assets/javascripts/components/markdown-editor.js
@@ -24,12 +24,15 @@ MarkdownEditor.prototype.init = function () {
   // Enable toggle bar
   this.$head.style.display = 'block'
 
-  // Handle events
+  // Handle button events
   this.$previewButton.addEventListener('click', $module.boundPreviewButtonClick)
   this.$editButton.addEventListener('click', $module.boundEditButtonClick)
 
-  // Bubble focus events
-  this.bubbleFocus(this.$input, this.$container)
+  // Reflect focus events
+  this.reflectFocusStateToContainer(this.$input, this.$container)
+  this.bubbleFocusEventToComponent(this.$input)
+  this.bubbleFocusEventToComponent(this.$editButton)
+  this.bubbleFocusEventToComponent(this.$previewButton)
 }
 
 MarkdownEditor.prototype.handlePreviewButton = function (event) {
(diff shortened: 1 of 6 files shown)


app/assets/javascripts/components/markdown-editor.js
```diff
@@ -122,6 +126,16 @@ MarkdownEditor.prototype.setTargetBlank = function (container) {
   }
 }
 
+// Bubble focus and blur events from specified element to container
+MarkdownEditor.prototype.BubbleFocus = function (element, container) {
```
kevindew: Is this method intentionally prefixed with a capital letter?

alex-ju: No, it should be in camel-case.

alex-ju:
@kevindew was planning to raise a separate PR to deal with how this component works with the guidance

app/assets/javascripts/components/markdown-editor.js
```diff
@@ -26,6 +27,9 @@ MarkdownEditor.prototype.init = function () {
   // Handle events
   this.$previewButton.addEventListener('click', $module.boundPreviewButtonClick)
   this.$editButton.addEventListener('click', $module.boundEditButtonClick)
+
+  // Bubble focus events
+  this.bubbleFocus(this.$input, this.$container)
```
benthorner: Could you rename this to something like `toggleContainerFocus`, as it made me think it was a generic method?

benthorner:
Might this fix https://trello.com/c/uiCA2sAo/523-scrolling-down-with-keyboard-and-then-up-hides-the-top-of-the-body-text ?

alex-ju:
@benthorner yes, it fixes that problem too. It's mentioned in the PR description so the card references this PR

app/assets/javascripts/components/markdown-editor.js
alex-ju: yep, is not as generic as the names suggests, but it's not a toggle thing either (as it can receive consecutive focus/blur events without alternating). `reflectElementFocus`?

app/assets/javascripts/components/markdown-editor.js
```diff
@@ -122,6 +128,18 @@ MarkdownEditor.prototype.setTargetBlank = function (container) {
   }
 }
 
+// Bubble/reflect focus and blur events and style to other element
+MarkdownEditor.prototype.bubbleFocus = function (element, reflectEventElement, reflectStyleElement) {
+  element.addEventListener('focus', function (event) {
+    reflectEventElement && reflectEventElement.dispatchEvent(new window.Event('focus'))
+    reflectStyleElement && reflectStyleElement.classList.add('app-c-markdown-editor__container--focused')
```
benthorner: Could we separate this behaviour out with a separate function/listener?

app/assets/javascripts/components/markdown-editor.js
```diff
@@ -26,6 +27,11 @@ MarkdownEditor.prototype.init = function () {
   // Handle events
   this.$previewButton.addEventListener('click', $module.boundPreviewButtonClick)
   this.$editButton.addEventListener('click', $module.boundEditButtonClick)
+
+  // Bubble focus events
+  this.bubbleFocus(this.$input, this.$module, this.$container)
```
benthorner: Could we avoid passing module here? I think this would create an opportunity for a more explanatory method name :-).

alex-ju:
@benthorner thanks for the review. Updated addressing your comments.

PR #674, opened 2019-01-14
Merge in versioned code

kevindew:

This merges the code under the Versioned namespace into the main namespace of the application to remove it.


app/models/versioning/edition.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+
+module Versioning
+  class Edition < ApplicationRecord
+    self.table_name = "versioned_editions"
+
+    attr_readonly :number, :document_id
+
+    # rubocop:disable Rails/InverseOf
```
1pretz1: Why not use inverse_of here?

app/services/preview_service.rb
```diff
@@ -1,72 +1,61 @@
...
 
-  def try_create_preview(args)
-    save_changes(args)
-    try_publish_draft(document) unless has_issues?
+  def try_create_preview
+    if has_issues?
+      edition.update!(revision_synced: false)
+    else
+      begin
+        upload_assets(edition)
```
emmabeynon: Just a small one - can we just call `create_preview` here instead of repeating these lines?

kevindew: Yup - good catch

kevindew: @emmabeynon refactored in 5a4d96cdbc4088848ffb97db4123b67e8f50debe

PR #675, opened 2019-01-17
Fix image revisions blocking revision deletion

kevindew:

Prior to this change you would be unable to delete entries out of the
revision table if there were any associations with image revisions. This
might make these prohibitively hard to delete, instead we decide since
revision is the more important concept that if you want to delete that
then lose the joins.

We don't however allow the ability to remove an image revision without
removing the join first, this is because image_revisions are part of the
state of a revision and these should only be removed (without removing
the revision) with caution and exceptional circumstance.

Like many of these on_delete behaviours these are not here for day to
day to usage of the application, and are mostly anticipated to get used
in debugging or development scenarios.

### DIFF
--- a/db/migrate/20190116194130_cascade_image_revision_joins_on_revision_delete.rb
+++ b/db/migrate/20190116194130_cascade_image_revision_joins_on_revision_delete.rb
@@ -1,0 +1,23 @@
+# frozen_string_literal: true
+
+class CascadeImageRevisionJoinsOnRevisionDelete < ActiveRecord::Migration[5.2]
+  def up
+    remove_foreign_key :versioned_revision_image_revisions,
+                       column: :revision_id
+
+    add_foreign_key :versioned_revision_image_revisions,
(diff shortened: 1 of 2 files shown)


db/migrate/20190116194130_cascade_image_revision_joins_on_revision_delete.rb
```diff
@@ -0,0 +1,23 @@
...
+                    :versioned_revisions,
+                    column: :revision_id,
+                    on_delete: :cascade
+  end
+
+  def down
+    remove_foreign_key :versioned_revision_image_revisions,
+                       column: :revision_id
+
+    add_foreign_key :versioned_revision_image_revisions,
```
benthorner: This looks like it's the same in both directions :-|.

kevindew: Crap, thanks

PR #677, opened 2019-01-17
Upload all images to asset manager

kevindew:

This undoes the change that only uploaded images to Asset Manager if the
file was the lead image. This is in preparation for inline images where
we may want to use any of these images.

As part of this the AssetManagerHelper for tests has been created to
provide access to the mocking methods that have thus far been set up
inconsistently across these tests.

We may well want to move those methods to GDS API Adapters.


kevindew:
@benthorner yup I've made a note of both.

PR #678, opened 2019-01-17
Update autocomplete component to support custom template for results

alex-ju:
This PR:
- updates the autocomplete component to add a variation with custom template for results. It uses a `data-hint` attribute as the source for the additional information provided for each result.
- updates contacts autocomplete to use the custom template variation (adds organisation name to `data-hint` and seths the `custom_template: true` attribute to use the custom template variation

## Preview
Preview example `/component-guide/autocomplete/autocomplete_with_custom_template`.

autocomplete-custom-template

This is an alternative approach to #571, which was using a character separator to extract the hint.
[Trello card](https://trello.com/c/05U8l7DF)


app/assets/javascripts/components/autocomplete.js
```diff
@@ -4,18 +4,63 @@ var $autocompletes = document.querySelectorAll('[data-module="autocomplete"]')
...
-      selectElement: $el,
+      selectElement: $select,
       minLength: 3,
       showNoOptionsFound: true,
       customAttributes: customAttributes
     })
   })
 }
+
+var $customTemplateAutocomplete = document.querySelector('[data-module="autocomplete-custom-template"]')
```
benthorner: I'm a little confused here, but hopefully just my misunderstanding. The name 'autocomplete-custom-template' made me think we would be passing an HTML template in somewhere. But it looks like the 'custom-template' just changes the HTML in a fixed way; therefore, could we call it 'autocomplete-with-hint' or similar, to express that it's specific look 'n feel?

app/assets/javascripts/components/autocomplete.js
```diff
@@ -4,18 +4,63 @@ var $autocompletes = document.querySelectorAll('[data-module="autocomplete"]')
...
+    element: $customTemplateAutocomplete.querySelector('.autocomplete__wrapper'),
+    id: $select.id,
+    source: function (query, syncResults) {
+      var results = []
+      $options.forEach(function ($el) {
+        results.push({value: $el.textContent, hint: $el.dataset.hint || ''})
+      })
+
+      syncResults(query
+        ? results.filter(function (result) {
```
kevindew: Minor but I think there's probably a bit too much code here for a ternary operator

alex-ju:
Update the name of this variation to be `autocomplete-with-hint-on-options`, not to be confused with an autocomplete with a hint element  - the grey text underneath the label that we have on every form component.

app/assets/javascripts/components/autocomplete.js
```diff
@@ -4,18 +4,63 @@ var $autocompletes = document.querySelectorAll('[data-module="autocomplete"]')
...
+    templates: {
+      inputValue: function (result) {
+        return result && result.value
+      },
+      suggestion: function (result) {
+        return result && result.value + '<span class="autocomplete__option-hint">' + result.hint + '</span>'
+      }
+    }
+  })
+}
```
kevindew: Should we have this file as a distinctly different JS file since it doesn't share code with the first part?

alex-ju: I consider this as not being a different component, but a file where we initialise the autocomplete components – in our case in two different ways.

alex-ju:
@kevindew I've introduced that issue on my last commit. Fixed it now:

With results
screen shot 2019-01-23 at 15 14 27

Without results
screen shot 2019-01-23 at 15 15 28

PR #681, opened 2019-01-18
Versioned images redux

kevindew:

After a long session with Ben Thorner chatting about concern/confusion
with the way we model Images with Image, ImageRevision, AssetManagerFile
and AssetManagerImageVariant, we came up with a different approach that
was more consistent with the modelling of other revisions in Content
Publisher.

In this we have:

Image - stores a couple of simple bits of continuous information such as
an id, who created it and when.

ImageRevision - stores references to image, file_revision and
metadata_revision, doesn't really too much - is just used to group data
together and be a common join point

ImageFileRevision - this stores data that changing means that the assets
created need to be changed. It stores filename and crop information.
Originally this was to be called CropRevision however it needed to be
expanded to include filename and various blob details.

ImageMetadataRevision - this stores data that can be updated without
needing to re-send the assets to Asset Manager, it stores information
such as alt text, credit and caption

ImageAsset - this has a many to one relationship with an
ImageFileRevision, there will be one of these for every resulting file
that is to be sent to asset manager. It stores variant which is a string
to indicate which type of file is sent, the state on asset manager and
the file_url from asset_manager. It also permits storing a superseded_by
reference for when a different image_asset replaces it. Finally there
are two unique indexes, one on file_url so we don't have two images
pointing to the same asset manager file and variant
image_file_revision_id compound so we can't store the same variant twice
for a file_revision.


db/migrate/20190117191703_create_versioned_image_asset.rb
```diff
@@ -0,0 +1,29 @@
...
+                                 on_delete: :nullify },
+                    index: false,
+                    null: true
+
+      t.string :variant, null: false
+      t.string :file_url
+      t.string :state, default: "absent", null: false
+      t.timestamps
+
+      t.index :file_url, unique: true
```
benthorner: Is this a problem if we have two absent assets, where the file_url is null?

app/models/versioned/image/file_revision.rb
```diff
@@ -0,0 +1,105 @@
...
+    belongs_to :blob, class_name: "ActiveStorage::Blob"
+
+    # rubocop:disable Rails/InverseOf
+    belongs_to :created_by,
+               class_name: "User",
+               optional: true,
+               foreign_key: :created_by_id
+    # rubocop:enable Rails/InverseOf
+
+    has_many :assets,
```
benthorner: So much nicer :-)

app/models/versioned/image/file_revision.rb
```diff
@@ -0,0 +1,105 @@
+# frozen_string_literal: true
+
+module Versioned
+  # This model stores for an Image::Revision that when changed will require
```
benthorner: Can you re-phrase this initial sentence?

app/models/versioned/image/metadata_revision.rb
```diff
@@ -0,0 +1,42 @@
+# frozen_string_literal: true
+
+module Versioned
+  # A Image::MetadataRevision stores the data for an Image::Revision that when
```
benthorner: Don't think we need to repeat the class name in the comment :-). Can the comment focus on what it does do, rather than what it doesn't?

app/models/versioned/image/revision.rb
```diff
@@ -0,0 +1,138 @@
+# frozen_string_literal: true
+
+module Versioned
+  # An image revision represents an edit of a particular image, it's data is
+  # stored across two associations: Image::FileRevisiona and
```
benthorner: typo

app/models/versioned/image/revision.rb
```diff
@@ -0,0 +1,138 @@
...
+  # Image::MetadataRevision. The file revision stores data that when changed
```
benthorner: Perhaps we can save these comments for the classes themselves?

app/models/versioned/image/revision.rb
```diff
@@ -0,0 +1,138 @@
...
+               class_name: "Versioned::Image::MetadataRevision",
+               foreign_key: :metadata_revision_id,
+               inverse_of: :revisions
+
+    delegate :alt_text,
+             :caption,
+             :credit,
+             to: :metadata_revision
+
+    delegate_missing_to :file_revision
```
benthorner: Can we specify exactly what we need to delegate?

spec/models/versioned/image/file_revision_spec.rb
```diff
@@ -0,0 +1,115 @@
...
+
+      expect(revision1.different_to?(revision2)).to be true
+    end
+
+    it "is false when content is the same and only timestamps differ" do
+      revision1 = create(:versioned_image_file_revision,
+                         crop_x: 5,
+                         created_at: 10.days.ago)
+      revision2 = revision1.dup
+      revision2.crop_x = 5
```
benthorner: Why are we doing this?

app/services/versioned/publish_service.rb
```diff
@@ -54,66 +54,74 @@ def set_first_published_at
...
 
     def retire_old_images
       return unless live_edition
 
+      current_revisions = current_edition.image_revisions
+      current_revisions_by_image_id = current_revisions.group_by(&:image_id)
+
       live_edition.image_revisions.each do |revision|
-        next if revision == current_edition.lead_image_revision
+        next if current_revisions.include?(revision)
```
benthorner: Do we need to specifically check the file revision here, otherwise won't we redirect when changing the alt text?

app/services/versioned/publish_service.rb
```diff
@@ -54,66 +54,74 @@ def set_first_published_at
...
-          revision.asset_manager_variants.each { |v| remove_image_variant(v) }
+          revision.assets.each { |a| remove_image_asset(a) }
         end
       end
     end
 
-    def remove_image_variant(variant)
-      return if variant.absent?
+    def redirect_images(old_revision, new_revision)
+      grouped_new_assets = new_revision.assets.group_by(&:variant)
```
benthorner: Just a refactoring thought: reckon it would be clearer to ask the revision for its variants, and then have a #asset(variant) method to see if each one exists, which I believe is what we're doing here with group_by/.first.

spec/services/versioned/publish_service_spec.rb
```diff
@@ -112,65 +109,67 @@
...
 
-        other_image.reload
-        expect(old_lead_image.asset_manager_variants.map(&:state).uniq)
-          .to eq(%w[live])
+        image_revision_to_keep.reload
+        expect(image_revision_to_keep.assets.map(&:state).uniq).to eq(%w[live])
       end
     end
 
     context "when the current edition has an updated version of the live editions lead image" do
```
benthorner: Is this just another 'it' of the previous context? Also, it's still referring to 'lead images' :-|.

db/migrate/20190117191703_create_versioned_image_asset.rb
kevindew: No it's not as a unique index ignores null values (which is one of my favourite features 😁)

app/models/versioned/image/revision.rb
kevindew: yes - but it is quite a lot may end up quite verbose.

spec/models/versioned/image/file_revision_spec.rb
kevindew: will remove.

app/services/versioned/publish_service.rb
kevindew: Sounds a good idea - will update.

spec/services/versioned/publish_service_spec.rb
kevindew: I'll update the previous context to make it more distinct, that one is checking we just remove the right images and this one checks we supersede images.

kevindew: Thanks re: lead image!

app/models/versioned/image/metadata_revision.rb
kevindew: Done.

app/models/versioned/image/revision.rb
kevindew: Thanks

app/models/versioned/image/revision.rb
benthorner: True, but I think this is another scenario that we could just pass the specific revision around e.g. crop.html.erb. At least expanding it out makes it more obvious to change, without the mystery of delegate_missing.

app/services/versioned/publish_service.rb
benthorner: Ah, that's nicer than I thought it would look :-).

spec/services/versioned/publish_service_spec.rb
benthorner: 👍

PR #684, opened 2019-01-18
 Replace draft column on versioned_editions

kevindew:

This replaces the string column on versioned_editions with a
publishing_api_synced column. This column is to be of most use in a
draft context where we allow state changes and edits while a document is
not on the Publishing API. Whereas once we are in the realm of live
states the Publishing API must be synced for the state to be valid.

An area of confusion this introduces is whether this still makes
semantic sense on old editions that have been replaced on Publishing
API, in theory though since to publish the next item they have been
synced so I guess it still makes sense.

### DIFF
--- a/app/models/versioned/edition.rb
+++ b/app/models/versioned/edition.rb
@@ -70,12 +70,6 @@ module Versioned
                             class_name: "Versioned::Revision",
                             join_table: "versioned_edition_revisions"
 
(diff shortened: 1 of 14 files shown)


kevindew:
Thanks @benthorner as discussed offline I've changed the column to be revision_synced as that seems better that publishing_api_synced

PR #691, opened 2019-01-21
Image filenames

benthorner:
https://trello.com/c/hAC9p5ae/534-auto-generate-unique-filename-govspeak-snippet-for-images

This adds an ImageFilenameService to ensure each uploaded image has a unique, normalised filename for its associated revision. We can then show a markdown snippet, using this filename as the ID. The snippet isn't styled or easy to copy, partly because it's unclear what the workflow is.

Showing the snippet is therefore behind a feature-flag and doesn't do anything at the moment - rendering inline image snippets will be the next piece of work.

I found some of the code a bit counter-intuitive to write e.g. `lead_image = edition.lead_image_revision` looks a bit confusing. Maybe we could do some aliasing here?


app/views/components/docs/image_meta.yml
```diff
@@ -13,17 +13,30 @@ examples:
   default:
     data:
       src: https://assets.publishing.service.gov.uk/government/assets/history/buildings/larry-the-cat-a47549e08bdbc6cd0e3e042eea943f65b7a4590d95642586e51acb44bb2dcea2.jpg
-      meta:
-        alt_text: Larry the cat
-        caption: Larry, Chief Mouser to the Cabinet Office
-        credit: Her Majesty's Government
+      alt_text: Larry the cat
```
emmabeynon: Do you know why the alt text is passed to the component twice?  I know you haven't introduced it in this change, but I'm curious as to why.

app/views/images/_image.html.erb
```diff
@@ -0,0 +1,47 @@
+<% actions = [] %>
+
+<% actions << link_to("Edit crop", crop_image_path(document, image.image_id), class: "govuk-link") %>
+<% actions << link_to("Edit details", edit_image_path(document, image.image_id), class: "govuk-link") %>
+<% actions << link_to('Download 960x640 image', download_image_path(document, image.image_id), class: "govuk-link") %>
```
emmabeynon: Just a small quotation mark inconsistency here.

app/views/images/_lead_image.html.erb
```diff
@@ -1,46 +1,42 @@
...
+</h2>
 
-<% if lead_image %>
-  <h2 class="govuk-heading-m">
-    <%= t("images.index.lead_image") %>
-  </h2>
+<% actions = [] %>
+<% actions << link_to("Edit crop", crop_image_path(document, lead_image.image_id), class: "govuk-link") %>
+<% actions << link_to("Edit details", edit_image_path(document, lead_image.image_id), class: "govuk-link") %>
+<% actions << link_to('Download 960x640 image', download_image_path(document, lead_image.image_id), class: "govuk-link") %>
```
emmabeynon: Just a small quotation mark inconsistency here.

app/services/image_filename_service.rb
```diff
@@ -0,0 +1,47 @@
...
+    @revision = revision
+  end
+
+  def call(suggested_name, mime_type)
+    filename = ActiveStorage::Filename.new(suggested_name)
+
+    base = filename.base.parameterize.slice 0...MAX_LENGTH
+    base = ensure_unique(base)
+
+    extension = Rack::Mime::MIME_TYPES.invert[mime_type]
```
kevindew: I think doing this here could have some unexpected consequences. Impact is likely low with images but could be worse for attachments.

First off the `Rack::Mime::MIME_TYPES` is a different database of mime types to what ActiveStorage uses which comes from [MimeMagic](https://github.com/minad/mimemagic)

As invert also just uses the first thing we could end up with some unlikely extensions:

E.g:

```
irb(main):019:0> Rack::Mime::MIME_TYPES[".svg"]
=> "image/svg+xml"
irb(main):018:0> Rack::Mime::MIME_TYPES.invert["image/svg+xml"]
=> ".svgz"
```

Frustratingly mime magic doesn't seem to have an aspect of the most common or preferred extension in it's tables which gives you:

```
irb(main):026:0> MimeMagic.new("image/jpeg").extensions
=> ["jpe", "jpeg", "jpg"]
```

As I feel like I'm opening up a rabbit hole here it might be worth having a quick chat to see what's simplest route forward.

benthorner: Sure - whenever you're free 👍 .

app/services/image_filename_service.rb
```diff
@@ -0,0 +1,47 @@
+# frozen_string_literal: true
+
+class ImageFilenameService
+  MAX_LENGTH = 65
```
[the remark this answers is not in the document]
benthorner: Sure - I've updated the commit that adds this service to explain it's an arbitrary length :-).

app/views/components/docs/image_meta.yml
benthorner: It's because the component renders an image, with alt_text; and some metadata lines, which happen to include the alt text. The previous version tried to magically copy them across, but that's not compatible with the new, more flexible version.

app/views/images/_lead_image.html.erb
benthorner: Fixed :-).

app/views/images/_image.html.erb
benthorner: Fixed again :-).

app/services/image_filename_service.rb
```diff
@@ -0,0 +1,47 @@
...
+private
+
+  attr_reader :revision
+
+  def image_revisions
+    @image_revisions ||= revision.image_revisions
+  end
+
+  def ensure_unique(base)
+    potential_conflicts = revision.image_revisions.map(&:filename)
```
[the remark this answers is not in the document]
benthorner: I don't think so, as a revision should fully represent all files that it has. If we make a new revision with more/fewer files, the available names will change accordingly - because the content of the document updates alongside changes to the images, the two should always be compatible.

app/services/image_filename_service.rb
benthorner: Updated to just preserve any extension that comes with the uploaded file.

app/controllers/images_controller.rb
```diff
@@ -26,10 +26,9 @@ def create
         return
       end
 
-      image_revision = ImageUploadService.new(params[:image]).call(current_user)
-
       current_edition = @document.current_edition
       current_revision = current_edition.revision
+      image_revision = ImageUploadService.new(params[:image], current_revision).call(current_user)
```
kevindew: I know this doesn't introduce this here so no biggie or blocker, but I'm pretty confused as to why some args in initialize and others in call. Is there a logic that I'm missing?

spec/services/image_filename_service_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+    end
+
+    it "copes if the file has no extension" do
+      revision = build :revision
+      name = ImageFilenameService.new(revision).call("file")
+      expect(name).to eq "file"
+    end
+
+    it "truncates lengthy base filenames" do
+      stub_const "ImageFilenameService::MAX_LENGTH", 3
```
kevindew: ooh I didn't know this method - looks handy.

PR #693, opened 2019-01-22
Rename retirement to withdrawal

emmabeynon:
For https://trello.com/c/oyI7eOlC/586-change-all-internal-references-to-retirement-withdrawal

This renames internal references to "retirement" to "withdrawal" ahead of building out the withdrawal workflow to ensure consistency and avoid future confusion.  This does not include renaming the `versioned_retirements` table, which will be renamed at a later date along with the other `versioned_` prefixed tables

### DIFF
--- a/config/locales/en/documents/history.yml
+++ b/config/locales/en/documents/history.yml
@@ -16,7 +16,7 @@ en:
         image_removed: "Removed image"
         new_edition: "New edition"
         create_preview: "Create preview"
-        retired: "Retired"
+        withdrawn: "Withdrawn"
         removed: "Removed"
         internal_note: "Internal note"
         draft_discarded: "Draft discarded"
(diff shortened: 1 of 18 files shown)


config/locales/en/documents/history.yml
```diff
@@ -16,7 +16,7 @@ en:
         image_removed: "Removed image"
         new_edition: "New edition"
         create_preview: "Create preview"
-        retired: "Retired"
+        withdrawn: "Retired"
```
kevindew: Did we decide at a product level not to change these? I just wonder if it's simpler to do them all particularly as bits of the naming creep out in URLs.

emmabeynon: No, I just thought maybe the user facing change should happen at the same time as the functionality itself.  Happy to change now though.

kevindew: Yeah, should be fine to go whole way with it.

emmabeynon: Now updated 👍

emmabeynon:
@kevindew good spot on the documentation, updated here https://github.com/alphagov/content-publisher/pull/693/commits/5d62c48b504492163d31c7be36db389cc8b71b0d

PR #699, opened 2019-01-23
Remove delegated current_edition methods from document

kevindew:

Using these didn't seem to add much value and seems to introduce some
confusion. Using a current_edition_title vs current_edition.title
doesn't save any characters.

So the main reason for these to exist is to conform to the Law of
Demeter, which we can always go back to if needed.


kevindew:
> Cool - looks like the live delegates weren't used anyway, too :-).

Yup, just there for consistency

PR #701, opened 2019-01-23
Various versioning related boolean updates

kevindew:

This PR comprises of a few updates in relation to ActiveRecord booleans.

It first adds in a `#editable?` method to Edition which returns the inverse of `#live?`.
It then uses a question mark suffix on the places we use ActiveRecord Booleans.
It removes a `#live?` method from Document so that's not confused with `#live?` on Edition.
And then updates tests for consistency.

### DIFF
--- a/app/controllers/editions_controller.rb
+++ b/app/controllers/editions_controller.rb
@@ -7,7 +7,7 @@ class EditionsController < ApplicationController
 
       current_edition = document.current_edition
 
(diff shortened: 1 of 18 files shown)


kevindew:
@benthorner yeah the .present? was just to coerce it to a boolean as it seemed weird for a `#.live?` to return a model or a nil.

PR #709, opened 2019-01-24
Rename ambiguous revision method names

kevindew:

We had two pairs of confusing method names on Revision: #status
and #statuses; and #revision and #revisions.

This attempts to make these more distinct by prefixing the many variants
with all. So we now have #all_statuses and #all_revisions.

The other option on this was to prefix the singular versions e.g. if we
had #current_status and #current_revision. However this feels
unnecessarily verbose since revision and status will be frequently
accessed whereas the collection equivalents are used for audit trails
and will be rarely used.

I'm not 100% sure the `all_` prefix is the best one, I thought a better
one for statuses would be #statuses_held but that didn't seem to fit
nicely with revisions.

### DIFF
--- a/app/models/edition.rb
+++ b/app/models/edition.rb
@@ -22,7 +22,7 @@ class Edition < ApplicationRecord
 
     # Used to keep an audit trail of all the revisions that have been
     # associated with an edition
-    revisions << revision unless revisions.include?(revision)
+    all_revisions << revision unless all_revisions.include?(revision)
   end
 
   attr_readonly :number, :document_id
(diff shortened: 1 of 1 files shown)


kevindew:
@benthorner That's a case where `statuses_held` would fit better. But there's not a good one for revisions. 

I stand by that it seems unnecessarily verbose to have current_revision and current_status when they're the frequent ones you access and the historic ones are the rarity so would much rather resolve it on the collection side.

1pretz1:
I agree with Ben in that in my mind having a 'current_' prefix is clearer and maybe worth the extra few characters typed unless a clearer name can be thought of for the collection cases.

emmabeynon:
Having spent a bit of time working with the new versioning changes in https://github.com/alphagov/content-publisher/pull/714 I found the use of `current_` really helpful in figuring out which type of edition to use, so I'd be in favour of keeping that prefix.

kevindew:
Bah! You're all wrong. :(

Will sort though. At least delegates will conceal it.

PR #714, opened 2019-01-24
Add withdraw workflow

emmabeynon:
For https://trello.com/c/GGxgDv3W/533-add-form-to-let-users-retire-a-document

- This adds a frontend interface so that users with managing editor permissions can withdraw a document.  
> withdraw_public_explanation_guidance_screen

- Users must enter a public explanation, otherwise an error will be displayed. The public explanation can contain markdown, which is converted to HTML before being sent to the Publishing API.
> withdraw_blank_public_explanation_screen

- We display an error if the Publishing API is down when the user is trying to withdraw a document or update the public explanation
> publishing_api_down

- Users without managing editor permissions will see a screen directing them to speak to their managing editor
> no_managing_editor_permissions

- After a document has been withdrawn, a notice is displayed on the document page with the time the document was withdrawn, who withdrew it, and the public explanation.  We also display a link to change the public explanation
> withdrawn_document_screen

- Clicking on the "Undo withdrawal" button takes users to a screen explaining that the feature hasn't been built yet
> unwithdraw

### DIFF
--- a/app/controllers/unwithdraw_controller.rb
+++ b/app/controllers/unwithdraw_controller.rb
@@ -1,0 +1,7 @@
+# frozen_string_literal: true
+
+class UnwithdrawController < ApplicationController
+  def index
+    @document = Document.with_current_edition.find_by_param(params[:id])
+  end
+end
--- a/app/controllers/withdraw_controller.rb
+++ b/app/controllers/withdraw_controller.rb
@@ -4,4 +4,16 @@ class WithdrawController < ApplicationController
   def new
     @document = Document.with_current_edition.find_by_param(params[:id])
   end
+
+  def create
+    Document.transaction do
+      document = Document.with_current_edition.lock!.find_by_param(params[:id])
+      public_explanation = params[:public_explanation]
+
+      #FIXME We should check that the edition is withdrawable before passing
+      # it to the UnpublishService
+      UnpublishService.new.withdraw(document.current_edition, public_explanation)
+      redirect_to document
+    end
+  end
 end
--- a/app/controllers/withdraw_controller.rb
+++ b/app/controllers/withdraw_controller.rb
@@ -6,6 +6,12 @@ class WithdrawController < ApplicationController
     edition = @document.current_edition
     @public_explanation =
       edition.withdrawn? ? edition.status.details.public_explanation : nil
+
+    if current_user.has_permission?(User::MANAGING_EDITOR_PERMISSION)
+      render :new
+    else
+      render :non_managing_editor, status: :forbidden
+    end
   end
 
   def create
--- a/app/controllers/withdraw_controller.rb
+++ b/app/controllers/withdraw_controller.rb
@@ -22,10 +22,17 @@ class WithdrawController < ApplicationController
 
         render :new
       else
-        #FIXME We should check that the edition is withdrawable before passing
-        # it to the UnpublishService
-        UnpublishService.new.withdraw(@document.current_edition, public_explanation)
-        redirect_to @document
+        begin
+          #FIXME We should check that the edition is withdrawable before passing
+          # it to the UnpublishService
+          UnpublishService.new.withdraw(@document.current_edition, public_explanation)
+          redirect_to @document
+        rescue GdsApi::BaseError => e
+          GovukError.notify(e)
+          redirect_to withdraw_path,
+            alert_with_description: t("withdraw.new.flashes.publishing_api_error"),
+            public_explanation: public_explanation
+        end
       end
     end
   end
--- a/app/controllers/withdraw_controller.rb
+++ b/app/controllers/withdraw_controller.rb
@@ -10,13 +10,23 @@ class WithdrawController < ApplicationController
 
   def create
     Document.transaction do
-      document = Document.with_current_edition.lock!.find_by_param(params[:id])
+      @document = Document.with_current_edition.lock!.find_by_param(params[:id])
       public_explanation = params[:public_explanation]
+      issues = Requirements::WithdrawalChecker.new(public_explanation).pre_withdrawal_issues
 
-      #FIXME We should check that the edition is withdrawable before passing
-      # it to the UnpublishService
-      UnpublishService.new.withdraw(document.current_edition, public_explanation)
-      redirect_to document
+      if issues.any?
+        flash["alert_with_items"] = {
+          "title" => I18n.t!("withdraw.new.flashes.requirements"),
+          "items" => issues.items,
+        }
+
+        render :new
+      else
+        #FIXME We should check that the edition is withdrawable before passing
+        # it to the UnpublishService
+        UnpublishService.new.withdraw(@document.current_edition, public_explanation)
+        redirect_to @document
+      end
     end
   end
 end
--- a/app/controllers/withdraw_controller.rb
+++ b/app/controllers/withdraw_controller.rb
@@ -3,6 +3,9 @@
 class WithdrawController < ApplicationController
   def new
     @document = Document.with_current_edition.find_by_param(params[:id])
+    edition = @document.current_edition
+    @public_explanation =
+      edition.withdrawn? ? edition.status.details.public_explanation : nil
   end
 
   def create
--- a/app/controllers/withdraw_controller.rb
+++ b/app/controllers/withdraw_controller.rb
@@ -25,7 +25,7 @@ class WithdrawController < ApplicationController
         begin
           #FIXME We should check that the edition is withdrawable before passing
           # it to the UnpublishService
-          UnpublishService.new.withdraw(@document.current_edition, public_explanation)
+          UnpublishService.new.withdraw(@document.current_edition, public_explanation, current_user)
           redirect_to @document
         rescue GdsApi::BaseError => e
           GovukError.notify(e)
--- a/app/controllers/withdraw_controller.rb
+++ b/app/controllers/withdraw_controller.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
-class UnpublishController < ApplicationController
-  def remove
+class WithdrawController < ApplicationController
+  def new
     @document = Document.with_current_edition.find_by_param(params[:id])
   end
 end
--- a/app/controllers/withdraw_controller.rb
+++ b/app/controllers/withdraw_controller.rb
@@ -7,6 +7,11 @@ class WithdrawController < ApplicationController
     @public_explanation =
       edition.withdrawn? ? edition.status.details.public_explanation : nil
 
+    if !current_user.has_permission?(User::PRE_RELEASE_FEATURES_PERMISSION)
+      render :withdraw
+      return
+    end
+
     if current_user.has_permission?(User::MANAGING_EDITOR_PERMISSION)
       render :new
     else
@@ -15,6 +20,12 @@ class WithdrawController < ApplicationController
   end
 
   def create
+    unless user_has_permissions?
+      # FIXME: this shouldn't be an exception but we've not worked out the
+      # right response - maybe bad request or a redirect with flash?
+      raise "Can't withdraw an edition without permissions"
+    end
+
     Document.transaction do
       @document = Document.with_current_edition.lock!.find_by_param(params[:id])
       public_explanation = params[:public_explanation]
@@ -41,5 +52,12 @@ class WithdrawController < ApplicationController
         end
       end
     end
+  end
+
+private
+
+  def user_has_permissions?
+    current_user.has_all_permissions?([User::MANAGING_EDITOR_PERMISSION,
+                                       User::PRE_RELEASE_FEATURES_PERMISSION])
   end
 end
--- a/app/services/unpublish_service.rb
+++ b/app/services/unpublish_service.rb
@@ -9,7 +9,7 @@ class UnpublishService
       GdsApi.publishing_api_v2.unpublish(
         edition.content_id,
         type: "withdrawal",
-        explanation: public_explanation,
+        explanation: format_govspeak(public_explanation),
         locale: edition.locale,
       )
 
@@ -97,6 +97,10 @@ private
     if document.current_edition != document.live_edition
       raise "Publishing API does not support unpublishing while there is a draft"
     end
+  end
+
+  def format_govspeak(text)
+    GovspeakDocument.new(text).to_html
   end
 
   def delete_assets(edition)
--- a/app/services/unpublish_service.rb
+++ b/app/services/unpublish_service.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 class UnpublishService
-  def withdraw(edition, explanatory_note)
+  def withdraw(edition, public_explanation)
     Document.transaction do
       edition.document.lock!
       check_unpublishable(edition)
@@ -9,11 +9,11 @@ class UnpublishService
       GdsApi.publishing_api_v2.unpublish(
         edition.content_id,
         type: "withdrawal",
-        explanation: explanatory_note,
+        explanation: public_explanation,
         locale: edition.locale,
       )
 
-      withdrawal = Withdrawal.new(explanatory_note: explanatory_note)
+      withdrawal = Withdrawal.new(public_explanation: public_explanation)
 
       edition.assign_status(:withdrawn, nil, status_details: withdrawal)
       edition.save!
--- a/app/services/unpublish_service.rb
+++ b/app/services/unpublish_service.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 class UnpublishService
-  def withdraw(edition, public_explanation)
+  def withdraw(edition, public_explanation, user = nil)
     Document.transaction do
       edition.document.lock!
       check_unpublishable(edition)
@@ -15,7 +15,7 @@ class UnpublishService
 
       withdrawal = Withdrawal.new(public_explanation: public_explanation)
 
-      edition.assign_status(:withdrawn, nil, status_details: withdrawal)
+      edition.assign_status(:withdrawn, user, status_details: withdrawal)
       edition.save!
 
       TimelineEntry.create_for_status_change(
--- a/app/services/unpublish_service.rb
+++ b/app/services/unpublish_service.rb
@@ -2,16 +2,10 @@
 
 class UnpublishService
   def withdraw(edition, public_explanation, user = nil)
-    Document.transaction do
+    Document.transaction(requires_new: true) do
       edition.document.lock!
       check_unpublishable(edition)
 
-      GdsApi.publishing_api_v2.unpublish(
-        edition.content_id,
-        type: "withdrawal",
-        explanation: format_govspeak(public_explanation),
-        locale: edition.locale,
-      )
 
       withdrawal = Withdrawal.new(public_explanation: public_explanation)
 
@@ -23,11 +17,18 @@ class UnpublishService
         status: edition.status,
         details: withdrawal,
       )
+
+      GdsApi.publishing_api_v2.unpublish(
+        edition.content_id,
+        type: "withdrawal",
+        explanation: format_govspeak(public_explanation),
+        locale: edition.locale,
+      )
     end
   end
 
   def remove(edition, explanatory_note: nil, alternative_path: nil)
-    Document.transaction do
+    Document.transaction(requires_new: true) do
       edition.document.lock!
       check_unpublishable(edition)
 
@@ -56,7 +57,7 @@ class UnpublishService
   end
 
   def remove_and_redirect(edition, redirect_path, explanatory_note: nil)
-    Document.transaction do
+    Document.transaction(requires_new: true) do
       edition.document.lock!
       check_unpublishable(edition)
 
--- a/app/views/documents/show/_withdrawn_notice.html.erb
+++ b/app/views/documents/show/_withdrawn_notice.html.erb
@@ -1,0 +1,10 @@
+<%= render "govuk_publishing_components/components/notice", {
+  title: I18n.t("documents.show.withdrawn.title",
+                document_type: @document.document_type.label.downcase,
+                withdrawn_date: @edition.status.created_at.strftime("%d %B %Y"))
+} do %>
+  <%= @edition.status.details.public_explanation %>
+  <p>
+  <p>
+  <%= link_to "Change public explanation", withdraw_path(@document), class: "govuk-link govuk-link--no-visited-state" %>
+<% end %>
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -38,6 +38,8 @@ Rails.application.routes.draw do
 
   get "/documents/:id/withdraw" => "withdraw#new", as: :withdraw
   post "/documents/:id/withdraw" => "withdraw#create"
+  get "/documents/:id/unwithdraw" => "unwithdraw#index", as: :unwithdraw
+
   get "/documents/:id/remove" => "unpublish#remove", as: :remove
 
   get "/documents/:document_id/images" => "images#index", as: :images
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -37,6 +37,7 @@ Rails.application.routes.draw do
   post "/documents/:id/create-preview" => "preview#create", as: :create_preview
 
   get "/documents/:id/withdraw" => "withdraw#new", as: :withdraw
+  post "/documents/:id/withdraw" => "withdraw#create"
   get "/documents/:id/remove" => "unpublish#remove", as: :remove
 
   get "/documents/:document_id/images" => "images#index", as: :images
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -36,7 +36,7 @@ Rails.application.routes.draw do
   get "/documents/:id/preview" => "preview#show", as: :preview_document
   post "/documents/:id/create-preview" => "preview#create", as: :create_preview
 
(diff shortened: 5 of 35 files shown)


app/views/documents/show/_withdrawn_notice.html.erb
```diff
@@ -0,0 +1,10 @@
+<%= render "govuk_publishing_components/components/notice", {
+  title: I18n.t("documents.show.withdrawn.title",
+                document_type: @document.document_type.label.downcase,
+                withdrawn_date: @edition.status .created_at.strftime("%d %B %Y"))
+} do %>
+  <%= @edition.status.details.public_explanation %>
+  <p>
```
alex-ju: As discussed, we can replace these two paragraphs with `<hr class="govuk-section-break--m">` (or `--l` modifier if we need more than space).

emmabeynon: Super, thank you 🙏

emmabeynon: As discussed, this element is being overridden so we're going to stick with two `<p>` tags for now.  @alex-ju to add to the backlog.

alex-ju: It's not working within govspeak at the moment because of a CSS overwrite in govspeak, I'll add this to a card to look into clashes between govspeak styles and govuk-frontend.

spec/factories/edition_factory.rb
```diff
@@ -87,5 +87,15 @@
...
       end
     end
+
+    trait :withdrawn do
+      summary { SecureRandom.alphanumeric(10) }
+      live { true }
+
+      after(:build) do |edition, evaluator|
+        edition.revision = evaluator.association(:revision)
+        edition.status = create(:withdrawal).status
```
kevindew: We should avoid using a create here as that will always do a db query.

I'd suggest the most consistent way to do what you're aiming for here though would be to have a withdrawn trait on the status factory, so something like:

```
trait :withdrawn do
  state { "withdrawn" }
  association :details, factory: :withdrawal
end
```

And then you can probably create your edition in your test with `create(:edition, live: true, status: build(:status, :withdrawal))`

Or if we have it in edition factory still, the trait could do 

```
      after(:build) do |edition, evaluator|
        edition.status = evaluator.association(
          :status,
          :withdrawal,
          created_by: edition.created_by,
        )
      end
```

we shouldn't need the revision part in this factory by the way.

Also looking at the polymorphic section on https://github.com/thoughtbot/factory_bot/blob/master/GETTING_STARTED.md#associations made me think there's probably some other ways we can clean up this edition factory.

spec/features/workflow/withdraw_document_publishing_api_down_spec.rb
```diff
@@ -3,6 +3,7 @@
 RSpec.feature "Withdraw a document when Publishing API is down" do
   scenario do
     given_there_is_a_published_document
+    given_i_have_the_managing_editor_permission
```
kevindew: should this be "and_i_have..."

spec/features/managing_editor_spec.rb
```diff
@@ -0,0 +1,73 @@
...
+
+    given_there_is_a_withdrawn_document
+    when_i_dont_have_the_managing_editor_permission
+    and_i_visit_the_withdrawn_document_page
+    and_i_click_on_change_public_explanation
+    then_i_see_a_message_to_ask_my_managing_editor_to_withdraw_content
+    when_i_do_have_the_managing_editor_permission
+    and_i_visit_the_withdrawn_document_page
+    and_i_click_on_change_public_explanation
+    then_i_can_see_a_public_explanation_form
```
kevindew: This looks like it's two scenarios here. We have an example of this already in: https://github.com/alphagov/content-publisher/blob/25bdfc25a19ad86cee8f0c422a7b9d94445e474a/spec/features/workflow/edition_spec.rb

I also wonder if we need to check that things do work when they have the permission as doesn't that mean we then replicate tests we already have?

app/controllers/unwithdraw_controller.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class UnwithdrawController < ApplicationController
+  def index
+    @document = Document.find_by_param(params[:id])
```
kevindew: We'll want to use with_current_edition to check that this is actually a usable document.

app/controllers/withdraw_controller.rb
```diff
@@ -36,4 +43,11 @@ def new
       redirect_to document
     end
   end
+
+private
+
+  def user_has_permissions?
+    current_user.has_all_permissions?([User::MANAGING_EDITOR_PERMISSION,
+                                       User::PRE_RELEASE_FEATURES_PERMISSION])
```
kevindew: Does this method do anything if the user doesn't have the permissions it looks like it would return a boolean?

db/schema.rb
```diff
@@ -173,7 +173,7 @@
   end
 
   create_table "versioned_retirements", force: :cascade do |t|
-    t.string "explanatory_note"
+    t.string "public_explanation"
```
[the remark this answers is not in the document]
emmabeynon: You're right - an error is flagged in the UI if a public explanation is not provided, but makes sense to enforce it in the database as well.

spec/features/workflow/withdraw_document_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+  scenario do
+    given_there_is_a_published_document
+    when_i_visit_the_document_page
+    and_i_click_on_withdraw
+    then_i_see_that_i_can_withdraw_the_document
+  end
+
+  def given_there_is_a_published_document
+    @document = create(:document, :with_live_edition)
+  end
```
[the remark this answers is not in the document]
emmabeynon: Yep, I noticed that PR after I raised this one 😄 I'm working through changing them all.

app/controllers/unpublish_controller.rb
```diff
@@ -4,8 +4,4 @@ class UnpublishController < ApplicationController
   def remove
     @document = Document.with_current_edition.find_by_param(params[:id])
   end
-
-  def withdraw
-    @document = Document.with_current_edition.find_by_param(params[:id])
-  end
```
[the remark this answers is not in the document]
emmabeynon: In d0e1620e35b6daa868e8a806e898e79ea032695b I've created a view for users without pre release permissions.  I figured it would be simpler to maintain one withdraw route than two and avoid adding another conditional in the actions partial.

app/controllers/withdraw_controller.rb
```diff
@@ -0,0 +1,53 @@
+# frozen_string_literal: true
+
+class WithdrawController < ApplicationController
+  before_action :user_has_permissions?, only: :new
+
+  def create
+    @document = Document.with_current_edition.find_by_param(params[:id])
+    @public_explanation = @document.current_edition.status.details&.public_explanation || params[:public_explanation]
```
benthorner: Would it be possible to split these dots into a variable or two?

app/controllers/withdraw_controller.rb
```diff
@@ -0,0 +1,53 @@
...
+      render :create
+    else
+      render :non_managing_editor
+    end
+  end
+
+  def new
+    document = Document.with_current_edition.find_by_param(params[:id])
+    public_explanation = params[:public_explanation]
+    issues = Requirements::PublicExplanationChecker.new(public_explanation).pre_withdrawal_issues
```
benthorner: Ideally we should name the requirements checkers after the feature, so it's clear where they're being used e.g. 'UnwithdrawChecker' or something.

app/controllers/withdraw_controller.rb
```diff
@@ -4,4 +4,11 @@ class WithdrawController < ApplicationController
   def create
     @document = Document.with_current_edition.find_by_param(params[:id])
   end
+
+  def new
+    document = Document.with_current_edition.find_by_param(params[:id])
+    public_explanation = params[:public_explanation]
+    UnpublishService.new.withdraw(document.current_edition, public_explanation)
+    redirect_to document
```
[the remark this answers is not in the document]
emmabeynon: The Unpublish Service checks that the edition is suitable for unpublishing so should be covered there https://github.com/alphagov/content-publisher/blob/bcdc97ebcb7aa6a6ce9d5c2467a9f71c6bad5430/app/services/unpublish_service.rb#L90

app/controllers/withdraw_controller.rb
```diff
@@ -3,6 +3,7 @@
 class WithdrawController < ApplicationController
   def create
     @document = Document.with_current_edition.find_by_param(params[:id])
+    @public_explanation = @document.current_edition.status.details&.public_explanation
```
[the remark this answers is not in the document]
emmabeynon: The intention here is that a public explanation would be shown in the text area if it's present.  I hadn't realised that details were optional for a status, so I'll change to `@document.current_edition.status&.details&.public_explanation`.

app/views/documents/show/_withdrawn_notice.html.erb
```diff
@@ -0,0 +1,10 @@
+<%= render "govuk_publishing_components/components/notice", {
+  title: I18n.t("documents.show.withdrawn.title",
+                document_type: @document.document_type.label.downcase,
+                withdrawn_date: @edition.status .created_at.strftime("%d %B %Y"))
+} do %>
+  <%= @edition.status.details.public_explanation %>
```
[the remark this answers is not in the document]
emmabeynon: We are showing this as raw govspeak as I figured there should be consistency with how we display body content.

app/controllers/withdraw_controller.rb
```diff
@@ -4,6 +4,11 @@ class WithdrawController < ApplicationController
   def create
     @document = Document.with_current_edition.find_by_param(params[:id])
     @public_explanation = @document.current_edition.status.details&.public_explanation || params[:public_explanation]
+    if current_user.has_permission?(User::MANAGING_EDITOR_PERMISSION)
+      render :create
+    else
+      render :non_managing_editor
```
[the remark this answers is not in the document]
emmabeynon: We wanted to display some guidance to non managing editors to direct them to speak to their managing editor, so I don't think returning a forbidden status code would be right here.

app/controllers/withdraw_controller.rb
emmabeynon: I've changed the new action to render a 403 status code if the user doesn't have these permissions.

app/controllers/withdraw_controller.rb
```diff
@@ -19,7 +19,15 @@ def new
 
       redirect_to withdraw_path, public_explanation: public_explanation
     else
-      UnpublishService.new.withdraw(document.current_edition, public_explanation)
+      begin
+        UnpublishService.new.withdraw(document.current_edition, public_explanation)
+      rescue GdsApi::BaseError => e
```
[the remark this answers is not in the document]
emmabeynon: Should we also ensure that the DB operations are rolled back if the Publishing API fails?  With the new ordering we end up in a situation where an edition is withdrawn is Content Publisher, but not in Publishing API.

app/controllers/withdraw_controller.rb
emmabeynon: Change to `WithdrawalChecker`

app/controllers/withdraw_controller.rb
```diff
@@ -3,13 +3,24 @@
...
-    redirect_to document
+    issues = Requirements::PublicExplanationChecker.new(public_explanation).pre_withdrawal_issues
+
+    if issues.any?
+      flash["alert_with_items"] = {
+        "title" => I18n.t!("withdraw.create.flashes.requirements"),
+        "items" => issues.items,
+      }
+
+      redirect_to withdraw_path, public_explanation: public_explanation
```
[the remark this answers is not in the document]
emmabeynon: Now rendering :create instead.

app/controllers/withdraw_controller.rb
```diff
@@ -0,0 +1,56 @@
+# frozen_string_literal: true
+
+class WithdrawController < ApplicationController
+  def create
+    @document = Document.with_current_edition.find_by_param(params[:id])
+    edition = @document.current_edition
+    @public_explanation = edition.status&.details&.public_explanation || params[:public_explanation]
```
kevindew: I don't think we need params[:public_explanation] anymore.

app/controllers/withdraw_controller.rb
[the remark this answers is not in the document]
kevindew: The db operations should roll back by an unhandled exception occurring, but we should check this.

app/views/documents/show/_withdrawn_notice.html.erb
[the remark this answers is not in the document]
kevindew: Can we check in with Sonia on this with a demo of it having a link?

app/controllers/withdraw_controller.rb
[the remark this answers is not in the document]
kevindew: We should be able to do both, worth confirming in integration that it works.

app/controllers/withdraw_controller.rb
```diff
@@ -0,0 +1,56 @@
...
+
+    if current_user.has_permission?(User::MANAGING_EDITOR_PERMISSION)
+      render :create
+    else
+      render :non_managing_editor
+    end
+  end
+
+  def new
+    render status: :forbidden if !user_has_permissions?
```
kevindew: I'm not sure what this is rendering? is it the new view? 

It might be more consistent with our app to have an exception with a fixme comment. See - editions and review controllers for example.

app/controllers/withdraw_controller.rb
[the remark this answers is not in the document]
emmabeynon: I've checked this locally and the database operations don't complete, so I should just need to update my tests 👍

app/views/documents/show/_withdrawn_notice.html.erb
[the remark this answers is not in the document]
emmabeynon: Confirmed with Sonia this showing raw govspeak is correct.

app/controllers/withdraw_controller.rb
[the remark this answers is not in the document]
emmabeynon: I had a think about this - is this something we could add to the Withdrawal checker and display an error on the page if the edition is not suitable for withdrawing?

app/controllers/withdraw_controller.rb
emmabeynon: The new action now raises an error if the user doesn't have these permissions.

app/controllers/withdraw_controller.rb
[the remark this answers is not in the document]
emmabeynon: We now check that the edition's status is withdrawn before attempting to access its details and public explanation.

app/controllers/withdraw_controller.rb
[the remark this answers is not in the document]
emmabeynon: Now returning 403 along with the view.

app/controllers/withdraw_controller.rb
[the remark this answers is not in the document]
kevindew: @emmabeynon could you elaborate on this a bit? I'm wondering if you mean a catching an exception or not?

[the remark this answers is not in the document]
emmabeynon: I was thinking instead of catching an exception, displaying the error to the user like we do if there's a blank public explanation.  Raising an exception might be simpler though.

config/routes.rb
```diff
@@ -37,6 +37,7 @@
   post "/documents/:id/create-preview" => "preview#create", as: :create_preview
 
   get "/documents/:id/withdraw" => "withdraw#create", as: :withdraw
+  post "/documents/:id/withdraw" => "withdraw#new", as: :new_withdrawal
```
1pretz1: May be being pedantic but shouldn't these two routes have each others action. e.g the get request for the form should be linked to the `new` action and the `post` request should be linked to the `create` action in the withdraw controller.

app/controllers/withdraw_controller.rb
emmabeynon: Now raising an exception if user does not have permissions.

config/routes.rb
emmabeynon: That Rails book is coming in handy! Good catch 👍

app/controllers/withdraw_controller.rb
```diff
@@ -0,0 +1,68 @@
+# frozen_string_literal: true
+
+class WithdrawController < ApplicationController
+  def create
+    @document = Document.with_current_edition.find_by_param(params[:id])
+    edition = @document.current_edition
+    @public_explanation =
+      edition.withdrawn? ? edition.status&.details&.public_explanation : nil
```
kevindew: Minor - but I think we can remove the safe navigation operators here as we should expect a withdrawn document to have a details and public explanation and seems reasonable to throw exception if not

app/controllers/withdraw_controller.rb
```diff
@@ -0,0 +1,68 @@
...
+      if issues.any?
+        flash["alert_with_items"] = {
+          "title" => I18n.t!("withdraw.create.flashes.requirements"),
+          "items" => issues.items,
+        }
+
+        render :create
+      else
+        begin
+          raise "Cannot withdraw an edition that is not live" unless withdrawable?(@document)
```
kevindew: I think as this is essentially same check we do in UnpublishService with same result in UnpublishService this is superfluous.

I think my previous comments are the reason this ended up here, so firstly, Sorry! and secondly,  it'd make sense for us to do something better - which we should eventually do.

Do we want something like this on the `create` method? maybe even just a comment that says `# FIXME should check document is withdrawable`?

app/controllers/withdraw_controller.rb
```diff
@@ -0,0 +1,68 @@
...
+
+          UnpublishService.new.withdraw(@document.current_edition, public_explanation, current_user)
+        rescue GdsApi::BaseError => e
+          GovukError.notify(e)
+          redirect_to withdraw_path,
+            alert_with_description: t("withdraw.create.flashes.publishing_api_error"),
+            public_explanation: public_explanation
+          return
+        end
+        redirect_to @document
```
kevindew: super minor but might want to have this inside the begin end then you can skip the return inside the rescue.

config/routes.rb
kevindew: It looks like this didn't actually get resolved? 

Also I don't think we need the `as: :new_withdrawal` as we could use `withdraw_path(@document)` in the place we call that as it's the same path.

db/migrate/20190130100821_rename_withdrawal_explanatory_note_to_public_explanation.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+
+class RenameWithdrawalExplanatoryNoteToPublicExplanation < ActiveRecord::Migration[5.2]
+  def change
+    rename_column :withdrawals, :explanatory_note, :public_explanation
+    change_column_null :withdrawals, :public_explanation, false
```
kevindew: ⭐️

spec/factories/withdrawal_factory.rb
```diff
@@ -2,6 +2,7 @@
 
 FactoryBot.define do
   factory :withdrawal do
-    explanatory_note { SecureRandom.alphanumeric }
+    public_explanation { SecureRandom.alphanumeric }
+    association :status, state: :withdrawn
```
kevindew: I don't think we want this status association here as it's not the owner of status. 

Consequences of this are:

When we do `FactoryBot.create(:status, :withdrawn)` a Status (id: 1) is created, then a Withdrawal is created via this factory, which then creates it's own Status (id: 2). So we end up with an extra status and Withdrawal being pointed at by 2 objects.

Simplest suggestion would be to just remove this from this factory, whereas more complex approaches would be to use the `after(:build)` block or to battle with `inverse_of` - but I doubt we're using this currently.

spec/features/workflow/withdraw_spec.rb
```diff
@@ -0,0 +1,86 @@
...
+    when_i_visit_the_summary_page
+    and_i_click_on_withdraw
+    then_i_see_that_i_can_withdraw_the_document
+    when_i_fill_in_the_public_explanation
+    and_click_on_withdraw_document
+    then_i_see_the_document_has_been_withdrawn
+
+    when_i_click_to_change_the_public_explanation
+    then_i_can_see_the_existing_public_explanation
+    and_i_can_edit_the_public_explanation
```
kevindew: I'd suggest that editing a withdrawal is a separate feature that we'd want to test rather than grouping them.

spec/features/managing_editor_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+    and_i_click_on_withdraw
+    then_i_see_a_message_to_ask_my_managing_editor_to_withdraw_content
+  end
+
+  scenario "withdrawn edition" do
+    when_there_is_a_withdrawn_edition
+    and_i_visit_the_withdrawn_document_summary_page
+    and_i_click_on_change_public_explanation
+    then_i_see_a_message_to_ask_my_managing_editor_to_withdraw_content
+  end
```
kevindew: Hmm since there's only one thing in common between these (the `given_i_dont_have_the_managing_editor_permission`) it might not actually be that nice that these share the same file. 

I think it's also a little counter intuitive that this is called managing editior permissions when it's all about the lack of them.

It doesn't seem immediately obvious what the best way to solve this is that is consistent with our codebase. 

I'd probably suggest that the most consistent way is "withdraw a document without managing editor permission" and "edit a withdrawal explanation without managing editor permission". Does feel like we end up with an awful lot of feature tests doing that but that probably plays into the wall of pain card of: https://trello.com/c/sQ7Go4dm/570-hard-to-work-out-how-to-test-off-the-happy-path

app/controllers/withdraw_controller.rb
emmabeynon: Sorry, not quite sure what I need to do here - just leave a comment and not raise?

kevindew: > just leave a comment and not raise?

Yup 👍

app/views/withdraw/create.html.erb
```diff
@@ -0,0 +1,43 @@
...
+    <%= link_to "Cancel", document_path(@document), class: "govuk-link govuk-link--no-visited-state" %>
+  </div>
+  <div class="govuk-grid-column-one-third">
+    <%= render "components/contextual_guidance", {
+      id: "withdrawal-public-explanation-guidance",
+      title: t("withdraw.create.public_explanation.guidance_title"),
+    } do %>
+      <p>Explain to users why the page has been withdrawn. You can link to alternative content with markdown.</p>
+      <pre class="app-c-markdown-guidance__code-snippet">
+        [linktext](http://www.gov.uk/example)
```
kevindew: painfully minor point, this should be https://

app/controllers/withdraw_controller.rb
emmabeynon: I can't do that because I get a Double Render error.

app/views/documents/show/_withdrawn_notice.html.erb
```diff
@@ -0,0 +1,10 @@
+<%= render "govuk_publishing_components/components/notice", {
+  title: I18n.t("documents.show.withdrawn.title",
+                document_type: @document.document_type.label.downcase,
+                withdrawn_date: @edition.status.created_at.strftime("%d %B %Y"))
+} do %>
+  <%= @edition.status.details.public_explanation %>
+  <p>
+  <p>
+  <%= link_to "Change public explanation", withdraw_path(@document), class: "govuk-link govuk-link--no-visited-state" %>
```
kevindew: Would we be able to move this outside of the govspeak block? the govspeak overrides the `govuk-link--no-visited-state`

see:

screen shot 2019-01-30 at 15 10 58

kevindew:
Testing on integration went good, only things I spotted were the two comments I raised about https and the no-visited link. Everything else went great 👍

I've also added managing_editor as a supported permission in integration and prod

app/controllers/withdraw_controller.rb
kevindew: I assume you get double render error if the redirect_to isn't moved to inside the begin?

spec/features/managing_editor_spec.rb
emmabeynon: All the other tests are updated to include the managing editor permissions, so maybe that's enough to test that the permission works, and we scrap this file?

app/views/documents/show/_withdrawn_notice.html.erb
kevindew: As discussed @alex-ju is planning to raise a card about this. Issue seems to be notice component only expects govspeak and not HTML: https://github.com/alphagov/govuk_publishing_components/blob/master/app/views/govuk_publishing_components/components/_notice.html.erb

app/services/unpublish_service.rb
```diff
@@ -2,16 +2,10 @@
 
 class UnpublishService
   def withdraw(edition, public_explanation, user = nil)
-    Document.transaction do
+    Document.transaction(requires_new: true) do
```
benthorner: Great spot - that was an interesting read! Hopefully we can refactor our controllers/services so that we're less likely to get caught-out by this sort of thing - it seems pretty niche.

PR #717, opened 2019-01-24
Add ADR to describe our DB deletion policy

benthorner:
https://trello.com/c/5JMGsuBj/581-follow-up-work-from-versioning-review

### DIFF
--- a/docs/adr/000x-restricted-deletion.md
+++ b/docs/adr/000x-restricted-deletion.md
@@ -1,0 +1,44 @@
+# X. Restricted Deletion
+
+Date: 2019-01-24
+
+## Context
+
+We use foreign keys to help enforce database integrity e.g.
+
+```
+# ensure creator_id on a document matches a row in the users table
+# when we try to delete a user, raise an error if its the creator of a document
+add_foreign_key "documents", "users", column: "creator_id", on_delete: :restrict
+```
+
+There are other `on_delete` behaviours instead of `restrict`.
+
+```
+# when we delete a user, set creator_id to NULL on all of its documents
+add_foreign_key "documents", "users", column: "creator_id", on_delete: :nullify
+
+# when we delete a user, also delete all of its document
+add_foreign_key "documents", "users", column: "creator_id", on_delete: :cascade
+```
+
+Different options are appropriate depending on the foreign key.
+
+   * `restrict` is a good choice in a child/parent scenario, where deleting a child should be restricted if it would make the parent invalid. For example, we should not be able to delete an ActiveStorage blob if its parent image still exists.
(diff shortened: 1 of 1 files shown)


docs/adr/000x-restricted-deletion.md
```diff
@@ -0,0 +1,43 @@
...
+
+> We may want to delete data in our development environments, but this can also be achieved by resetting the database. We should avoid making changes to the production schema for the convience of development.
+
+## Status
+
+Accepted
+
+## Consequences
+
+Ensure all existing foreign keys are set to `restrict` on deletion.
```
1pretz1: So with this in mind, it means we cant delete a user if they wanted to be removed from the db if they had created documents?

benthorner: Yup - we don't have a concrete use-case for doing that right now. If and when it comes along, then changing these will make more sense, as we'll be able to refer to an example.

docs/adr/000x-restricted-deletion.md
```diff
@@ -0,0 +1,43 @@
...
+
+Different options are appropriate depending on the foreign key.
+
+   * `restrict` is a good choice in a child/parent scenario, where deleting a child should be restricted if it would make the parent invalid. For example, we should not be able to delete an ActiveStorage blob if its parent image still exists.
+   * `nullify` works well if the relationship is optional. For example, our code doesn't currently require a document to have a creator, so it might be reasonable to `creator_id` on document to NULL if we delete a user for some reason.
+   * `cascade` is a good choice in a parent/child scenario, where it makes sense to delete the child when we delete the parent. For example, if we delete a document for some reason, it also makes sense to delete the editions.
+
+## Decision
+
+We do not expect to delete anything from the database. Therefore, we should use `restrict` on all foreign key delete actions. `restrict` is actually the default, but we should continue to specify it explicitly to indicate a conscious decision.
```
kevindew: I'd maybe rephrase "We do not expect to delete anything from the database" to something along the lines of: "We haven't anticipated any real world scenarios that will require us to delete records from the database."

docs/adr/000x-restricted-deletion.md
```diff
@@ -0,0 +1,43 @@
...
+
+> We may want to delete data in our development environments, but this can also be achieved by resetting the database. We should avoid making changes to the production schema for the convience of development.
```
kevindew: I'd probably put this as a consequence rather than part of the decision and phrase it more that deleting stuff in a development or test environment will be difficult and thus resetting the database is encouraged as an alternative.

benthorner:
@kevindew I've updated to expand on the consequences. Could you take another look?

PR #718, opened 2019-01-24
Remove old draft assets from Asset Manager when they are no longer used

kevindew:

This uses the point of creating a preview as a point to delete assets
from Asset Manager which are no longer used.

When an edition is sent to the preview service the previous revision
assets are compared with the current revision and any that are in a
draft state are then deleted from Asset Manager.

This doesn't feel like the most robust of solutions as if assets are not
removed at this point they will remain in draft forever. The approach
that would resolve that would be to have some form of scheduled task
that performs a cleanup of assets periodically. However since we don't
have prior art currently in scheduled tasks on this app I have not
implemented it.

### DIFF
--- a/spec/services/preview_service_spec.rb
+++ b/spec/services/preview_service_spec.rb
@@ -1,0 +1,126 @@
+# frozen_string_literal: true
+
+RSpec.describe PreviewService do
+  let(:draft_asset_cleanup_service) do
+    instance_double("DraftAssetCleanupService", call: nil)
+  end
+
+  before do
+    stub_any_publishing_api_put_content
+    stub_any_asset_manager_call
+    allow(DraftAssetCleanupService).to receive(:new) { draft_asset_cleanup_service }
+  end
+
+  describe "#create_preview" do
+    it "updates the Publishing API" do
+      edition = create(:edition)
+      request = stub_publishing_api_put_content(edition.content_id, {})
+      PreviewService.new(edition).create_preview
+      expect(request).to have_been_requested
+    end
+
+    it "marks the edition as 'revision_synced'" do
+      edition = create(:edition, revision_synced: false)
+      PreviewService.new(edition).create_preview
+      expect(edition.reload.revision_synced).to be(true)
+    end
+
+    it "delegates cleaning up draft assets" do
+      edition = create(:edition)
+      expect(draft_asset_cleanup_service).to receive(:call).with(edition)
+      PreviewService.new(edition).create_preview
+    end
+
+    context "when there are assets that aren't on Asset Manager" do
+      it "uploads the assets to Asset Manager" do
+        image_revision = create(:image_revision)
+        edition = create(:edition, image_revisions: [image_revision])
+
+        request = stub_asset_manager_receives_an_asset
+        PreviewService.new(edition).create_preview
+
+        expect(request).to have_been_requested.at_least_once
+        expect(image_revision.assets.map(&:state).uniq).to match(%w[draft])
+      end
+    end
+
+    context "when there are assets that are on Asset Manager" do
+      it "doesn't upload the assets to asset manager" do
+        image_revision1 = create(:image_revision, :on_asset_manager)
+        image_revision2 = create(:image_revision, :on_asset_manager, state: :live)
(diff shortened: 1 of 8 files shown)


kevindew:
@benthorner following on from our IRL chat on Friday.

Like discussed, a scheduled task is out of the scope from this card. May well prioritise once we have some scheduling art off the back of scheduling work.

I've put together the DraftAssetCleanupService as discussed and have slimmed down the tests for `create_preview` and `try_create_preview`. There's still a bit of repetition since the try_create_preview is close to being `create_preview` with a boolean (although not quite because of the `has_issues?` inconsistency)

I also moved around the exception catching to be done at method level.

spec/services/preview_service_spec.rb
```diff
@@ -0,0 +1,117 @@
...
+        expect { PreviewService.new(edition).try_create_preview }
+          .not_to raise_error
+        expect(edition.revision_synced).to be(false)
+      end
+    end
+
+    context "when there are pre-preview issues" do
+      let(:edition) { create(:edition, title: "", revision_synced: true) }
+
+      it "marks the edition as revision not synced and doesn't sent to the Publishing API" do
```
benthorner: I think this would be a bit easier to read it the 'and's were separate 'it's. We seem to do that pretty consistently everywhere else. What do you think?

spec/services/preview_service_spec.rb
```diff
@@ -0,0 +1,117 @@
...
+      expect(request).to have_been_requested
+    end
+
+    it "marks the edition as revision synced" do
+      edition = create(:edition, revision_synced: false)
+      PreviewService.new(edition).create_preview
+      expect(edition.reload.revision_synced).to be(true)
+    end
+
+    it "cleans up draft assets" do
```
benthorner: There's a slight inconsistency between this and the 'try_create_preview' description. How about 'delegates cleaning up draft assets', just to make the content of the test more quickly identifiable?

spec/services/preview_service_spec.rb
```diff
@@ -0,0 +1,117 @@
...
+
+  describe "#create_preview" do
+    it "updates the Publishing API" do
+      edition = create(:edition)
+      request = stub_publishing_api_put_content(edition.content_id, {})
+      PreviewService.new(edition).create_preview
+      expect(request).to have_been_requested
+    end
+
+    it "marks the edition as revision synced" do
```
benthorner: Looking at the test for the new service, how about 'revision_synced'?

spec/services/preview_service_spec.rb
kevindew: Agreed have amended

spec/services/preview_service_spec.rb
kevindew: Sure

spec/services/preview_service_spec.rb
kevindew: I'm not quite sure what you mean here, do you mean `it "marks the edition as 'revision_synced'"`?

kevindew:
> However, shouldn't we be stubbing the DraftAssetCleanupService globally in the PreviewService spec, so it remains a unit test? 

I'd suggest not myself. I'd argue that interfacing with that object is part of the unit of this test. It's also not a common approach we've used. 

Personal preference on this is normally to check that the method does what it's expected which allows the tests to not be concerned with the how, but just not to sweat the edge cases as they'll part of the other service's tests. So doing `expect_any_instance_of(DraftAssetCleanupService).to receive(:call)` feels a bit funky for me - but the setup requirements of avoiding that are pretty high, nor does it feel like a dependency we'd want to inject (where a stub would feel more appropriate).

> Related to that: the asset manager down case is now split in two: between upload_assets and a stubbed exception raised by this new service. I suppose it depends on the testing approach, though...

Yup, either could cause it - we would have same coverage though if that was changed. Given the wording of asset manager down it feels fine to be either case.

If we were handling it different I imagine it'd make sense to have separate whens. Would quite possibly make sense for DraftAssetCleanupService to raise an exception of it's own.

kevindew:
@benthorner lets chat face-to-face once you're back.

kevindew:
@benthorner sorry - just done those two small things.

PR #719, opened 2019-01-25
Update model comments

kevindew:

This aims to make these a little better and clearer. Hopefully they add
some value.

### DIFF
--- a/app/models/document.rb
+++ b/app/models/document.rb
@@ -1,6 +1,11 @@
 # frozen_string_literal: true
 
-# Represents all the versions of a piece of content in a particular locale
+# Represents all versions of a piece of content in a particular locale.
+# The version of content that is draft or been on GOV.UK is represented as an
+# edition model. Each edit a user has made to content is represented through a
+# revision model on an edition.
+#
+# This model is mutable
 class Document < ApplicationRecord
   self.table_name = "versioned_documents"
 
--- a/app/models/edition.rb
+++ b/app/models/edition.rb
@@ -2,9 +2,9 @@
 
 # Respresents the current state of a piece of content that was once or is
 # expected to be published on GOV.UK.
-# It is a mutable concept that is associated with a variety of immutable
-# models such as revisions and status which represent the current & past
-# information on the content.
+#
+# It is a mutable concept that is associated with a revision model and status
+# model to represent the current content and state of the edition.
 class Edition < ApplicationRecord
   self.table_name = "versioned_editions"
 
--- a/app/models/image.rb
+++ b/app/models/image.rb
@@ -1,8 +1,9 @@
 # frozen_string_literal: true
 
-# A basic model that is used to join together a collection of image revisions
-# with the item they represent. Thus two image_revisions that link to the
-# same Image are considered two revision of that image.
+# Represents all versions of a particular image. Each change is represented
+# as an ImageRevision.
+#
+# This model is immutable
 class Image < ApplicationRecord
   self.table_name = "versioned_images"
 
--- a/app/models/removal.rb
+++ b/app/models/removal.rb
@@ -1,5 +1,8 @@
 # frozen_string_literal: true
 
+# Stores the specific data for a removed status.
+#
+# This model is immutable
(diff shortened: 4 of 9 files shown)


app/models/document.rb
```diff
@@ -1,6 +1,11 @@
 # frozen_string_literal: true
 
-# Represents all the versions of a piece of content in a particular locale
+# Represents all versions of a piece of content in a particular locale.
+# The version of content that is draft or been on GOV.UK is represented as an
+# edition. Each edit a user has made to content is represented through a
```
benthorner: ...where an edition is a separate model...

app/models/edition.rb
```diff
@@ -2,6 +2,7 @@
 
 # Respresents the current state of a piece of content that was once or is
 # expected to be published on GOV.UK.
+#
 # It is a mutable concept that is associated with a variety of immutable
```
benthorner: Slightly ambiguous about what's providing the past information - might be clearer to say 'succession' instead of 'variety', and 'that together represents successive snapshots of the content'.

app/models/image.rb
```diff
@@ -1,8 +1,9 @@
 # frozen_string_literal: true
 
-# A basic model that is used to join together a collection of image revisions
-# with the item they represent. Thus two image_revisions that link to the
-# same Image are considered two revision of that image.
+# Represents all versions of a particular image. Each change is represented
+# as an ImageRevision.
+#
+# This is an immutable model
```
benthorner: Picky, but slightly inconsistent phrasing here :-).

app/models/removal.rb
```diff
@@ -1,5 +1,8 @@
 # frozen_string_literal: true
 
+# Stores the specific data for the a removed status.
```
benthorner: 'the' + 'a' :-|

app/models/revision.rb
```diff
@@ -1,12 +1,13 @@
 # frozen_string_literal: true
 
-# A model that represents a particular revision of a document. An edition
-# always has a revision for the current state of it and then there are past
-# revisions that represent all the changes a document has been through.
+# Represents a particular revision of a document - acting as a snapshot to
+# a particular users edit.
 #
 # This model aims to store as little data as possible (since there are many
```
benthorner: Not sure the argument is clear from saying 'there are many in the database'. Perhaps: 'This model stores as little data as possible by having associations to more specific types of revision and delegating its methods to them.'

app/models/withdrawal.rb
```diff
@@ -1,5 +1,8 @@
 # frozen_string_literal: true
 
+# Stores the specific data for the a withdrawn status.
```
benthorner: 'the' + 'a' again :-|

app/models/removal.rb
kevindew: d'oh - thanks

app/models/revision.rb
kevindew: Sounds good.

kevindew:
Thanks @benthorner, all comments addressed in ffbe7c3

app/models/document.rb
```diff
@@ -1,6 +1,11 @@
 # frozen_string_literal: true
 
-# Represents all the versions of a piece of content in a particular locale
+# Represents all versions of a piece of content in a particular locale.
+# The version of content that is draft or been on GOV.UK is represented as an
+# edition model. Each edit a user has made to content is represented through a
+# revision model.
```
benthorner: (...on an edition)

app/models/revision.rb
```diff
@@ -1,12 +1,12 @@
 # frozen_string_literal: true
 
-# A model that represents a particular revision of a document. An edition
-# always has a revision for the current state of it and then there are past
-# revisions that represent all the changes a document has been through.
+# Represents a particular revision of a document - acting as a snapshot to
+# a particular users edit.
```
benthorner: (user's)

kevindew:
@benthorner Thanks I've put those changes in.

PR #720, opened 2019-01-25
Counter proposal on current_status and current_revision

kevindew:

Off the back of https://github.com/alphagov/content-publisher/pull/709 I went down the road of trying out current_revision and current_status in https://github.com/alphagov/content-publisher/compare/make-names-verbose-and-current-repetitive

Doing that actually increased my concerns. But do have a look through and see if you feel that is actually a benefit or makes the code easier to understand.

The reasons I think adding the current prefix isn't a good idea are:

- It's implied and superfluous
- It starts a spread of increasing use of "current_" in various parts of an edition
  - renames assign_status to assign_current_status
  - renamed assign_revision to assign_current_revision
  - should then rename revision_synced to current_revision_synced
  - should rename status's revision_created_at to current_revision_created_at
- Causes things to seem a bit confusing when you compare `document.current_edition.current_status` and `document.live_edition.current_status`
- It involves making a lot of changes to disambiguate from something that is an audit trail.

Since the concerns seemed to be related to the naming clash of revision / revisions and status / statuses and that since the many's are used as an audit trail I felt it might be simpler to just remove the collection equivalents.

If/when we determine that these may be called on an edition I think options we could follow are:

- See if we can find a better name than all_*
- Call them past_statuses, past_revisions and use a lambda scope to cut out the current one.

Anyway, see what you think.

### DIFF
--- a/app/models/edition.rb
+++ b/app/models/edition.rb
@@ -18,9 +18,12 @@ class Edition < ApplicationRecord
     # Used to keep an audit trail of statuses a revision has held
     revision.statuses << status unless revision.statuses.include?(status)
 
-    # Used to keep an audit trail of all the revisions that have been
-    # associated with an edition
-    revisions << revision unless revisions.include?(revision)
+    # An edition points to a single revision, however we want to mantain a log
+    # of all joins between revision and edition. Revision has a many-to-many
+    # edition association that we use for storing this (to avoid the complexity
+    # of an edition having revision and revsions methods). Typically a revision
+    # would only be associated with a single edition.
+    revision.editions << self unless revision.editions.include?(self)
   end
 
   attr_readonly :number, :document_id
@@ -35,13 +38,9 @@ class Edition < ApplicationRecord
 
   belongs_to :status
 
-  has_many :statuses
-
   has_many :timeline_entries
 
   has_many :internal_notes
-
-  has_and_belongs_to_many :revisions, -> { order("revisions.number DESC") }
 
   delegate :content_id, :locale, :document_type, :topics, to: :document
 


kevindew:
@emmabeynon @1pretz1 do you feel the same as Ben?

1pretz1:
I do think for readability using `current` is clearer as we use `current_edition` to signify the 'active' edition so following that naming convention we should use `current` for the 'active' revision and any thing else that is thought of as 'active' when its part of a collection of records.

Unless we rename `current_edition` to `edition` and imply that any query that uses a singular word (e.g edition, revision, status etc) is referring to the most recent record.

kevindew:
> Erm, I actually quite like all the examples you're thinking of as problems - I think it reads much clearer that way. The problem to me is that `edition.revisions` does actually make sense, but `edition.revision` is ambiguous. Thanks for trying...sorry for being resolute :-|.

The bit that I find a bit off on that is that edition.revisions feels a bit off as that's the history of it rather than properties it has. 

> I do think for readability using `current` is clearer as we use `current_edition` to signify the 'active' edition so following that naming convention we should use `current` for the 'active' revision and any thing else that is thought of as 'active' when its part of a collection of records.
> 
> Unless we rename `current_edition` to `edition` and imply that any query that uses a singular word (e.g edition, revision, status etc) is referring to the most recent record.

For document we have the case of live_edition as well, which is why we distinguish current_edition from it. 

I'd argue that by needing a prefix of current_ on anything where we want to keep a history is that we're pushing that need to know we have a history throughout the application, whereas it's the edge case that we dig into the history.

emmabeynon:
I agree that `document.current_edition.current_status` does look a bit weird.  For me, `.status` is less problematic than `.edition` as, for example, I would expect `edition.status` to tell me that edition's current status.  I do still feel that `document.edition` is not as clear as I would like and I'd prefer `document.current_edition`.  Then you'd have `document.current_edition.status` which makes sense to me.

As for the collections, I think using `past_` is certainly more explicit, though in my mind calling something like `current_edition.revisions` implies to me that I'd get all the revisions associated with that edition rather than some unspecificed subset.

kevindew:
@benthorner I've updated comment as discussed.

kevindew:
🇯🇵

PR #726, opened 2019-01-28
Fix missing words in test "it"

kevindew:
This test didn't make much sense.


kevindew:
@benthorner oh darn foiled again by grammar - thanks

PR #729, opened 2019-01-29
ADR on modelling history redux

kevindew:
Trello: https://trello.com/c/Xt19LlP7/484-an-adr-for-editioning-versioning-documents

👉 [View rendered version](https://github.com/alphagov/content-publisher/blob/modelling-history-adr/docs/adr/0009-modelling-history.md) 👈

I've decided to close the previous PR as there was lots of discussion there that was now somewhat dated. I've updated various parts of this ADR to reflect changes that were made and I've removed the numbering aspect of it.


docs/adr/000x-modelling-history.md
```diff
@@ -0,0 +1,314 @@
...
+
+### Must
+
+- Store full information on current and live versions of a document;
+  *because* we may need to re-present either to the Publishing API as Content
+  Publisher is the source of truth.
+- Consider how a revision of a document and aspects of it have state that may
+  change;
+  *because* the concept of revision could become convoluted if it stores
+  information that a user does not directly impact.
```
emmabeynon: Could you give an example of such information?

docs/adr/000x-modelling-history.md
```diff
@@ -0,0 +1,314 @@
...
+A [MoSCoW](https://en.wikipedia.org/wiki/MoSCoW_method) breakdown of this
+functionality is defined as:
+
+### Must
+
+- Store full information on current and live versions of a document;
+  *because* we may need to re-present either to the Publishing API as Content
+  Publisher is the source of truth.
+- Consider how a revision of a document and aspects of it have state that may
+  change;
```
1pretz1: Don't really understand this point?

docs/adr/000x-modelling-history.md
```diff
@@ -0,0 +1,314 @@
...
+revision object is created and the edition is updated to reference that
+revision. A revision itself cannot be edited and is considered immutable. A
+revision may have a association with a collection of items, such as image
+revisions, when a new revision is created new image revisions are not created
+but a new join between them is. A revision has a number of components which are
+described further in this document.
+
+This modelling now has a concept of:
+
+**Status**: Represents a particular singular state that an edition can
```
1pretz1: Maybe lose the word 'singular' as I think its clearer to just put `Represents a particular state that an edition can hold ...`

docs/adr/000x-modelling-history.md
```diff
@@ -0,0 +1,314 @@
...
+
+### Timeline entry
+
+A timeline entry represents an event that should be shown to a user as part
+of a documents history. This stores the event which occurred and associations
+to relevant models. The sole purpose of this is for representing a
+timeline to users visually and not for storing historical metadata.
+
+The TimelineEntry model should not store data which could not be
+derived from other aspects of a document. This is to maintain
```
1pretz1: `The TimelineEntry model should not store data which could not be derived from other aspects of a document` -  could you give an example of what data this is referring to?

docs/adr/000x-modelling-history.md
```diff
@@ -0,0 +1,314 @@
...
+### Mo' models, mo' problems
+
+With an increase in the number of models there is more for a developer to
+understand and a higher barrier for entry. For every action there are either
+more database queries or a more complicated single query to look up information.
+
+Some ideas to alleviate this are:
+
+- the use of [delegate][] with ActiveRecord to make it simpler to access data
+  without interacting with individual models;
```
1pretz1: A note on this - I agree with the use of delegate but can be quite confusing when it's nested through multiple models

docs/adr/000x-modelling-history.md
1pretz1: +1 to the above comment as I don't really understand this point?

docs/adr/000x-modelling-history.md
```diff
@@ -0,0 +1,314 @@
...
+  debugging information.
+- Avoid relying on the TimelineEntry model as the main answer to any
+  historical queries;
+  *because* this model is close to the user and is likely to be iterated. The
+  deeper the coupling to a document the more difficult this becomes.
+
+
+### Could
+
+- Link TimelineEntry to changes rather than storing the information twice;
```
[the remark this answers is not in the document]
kevindew: Not necessarily - you could change status for instance which isn't a revision change.

docs/adr/000x-modelling-history.md
```diff
@@ -0,0 +1,314 @@
...
+  *because* this would reduce concerns and opportunities for error.
+- Consider states to be something that may change in a background job;
+  *because* publishing can be a slow task that could timeout a single
+  HTTP request, considering this up-front will be less painful than bolting it
+  on.
+- Enable a future feature to enable reverting to different points in time;
+  *because* this could enable users to make minor changes to a live document
+  and then recover a draft.
+- Separate related data to different tables to reduce data usage;
```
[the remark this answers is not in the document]
kevindew: Sure, this'll be another area where I struggled to be succinct.

It's based on the idea of storing bits of revisions in different tables that are unlikely to change at the same time so that a revision itself isn't a huge amount of data.

I'll have a think about better ways to phrase it.

docs/adr/000x-modelling-history.md
kevindew: I went for the word "singular" to represent that it's meant to be just a single word that describes that edition as a whole. 

So if an edition is a "draft" it can't be "submitted for review", they're separate things. Whereas another aspect of state is `revision_synced` but that's not a singular thing as it can be applied to multiple statuses.

Though as I type this I don't think I'm doing too good a job of explaining here so it might be actually better to try work out how to rephrase it. I think what I'm trying to explain is that compared to other stateful aspects of an edition a status is special because it's 1) shown to the user 2) used for filtering and 3) can only be in one of them at a time.

docs/adr/000x-modelling-history.md
kevindew: We had a previous example of this where Removal, Retire and InternalNote were associated to a TimelineEntry rather than to a document directly.

This meant that when we wanted to show that a document was removed we had to query timeline entry to get the first removal. So in this instance TimelineEntry was being queried to find out what the removal was for an Edition

docs/adr/000x-modelling-history.md
kevindew: Yup absolutely, they're definitely a trade off and with each level they get harder to trace to origin.

kevindew:
Thanks @emmabeynon and @1pretz1. @emmabeynon I really appreciated your English Teacher hat 👍.

docs/adr/000x-modelling-history.md
```diff
@@ -0,0 +1,314 @@
...
+
+- Transitioning from `draft` -> `submitted_for_review`. We'd consider the
+  document to instantly be in `submitted_for_review` status and no need to
+  update the Publishing API.
+- Transitioning from `scheduled` -> `published`. We'd consider the document to not
+  be considered published until we had successfully synced the resources with
+  the Publishing API. We'd want to inform the user that a transition was in
+  progress and whether that failed.
+
+An approach to resolve this would be to provide the ability for an edition to
```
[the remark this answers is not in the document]
kevindew: This is something that the design/product side are having to consider with regards to scheduling so the decision on that will be likely dictated from that.

My logic is that you're in the same status until the status change completes. E.g. if you're scheduled you're still there until publishing completes, if you're unpublishing you're published until that completes. These might not all hold though.

Pending kind of works unless the pending fails and then you end up in a state of published but not actually published. It could even be that there's a concept of a cover all "failed send help" status that would then allow a pending concept. 

Key outcome I'm hoping for with the scheduling work is that we're able to find a pattern that is suitable to scale to asynchronously allowing publishing/unpublishing/etc.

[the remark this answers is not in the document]
benthorner: Thanks for explaining more :-). I think our logic is quite similar; either way we end up with a piece of state - whether it's a pending flag on a status, or a separate pending status, that we need to do something with. Since we have to do work either way in case of failure/success, it seems to me like the flag approach is simpler, but perhaps I'm missing some advantage?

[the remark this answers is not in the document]
kevindew: No no, you're correct that a single status with a flag would be simpler.

It does feel a bit weird to me that you'd have something with a status of "published" that isn't actually published but if that fits into how product understands it then that'd be fine.

But yeah to me the key bit is what something is when a status transition fails, that approach there will probably dictate what option is applicable.

Btw I'm planning to drop this section from the ADR, but very happy to continue discussing our best options for this as this would be a really good problem to solve.

docs/adr/000x-modelling-history.md
emmabeynon: Ok got it.  Perhaps we could say something like "the concept of revision could become convoluted if it stores information that is not the content itself, such as internal state data"?

kevindew: Yup that sounds good - I've now dropped this section as it generated a lot of comments/confusion and hopefully is superfluous to the overall document.

docs/adr/000x-modelling-history.md
```diff
@@ -0,0 +1,314 @@
...
+in time. It has a number to indicate which revision of the document it is and
+stores who created it. Any time a user changes the data of a document a new
+revision object is created and the edition is updated to reference that
+revision. A revision itself cannot be edited and is considered immutable. A
+revision may have a association with a collection of items, such as image
+revisions, when a new revision is created new image revisions are not created
+but a new join between them is. A revision has a number of components which are
+described further in this document.
+
+This modelling now has a concept of:
```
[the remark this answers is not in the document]
kevindew: I've managed to get rid of it 👍

docs/adr/000x-modelling-history.md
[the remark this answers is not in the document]
kevindew: I've dropped this section now

docs/adr/000x-modelling-history.md
kevindew: I've had a go at rephrasing this section now.

docs/adr/000x-modelling-history.md
kevindew: I've included that information in the doc now.

docs/adr/000x-modelling-history.md
```diff
@@ -0,0 +1,314 @@
...
+Content Publisher currently has a concept of a document and not editions or
+revisions. A consequence of this is that an edit to a document overwrites the
+previous state of the document. Since a document can exist in two forms, draft
+and live, we can only know about one of these at any time and not of any
+previous versions.
+
+The suggestion is that we enhance Content Publisher to support the concepts of
+Document, Edition and Revision.
+
+For a document with a live version and a draft this could enable:
```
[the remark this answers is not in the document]
kevindew: Yup, good point. I've reproached this section.

docs/adr/000x-modelling-history.md
```diff
@@ -0,0 +1,314 @@
...
+versions;
+- **Revision**: A particular edit of a piece of content, represents a document at
+a specific point in time;
+- **Edition**: A particular revision that was published to GOV.UK or is the
+current draft of a document.
+
+Content Publisher currently has a concept of a document and not editions or
+revisions. A consequence of this is that an edit to a document overwrites the
+previous state of the document. Since a document can exist in two forms, draft
+and live, we can only know about one of these at any time and not of any
```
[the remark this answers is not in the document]
kevindew: I've rewrote this section.

docs/adr/000x-modelling-history.md
```diff
@@ -0,0 +1,314 @@
...
+previous versions.
```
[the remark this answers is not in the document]
kevindew: Yup good point, I've added that into the re-approach of context.

docs/adr/000x-modelling-history.md
```diff
@@ -0,0 +1,314 @@
...
+- Ability to compare two revisions of a document to show changes;
+- Storing a full history of a document - to enable importing Whitehall
+  documents with histories pre-dating the Publishing API.
+
+A [MoSCoW](https://en.wikipedia.org/wiki/MoSCoW_method) breakdown of this
+functionality is defined as:
+
+### Must
+
+- Store full information on current and live versions of a document;
```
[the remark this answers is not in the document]
kevindew: I've dropped this section now. I think I make use of your snapshot suggestion later though.

docs/adr/000x-modelling-history.md
```diff
@@ -0,0 +1,314 @@
...
+  *because* we may need to re-present either to the Publishing API as Content
+  Publisher is the source of truth.
+- Consider how a revision of a document and aspects of it have state that may
```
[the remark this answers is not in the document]
kevindew: Good points, I've dropped this bit now. It was much more relevant as part of scoping that the actual decision.

docs/adr/000x-modelling-history.md
```diff
@@ -0,0 +1,314 @@
...
+  *because* the concept of revision could become convoluted if it stores
+  information that a user does not directly impact.
+
+### Should
+
+- Have the full information for every edition that was published on GOV.UK;
+  *because* this will allow us the means to present what changed between
+  published editions - which is needed for a variety reasons, including, in
+  some cases, legal.
+- Record every change a publisher makes and allow for that information to be
```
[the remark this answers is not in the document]
kevindew: I think I was badly referencing how hard it would have been to access that data through papertrail. That's gone now.

docs/adr/000x-modelling-history.md
```diff
@@ -0,0 +1,314 @@
...
+  accessible;
+  *because* this maintains options for TimelineEntry and provides rich
```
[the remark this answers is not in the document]
kevindew: Dropped this section now.

docs/adr/000x-modelling-history.md
```diff
@@ -0,0 +1,314 @@
...
+  debugging information.
+- Avoid relying on the TimelineEntry model as the main answer to any
```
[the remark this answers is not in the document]
kevindew: I've dropped this section now, hopefully I've captured the essence elsewhere.

docs/adr/000x-modelling-history.md
```diff
@@ -0,0 +1,314 @@
...
+  historical queries;
+  *because* this model is close to the user and is likely to be iterated. The
+  deeper the coupling to a document the more difficult this becomes.
+
+
+### Could
+
+- Link TimelineEntry to changes rather than storing the information twice;
+  *because* this would reduce concerns and opportunities for error.
```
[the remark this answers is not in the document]
kevindew: Dropped section.

docs/adr/000x-modelling-history.md
```diff
@@ -0,0 +1,314 @@
...
+
+The proposed data model expands on the definitions of entities accordingly:
+
+**Document**: A record that represents all versions of a piece of content in a
+particular locale. It has many editions and many revisions. At any time it will
+have a current edition -  shown on Content Publisher index - and potentially a
+live edition which is currently on GOV.UK. The live and current edition can be
+the same. Document serves as a mutable entity that tracks data across all
+editions (such as first publishing date) and it is expected to be a joining
+point for document related data that is not associated with a particular
```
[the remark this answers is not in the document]
kevindew: Thanks - got it.

docs/adr/000x-modelling-history.md
```diff
@@ -0,0 +1,314 @@
...
+edition.
+
+**Edition**: A numbered version of a document that has been, or is
+expected to be, published on GOV.UK. It is associated with a revision
+and a status. It is mutable and is expected as a place where any edition level
```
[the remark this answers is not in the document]
kevindew: Thanks - done so.

docs/adr/000x-modelling-history.md
```diff
@@ -0,0 +1,314 @@
...
+persisted to the database once and never updated or deleted, any need to change
+them requires creating a new record. This allows us to store a full history
+by only adding to the database.
+
+Accessing immutable models is intended to always be done by a foreign key that
+is updated and not by the usage of `SELECT MAX` style queries - which are a
+common way to interact with append-only database models. This is done for
+simplicity, performance and for consistency with Rails idioms. To achieve this
+we maintain a number of mutable models (such as Document or Edition) that are
+updated to the most recent foreign keys which makes them not suitable vessels
```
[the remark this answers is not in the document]
kevindew: I've rephrased this bit now.

docs/adr/000x-modelling-history.md
```diff
@@ -0,0 +1,314 @@
...
+### Breakdown of Revision
+
+Revision is a model that stores little data itself and joins to other models
+that store the data. This can be visualised as:
+
+![Revision breakdown](000x/revision-diagram.png)
+
+The intention of breaking this up is to be conservative with the amount of data
+duplicated for both storage and maintenance reasons. Thus when a user edits
+the title of an edition a new ContentRevision object is created and the other
```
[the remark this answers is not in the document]
kevindew: Thanks 👍

docs/adr/000x-modelling-history.md
```diff
@@ -0,0 +1,314 @@
...
+
+![Image Revision breakdown](000x/image-revision-diagram.png)
+
+The Image model itself is used for continuation between image revisions, we know
+that two Image::Revisions are versions of the same item if they share the same
+Image association.
+
+The data of an Image::Revision is stored between an Image::FileRevision and an
+Image::MetadataRevision. These two objects differ by the fact that any change to
+Image::FileRevision requires changes to the resultant files (such as crop
```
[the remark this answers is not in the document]
kevindew: I've changed it to Asset Manager files. I felt "asset" was also rather ambiguous.

docs/adr/000x-modelling-history.md
```diff
@@ -0,0 +1,314 @@
...
+dimensions), whereas Image::MetadataRevision stores accompanying data that
+doesn't affect the files (such as alt text).
+
+Image::FileRevision is associated with an ActiveStorage::Blob object that is
+responsible for managing the storage of the source file. It also has a one to
+many association with Image::Asset, each of these represent resulting files that
+are uploaded to Asset Manager. For a single image uploaded we'd likely upload
+multiple sizes to Asset Manager - these different files are named variants. An
+Image::Asset object stores whether the asset is on Asset Manager, what state it
```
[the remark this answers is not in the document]
kevindew: I've rephrased this.

docs/adr/000x-modelling-history.md
```diff
@@ -0,0 +1,314 @@
...
+is in and the URL.
+
+### Timeline entry
+
+A timeline entry represents an event that should be shown to a user as part
+of a documents history. This stores the event which occurred and associations
+to relevant models. The sole purpose of this is for representing a
```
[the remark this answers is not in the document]
kevindew: Good point, I've amended this.

docs/adr/000x-modelling-history.md
```diff
@@ -0,0 +1,314 @@
...
+Initially this object is intended to be immutable although this may change in
+future if status changes are not synchronous.
+
+These are represented in the following diagram:
+
+![Main concepts](000x/main-concepts-diagram.png)
+
+Further details are then explained in the following sections.
+
+### Approach to mutability/immutability
```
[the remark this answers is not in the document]
kevindew: Yeah I can understand you got lost. I didn't really like the "Decision point: " prefix but I've hopefully improved this by reducing the number of headers and overall scope of the document.

docs/adr/000x-modelling-history.md
```diff
@@ -0,0 +1,314 @@
+# X. Modelling history
+
+Date: 2018-12-17  
+Amended: 2019-01-29
+
+## Context
+
+Using terminology as defined in [ADR-3](0003-initial-domain-modelling.md):
+
+- **Document**: A piece of content in a particular locale that can have many
```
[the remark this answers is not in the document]
kevindew: Sure, I've removed it now.

kevindew:
@emmabeynon @1pretz1 @benthorner I've done a fair re-working of this (which probably means there's a fresh batch of English errors 😬) let me know what you think.

docs/adr/000x-modelling-history.md
```diff
@@ -0,0 +1,250 @@
...
+  of a document;
+- users cannot edit or remove images on a document once the first
+  edition is published;
+- the TimelineEntry model stores aspects of a documents state, resulting in it
+  needed to be queried outside a timeline context.
+
+And this prevents a number of intended features for Content Publisher:
+
+- comparing different editions of a document;
+- republishing live content if there is any problems - which is a frequent
```
emmabeynon: Suggested rephrasing:
"republishing live content if there are any problems (currently a common support task for Whitehall publisher)"

docs/adr/000x-modelling-history.md
```diff
@@ -0,0 +1,250 @@
...
+
+### Core Concepts
+
+![Main concepts](000x/main-concepts-diagram.png)
+
+**Document**: A record that represents all versions of a piece of content in a
+particular locale. It has many editions and at any time it will have a current
+edition - shown on Content Publisher index - and potentially a live edition
+which is currently on GOV.UK. The live and current edition can be
+the same. Each iteration of a documents content is represented as a revision,
```
emmabeynon: "documents" => "document's"

docs/adr/000x-modelling-history.md
```diff
@@ -0,0 +1,250 @@
...
+timeline would show, and therefore it wasn't clear exactly how
+best to model them. Because of this TimelineEntry will be modelled in a
+speculative way with a number of references to relevant data.
+
+### Topics
+
+No data related to topics (otherwise known as GOV.UK taxonomy) is intended to
+be stored in Content Publisher at the current point in time. This is due to
+topics being accessed and edited by directly interacting with the Publishing
+API. This is because other applications, notably Content Tagger, can also edit
```
emmabeynon: Delete "This because"

docs/adr/000x-modelling-history.md
```diff
@@ -0,0 +1,250 @@
...
+these which has the consequence that the Publishing API is the source of truth
```
emmabeynon: "these" => "topics"

docs/adr/000x-modelling-history.md
```diff
@@ -0,0 +1,250 @@
...
+Amended: 2019-01-29
+
+## Context
+
+The Content Publisher domain model is focused on storing current
+information. There is a documents table which stores the most recent content of
+a document and an images table which stores the most recent version of
+a document's images. History of these changes is stored using [Papertrail][],
+which is not [intended as a permanent store][papertrail-pr]. The presentation
+of a documents history is done via a TimelineEntry model which stores that a
```
emmabeynon: "documents" => "document's"

kevindew:
Thanks @emmabeynon - I've made those changes.

docs/adr/000x-modelling-history.md
1pretz1: Nice!

docs/adr/000x-modelling-history.md
```diff
@@ -0,0 +1,250 @@
...
+a document's images. History of these changes is stored using [Papertrail][],
+which is not [intended as a permanent store][papertrail-pr]. The presentation
+of a document's history is done via a TimelineEntry model which stores that a
+user did something, but not what they did. In places where richer information
+was required there have been models such as Removal or Withdrawal associated
+with a TimelineEntry.
+
+This has led to a number of pain points:
+
+- users cannot discard a draft of a document that has between previously
```
1pretz1: I think you mean `been` here instead of `between` or `has previously been published` ?

docs/adr/000x-modelling-history.md
```diff
@@ -0,0 +1,250 @@
...
+It is intended that [delegation][delegate] be used when interfacing with a
+revision so that the caller need not be concerned with which sub-revision
+stores particular fields. This allows a revision to have a rich interface
+despite storing a low amount of data directly.
+
+### Image modelling
+
+Content Publisher supports a user uploading image files and referencing them
+in a revision of a document. They have metadata and editable properties that a
+user can change, of which a history stored. A single image file uploaded
```
1pretz1: I don't think this reads quite right maybe `which a history stored` -> `which a history is stored`

1pretz1:
@kevindew Good stuff, this is much clearer / easier to understand! I've just left a few comments other then that it LGTM 👍

docs/adr/000x-modelling-history.md
kevindew: Oops! thanks 👍

docs/adr/000x-modelling-history.md
kevindew: Oops again - thanks. I don't know why I'm so bad at English 😞

kevindew:
Thanks  @1pretz1 I've made those changes.

docs/adr/000x-modelling-history.md
```diff
@@ -0,0 +1,250 @@
...
+can only have one status at any one time. It is intended to associate with
+specialist data for a state such as a Removal or Withdrawal, which allows the
+history of these to no longer be the responsibility of TimelineEntry. Initially
+this object is intended to be immutable however this may be changed if status
+changes become asynchronous operations, which may be the case with scheduling.
+
+### Approach to mutability/immutability
+
+A number of the models in Content Publisher are defined as immutable, most
+significantly Revision and its associated types. These models should be
```
benthorner: I think we should be more specific about 'type' here, or perhaps link to the section that explains more. Fortunately, 'type' isn't currently a concept we're using in the app - long may this last!

docs/adr/000x-modelling-history.md
```diff
@@ -0,0 +1,250 @@
...
+persisted to the database once and never updated or deleted, any need to change
+them requires creating a new record. This allows us to store a full history
+by only appending to the database.
+
+Accessing immutable models is intended to always be done by a foreign key and
+not by the usage of `SELECT MAX` style queries. This is done for
+simplicity, performance and for consistency with Rails idioms. To achieve this
+there are a some mutable models, eg Edition, that reference immutable models,
```
benthorner: eg -> like (or e.g.)

docs/adr/000x-modelling-history.md
```diff
@@ -0,0 +1,250 @@
...
+
+### Approach to mutability/immutability
+
+A number of the models in Content Publisher are defined as immutable, most
+significantly Revision and its associated types. These models should be
+persisted to the database once and never updated or deleted, any need to change
+them requires creating a new record. This allows us to store a full history
+by only appending to the database.
+
+Accessing immutable models is intended to always be done by a foreign key and
```
benthorner: Could we think about rephrasing this paragraph, so it explains the problem before the solution? The wording is also a bit strange, as it suggests we 'invented' the Edition model to solve this particular problem, whereas I think it's intended for illustration purposes.

docs/adr/000x-modelling-history.md
```diff
@@ -0,0 +1,250 @@
...
+not by the usage of `SELECT MAX` style queries. This is done for
+simplicity, performance and for consistency with Rails idioms. To achieve this
+there are a some mutable models, eg Edition, that reference immutable models,
+eg Revision. When the content for an edition is revised the Edition model
+references the new Revision. Mutable models should not be used to store fields
+where there is a need to know how the value of that field changed over time.
```
benthorner: Could we simply say "Mutable models should not store data that can have a history"? Earlier, we talked about whether something is "changeable by the user directly" - is that appropriate to re-iterate here?

docs/adr/000x-modelling-history.md
```diff
@@ -0,0 +1,250 @@
...
+
+The choice of this immutability strategy is to store both present and
+historical concerns in the same data stores, thus ensuring history remains a
```
benthorner: I read 'data store' as the choice of database. How about "is to model current and historical data in the same way"?

docs/adr/000x-modelling-history.md
```diff
@@ -0,0 +1,250 @@
...
+first class citizen. A nice side effect of having immutable models is
+this opens options for caching. Since data for that
+model will never change it can effectively be cached forever.
+
+### Breakdown of Revision
+
+Revision is a model that stores little data itself and joins to other models
```
benthorner: Would be good to emphasise the immutability here, as this is key to the following argument...

docs/adr/000x-modelling-history.md
```diff
@@ -0,0 +1,250 @@
...
+that store the data. This can be visualised as:
+
+![Revision breakdown](000x/revision-diagram.png)
+
+The intention of breaking this up is to be conservative with the amount of data
+duplicated for both storage and maintenance reasons. Thus when a user edits
```
benthorner: "for both storage and maintenance reasons" -> "when the content of the revision changes".

benthorner: "Thus" -> "For example" (there's no particular implication here)

docs/adr/000x-modelling-history.md
```diff
@@ -0,0 +1,250 @@
...
+the title of an edition a new ContentRevision object is created and the existing
+objects (TagsRevision, MetadataRevision and ImageRevisions) are associated with
```
benthorner: How about just "and the existing TagsRevision, ..."? I wonder whether it's worth casually introducing the term 'sub-revision', as 'object' seems overly generic for a model.

docs/adr/000x-modelling-history.md
```diff
@@ -0,0 +1,250 @@
...
+It is intended that [delegation][delegate] be used when interfacing with a
+revision so that the caller need not be concerned with which sub-revision
+stores particular fields. This allows a revision to have a rich interface
+despite storing a low amount of data directly.
+
+### Image modelling
+
+Content Publisher supports a user uploading image files and referencing them
+in a revision of a document. They have metadata and editable properties that a
+user can change, of which a history is stored. A single image file uploaded
```
benthorner: Could we try "gets turned into multiple assets (byte streams)" instead? I think it's OK to be a bit more specific here, as all the concepts are quite meta ;-).

docs/adr/000x-modelling-history.md
```diff
@@ -0,0 +1,250 @@
...
+produces multiple files that are uploaded to Asset Manager for different sizing
+variations. Images are modelled in a similar way to Revision with an
+Image::Revision model, as represented below:
+
+![Image Revision breakdown](000x/image-revision-diagram.png)
+
+The Image model itself is used for continuation between image revisions. It is
```
benthorner: Although this is true, it might be a good idea to give an example of why we need to know this.

docs/adr/000x-modelling-history.md
```diff
@@ -0,0 +1,250 @@
...
+known that two Image::Revisions are versions of the same item if they share the
+same Image association.
+
+The data of an Image::Revision is stored between an Image::FileRevision and an
+Image::MetadataRevision. These two objects differ by the fact that any change to
```
benthorner: 'two objects' -> 'two models'

benthorner: Could we rephrase to say the less important thing first? Something like 'Both have the same immutable behaviour as the sub-revisions of a Revision, but any change to...'

benthorner: Could we use 'assets' instead of 'Asset Manager files', as the latter doesn't seem to be defined - it's at least unclear whether it's a concept in our app or external (or both ;-).

docs/adr/000x-modelling-history.md
```diff
@@ -0,0 +1,250 @@
...
+Image::FileRevision requires changes to the resultant Asset Manager files (such
+as crop dimensions), whereas Image::MetadataRevision stores accompanying data
+that doesn't affect the Asset Manager files (such as alt text).
+
+Image::FileRevision is associated with an ActiveStorage::Blob object that is
+responsible for managing the storage of the source file. It also has a one to
+many association with Image::Asset, each of these represent resulting files that
+are uploaded to Asset Manager. The Image::Asset model stores the URL to the
```
benthorner: Probably good to reiterate the size thing here somewhere, to clarify why it's a one-to-many.

docs/adr/000x-modelling-history.md
```diff
@@ -0,0 +1,250 @@
...
+Asset Manager file and what state the file is on Asset Manager.
+
+### Timeline
+
+The TimelineEntry model represents an event that should be shown to a user as
+part of a visual timeline of a document's history. It stores the event which
+occurred and associations to relevant models. The sole purpose of this model
```
benthorner: Perhaps add 'polymorphic' here to indicate the association is very flexible.

docs/adr/000x-modelling-history.md
```diff
@@ -0,0 +1,250 @@
...
+derived from other aspects of a document. This is to maintain
+flexibility for timeline to be an evolving and iterated part of the
+application. Previously models such as Removal and Withdrawal were associated
+directly with a TimelineEntry which led to using this model outside of
+displaying a timeline. These are now suggested to be associated with a Status
+model.
+
+At the time of writing it wasn't yet determined what the
+timeline would show, and therefore it wasn't clear exactly how
+best to model them. Because of this TimelineEntry will be modelled in a
```
benthorner: 'model them' -> 'model an entry for it'

docs/adr/000x-modelling-history.md
benthorner: The reasoning is getting a bit lost here in trying to explain how things used to be. I think this change might help: 

'The sole purpose...' -> 'At the time of writing it wasn't yet determined what the timeline would show, and therefore it wasn't clear exactly how best to model them. In order to support future iteration on this model, it should not be used outside of the context of displaying a timeline'.

We could then modify the following paragraph to explain one of the consequences of this is/would-be to re-think the associations for Removals and Withdrawals.

docs/adr/000x-modelling-history.md
```diff
@@ -0,0 +1,250 @@
...
+speculative way with a number of references to relevant data.
+
+### Topics
+
+No data related to topics (otherwise known as GOV.UK taxonomy) is intended to
+be stored in Content Publisher at the current point in time. This is due to
```
benthorner: 'the' -> 'this'

docs/adr/000x-modelling-history.md
```diff
@@ -0,0 +1,250 @@
...
+topics being accessed and edited by directly interacting with the Publishing
+API. Other applications, notably Content Tagger, can also edit topics which has
+the consequence that the Publishing API is the source of truth for this data
+rather than Content Publisher.
+
+This inconsistency makes it difficult to store the history of topics in a
+reliable way. Thus, until needs are determined to store this data, Content
+Publisher will continue to query the Publishing API rather than storing this
+data.
```
benthorner: ...and history for topics is not supported. Otherwise, with the previous sentence, it sounds like we're currently querying for history.

docs/adr/000x-modelling-history.md
```diff
@@ -0,0 +1,250 @@
...
+![Revision breakdown](000x/revision-diagram.png)
+
+The intention of breaking this up is to be conservative with the amount of data
+duplicated for both storage and maintenance reasons. Thus when a user edits
+the title of an edition a new ContentRevision object is created and the existing
+objects (TagsRevision, MetadataRevision and ImageRevisions) are associated with
+the next revision. An ImageRevision is modelled in a similar way to a Revision
+and this is explained further in [Image modelling](#image-modelling).
+
+It is intended that [delegation][delegate] be used when interfacing with a
```
benthorner: Should we also mention that we intend to delegate update operations, for the same reason?

docs/adr/000x-modelling-history.md
```diff
@@ -0,0 +1,250 @@
...
+of reading with noticeable omissions being a User model, since most models have
+a creator, and the speculative joins of a TimelineEntry.
+
+![Domain model diagram](000x/domain-model-diagram.png)
+
+## Status
+
+Accepted
+
+## Consequences
```
benthorner: I wonder if this section should discuss ideas for migration to the new modelling, as this was definitely a design decision. It does feel somewhat pointless now that it's actually implemented, but it's curious to have so few consequences for such a big proposal; the missing piece seems to be the implementation/migration strategy.

docs/adr/000x-modelling-history.md
```diff
@@ -0,0 +1,250 @@
...
+Date: 2018-12-17  
+Amended: 2019-01-29
+
+## Context
+
+The Content Publisher domain model is focused on storing current
+information. There is a documents table which stores the most recent content of
+a document and an images table which stores the most recent version of
+a document's images. History of these changes is stored using [Papertrail][],
+which is not [intended as a permanent store][papertrail-pr]. The presentation
```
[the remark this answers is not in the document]
kevindew: I'm not sure what you're suggesting?

docs/adr/000x-modelling-history.md
```diff
@@ -0,0 +1,250 @@
...
+of a document's history is done via a TimelineEntry model which stores that a
+user did something, but not what they did. In places where richer information
+was required there have been models such as Removal or Withdrawal associated
+with a TimelineEntry.
+
+This has led to a number of pain points:
+
+- users cannot discard a draft of a document that has between previously
+  published because there isn't a data representation it can be reverted to;
+- Content Publiser can't show an accurate link or status for the live edition
```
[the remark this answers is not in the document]
kevindew: You missed by big "Publiser" typo 😁 

I've amended this with different wording than what you suggested.

docs/adr/000x-modelling-history.md
```diff
@@ -0,0 +1,250 @@
...
+  of a document;
+- users cannot edit or remove images on a document once the first
+  edition is published;
+- the TimelineEntry model stores aspects of a documents state, resulting in it
```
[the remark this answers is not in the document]
kevindew: Thanks - my English is shocking 😢 

I've added a quick "which.." to this to clarify. It's a bit of a balancing act trying to make the list succinct when there's lot of detail that could be contained.

docs/adr/000x-modelling-history.md
```diff
@@ -0,0 +1,250 @@
...
+  needed to be queried outside a timeline context.
+
+And this prevents a number of intended features for Content Publisher:
+
+- comparing different editions of a document;
+- republishing live content if there are any problems (currently a common
+  support task for Whitehall publisher);
+- showing users what changes a user made in a particular edit.
```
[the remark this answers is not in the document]
kevindew: Only in the sense that to us they're both different ways of comparing a revision.

In most of our publishing apps these are quite distinct features though where edition to edition comparison is common and revision to revision change is done manually (for example mainstream publisher, where a publisher writes out the diff).

docs/adr/000x-modelling-history.md
```diff
@@ -0,0 +1,250 @@
...
+
+## Decision
+
+This ADR proposes changes to the domain model to resolve the aforementioned
+pain points and provide a means to support the future intended features. It
+defines a number of [core concepts](#core-concepts) that underpin the
```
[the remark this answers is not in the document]
kevindew: It's not intended to be a link to the image it's intended to be a link to the section. This section is supposed to introduce and summarise the ADR. I realise linking to the first thing isn't particularly useful so stripped that out.

docs/adr/000x-modelling-history.md
```diff
@@ -0,0 +1,250 @@
...
+each revision of the content of a document and each status an edition has held.
+As per [ADR-3](0003-initial-domain-modelling.md) it does not consider the
+option of sharing data between translations of a document as there are not
+the appropriate product decisions for this.
+
+A common theme in this modelling is the defining of
+[whether models are mutable](#approach-to-mutabilityimmutability) once
+persisted. Immutability is favoured as a means to preserve a history since
+each change requires storing a new record, implicitly creating a history.
+Mutability is used as a means to simplify interacting with models and preserve
```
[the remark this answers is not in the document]
kevindew: Initially I made your edit like you suggested, however it bugged me as this was the one use of "we" in the document. 

In the end I've only rephrased this a bit, hopefully you deem it an improvement.

docs/adr/000x-modelling-history.md
```diff
@@ -0,0 +1,250 @@
...
+API. Other applications, notably Content Tagger, can also edit topics which has
+the consequence that the Publishing API is the source of truth for this data
+rather than Content Publisher.
+
+This inconsistency makes it difficult to store the history of topics in a
+reliable way. Thus, until needs are determined to store this data, Content
+Publisher will continue to query the Publishing API rather than storing this
+data.
+
+### Overview
```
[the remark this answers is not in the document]
kevindew: Yeah summary has the same problem has overview in that it's frequently used at the introduction rather than conclusion of a document. I purposely opted out of summary since that carried connotations of brief and simplified since this is probably the most complicated part.

For lack of a better term I've gone with "Collated Diagram" which may well be the worst of all worlds. If you have any other suggestions that'd be great. I'm looking for something that reflects that it's the combined picture which I was what I hoped "Overview" portrayed.

docs/adr/000x-modelling-history.md
```diff
@@ -0,0 +1,250 @@
...
+the appropriate product decisions for this.
+
+A common theme in this modelling is the defining of
+[whether models are mutable](#approach-to-mutabilityimmutability) once
+persisted. Immutability is favoured as a means to preserve a history since
+each change requires storing a new record, implicitly creating a history.
+Mutability is used as a means to simplify interacting with models and preserve
+Rails idioms.
+
+This ADR then considers the impact of history for [timeline](#timeline) and
```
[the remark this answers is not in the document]
kevindew: It's meant to act as a way of summarises all that's covered in the decision. I think without it this introduction loses flow.

Maybe there is just too much going on in the mutability paragraph before, that might be what makes it a less smooth read?

docs/adr/000x-modelling-history.md
```diff
@@ -0,0 +1,250 @@
...
+### Core Concepts
+
+![Main concepts](000x/main-concepts-diagram.png)
+
+**Document**: A record that represents all versions of a piece of content in a
+particular locale. It has many editions and at any time it will have a current
+edition - shown on Content Publisher index - and potentially a live edition
+which is currently on GOV.UK. The live and current edition can be
+the same. Each iteration of a document's content is represented as a revision,
+thus a document has many revisions. Document is a mutable entity that tracks
```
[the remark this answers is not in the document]
kevindew: I've rephrased the use of track. I think it's useful to say that it is intended to store data common to the Document.

docs/adr/000x-modelling-history.md
```diff
@@ -0,0 +1,250 @@
...
+data across all editions (such as first publishing date) and it is expected to
+be a joining point for document-related data that is not associated with a
+particular edition.
+
+**Edition**: A numbered version of a document that has been, or is
+expected to be, published on GOV.UK. It is associated with a revision
+and a status. It is mutable and is a place where any edition level
+database constraints can be placed, such as only one live edition can exist for
+a document. It is possible that two editions of the same document share the
+same revision, this is in the case that they have identical content.
```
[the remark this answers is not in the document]
kevindew: I struggled how to resolve this concern of yours for a while.

I don't think imply is the correct term as it's explicit rather than implicit that the content is the same. I've tried to capture this in a sentence.

docs/adr/000x-modelling-history.md
```diff
@@ -0,0 +1,250 @@
...
+which is currently on GOV.UK. The live and current edition can be
+the same. Each iteration of a document's content is represented as a revision,
+thus a document has many revisions. Document is a mutable entity that tracks
+data across all editions (such as first publishing date) and it is expected to
+be a joining point for document-related data that is not associated with a
+particular edition.
+
+**Edition**: A numbered version of a document that has been, or is
+expected to be, published on GOV.UK. It is associated with a revision
+and a status. It is mutable and is a place where any edition level
```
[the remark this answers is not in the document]
kevindew: I've added a clause in to say why.

docs/adr/000x-modelling-history.md
```diff
@@ -0,0 +1,250 @@
...
+database constraints can be placed, such as only one live edition can exist for
+a document. It is possible that two editions of the same document share the
+same revision, this is in the case that they have identical content.
+
+**Revision**: Represents an immutable snapshot of the content of a document at a
+particular point in time. It has a number to indicate which revision of the
+document it is and stores who created it. Any request by a user that changes
+content should result in a single new revision, and only data that is
+changeable by a user directly should be stored as part of a revision. This is
```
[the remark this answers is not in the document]
kevindew: Fair point.

I was referring to Revisions should not be used to hold data that may change independently of a users request. For example, were we to store Asset Manager state inside a revision and created new revisions each time we changed data in Asset Manager. That would be a violation of a Revision to me as it would mean that our history is confused by events that a user created and events our system created. This would prevent us from ever presenting revision history to users or revision numbers.

I've tried to rephrase this based around what we do forbid rather than what we should include.

docs/adr/000x-modelling-history.md
```diff
@@ -0,0 +1,250 @@
...
+revises a document. The [anatomy of a Revision model](#breakdown-of-revision)
+is explored further in this document.
+
+**Status**: Represents a state that an edition can hold such as: "draft" or
+"submitted for review". This model is coupled to the concept of status that is
+shown and changed by a user. Each time a user changes the status of an edition
+a new Status model is created and the user who created it stored. An edition
+can only have one status at any one time. It is intended to associate with
+specialist data for a state such as a Removal or Withdrawal, which allows the
+history of these to no longer be the responsibility of TimelineEntry. Initially
```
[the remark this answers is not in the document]
kevindew: I don't think that's necessary, the content here represents our current understanding and changing that wouldn't be an accurate reflection on the intentions of the model.

docs/adr/000x-modelling-history.md
```diff
@@ -0,0 +1,250 @@
...
+changeable by a user directly should be stored as part of a revision. This is
+to directly map the concept of a revision to each time a user consciously
+revises a document. The [anatomy of a Revision model](#breakdown-of-revision)
+is explored further in this document.
+
+**Status**: Represents a state that an edition can hold such as: "draft" or
+"submitted for review". This model is coupled to the concept of status that is
+shown and changed by a user. Each time a user changes the status of an edition
+a new Status model is created and the user who created it stored. An edition
+can only have one status at any one time. It is intended to associate with
```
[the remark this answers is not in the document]
kevindew: There isn't a section that explains more, I've added more information here though to try make it less mysterious.

docs/adr/000x-modelling-history.md
kevindew: swapped "type" for "models"

docs/adr/000x-modelling-history.md
kevindew: eg is GOV.UK style

I've rephrased this section and it no longer has the eg

docs/adr/000x-modelling-history.md
kevindew: I've added a paragraph covering this.

docs/adr/000x-modelling-history.md
kevindew: I've gone with naming the individual models.

docs/adr/000x-modelling-history.md
kevindew: Asset is more ambiguous term than file and would do with qualifying. I understand you've tried to do that with the "byte streams" suggestion but I think that ends up opening a new question of "what is the relevance of a byte stream"?

docs/adr/000x-modelling-history.md
kevindew: I've changed this around a bit.

As per other comment, you have a point but I think "asset" requires more qualifying and comparing the terms in context we mention "files that are uploaded to Asset Manager" in the previous sentence so this terminology follows.

docs/adr/000x-modelling-history.md
kevindew: I've changed this around.

I've amended your suggestion though as that implies that we don't intend to use it for other reasons because we don't know the future iteration. Whereas that is only one of the reasons, the other being separation of concerns.

kevindew:
With much concern that this will only lengthen this rather protracted review cycle I've added a couple of new consequences section in.

@emmabeynon I've added a new section to the consequences about the wrapping of actions inside transactions. Do let me know if that reads ok to you and if it helps explain the "why"

@benthorner You're right about the missing consequence, it's a part that is rather weird in hindsight. I've added an extra section in that addresses the tech debt part as that seems to be the part that is currently relevant.

benthorner:
Nearly there, I think :-). The new consequences section looks good - most of my comments are follow-ups on the sections before.

emmabeynon:
> @emmabeynon I've added a new section to the consequences about the wrapping of actions inside transactions. Do let me know if that reads ok to you and if it helps explain the "why"

Reads well to me 👍

docs/adr/000x-modelling-history.md
```diff
@@ -0,0 +1,343 @@
...
+the same. Each iteration of a document's content is represented as a revision
+on the current edition, thus a document has many revisions. Document is a
+mutable entity that is used to store data common across all editions (such as
+first publishing date) and it is expected to be a joining point for
+document-related data that is not associated with a particular edition.
+
+**Edition**: A numbered version of a document that has been, or is
+expected to be, published on GOV.UK. It is associated with a revision
+and a status. It is mutable so that it can be a consistent object that
+joins to immutable data. It is a place where any edition-level
```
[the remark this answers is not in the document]
kevindew: I don't think that's the reason, as we could do the same through a series of immutable models. To me it is just so that it is a singular thing and can manage constraints.

docs/adr/000x-modelling-history.md
```diff
@@ -0,0 +1,343 @@
...
+database constraints can be placed, such as only one live edition can exist for
+a document. It is supported that two editions of the same document share the
```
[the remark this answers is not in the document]
kevindew: I don't think the scenario you've picked there is a particularly pertinent one as withdrawn doesn't make a difference. I've stayed away from referring to last revision of live edition being equal to first revision of next draft because of the change note stuff, so instead I've referenced the potential future revert aspect.

docs/adr/000x-modelling-history.md
kevindew: Oh sorry - I thought this was in reference to the models section in consequences which I updated.

Looking at this I don't think it needs changing as it only references interfacing which can refer to both reading or writing.

kevindew:
Thanks @emmabeynon 

@benthorner I've addressed most of your comments in https://github.com/alphagov/content-publisher/pull/729/commits/a3bab976c006df3b639b33f17d858586b7b831b1 and replied to ones I've not addressed.

docs/adr/000x-modelling-history.md
```diff
@@ -93,27 +93,31 @@ this document.
...
 a full history by only appending to the database.
 
 For simplicity, performance and consistency with Rails idioms the accessing
 of immutable models is intended to be done by foreign key and not by the usage
-of `SELECT MAX` style queries. An example of this is the Edition model which
-references an immutable model, Revision, that stores the content. This allows
-accessing an edition by a consistent primary key and the access of it's
-revision by a foreign key on the edition.
+of `SELECT MAX` style queries. This maintains the ability to use the regular
+approach to ActiveRecord associations and the means to require an associations
```
benthorner: "associations" -> "association's"

docs/adr/000x-modelling-history.md
```diff
@@ -172,10 +176,10 @@ Manager files (such as crop dimensions), whereas Image::MetadataRevision stores
 accompanying data that doesn't affect the Asset Manager files (such as alt
 text).
 
-Image::FileRevision is associated with an ActiveStorage::Blob object that is
-responsible for managing the storage of the source file. It also has a one to
-many association with Image::Asset, each of these representing resultant files
-that are uploaded to Asset Manager for the various image sizes. The
+Each Image::FileRevision is associated with an ActiveStorage::Blob object that
+is responsible for managing the storage of the source file. It also has a one
+to many association with Image::Asset, each Image::Asset represents resultant
```
benthorner: New sentence after ','

kevindew:
Finally done :tada: just shy of the 200 comments

PR #736, opened 2019-01-31
Enable factory bot linting

1pretz1:
FactoryBot offers a lint method to check the validity of each of the factories we have. This should be enabled to run as part of the default rake task and as part of our Jenkins build.

Trello:
https://trello.com/c/J48ZqCto/593-enable-factorybot-linting

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -12,6 +12,9 @@ node {
         sh("yarn")
         sh("yarn run lint")
       }
+      stage("Lint FactoryBot") {
+        sh("bundle exec rake factorybot:lint RAILS_ENV='test'")
+      }
     },
     rubyLintDiff: false,
     rubyLintDirs: "",
--- a/Rakefile
+++ b/Rakefile
@@ -6,4 +6,4 @@
 require_relative "config/application"
 
 Rails.application.load_tasks
-task default: %i(spec jasmine:ci lint brakeman)
+task default: %i(jasmine:ci lint brakeman)
--- a/lib/tasks/factorybot.rake
+++ b/lib/tasks/factorybot.rake
@@ -1,0 +1,16 @@
+# frozen_string_literal: true
+
+desc "Run FactoryBot linter"
+namespace :factorybot do
+  task lint: :environment do
+    if Rails.env.test?
+      ActiveRecord::Base.transaction do
+        FactoryBot.lint(traits: true)
+        raise ActiveRecord::Rollback
+      end
+    else
+      system("bundle exec rake factorybot:lint RAILS_ENV='test'")
+      fail if $?.exitstatus.nonzero?
+    end
+  end
+end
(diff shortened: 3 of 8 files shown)


Rakefile
```diff
@@ -6,4 +6,4 @@
 require_relative "config/application"
 
 Rails.application.load_tasks
-task default: %i(spec jasmine:ci lint brakeman)
+task default: %i(jasmine:ci lint brakeman)
```
kevindew: Nice catch 👍

lib/tasks/factorybot.rake
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+desc "Run Factroybot linter"
```
kevindew: minor typo

lib/tasks/factorybot.rake
```diff
@@ -0,0 +1,14 @@
...
+task factorybot_lint: :environment do
```
kevindew: I think we could make this more conventional to rake if this was a task of lint under the namespace of factorybot. So task becomes `factorybot:lint`.

If you check our tasks with `rake -T` to compare to others that might help

lib/tasks/factorybot.rake
```diff
@@ -0,0 +1,14 @@
...
+  if Rails.env.test?
+    DatabaseCleaner.clean_with(:deletion)
+    DatabaseCleaner.cleaning do
```
kevindew: It makes sense why you're using DatabaseCleaner - nice that it was explained in the commit ⭐️ 

However I do wonder if we can avoid the need for the gem by running this in a transaction, we can have a go at that together if you like?

lib/tasks/lint.rake
```diff
@@ -1,8 +1,9 @@
 # frozen_string_literal: true
 
-desc "lint Ruby, Sass and Javascript"
+desc "lint Ruby, Factorybot, Sass and Javascript"
 task "lint" do
   sh "govuk-lint-ruby --format clang"
+  sh "rake factorybot_lint"
```
kevindew: we probably should have a bundle exec here too, might want to see RAILS_ENV=test to avoid that call then shelling out again as per https://github.com/alphagov/content-publisher/pull/736/files#diff-8b9ad8593305e1b3717d3f0d09c96f9bR11

Jenkinsfile
```diff
@@ -12,6 +12,9 @@ node {
         sh("yarn")
         sh("yarn run lint")
       }
+      stage("Lint Factorybot") {
+        govuk.runTests("factorybot_lint")
```
kevindew: Since we're not actually running tests here we should probably just use the sh function. Otherwise that's going to conflict with the statsd timing in: https://github.com/alphagov/govuk-jenkinslib/blob/d2bb2e55c779dacdead1b290031c2c3b07bffa1f/vars/govuk.groovy#L652

spec/factories/tag_field_factory.rb
```diff
@@ -2,6 +2,8 @@
 
 FactoryBot.define do
   factory :tag_field, class: DocumentType::TagField do
+    skip_create
```
benthorner: Waaah - this has been in the gem for 7 years! How did I not see this before? I love this :-).

PR #744, opened 2019-02-01
Allow inserting contacts again

kevindew:
Trello: https://trello.com/c/za4TTu3h/614-fix-issue-blocking-contact-inserts

See commits for details.

### DIFF
--- a/app/views/contacts/search.html.erb
+++ b/app/views/contacts/search.html.erb
@@ -19,7 +19,7 @@
       text: %{<h1 class="govuk-heading-l">#{t("contacts.search.title")}</h1>}.html_safe,
     },
     id: "contact-id",
-    options: ["", ""] + contact_options,
+    options: [["", ""]] + contact_options,
     data_module: "autocomplete-with-hint-on-options",
   } %>
 


benthorner:
I see what you mean about the test - it is a bit whacky, and not much concerned with the feature. Do you think it would be reasonable to move this into a new JavaScript spec for this component (wrapper)?

spec/support/accessible_autocomplete_helper.rb
```diff
@@ -0,0 +1,22 @@
...
+    # select the option
+    parent_node = autocomplete_input.find(:xpath, "..")
+    parent_node.first(".autocomplete__option").click
+
+    # wait until option is selected
+    within("##{for_id}-select", visible: false) do
+      if value
+        find(%{option[value="#{value}"]:checked}, visible: false)
+      else
+        find("option:checked", visible: false, text: option)
```
benthorner: Looks good - not sure why we need this else clause right now, though?

kevindew: Yeah - it isn't needed here. It was a bit of a pain to work out so put it in. I think we should use it in other places. I've pulled it out for now. Will try bring it back soon though before it slips my mind.

kevindew:
@benthorner changed to data_module

app/views/components/docs/autocomplete.yml
```diff
@@ -88,7 +88,7 @@ examples:
       name: autocomplete-with-hint-on-options
       label:
         text: Select your country
-      custom_template: true
+      js_module: autocomplete-with-hint-on-options
```
benthorner: Missed a bit...

kevindew: d'oh thanks!

PR #749, opened 2019-02-05
Allow users to undo withdraw

1pretz1:
Having implemented the [withdrawal workflow](https://trello.com/c/GGxgDv3W/533-add-form-to-let-users-withdraw-a-document), we now need to make it possible for managing editors to revert the withdrawal of a document. 

We need to figure out what the status of the documents was before it got withdrawn so that we can set it back to that state after the withdrawal has been reverted.  While we know a document would have been published, it's possible that it still needed to be 2i'd, so we need to make sure the right state is set. We need to do this so a withdrawn document has the ability to be reinstated.

Trello:
https://trello.com/c/yhcKNEWr/587-allow-users-undo-withdraw


app/controllers/unwithdraw_controller.rb
```diff
@@ -1,7 +1,31 @@
 # frozen_string_literal: true
 
 class UnwithdrawController < ApplicationController
-  def index
+  def confirm_unwithdraw
```
kevindew: Since this is the Unwithdraw controller I think we could just call this confirm

app/controllers/unwithdraw_controller.rb
```diff
@@ -1,7 +1,31 @@
...
+
+    if current_user.has_permission?(User::MANAGING_EDITOR_PERMISSION)
+      redirect_to document_path(@document), confirmation: "unwithdraw/confirm"
+    else
+      render :non_managing_editor
+    end
+  end
+
+  def unwithdraw
+    document = Document.with_current_edition.lock!.find_by_param(params[:id])
```
kevindew: For this lock to work it needs to be inside a transaction

app/services/unpublish_service.rb
```diff
@@ -5,9 +5,7 @@ def withdraw(edition, public_explanation, user = nil)
     Document.transaction(requires_new: true) do
       edition.document.lock!
       check_unpublishable(edition)
-
-
-      withdrawal = Withdrawal.new(public_explanation: public_explanation)
+      withdrawal = new_withdrawal(edition, public_explanation)
```
kevindew: minor but since we don't use .new on Withdrawal anymore it might be clearer to call this create_withdrawal

app/services/unpublish_service.rb
```diff
@@ -88,6 +86,21 @@ def remove_and_redirect(edition, redirect_path, explanatory_note: nil)
 
 private
 
+  def new_withdrawal(edition, public_explanation)
+    if edition.status.withdrawn?
```
kevindew: You can just run `edition.withdrawn?` here

app/services/unwithdraw_service.rb
```diff
@@ -0,0 +1,32 @@
+# frozen_string_literal: true
+
+class UnwithdrawService
+  def process(edition, user = nil)
```
kevindew: If this is just a generic "do your thing" method then going with "call" is probably the most consistent one as that has a special meaning in Ruby and is used in quite a lot of our single purpose services.

app/services/unwithdraw_service.rb
```diff
@@ -0,0 +1,32 @@
...
+    Document.transaction do
```
kevindew: We'll need to set `requires_new: true` on this as per: https://github.com/alphagov/content-publisher/commit/0bcb0cddf0eba7acbde9aaf8a9cc5b5baac9c072

app/services/unwithdraw_service.rb
```diff
@@ -0,0 +1,32 @@
...
+      edition.document.lock!
```
kevindew: We don't really need to do this lock as it's not called from a rake task like the unpublishing service is, I'd remove this until we have a secondary use for this class.

app/services/unwithdraw_service.rb
```diff
@@ -0,0 +1,32 @@
...
+      check_publishable(edition)
```
kevindew: This should probably be check_unwithdrawable, but you can probably skip having this as a method and just have `raise "edition must be withdrawn to be unwithdrawn" unless edition.withdrawn?` on this line.

app/services/unwithdraw_service.rb
```diff
@@ -0,0 +1,32 @@
...
+      edition.assign_status(published_status.state, user)
+      edition.save!
+
+      TimelineEntry.create_for_status_change(
+        entry_type: :unwithdrawn,
+        status: edition.status,
+      )
+
+      GdsApi.publishing_api_v2.republish(
+        edition.content_id,
```
kevindew: We should bung locale in here too for consistency with other Publishing API calls.

app/views/unwithdraw/_confirm.html.erb
```diff
@@ -0,0 +1,20 @@
...
+<%= form_tag(unwithdraw_path(@document), method: :post) do %>
+
+  <p><%= t("documents.show.unwithdraw.title") %></p>
+
+  <%= render "govuk_publishing_components/components/button", {
+    text: t("documents.show.unwithdraw.confirm"),
+    destructive: true
+  } %>
+
+  <div style="display: inline-block; padding-left: 15px; padding-top: 8px">
```
kevindew: I think you're already aware of this, but this is something that we already changed on discard draft and this should be updated for it.

config/locales/en/documents/show.yml
```diff
@@ -83,3 +87,6 @@ en:
         topic_update_error:
           title: Something has gone wrong
           description_govspeak: Something went wrong when saving your changes. Please try again or [raise a support request](https://support.publishing.service.gov.uk/technical_fault_report/new).
+        unwithdraw_error:
+          title: Something has gone wrong
+          description_govspeak: Something went wrong when trying to publish the withdrawn document. Please try again or [raise a support request](https://support.publishing.service.gov.uk/technical_fault_report/new).
```
kevindew: should this be "republish the withdrawn document" ? as per the republish terminology used in line 15

config/locales/en/unwithdraw/unwithdraw.yml
```diff
@@ -1,4 +1,7 @@
 en:
   unwithdraw:
+    no_managing_editor_permission:
+      title: Only a managing editor can unwithdraw a document
```
kevindew: I'm not sure if this bit of text has been through a content designer. Assuming it hasn't I think it'd be more consistent with other text if it said "Only a managing editor can undo withdraw"

config/locales/en/unwithdraw/unwithdraw.yml
```diff
@@ -1,4 +1,7 @@
...
+      content: If you think this content should be unwithdrawn on GOV.UK, speak to the managing editor for your organisation.
```
kevindew: Similar vein as above comment:

"If you think this content should have the withdrawn banner removed, speak to the managing editor for your organisation."

spec/factories/status_factory.rb
```diff
@@ -16,4 +16,8 @@
     state { :withdrawn }
     association :details, factory: :withdrawal
   end
+
+  trait :published do
+    state { :published }
+  end
```
kevindew: We should move this into the factory block (and kill off the rogue withdrawn one)

spec/features/workflow/unwithdraw_spec.rb
```diff
@@ -3,24 +3,38 @@
...
     click_on "Undo withdrawal"
+    click_on(I18n.t!("documents.show.unwithdraw.confirm"))
   end
 
-  def then_i_see_the_feature_is_currently_unavailable
-    expect(page).to have_content("Sorry, this hasn't been built yet")
+  def then_i_see_the_document_is_now_unwithdrawn
+    @withdrawn_edition.reload
+
+    expect(@withdrawn_edition.status.state).to eq('published')
```
kevindew: minor but this could just be `expect(@withdrawn_edition).to be_published`

spec/features/workflow/unwithdraw_without_managing_editor_permissions.spec.rb
```diff
@@ -0,0 +1,39 @@
+# frozen_string_literal: true
+
+RSpec.feature "Undo a withdraw without managing editor permission" do
+  background do
+    given_i_dont_have_the_managing_editor_permission
+  end
```
kevindew: Rather than having this in a background block I think this would make sense in the scenario since it's not used to apply to multiple scenarios

spec/services/unwithdraw_service_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+    it "raises an error if current edition is published" do
+      published_edition = create(:edition, :published)
+      stub_publishing_api_republish(published_edition.content_id, {})
+
+      expect { UnwithdrawService.new.process(published_edition, @user) }.to raise_error(
+        "edition must be withdrawn to be unwithdrawn"
+      )
+    end
+
+    it "raises an error if current edition is draft" do
```
kevindew: minor but the condition here is probably that we care that the item is "not withdrawn" rather than it's specifically a "draft"

config/locales/en/documents/show.yml
1pretz1: yep nice spot

config/locales/en/unwithdraw/unwithdraw.yml
1pretz1: Yep, none of the text has really so that's probably the next thing to do

spec/services/unwithdraw_service_spec.rb
1pretz1: Yep i've removed this test now as its actually just a duplicate of the one above

1pretz1:
Still to do - Create a Publishing API down spec

kevindew:
@1pretz1 This looks good to me. I've added the feature test in now.

spec/features/workflow/unwithdraw_without_managing_editor_permissions.spec.rb
```diff
@@ -0,0 +1,36 @@
...
+RSpec.feature "Undo a withdraw without managing editor permission" do
+  scenario "withdrawn edition" do
+    given_i_dont_have_the_managing_editor_permission
+    when_there_is_a_withdrawn_edition
+    and_i_visit_the_withdrawn_document_summary_page
+    and_i_click_on_undo_withdrawal
+    then_i_see_a_message_to_ask_my_managing_editor_to_unwithdraw_content
+  end
+
+  def given_i_dont_have_the_managing_editor_permission
```
benthorner: Can we just use the default here?

spec/features/workflow/unwithdraw_spec.rb
```diff
@@ -3,24 +3,38 @@
...
+    click_on(I18n.t!("documents.show.unwithdraw.confirm"))
   end
 
-  def then_i_see_the_feature_is_currently_unavailable
-    expect(page).to have_content("Sorry, this hasn't been built yet")
+  def then_i_see_the_document_is_now_unwithdrawn
+    @withdrawn_edition.reload
+
+    expect(@withdrawn_edition).to be_published
+    expect(page.body).to_not include("This #{@withdrawn_edition.document_type.label.downcase} was withdrawn")
```
benthorner: Could we make this test more robust? It feels brittle, as there's nothing in this spec that checks the text was there in the first place.

benthorner: Also, can we use I18n.t! here?

app/controllers/unwithdraw_controller.rb
```diff
@@ -1,7 +1,33 @@
 # frozen_string_literal: true
 
 class UnwithdrawController < ApplicationController
-  def index
+  def confirm
     @document = Document.with_current_edition.find_by_param(params[:id])
+
+    if !current_user.has_permission?(User::PRE_RELEASE_FEATURES_PERMISSION)
+      render :index
```
benthorner: This `render` might be brittle, as it relies on us setting all the same instance variable for the `index` action. I know it's not going to be there for that long, but maybe a `raise` would be simpler?

app/services/unpublish_service.rb
```diff
@@ -88,6 +86,21 @@ def remove_and_redirect(edition, redirect_path, explanatory_note: nil)
 
 private
 
+  def create_withdrawal(edition, public_explanation)
```
benthorner: I thought we already allowed people to edit the withdrawal explanation? I can't see a commit for this, so just wondering how this came about...

app/services/unpublish_service.rb
```diff
@@ -88,6 +86,21 @@ def remove_and_redirect(edition, redirect_path, explanatory_note: nil)
...
+    if edition.withdrawn?
+      Withdrawal.create!(
+        public_explanation: public_explanation,
+        published_status: edition.status.details.published_status,
```
benthorner: Would it be more consistent to use delegation here? `edition.published_status` has a certain ring to it ;-).

benthorner: Actually, I see how that wouldn't make much sense outside of the context of a withdrawal. Perhaps this would read more clearly if `edition.status` were clarified to be a withdrawal?

app/services/unwithdraw_service.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
+
+class UnwithdrawService
+  def call(edition, user = nil)
+    Document.transaction(requires_new: true) do
+      raise "edition must be withdrawn to be unwithdrawn" unless edition.withdrawn?
+
+      published_status = edition.status.details.published_status
```
benthorner: Same as the previous comment - it's a little vague that the current status is for a withdrawal, due to the separation of `edition.withdrawn?` and `edition.status`.

app/views/unwithdraw/_confirm.html.erb
```diff
@@ -0,0 +1,21 @@
+<% confirmation = capture do %>
+<%= form_tag(unwithdraw_path(@document), method: :post) do %>
```
benthorner: Missing some indentation here...

spec/services/unwithdraw_service_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+
+      UnwithdrawService.new.call(withdrawn_edition, @user)
+      withdrawn_edition.reload
+
+      expect(withdrawn_edition.status.state).to eq(previously_published_state)
+    end
+
+    it "raises an error if current edition is not withdrawn" do
+      published_edition = create(:edition, :published)
+      stub_publishing_api_republish(published_edition.content_id, {})
```
benthorner: Do we need this stub here?

spec/services/unwithdraw_service_spec.rb
```diff
@@ -0,0 +1,39 @@
+# frozen_string_literal: true
+
+RSpec.describe UnwithdrawService do
+  before do
+    @user = create(:user)
+  end
+
+  describe "#call" do
```
benthorner: The timeline and status stuff is lovely, but are we missing a test for the actual call to republish?

spec/services/unwithdraw_service_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+      UnwithdrawService.new.call(withdrawn_edition, @user)
+
+      expect(withdrawn_edition.timeline_entries.first.entry_type).to eq("unwithdrawn")
+    end
+
+    it "creates a new status with the same state as the previously published version" do
+      withdrawn_edition = create(:edition, :withdrawn)
+      stub_publishing_api_republish(withdrawn_edition.content_id, {})
+
+      previously_published_state = withdrawn_edition.status.details.published_status.state
```
benthorner: Perhaps it would also be good to check the statuses are actually different rows here (vs. the behaviour for editing the withdrawing).

spec/services/unpublish_service_spec.rb
```diff
@@ -47,6 +47,22 @@
       expect(edition.status.details.public_explanation).to eq(public_explanation)
     end
 
+    it "saves the published_status to the withdrawal record" do
+      UnpublishService.new.withdraw(edition, public_explanation, user)
+      edition.reload
+
+      expect(edition.status.details.published_status.state).to eq("published")
```
benthorner: I think this test would be a bit easier to follow if we compared the status before/after, instead of the state. Otherwise, it's a bit surprising to see the literal `published` appearing here.

spec/services/unpublish_service_spec.rb
```diff
@@ -47,6 +47,22 @@
...
+    end
+
+    it "saves the correct published_status when the document is already withdrawn" do
```
benthorner: Again, would it be better to compare the statuses here?

spec/features/workflow/unwithdraw_spec.rb
```diff
@@ -3,24 +3,38 @@
...
+  def and_i_click_on_undo_withdraw_and_confirm
+    stub_publishing_api_republish(@withdrawn_edition.content_id, {})
+
     click_on "Undo withdrawal"
+    click_on(I18n.t!("documents.show.unwithdraw.confirm"))
   end
 
-  def then_i_see_the_feature_is_currently_unavailable
-    expect(page).to have_content("Sorry, this hasn't been built yet")
+  def then_i_see_the_document_is_now_unwithdrawn
```
benthorner: Should we be checking the request happened here?

app/services/unpublish_service.rb
1pretz1: We do, well we allow them to edit a withdrawal on the live site and the banner but it creates a new withdrawal and status in the app, here i'm only changing the process of submitting / editing a withdrawal to include saving the `published_status` of the document

app/services/unpublish_service.rb
1pretz1: The if statement suggests the edition status is withdrawn however if this is not clear i could assign edition.status to a variable e.g
```
if edition.withdrawn?
withdrawal = edition.status
Withdrawal.create!(
        public_explanation: public_explanation,
        published_status: withdrawal.details.published_status
``` ?

**Comment by kevindew** [comment]
> * When I try to change the withdrawal explanation, the button still says 'Withdraw' and a new timeline entry comes up saying 'Withdrawn', even though it's already withdrawn - is this intentional?
> * Edge case: if I don't change the explanation but click to 'Withdraw', it still does something :-|.

Oh darn good catch on that edge case. 

Both those things are outside of this PR so I'll have a look/chat with you about them separately.

File: spec/services/unpublish_service_spec.rb (hunk 28 above)
Comment (1pretz1) [reply spec/services/unpublish_service_spec.rb]: As i'm not testing here whether the status is being set to withdrawn, i'm only checking whether the `published_status` has been set correctly when a withdrawal is created. 

Checking the state is 'published' comes from the factory creation at the top of the page `let(:edition) { create(:edition, :published) }` as the `published_status` is set to the prior status of the edition. However if this isn't clear reading it

I could do:
```
   it "saves the published_status to the withdrawal record" do
      previous_published_status = edition.status
      UnpublishService.new.withdraw(edition, public_explanation, user)
      edition.reload

       expect(edition.status.details.published_status).to eq(previous_published_status)

spec/services/unpublish_service_spec.rb
1pretz1: See above comment ^

app/views/unwithdraw/_confirm.html.erb
1pretz1: Good spot!

spec/features/workflow/unwithdraw_without_managing_editor_permissions.spec.rb
1pretz1: Yep we sure can!

app/controllers/unwithdraw_controller.rb
1pretz1: I'll change the view from `index.html.erb` to `no_pre_release.html.erb` as this may be clearer, and the only instance variable the view needs is `@document` which is set before this view is rendered and this is only going to be here for a matter of days so think it's okay ?

app/services/unwithdraw_service.rb
1pretz1: Yep i agree, i've updated to use a withdrawal variable.

spec/services/unwithdraw_service_spec.rb
1pretz1: Nope, nice catch!

spec/services/unwithdraw_service_spec.rb
1pretz1: Yep sure!

spec/services/unwithdraw_service_spec.rb
1pretz1: I've added one in

spec/features/workflow/unwithdraw_spec.rb
1pretz1: I agree and done

app/controllers/unwithdraw_controller.rb
benthorner: Ah, I see - that's much clearer :-).

app/services/unpublish_service.rb
benthorner: Oh yeah, I see :-).

app/services/unpublish_service.rb
benthorner: 👍

spec/services/unpublish_service_spec.rb
benthorner: Yeah, I think that would be clearer, as it also matches the description in the 'it' ;-).

spec/services/unpublish_service_spec.rb
benthorner: See above comment below your comment ^˅ ;-).

spec/services/unpublish_service_spec.rb
1pretz1: done

spec/services/unpublish_service_spec.rb
1pretz1: done the comment above below your comment ;)

benthorner: Should we actually be checking the `withdrawn_edition.status.details.published_status` here? I think what might help is if we clarify the 'it' for this spec, as 'correct' is a bit ambiguous :-|.

benthorner:
@1pretz1 could we have a quick chat about this before we merge? @kevindew has got another PR (https://github.com/alphagov/content-publisher/pull/758) that might be sensible to merge into this one, as opposed to master.

app/services/unpublish_service.rb
```diff
@@ -88,6 +86,21 @@ def remove_and_redirect(edition, redirect_path, explanatory_note: nil)
 
 private
 
+  def create_withdrawal(edition, public_explanation)
+    if edition.withdrawn?
```
benthorner: This is similar to the work @kevindew is doing to preserve the withdrawal time when updating the public explanation. I wonder if it would be clearer to just `dup` any previous withdrawal and update/set the public explanation.

1pretz1: Yep sure, i've amended the code to reflect this suggestion.

PR #750, opened 2019-02-05
560 support rendering inline images

kgarwood:
https://trello.com/c/ZYFBN0ns/560-support-rendering-inline-images

These changes relate to supporting inline images in a way that supports both preview and payload contexts.


app/assets/stylesheets/components/_markdown-editor.scss
```diff
@@ -82,6 +82,10 @@ $toolbar-button-size: 50px;
   .govuk-textarea {
     overflow: scroll;
   }
+
+  figure {
```
kevindew: I assume this is for within the govspeak rendering?

I'd suggest having it underneath the govspeak class rather than (or within) the markdown editor to indicate we're specifying styles on what govspeak outputs. Also can we target this by .image as per https://github.com/alphagov/govspeak/blob/master/README.md#images ?

app/services/govspeak_document/in_app_options.rb
```diff
@@ -0,0 +1,31 @@
+# frozen_string_literal: true
+
+class GovspeakDocument::InAppOptions < GovspeakDocument::Options
+  include Rails.application.routes.url_helpers
+
+  def to_h
+    super.to_h.merge(images: in_app_images)
```
kevindew: The `to_h` here is superfluous. `super` will actually be calling `to_h` on the superclass and then we'd be calling `to_h` again on that hash.

app/services/govspeak_document/in_app_options.rb
```diff
@@ -0,0 +1,31 @@
...
+  end
+
+private
+
+  def in_app_images
+    edition.revision.image_revisions.map do |image_revision|
+      image_attributes(image_revision).merge(
+        # This is the same as url_for(image_revision.crop_variant) in the view, or
+        # image_revision.crop_variant.service_url, except that these don't work outside
+        # of a request context which normally includes the request host.
```
kevindew: This seems quite a big comment to have in a merge, it'd probably be easier to follow if we assigned a variable to this with the comment and then injected it in. 

For what it's worth though I don't think this comment is that necessary as I don't really know that our usage of `service_url`  is advised behaviour by activestorage and I don't think we actually use it in the app already (based on a quick grep)

app/services/govspeak_document/in_app_options.rb
```diff
@@ -0,0 +1,31 @@
...
+        url: rails_representation_path(image_revision.crop_variant, only_path: true),
```
kevindew: Do we need only_path here?

app/services/govspeak_document/in_app_options.rb
```diff
@@ -0,0 +1,31 @@
...
+      )
+    end
+  end
+
+  def image_attributes(image_revision)
```
kevindew: Since the data here isn't complete until we merge in a url parameter I think this method ends up being a bit misleadingly named.

I'd suggest though that you probably don't need it to be a separate method and could just have the whole hash in the in_app_images method to save having to context switch when reading the code.

app/services/unpublish_service.rb
```diff
@@ -100,8 +100,8 @@ def check_unpublishable(edition)
     end
   end
 
-  def format_govspeak(text)
-    GovspeakDocument.new(text).to_html
+  def format_govspeak(edition, text)
```
kevindew: minor, but flipping the order of these variables around to be "text, edition" would make it a slightly nicer read.

config/routes.rb
```diff
@@ -67,7 +67,7 @@
   get "/beta-capabilities" => "publisher_information#beta_capabilities", as: :beta_capabilities
   get "/publisher-updates" => "publisher_information#publisher_updates", as: :publisher_updates
 
-  post "/govspeak-preview" => "govspeak_preview#to_html"
+  post "/documents/:document_id/govspeak-preview" => "govspeak_preview#to_html", as: :govspeak_preview
```
kevindew: This would be more consistent with our other document routes if it was "/documents/:id/govspeak-preview"

spec/services/govspeak_document/in_app_options_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+                      image_revisions: [image_revision])
+
+      in_app_options = GovspeakDocument::InAppOptions.new("govspeak", edition)
+      actual_image_options = in_app_options.to_h[:images].first
+      expect(actual_image_options[:alt_text]).to eq("Some alt text")
+      expect(actual_image_options[:caption]).to eq("An optional caption")
+      expect(actual_image_options[:credit]).to eq("An optional credit")
+      expect(actual_image_options[:url]).to match("/representations")
+      expect(actual_image_options[:url]).to match("filename.png")
+      expect(actual_image_options[:id]).to eq("filename.png")
```
kevindew: Looks like we're checking a hash here? maybe we do that in just a single expectation?

app/services/govspeak_document/in_app_options.rb
benthorner: Oh yeah - removed :-).

app/assets/stylesheets/components/_markdown-editor.scss
benthorner: Sure - done.

app/services/govspeak_document/in_app_options.rb
benthorner: I think it's necessary, as it's not something that's covered in the documentation - I've updated the comment to make this clearer.

app/services/govspeak_document/in_app_options.rb
benthorner: Turns out we do :-(.

app/services/unpublish_service.rb
benthorner: Sure - done.

config/routes.rb
benthorner: Indeed - done.

spec/services/govspeak_document/in_app_options_spec.rb
benthorner: I think we need the individual ones, as we're doing a 'match' on the url part, to avoid having cumbersome test setup where we generate the full thing, which seems unnecessary.

app/services/govspeak_document/in_app_options.rb
benthorner: Agreed - done.

benthorner:
@kevindew I think I've addressed all your comments (including the commit typo). With the refactoring, I feel that what we have now is better than pushing all the adapter logic into one class; @kgarwood also commented that they feel the tests are now in a better state, after taking the options approach.

Is there an alternative approach you would have taken?

spec/services/govspeak_document/in_app_options_spec.rb
kevindew: oh darn didn't spot they were matches.

kevindew:
> @kevindew I think I've addressed all your comments (including the commit typo). 

Awesome thanks

> With the refactoring, I feel that what we have now is better than pushing all the adapter logic into one class; @kgarwood also commented that they feel the tests are now in a better state, after taking the options approach.
> 
> Is there an alternative approach you would have taken?

I'm intrigued because we've separated a file that could be understand as a whole into 4 files with inheritance to consider and the justification has entirely been around that it makes the tests better. So I was wondering whether you felt it benefits the code as a whole or just the testing?

benthorner:
@kevindew I agree there's a compromise on the number of files and the inheritance, but I think the new approach also offers a few advantages over the old one:

   * The GovspeakDocument class is now smaller, bringing us closer to removing the confusion between GovspeakDocument and Govspeak::Document.
   * The new options classes concentrate on a single context (in-app or payload); having all the code in one class made it harder to concentrate on the context of interest.

Another approach I considered was splitting GovspeakDocument into multiple classes, but this wouldn't allow for sharing the logic to determine the contact options.

Taken together with the reduced duplication in the tests, I'm just about happier with the new approach. I also reckon this will continue to benefit us as we add more options.

PR #757, opened 2019-02-07
Don't store identical withdrawal submissions and withdrawn_update timeline entry

kevindew:
Trello: https://trello.com/c/GGxgDv3W/533-add-form-to-let-users-withdraw-a-document

This addresses comments raised by Ben in: https://github.com/alphagov/content-publisher/pull/749#pullrequestreview-201081495

It redirects when a user submits the same explanation for updating a withdrawal.

It updates the button on the form to update an explanation. 

See:
screen shot 2019-02-07 at 16 40 37

And changes what's shown on the timeline entry:

See:
screen shot 2019-02-07 at 16 42 48

There's a secondary issue I noticed while doing this that the withdrawal date can go wrong so I'm going to fix that in a separate PR.

### DIFF
--- a/app/services/unpublish_service.rb
+++ b/app/services/unpublish_service.rb
@@ -6,25 +6,16 @@ class UnpublishService
       edition.document.lock!
       check_unpublishable(edition)
 
-      previous_withdrawal = edition.withdrawn? && edition.status.details
-
       return if withdrawn_with_public_explanation?(edition, public_explanation)
 
-      withdrawal = if previous_withdrawal
-                     previous_withdrawal.dup.tap do |w|
-                       w.assign_attributes(public_explanation: public_explanation)
-                     end
-                   else
-                     Withdrawal.new(public_explanation: public_explanation,
-                                    published_status: edition.status,
-                                    withdrawn_at: Time.current)
-                   end
+      withdrawal = build_withdrawal(edition, public_explanation)
 
+      already_withdrawn = edition.withdrawn?
       edition.assign_status(:withdrawn, user, status_details: withdrawal)
       edition.save!
 
       TimelineEntry.create_for_status_change(
-        entry_type: previous_withdrawal ? :withdrawn_updated : :withdrawn,
+        entry_type: already_withdrawn ? :withdrawn_updated : :withdrawn,
         status: edition.status,
         details: withdrawal,
       )
@@ -116,6 +107,19 @@ private
 
     withdrawal = edition.status.details
     withdrawal.public_explanation == public_explanation
+  end
+
+  def build_withdrawal(edition, public_explanation)
+    if edition.withdrawn?
+      withdrawal = edition.status.details.dup
+      withdrawal.tap do |w|
+        w.assign_attributes(public_explanation: public_explanation)
+      end
+    else
+      Withdrawal.new(public_explanation: public_explanation,
+                     published_status: edition.status,
+                     withdrawn_at: Time.current)
+    end
   end
 
   def format_govspeak(text, edition)
--- a/app/services/unpublish_service.rb
+++ b/app/services/unpublish_service.rb
@@ -21,7 +21,7 @@ class UnpublishService
       edition.save!
 
       TimelineEntry.create_for_status_change(
-        entry_type: :withdrawn,
+        entry_type: previous_withdrawal ? :withdrawn_updated : :withdrawn,
         status: edition.status,
         details: withdrawal,
       )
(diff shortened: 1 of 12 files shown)


kevindew:
Thanks for the review @emmabeynon 

I've just rebased this off master and I've made the change you requested @benthorner it felt a bit weird having a guard clause in the middle of a method but seemed most appropriate.

spec/features/workflow/edit_withdrawal_spec.rb
```diff
@@ -33,14 +35,20 @@ def then_i_can_see_the_existing_public_explanation
   end
 
   def and_i_can_edit_the_public_explanation
```
benthorner: Should this be a `when`?

app/services/unpublish_service.rb
```diff
@@ -7,21 +7,17 @@ def withdraw(edition, public_explanation, user = nil)
...
-                       w.assign_attributes(public_explanation: public_explanation)
-                     end
-                   else
-                     Withdrawal.new(public_explanation: public_explanation,
-                                    published_status: edition.status,
-                                    withdrawn_at: Time.current)
-                   end
+
+      # Don't do anything if nothing has changed
+      return if previous_withdrawal && previous_withdrawal.public_explanation == public_explanation
```
benthorner: This plus comment suggests a private method to me ;-).

spec/features/workflow/edit_withdrawal_spec.rb
```diff
@@ -33,14 +35,20 @@ def then_i_can_see_the_existing_public_explanation
...
+    fill_in "public_explanation", with: @new_explanation
+    click_on "Update explanation"
+  end
+
+  def then_i_can_see_the_updated_explanation
+    expect(page).to have_content(@new_explanation)
+  end
 
-    expect(@edition.reload.status.details.public_explanation).to eq(new_explanation)
+  def and_i_can_see_the_history_entry
```
benthorner: Minor one: I think we usually fold this step into the previous one; as most features involve a timeline entry, it seems better to avoid the extra step everywhere when reading the scenario. Or maybe we should make it explicit everywhere?

spec/features/workflow/edit_withdrawal_spec.rb
kevindew: D'oh yes - think I rushed that.

spec/features/workflow/edit_withdrawal_spec.rb
kevindew: Ah I wasn't aware of that convention, will do as you suggest.

app/services/unpublish_service.rb
kevindew: Makes sense 👍

app/views/documents/show/_history_tab.html.erb
```diff
@@ -24,7 +24,7 @@
           <%= simple_format(escape_and_link(entry.details.body)) %>
         <% end %>
 
-        <% if entry.withdrawn? && entry.details %>
+        <% if (entry.withdrawn? || entry.withdrawn_updated?) && entry.details %>
```
1pretz1: Is `&& entry.details` necessary here as if the document has been withdrawn the `details` for a TimelineEntry would have been set to the withdrawal?

1pretz1: As per the UnpublishService

kevindew: It's not essential it's just defensive. I guess it depends on how deeply we want to guarantee that a details object exists for certain TimelineEntry types. 

It's consistent with all the other timeline entry though, we could indeed remove all of them as our current doesn't present opportunities that they could be a miss - but on the other side we have nothing to guarantee their presence.

kevindew:
@benthorner I've made those changes now. I plan to separate UnpublishService into two after this PR - I didn't want to introduce that into this PR so late in the process.

app/views/documents/show/_history_tab.html.erb
1pretz1: Hmm yep i see what you mean, however I'd be in favour of potentially removing them as from the code you couldn't create an entry without associating it with a `detail` as an exception would be thrown and the entry would never of been created in the first place?

This is maybe something out of scope for this pr though

kevindew: Yeah no problems if you want to remove it in a separate PR.

I personally tend to veer towards defensive code - though I don't think I wrote it originally - so it doesn't feel too concerning to me but no qualms with it leaving either.

PR #758, opened 2019-02-07
Resolve issue where withdrawal time changes

kevindew:
Trello: https://trello.com/c/GGxgDv3W/533-add-form-to-let-users-withdraw-a-document

We previously used the time a status was created to show a user when a
piece of content was withdrawn, however this is not accurate if you
update the explanation of a withdrawal as this creates a new status.

Thus the way I've approached resolving this is by passing a timestamp
among the details of Withdrawal objects so that this data is persisted.
If a user undoes a withdraw and re-withdraws this timestamp will be
reset.

Note: this migration relies on there not being any withdrawal records persisted, as is the case in production. In a dev environment it may well require a db:reset. For integration it should be straightforward if the data sync has ran, if not it’ll need any documents that have been withdrawn to be deleted or statuses changed

### DIFF
--- a/app/services/unpublish_service.rb
+++ b/app/services/unpublish_service.rb
@@ -6,8 +6,15 @@ class UnpublishService
       edition.document.lock!
       check_unpublishable(edition)
 
-
-      withdrawal = Withdrawal.new(public_explanation: public_explanation)
+      previous_withdrawal = edition.withdrawn? && edition.status.details
+      withdrawal = if previous_withdrawal
+                     previous_withdrawal.dup.tap do |w|
+                       w.assign_attributes(public_explanation: public_explanation)
+                     end
+                   else
+                     Withdrawal.new(public_explanation: public_explanation,
+                                    withdrawn_at: Time.current)
+                   end
 
       edition.assign_status(:withdrawn, user, status_details: withdrawal)
       edition.save!
--- a/spec/services/unpublish_service_spec.rb
+++ b/spec/services/unpublish_service_spec.rb
@@ -1,6 +1,8 @@
 # frozen_string_literal: true
 
 RSpec.describe UnpublishService do
+  include ActiveSupport::Testing::TimeHelpers
+
   let(:edition) { create(:edition, :published) }
   let(:edition_with_image) do
     create(:edition, :published, lead_image_revision: image_revision)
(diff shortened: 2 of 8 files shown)


spec/services/unpublish_service_spec.rb
```diff
@@ -67,6 +73,21 @@
...
+
+    context "when the edition is already withdrawn" do
+      it "updates the public explanation" do
+        edition = create(:edition, :withdrawn)
+        expect { UnpublishService.new.withdraw(edition, public_explanation, user) }
+          .to change { edition.reload.status.details.public_explanation }
+          .to(public_explanation)
+      end
+
+      it "maintains the withdrawn timestamp" do
```
benthorner: I wonder if this is actually testing anything, as it's very possible the system time hasn't advanced enough to create any potential difference to test against; probably we need to specify a time that's definitely older in order to test this.

spec/services/unpublish_service_spec.rb
```diff
@@ -40,11 +42,15 @@
     end
 
     it "updates the edition status" do
```
benthorner: Might be clearer to say what it updates the status to here.

app/services/unpublish_service.rb
```diff
@@ -6,8 +6,15 @@ def withdraw(edition, public_explanation, user = nil)
       edition.document.lock!
       check_unpublishable(edition)
 
-
-      withdrawal = Withdrawal.new(public_explanation: public_explanation)
+      previous_withdrawal = edition.withdrawn? && edition.status.details
```
benthorner: This is similar to what @1pretz1 has been doing with the 'published_status'. I wonder if it would be simpler to just `dup` any existing withdrawal, to make it clear that we're just changing the explanation?

spec/services/unpublish_service_spec.rb
kevindew: Well yeah time will always have moved forward by then. I can make it clearer for you though.

app/services/unpublish_service.rb
kevindew: Yup, dup sounds good.

kevindew:
@benthorner thanks for review, all comments addressed

PR #759, opened 2019-02-08
[WIP] Replace revision build_ methods with services

benthorner:

### DIFF
--- a/app/controllers/contacts_controller.rb
+++ b/app/controllers/contacts_controller.rb
@@ -10,7 +10,7 @@ class ContactsController < ApplicationController
   end
 
   def insert
-    Document.transaction do # rubocop:disable Metrics/BlockLength
+    Document.transaction do
       document = Document.with_current_edition.lock.find_by_param(params[:id])
 
       redirect_location = edit_document_path(document) + "#body"
@@ -21,6 +21,7 @@ class ContactsController < ApplicationController
       end
 
       contact_markdown = "[Contact:#{params[:contact_id]}]\n"
+      current_edition = document.current_edition
       current_revision = document.current_edition.revision
 
       body = current_revision.contents.fetch("body", "").chomp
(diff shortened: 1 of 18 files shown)


benthorner:
@kevindew the description doesn't really cover one decision point, as I feel it's more minor, but I'd like to discuss it here: the choice between specifying ignored attributes vs. allowed attributes.

The first thing I noticed is that we already take the 'allowed attributes' approach when assigning attributes to sub-revisions, using `Hash#slice`. I don't think there's a simple way to get the same result using the 'ignored attributes' approach, and it's confusing to use 'allowed attributes' when assigning stuff vs. 'ignored attributes' when we're checking if the revision has changed.

The second problem I had with the 'ignored attributes' approach was a discrepancy between the attribute keys we use for assignment and the ones we get when generating a list of attributes that have changed. For example, we (have to) use 'lead_image_revision' to assign a new lead image (without an id); taking an 'ignored attributes' approach means we only get to check 'lead_image_revision_id', which doesn't accurately reflect the change (<revision> => <nil as no id yet> vs. <revision1> => <revision2>). On the other hand, using an 'allowed attributes' approach with `#public_send` allows us to inspect 'virtual attributes' like `image_revisions`, which resolves the discrepancy between assign/changed and yields accurate results from `changed_attributes`.

kevindew:
Thanks @benthorner for putting together and writing it up. It's looking really promising. 👍 

First thing I'd like us to do with this is split this out from the underlying story: https://trello.com/c/JcQEznco/605-add-checkbox-to-select-unselect-a-lead-image-on-metadata-page since we're probably in a position now where we can complete the story this sprint without needing to merge these changes.

Things where I think this is going well:

- I like having that we have the two classes RevisionUpdater and ImageRevisionUpdater
- Using the `updater.changed?` is nicer than the equality
- The updaters feel natural places for other common questions

Areas where I feel like we could use more work/regressions:

- It's a bit strange to be instantiate a `Base*` class - since that is terminology commonly used to reference an abstract class in Ruby due to it's lack of the feature
- The frequent use of super in methods on RevisionUpdater and ImageRevisionUpdater makes them harder to understand than the code they replace - basic suggestion would be that these don't inherit
- It seems a bit strange we've moved more stuff to the ImagesController
- Assuming you haven't got to it yet after our discussion but do still want to see ActiveRecord type casting
- Feel like the move to controller methods on ImagesController is something of a complexity regression, since that adds more code to an already large multi-responsibility entity

Re: attribute whitelist vs blacklist

So maybe it got lost in various other discussions, but the concern I was trying to iterate with the approach of slice and ignore was that if we pass in keys that aren't understood then the whitelist approach will just take things it understands and discard others and therefore concealing changes. That was the aspect that I was concerned about, I notice even looking at current Revision and Image::Revision classes that this is possible which is an oversight of mine 😞 but that's the thing I wanted to achieve with the whitelist/blacklist part. 

So in a nutshell I don't really mind which approach we take there but it'd be good if something blew up if unknown attributes are passed in.

Finally, apologies for a big question here, but something I'm unsure of is what the goal/success criteria for this work is?

benthorner:
Superseded by https://github.com/alphagov/content-publisher/pull/853.

Issue #740, opened 2019-02-01
Our controllers are long and we're not sure where to move the code


kevindew:
## Context

The actions in our controllers have become rather long in a [number](https://github.com/alphagov/content-publisher/blob/1eb067d35d557982f05601cde33c93f9ebea5694/app/controllers/images_controller.rb#L55-L101) [of](https://github.com/alphagov/content-publisher/blob/1eb067d35d557982f05601cde33c93f9ebea5694/app/controllers/documents_controller.rb#L45-L83) [places](https://github.com/alphagov/content-publisher/blob/1eb067d35d557982f05601cde33c93f9ebea5694/app/controllers/contacts_controller.rb#L12-L49). Testing them too can be rather verbose with lots of distinct feature tests: [1](https://github.com/alphagov/content-publisher/blob/1eb067d35d557982f05601cde33c93f9ebea5694/spec/features/workflow/delete_draft_after_publishing_spec.rb), [2](https://github.com/alphagov/content-publisher/blob/1eb067d35d557982f05601cde33c93f9ebea5694/spec/features/workflow/delete_draft_asset_manager_down_spec.rb), [3](https://github.com/alphagov/content-publisher/blob/1eb067d35d557982f05601cde33c93f9ebea5694/spec/features/workflow/delete_draft_publishing_api_down_spec.rb), [4](https://github.com/alphagov/content-publisher/blob/1eb067d35d557982f05601cde33c93f9ebea5694/spec/features/workflow/delete_draft_spec.rb).

We can anticipate that our controller needs are likely to grow rather than shrink in future for the following reasons:

- [We let users access pages they shouldn't have access to](https://trello.com/c/eAGtt4ry/322-we-dont-enforce-our-view-restrictions-in-our-controllers)
- [We raise 500 errors when users perform actions they shouldn't](https://github.com/alphagov/content-publisher/blob/1eb067d35d557982f05601cde33c93f9ebea5694/app/controllers/review_controller.rb#L41), this shouldn't be an internal server error
- We are expecting to add access limiting functionality
- We are hoping to iterate further on permissions

We want to avoid our controllers becoming hugely complicated as [has managed to occur in Whitehall](https://github.com/alphagov/whitehall/blob/27f205b6c3d151e2895c40053fc35645f31e1687/app/controllers/admin/editions_controller.rb)

[...]

PR #764, opened 2019-02-11
Lead image checkbox

benthorner:
https://trello.com/c/JcQEznco/605-add-checkbox-to-select-unselect-a-lead-image-on-metadata-page

As well as adding the checkbox, this makes a few changes to the TimelineEntry types we use for these actions, as well as tidying-up some of the tests. Please see the commits for more info.

I feel like this has made the ImagesController even harder to understand, which I'm hoping will be addressed by the work on https://github.com/alphagov/content-publisher/issues/740, as the tight coupling makes it hard to extract any services here.

Since the keys for some of the TimelineEntry types have changed, we will need to consider how to roll this change out. A multi-phase approach where we support the old ones for a time might work.


app/controllers/images_controller.rb
```diff
@@ -244,4 +254,26 @@ def update_crop_params
     # FIXME: this will raise a warning because of unpermitted paramaters
     params.permit(:crop_x, :crop_y, :crop_width).merge(crop_height: crop_height.to_i)
   end
+
+  def next_lead_image_revision(revision, image_revision, selected)
+    if revision.lead_image_revision&.image_id == image_revision.image_id
+      return nil unless selected
```
kevindew: This would be easier to read if the selected was part of the if statement - it's a little confusing to enter the block and then do nothing

Also, minor but the `nil` is redundant here

app/controllers/images_controller.rb
```diff
@@ -244,4 +254,26 @@ def update_crop_params
...
+    end
+
+    if revision.lead_image_revision&.image_id != image_revision.image_id
+      return image_revision if selected
+    end
+
+    revision.lead_image_revision
+  end
```
kevindew: I found this method a bit tricky to read

is it along the lines of:

```
return image_revision if selected

currently_lead = revision.lead_image_revision&.image_id == image_revision.image_id
if currently_lead && !selected
  nil
else
  revision.lead_image_revision
end
```

Or am I missing something?

benthorner: That's much better :-). Originally, we also had to determine if there had been a change, and I didn't see the refactoring opportunity now that logic is separated.

benthorner:
@kevindew thanks for reviewing :-). I forgot about the WIP commit - this was when I branched off to look at the refactoring we've been discussing. I've amended it with a proper description now.

As for the comments, they're intentional as I was trying to make a series of small changes, keeping most of the tests passing, but without deleting code I'd need later on. The alternative seemed to be to have larger or failing commits. This was a tricky change, so I was trying to make it easier to follow.

I think we may as well go with the first strategy for the TimelineEntry type, as if we decide to back out of this for some reason, then it would be better not to have to undo some preparatory work too.

benthorner:
@kevindew I've added a commit to preserve the old entries, and some tasks on the card to do the migration, so we don't forget about them.

PR #766, opened 2019-02-11
Add user ID to context for Sentry errors

benthorner:
https://sentry.io/govuk/app-content-publisher/issues/882726330/events/45d88c2a6e914dd4a5c2206d628d95b9/

Previously we've seen errors arising with multiple users interacting
with the same document. Having the user ID associated with the error
will help us to better debug these issues.


benthorner:
Sure - I didn't think it'd make much difference either way.

PR #771, opened 2019-02-12
Rename image TimelineEntry types

benthorner:
https://trello.com/c/JcQEznco/605-add-checkbox-to-select-unselect-a-lead-image-on-metadata-page

Previously we added support for 'lead_image_selected' and
'image_deleted' to replace 'lead_image_updated' and 'image_removed'.

This migrates the existing data to use the new entry types, after which
we will remove support for the old entry types.

### DIFF
--- a/db/migrate/20190212110438_rename_image_timeline_entry_types.rb
+++ b/db/migrate/20190212110438_rename_image_timeline_entry_types.rb
@@ -1,0 +1,19 @@
+# frozen_string_literal: true
+
+class RenameImageTimelineEntryTypes < ActiveRecord::Migration[5.2]
+  def up
+    TimelineEntry.where(entry_type: "lead_image_updated")
+      .update_all(entry_type: "lead_image_selected")
+
+    TimelineEntry.where(entry_type: "image_removed")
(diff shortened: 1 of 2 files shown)


db/migrate/20190212110438_rename_image_timeline_entry_types.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+class RenameImageTimelineEntryTypes < ActiveRecord::Migration[5.2]
+  def up
+    TimelineEntry.where(entry_type: "lead_image_updated")
+      .update!(entry_type: "lead_image_selected")
```
kevindew: should these be update_all?

benthorner: Good point - fixed :-).

kevindew: Thanks - not sure if you caught my other comment about db/schema.rb

benthorner: Oops - it was also modifying another line and I checked-it-out by mistake.

PR #796, opened 2019-02-19
Set scheduling date and time

emmabeynon:
For https://trello.com/c/4Ycdjj3C/640-allow-users-to-set-a-future-date-and-time-for-content-to-be-published

This PR enables users to set a scheduled date and time for content to be published.  A default time appears in the form and users can set and update their chosen scheduled date and time.  The input users choose is validated using the ScheduledDateTimeChecker service and errors are displayed in the event a user inputs invalid data.  

This PR does not handle the process of scheduling itself, which will be covered in a subsequent card.

The current usage of input fields for the date is a temporary measure.  There is a card in the backlog that will look at creating a date input component to replace these.  It is also expected that the time selection dropdown will eventually be replaced by a more dynamic component that will allow users to select a more granular minutes value.

screen shot 2019-02-19 at 11 51 11

screen shot 2019-02-21 at 15 53 02

screen shot 2019-02-19 at 11 52 33

### DIFF
--- a/app/controllers/schedule_controller.rb
+++ b/app/controllers/schedule_controller.rb
@@ -6,8 +6,18 @@ class ScheduleController < ApplicationController
       document = Document.with_current_edition.lock.find_by_param(params[:id])
       edition = document.current_edition
 
+      checker = Requirements::ScheduledDatetimeChecker.new(permitted_params)
+      issues = checker.pre_submit_issues
+
+      if issues.any?
+        flash[:scheduled_publishing_datetime_issues] = issues.items_for(:scheduled_datetime)
+        flash[:scheduled_publishing_params] = permitted_params
+        redirect_to document_path(document)
+        return
+      end
+
       new_revision = edition.revision.build_revision_update(
-        { scheduled_publishing_datetime: format_datetime },
+        { scheduled_publishing_datetime: checker.parsed_datetime },
         current_user,
       )
       edition.assign_revision(new_revision, current_user).save!
@@ -17,10 +27,6 @@ class ScheduleController < ApplicationController
   end
 
 private
-
-  def format_datetime
-    Time.zone.parse(permitted_params.values.join("-"))
-  end
 
   def permitted_params
     params.require(:scheduled).permit(:year, :month, :day, :time)
--- a/app/controllers/schedule_controller.rb
+++ b/app/controllers/schedule_controller.rb
@@ -1,0 +1,28 @@
+# frozen_string_literal: true
+
+class ScheduleController < ApplicationController
+  def save_scheduled_publishing_datetime
+    Document.transaction do
+      document = Document.with_current_edition.lock.find_by_param(params[:id])
+      edition = document.current_edition
+
+      new_revision = edition.revision.build_revision_update(
+        { scheduled_publishing_datetime: format_datetime },
+        current_user,
+      )
+      edition.assign_revision(new_revision, current_user).save!
+
+      redirect_to document_path(document)
+    end
+  end
+
+private
+
+  def format_datetime
+    Time.zone.parse(permitted_params.values.join("-"))
+  end
+
+  def permitted_params
+    params.require(:scheduled).permit(:year, :month, :day, :time)
+  end
+end
--- a/app/services/requirements/scheduled_datetime_checker.rb
+++ b/app/services/requirements/scheduled_datetime_checker.rb
@@ -1,0 +1,89 @@
+# frozen_string_literal: true
+
+module Requirements
+  class ScheduledDatetimeChecker
+    attr_reader :day, :month, :year, :time
+
+    FUTURE_TIME_PERIOD = { months: 14 }.freeze
+
+    def initialize(params)
+      @day = params[:day]
+      @month = params[:month]
+      @year = params[:year]
+      @time = params[:time]
+    end
+
+    def date_issues
+      issues = []
+
+      begin
+        parsed_date
+      rescue ArgumentError
+        issues << Issue.new(:scheduled_datetime, :invalid, field: "Date")
+      end
+
+      CheckerIssues.new(issues)
+    end
+
+    def time_issues
+      issues = []
+
+      if parsed_time.nil?
+        issues << Issue.new(:scheduled_datetime, :invalid, field: "Time")
+      end
+
+      CheckerIssues.new(issues)
+    end
+
+    def datetime_issues
+      issues = []
+
+      if in_the_past?
+        issues << Issue.new(:scheduled_datetime, :in_the_past)
+      end
+
+      if too_far_in_the_future?
+        time_period = FUTURE_TIME_PERIOD.map { |k, v| "#{v} #{k}" }.join(" & ")
+        issues << Issue.new(:scheduled_datetime,
+                            :too_far_in_future,
+                            time_period: time_period)
+      end
+
+      CheckerIssues.new(issues)
+    end
+
+    def pre_submit_issues
+      issues = []
+      issues += date_issues.to_a
+      issues += time_issues.to_a
+      if date_issues.items.empty? && time_issues.items.empty?
+        issues += datetime_issues.to_a
+      end
+      CheckerIssues.new(issues)
+    end
+
+    def parsed_datetime
+      @parsed_datetime ||= DateTime.parse("#{parsed_date} #{parsed_time}").in_time_zone
+    end
+
+  private
+
+    def parsed_date
+      @parsed_date ||= Date.parse("#{year}-#{month}-#{day}")
+    end
+
+    def parsed_time
+      @parsed_time ||= Time.zone.parse(time)
+    end
+
+    def in_the_past?
+      now = Time.zone.now
+      parsed_datetime <= now
+    end
+
+    def too_far_in_the_future?
+      now = Time.zone.now
+      parsed_datetime > now.advance(FUTURE_TIME_PERIOD).end_of_day
+    end
+  end
+end
--- a/app/views/documents/show/_scheduling_form.html.erb
+++ b/app/views/documents/show/_scheduling_form.html.erb
@@ -13,16 +13,19 @@
       label: { text: "Day" },
       name: "scheduled[day]",
       maxlength: 2,
+      value: publish_datetime&.day,
     } %>
     <%= render "govuk_publishing_components/components/input", {
       label: { text: "Month" },
       name: "scheduled[month]",
       maxlength: 2,
+      value: publish_datetime&.month,
     } %>
     <%= render "govuk_publishing_components/components/input", {
       label: { text: "Year" },
       name: "scheduled[year]",
       maxlength: 4,
+      value: publish_datetime&.year,
     } %>
     <%= render "govuk_publishing_components/components/select", {
       label: "Time",
--- a/app/views/documents/show/_scheduling_form.html.erb
+++ b/app/views/documents/show/_scheduling_form.html.erb
@@ -32,8 +32,14 @@
       id: "scheduled[time]",
       options: time_options(publish_datetime&.strftime("%l:%M%P")&.strip),
     } %>
-    <%= render "govuk_publishing_components/components/button", {
-      text: "Save", margin_bottom: true
-    } %>
+    <% if publish_datetime %>
+      <%= render "govuk_publishing_components/components/button", {
+        text: "Update"
+      } %>
+    <% else %>
+      <%= render "govuk_publishing_components/components/button", {
+        text: "Save"
+      } %>
+    <% end %>
   <% end %>
 <% end %>
--- a/app/views/documents/show/_scheduling_form.html.erb
+++ b/app/views/documents/show/_scheduling_form.html.erb
@@ -1,0 +1,40 @@
+<% publish_datetime = @edition.revision.scheduled_publishing_datetime %>
+
+<%= form_tag(save_scheduled_publishing_datetime_path(@edition.document)) do %>
+  <%= render "govuk_publishing_components/components/details", {
+    title: if publish_datetime
+             simple_format("Publish date: \n" + publish_datetime.strftime("%-d %B %Y - %l:%M%P"))
+           else
+             "Schedule publishing"
+           end
+  } do %>
+    <%= render "govuk_publishing_components/components/input", {
+      label: { text: "Day" },
+      name: "scheduled[day]",
+      maxlength: 2,
+    } %>
+    <%= render "govuk_publishing_components/components/input", {
+      label: { text: "Month" },
+      name: "scheduled[month]",
+      maxlength: 2,
+    } %>
+    <%= render "govuk_publishing_components/components/input", {
+      label: { text: "Year" },
+      name: "scheduled[year]",
+      maxlength: 4,
+    } %>
+    <%= render "govuk_publishing_components/components/select", {
+      label: "Time",
+      id: "scheduled[time]",
+      options: [
+        {
+          text: "12:01am",
+          value: "12:01am"
+        }
+      ]
+    } %>
+    <%= render "govuk_publishing_components/components/button", {
+      text: "Save", margin_bottom: true
+    } %>
+  <% end %>
+<% end %>
--- a/app/views/documents/show/_scheduling_form.html.erb
+++ b/app/views/documents/show/_scheduling_form.html.erb
@@ -1,4 +1,5 @@
 <% publish_datetime = @edition.revision.scheduled_publishing_datetime %>
+<% default_date = Time.zone.now.tomorrow %>
 
 <%= form_tag(save_scheduled_publishing_datetime_path(@edition.document)) do %>
   <%= render "govuk_publishing_components/components/details", {
@@ -26,12 +27,7 @@
     <%= render "govuk_publishing_components/components/select", {
       label: "Time",
       id: "scheduled[time]",
-      options: [
-        {
-          text: "12:01am",
-          value: "12:01am"
-        }
-      ]
+      options: time_options(publish_datetime&.strftime("%l:%M%P")&.strip),
     } %>
     <%= render "govuk_publishing_components/components/button", {
       text: "Save", margin_bottom: true
--- a/app/views/documents/show/_scheduling_form.html.erb
+++ b/app/views/documents/show/_scheduling_form.html.erb
@@ -1,38 +1,48 @@
-<% publish_datetime = @edition.revision.scheduled_publishing_datetime %>
+<% current_scheduled_publish_datetime = @edition.revision.scheduled_publishing_datetime %>
 <% default_date = Time.zone.now.tomorrow %>
+<% publish_date = current_scheduled_publish_datetime || default_date %>
+<% submitted_values = flash[:scheduled_publishing_params] %>
+<% selected_time = submitted_values&.fetch("time") ||  current_scheduled_publish_datetime&.strftime("%l:%M%P")&.strip %>
 
 <%= form_tag(save_scheduled_publishing_datetime_path(@edition.document)) do %>
   <%= render "govuk_publishing_components/components/details", {
-    title: if publish_datetime
-             simple_format("Publish date: \n" + publish_datetime.strftime("%-d %B %Y - %l:%M%P"))
+    title: if current_scheduled_publish_datetime
+             simple_format("Publish date: \n" + current_scheduled_publish_datetime.strftime("%-d %B %Y - %l:%M%P"))
            else
              "Schedule publishing"
            end
   } do %>
+    <% if flash[:scheduled_publishing_datetime_issues] %>
+      <% flash[:scheduled_publishing_datetime_issues].each do |issue| %>
+        <%= render "govuk_publishing_components/components/error_message", {
+          text: issue["text"]
+        } %>
+      <% end %>
+    <% end %>
     <%= render "govuk_publishing_components/components/input", {
       label: { text: "Day" },
       name: "scheduled[day]",
       maxlength: 2,
-      value: publish_datetime&.day,
+      value: submitted_values&.fetch("day") || publish_date&.day,
     } %>
     <%= render "govuk_publishing_components/components/input", {
       label: { text: "Month" },
       name: "scheduled[month]",
       maxlength: 2,
-      value: publish_datetime&.month,
+      value: submitted_values&.fetch("month") || publish_date&.month,
     } %>
     <%= render "govuk_publishing_components/components/input", {
       label: { text: "Year" },
       name: "scheduled[year]",
       maxlength: 4,
-      value: publish_datetime&.year,
+      value: submitted_values&.fetch("year") || publish_date&.year,
     } %>
     <%= render "govuk_publishing_components/components/select", {
       label: "Time",
       id: "scheduled[time]",
-      options: time_options(publish_datetime&.strftime("%l:%M%P")&.strip),
+      options: time_options(selected_time),
     } %>
-    <% if publish_datetime %>
+    <% if current_scheduled_publish_datetime %>
       <%= render "govuk_publishing_components/components/button", {
         text: "Update"
       } %>
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -12,6 +12,8 @@ Rails.application.routes.draw do
   post "/documents/:id/publish" => "publish#publish"
   get "/documents/:id/published" => "publish#published", as: :published
 
+  post "documents/:id/save-scheduled-publishing-datetime" => "schedule#save_scheduled_publishing_datetime", as: :save_scheduled_publishing_datetime
+
   get "/documents" => "documents#index"
   get "/documents/:id/edit" => "documents#edit", as: :edit_document
   patch "/documents/:id" => "documents#update", as: :document
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2019_02_12_110438) do
+ActiveRecord::Schema.define(version: 2019_02_14_120136) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
@@ -154,6 +154,7 @@ ActiveRecord::Schema.define(version: 2019_02_12_110438) do
     t.text "change_note"
     t.datetime "created_at"
     t.bigint "created_by_id"
+    t.datetime "scheduled_publishing_datetime"
   end
 
   create_table "removals", force: :cascade do |t|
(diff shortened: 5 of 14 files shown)


spec/features/workflow/schedule_spec.rb
```diff
@@ -0,0 +1,73 @@
+# frozen_string_literal: true
+
+RSpec.feature "Scheduling an edition" do
+  scenario do
+    given_there_is_an_edition_submitted_for_review
```
benthorner: Can we show scheduling if it's not submitted for review?

spec/features/workflow/schedule_requirements_spec.rb
```diff
@@ -0,0 +1,164 @@
+# frozen_string_literal: true
+
+RSpec.feature "Scheduling requirements" do
+  scenario do
```
benthorner: I don't think we need to cover all the cases here - the pattern in the other requirements features is just to test the UI works when just one thing is wrong, and leave the rest to the unit spec for the checker.

app/services/requirements/datetime_checker.rb
```diff
@@ -0,0 +1,138 @@
+# frozen_string_literal: true
+
+module Requirements
+  class DatetimeChecker
```
benthorner: Could we call this something more specific, like 'SchedulingChecker'? This is the pattern we've got with the other checkers. Although the logic is a universal truth, we might get a different format of input, so I don't think we can have a single 'DatetimeChecker'.

app/services/requirements/datetime_checker.rb
```diff
@@ -0,0 +1,138 @@
...
+    MAXIMUM_MONTH = 12
+    MINIMUM_MONTH = 1
+    MINIMUM_YEAR = Time.zone.now.year
+    MAXIMUM_HOUR = 12
+    MINIMUM_HOUR = 1
+    MAXIMUM_MINUTES = 59
+    MERIDIEN_VALUES = %w(am pm).freeze
+    FUTURE_TIME_PERIOD = { months: 14 }.freeze
+
+    def initialize(params)
```
benthorner: Do we have a requirement to give specific errors for each field? Especially as the fields are only on a couple of lines in the final design, I think it's probably enough to use `Date.parse` and `Time.parse` and just tell them they've entered an invalid date/time.

app/services/scheduling_service.rb
```diff
@@ -0,0 +1,38 @@
...
+
+class SchedulingService
+  DEFAULT_TIME = "9:00am"
+  DEFAULT_DATE = Time.zone.now.tomorrow
+
+  def self.time_options(selected = nil)
+    options = times.map { |time| { text: time, value: time } }
+    if selected
+      options.each do |option|
+        if option[:text] == selected
```
benthorner: Could you lose the `if` by doing `== (selected || DEFAULT_TIME)`?

app/services/scheduling_service.rb
```diff
@@ -0,0 +1,38 @@
+# frozen_string_literal: true
+
+class SchedulingService
+  DEFAULT_TIME = "9:00am"
+  DEFAULT_DATE = Time.zone.now.tomorrow
```
benthorner: I think we can't make this a constant, as it'll start being in the past :-|.

app/controllers/schedule_controller.rb
```diff
@@ -0,0 +1,72 @@
...
+class ScheduleController < ApplicationController
+  def save
+    @document = Document.with_current_edition.find_by_param(params[:id])
+    @edition = @document.current_edition
+    current_revision = @edition.revision
+    date_issues = Requirements::DatetimeChecker.new(date_params).date_issues
+    datetime_issues = Requirements::DatetimeChecker.new(datetime_params).datetime_issues
+
+    if date_issues.any? || datetime_issues.any?
+      flash[:day_issues] = date_issues.items_for(:scheduled_day)
```
benthorner: We could assign `@issues` here and do the extraction in the view - I think that's how the other views do it.

app/controllers/schedule_controller.rb
```diff
@@ -0,0 +1,72 @@
...
+      current_user,
+    )
+    @edition.assign_revision(new_revision, current_user).save!
+
+    redirect_to document_path(@document)
+  end
+
+private
+
+  def format_datetime
```
benthorner: Perhaps `parse` would be more accurate here ;-).

app/controllers/schedule_controller.rb
```diff
@@ -0,0 +1,72 @@
...
+    time = params[:scheduled_time]
+
+    Time.zone.parse("#{year}-#{month}-#{day}-#{time}")
+  end
+
+  def date_params
+    date_components = %I(day month year)
+    date_components.each_with_object({}) do |component, hash|
+      param = params["scheduled_#{component}".to_sym]
+      hash[component] = is_numeric?(param) ? param.to_i : param
```
benthorner: Could we let the checker handle this instead? That way, we can avoid having extra logic in the controller that's only used for the checker.

spec/features/workflow/schedule_spec.rb
```diff
@@ -0,0 +1,73 @@
...
+    when_i_visit_the_summary_page
+    and_i_click_on_schedule_publishing
+    then_i_see_a_default_date_and_time_selected
+
+    when_i_fill_in_a_valid_date
+    and_i_select_a_valid_time
+    and_i_click_save
+    then_i_see_the_edition_has_been_scheduled
```
benthorner: I think we should use a term other than 'scheduled', as (at least right now) the user has only set the date/time by doing this.

spec/features/workflow/schedule_spec.rb
```diff
@@ -0,0 +1,73 @@
+# frozen_string_literal: true
+
+RSpec.feature "Scheduling an edition" do
+  scenario do
+    given_there_is_an_edition_submitted_for_review
+    when_i_visit_the_summary_page
+    and_i_click_on_schedule_publishing
+    then_i_see_a_default_date_and_time_selected
+
+    when_i_fill_in_a_valid_date
```
benthorner: Could we combine these to make them more descriptive, like 'when_i_fill_in_the_schedule_date_and_time'

spec/features/workflow/edit_schedule_spec.rb
```diff
@@ -0,0 +1,76 @@
+# frozen_string_literal: true
+
+RSpec.feature "Edit the scheduling of an edition" do
```
benthorner: I wonder if we need this feature? Comparing to something like tags, where we also set an automatic default, we only check that we can edit in general. The main feature spec for this seems to check the value is saved, so I think that covers it :-).

app/services/scheduling_service.rb
```diff
@@ -0,0 +1,38 @@
...
+
+  def self.times
+    hours = [12] + (1..11).to_a
+    incremented_hours = hours.map do |hour|
+      [hour.to_s + ":00", hour.to_s + ":30"]
+    end
+
+    meridiem_hours = incremented_hours.flatten!.map { |hour| hour + "am" }
+    incremented_hours.each { |hour| meridiem_hours << hour + "pm" }
+    meridiem_hours[0] = "12:01am"
```
kevindew: I assume this is from the design, but is there any reason why midnight is special and is at 01 minutes?

spec/features/workflow/schedule_spec.rb
emmabeynon: Yes, if it's in draft.  It feels repetitive to cover both states but let me know what you think.

app/services/requirements/datetime_checker.rb
emmabeynon: I just felt like it would be helpful for users to know exactly what is wrong with their input so they can correct it.

app/services/scheduling_service.rb
emmabeynon: I think we could end up with a scenario where two options are selected if the default time is found before the selected time.

db/schema.rb
```diff
@@ -154,6 +154,7 @@
     t.text "change_note"
     t.datetime "created_at"
     t.bigint "created_by_id"
+    t.datetime "scheduled_datetime"
```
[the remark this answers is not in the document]
emmabeynon: `scheduled_publishing_datetime`?

app/controllers/schedule_controller.rb
```diff
@@ -0,0 +1,72 @@
+# frozen_string_literal: true
+
+class ScheduleController < ApplicationController
```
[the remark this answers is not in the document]
emmabeynon: Yes we figured this would handle the setting of the date and the scheduling itself.

spec/features/workflow/schedule_spec.rb
benthorner: I think the problem from a feature point of view is that we should either explicitly test both or none. At the moment, we're only testing one case - I think it would be fine if we just said 'there is an edition' (that's in draft by default) and wait for the need for more tests to come along :-).

app/services/scheduling_service.rb
benthorner: I should be fine - doing `(selected || DEFAULT_TIME)` will consistently use `selected` if it's defined, so there's no chance of accidentally choosing the `DEFAULT_TIME` if `selected` was passed in.

app/services/scheduling_service.rb
emmabeynon: Yes it's from the design.

app/services/scheduling_service.rb
```diff
@@ -0,0 +1,38 @@
+# frozen_string_literal: true
+
+class SchedulingService
```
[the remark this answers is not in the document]
emmabeynon: We decided to use a service as there isn't really any precedent for view helpers in this app, and we wouldn't expect it to be reused by any other views.

app/controllers/schedule_controller.rb
```diff
@@ -0,0 +1,72 @@
...
+  def save
+    @document = Document.with_current_edition.find_by_param(params[:id])
```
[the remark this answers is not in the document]
emmabeynon: This is causing a undefined method `created_at' for nil:NilClass` error in the metadata partial when invalid data is submitted.  It's expecting @document so I think we need to keep it.

app/controllers/schedule_controller.rb
emmabeynon: Opting for Kevin's suggestion of using a redirect + flash below 👇 n

db/schema.rb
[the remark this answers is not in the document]
kevindew: sounds good 👍

app/controllers/schedule_controller.rb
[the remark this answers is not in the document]
kevindew: Cool - might want to ponder more distinct action names as save feels quite ambiguous for both

app/services/scheduling_service.rb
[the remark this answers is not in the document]
kevindew: Sure, I don't think that reason justifies this being a service though, just seems to bring us more towards service being the utility directory.

I think it is fine for a single view usage to justify a view helper after all their role is to extract logic out of controllers - we do already have a few too but only global ones. However an alternative to a view helper could be to put the method in the controller and make that accessible as we do here: https://github.com/alphagov/content-publisher/blob/9b0175e4cb7fc4b4673cde7fe085704e6d30334b/app/controllers/debug_controller.rb#L5

I imagine we'd be hoping that a SchedulingService would be used to do the scheduling state transition similar to the WithdrawService or PublishService. 

I do wonder if it's better that we give all our services a verb name to indicate they should be used for doing a particular thing - but that's more in the scope of  https://github.com/alphagov/content-publisher/issues/740

app/services/scheduling_service.rb
kevindew: How weird :-/

app/services/requirements/datetime_checker.rb
kevindew: I think we'll struggle to display that many errors. If a user left everything blank they'd have at least 3 errors given which could all be summarised with "Enter a valid date". Particularly as we have very little space.

Might be worth checking with design but I think we could probably just have "Enter a valid date" for blank or any odd input, "Enter a date and time in the future" for that scenario and the time period one. 

The aspect that I think could be useful with this level of error messages is informing a user they entered the wrong format (e.g. text when we expected a number) - but then I don't really feel validation errors is the way to inform them how they should enter text and if that's ambiguous it should be clearer in the UI. Sidenote - I wonder if we want to use input type=number, datalist or both which may make that clearer.

spec/features/workflow/schedule_spec.rb
emmabeynon: Changed to `given_there_is_an_edition`

spec/features/workflow/schedule_requirements_spec.rb
emmabeynon: Now only testing one scenario

app/services/requirements/datetime_checker.rb
emmabeynon: I've renamed to `SchedulingDatetimeChecker` just to make it clear that it's checking the datetime specifically.  There will be another process dealing with the scheduling itself.

app/services/requirements/datetime_checker.rb
emmabeynon: I've reduced this down to just checking the validity of the date and time, and that the datetime provided isn't in the past or too far in the future.

app/services/scheduling_service.rb
emmabeynon: Done 👍

app/services/scheduling_service.rb
emmabeynon: This is now set as a variable in the view.

app/controllers/schedule_controller.rb
emmabeynon: This has now been moved to the `ScheduledDatetimeChecker` as `#parsed_datetime`.

spec/features/workflow/schedule_spec.rb
emmabeynon: Changed to `when_i_choose_a_scheduled_date_and_time`

spec/features/workflow/edit_schedule_spec.rb
emmabeynon: Now deleted

app/controllers/schedule_controller.rb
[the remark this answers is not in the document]
emmabeynon: Changed `save` => `save_scheduled_publishing_datetime`

app/controllers/schedule_controller.rb
```diff
@@ -0,0 +1,72 @@
+# frozen_string_literal: true
+
+class ScheduleController < ApplicationController
+  def save
```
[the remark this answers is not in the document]
emmabeynon: Now in a transaction and the document is locked.

app/controllers/schedule_controller.rb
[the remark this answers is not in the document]
emmabeynon: This is no longer an instance variable now that we're redirecting to document_path instead of rendering the view.

app/controllers/schedule_controller.rb
```diff
@@ -0,0 +1,72 @@
...
+      flash.now[:day_issues] = date_issues.items_for(:scheduled_day)
+      flash.now[:month_issues] = date_issues.items_for(:scheduled_month)
+      flash.now[:year_issues] = date_issues.items_for(:scheduled_year)
+      flash.now[:datetime_issues] = datetime_issues.items_for(:scheduled_datetime)
+
+      render "documents/show"
+      return
+    end
+
+    new_revision = current_revision.build_revision_update(
```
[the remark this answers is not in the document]
emmabeynon: Done

app/controllers/schedule_controller.rb
```diff
@@ -0,0 +1,72 @@
...
+      { scheduled_datetime: format_datetime },
```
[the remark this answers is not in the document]
emmabeynon: `format_datetime` is now `parsed_datetime` in the Time Options Helper

config/routes.rb
```diff
@@ -12,6 +12,8 @@
   post "/documents/:id/publish" => "publish#publish"
   get "/documents/:id/published" => "publish#published", as: :published
 
+  post "documents/:id/save_schedule" => "schedule#save", as: :save_schedule
```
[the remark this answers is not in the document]
emmabeynon: Done

db/schema.rb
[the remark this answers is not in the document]
emmabeynon: Now changed to `scheduled_publishing_datetime`

app/controllers/schedule_controller.rb
```diff
@@ -0,0 +1,72 @@
...
+  def format_datetime
+    year = params[:scheduled_year]
+    month = params[:scheduled_month]
+    day = params[:scheduled_day]
+    time = params[:scheduled_time]
+
+    Time.zone.parse("#{year}-#{month}-#{day}-#{time}")
+  end
+
+  def date_params
```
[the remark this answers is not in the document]
emmabeynon: These have been combined under `datetime_params` which uses `params.require(:scheduled).permit(:year, :month, :day, :time)`

app/controllers/schedule_controller.rb
```diff
@@ -0,0 +1,72 @@
...
+    time = params[:scheduled_time].dup
+    meridien = time.scan(/am|pm/)&.first
+    time_values = time.gsub!(meridien, "").split(":")
+    hour = time_values.first
+    minutes = time_values.last
+
+    time_params = {}
+    time_params[:hour] = is_numeric?(hour) ? hour.to_i : hour
+    time_params[:minutes] = is_numeric?(minutes) ? minutes.to_i : minutes
+    time_params[:meridien] = meridien
```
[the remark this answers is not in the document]
emmabeynon: Turns out that we can pass the time as a string e.g. "11:00am" and `Time.parse` can handle that, so no need to pre-process it 👍

app/controllers/schedule_controller.rb
```diff
@@ -0,0 +1,72 @@
+# frozen_string_literal: true
+
+class ScheduleController < ApplicationController
+  def save
+    @document = Document.with_current_edition.find_by_param(params[:id])
+    @edition = @document.current_edition
+    current_revision = @edition.revision
+    date_issues = Requirements::DatetimeChecker.new(date_params).date_issues
+    datetime_issues = Requirements::DatetimeChecker.new(datetime_params).datetime_issues
```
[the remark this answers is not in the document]
emmabeynon: Done

app/controllers/schedule_controller.rb
```diff
@@ -0,0 +1,72 @@
...
+
+    if date_issues.any? || datetime_issues.any?
+      flash.now[:day_issues] = date_issues.items_for(:scheduled_day)
+      flash.now[:month_issues] = date_issues.items_for(:scheduled_month)
+      flash.now[:year_issues] = date_issues.items_for(:scheduled_year)
+      flash.now[:datetime_issues] = datetime_issues.items_for(:scheduled_datetime)
+
+      render "documents/show"
```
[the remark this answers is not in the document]
emmabeynon: Now redirecting to `document_path` with the errors stored in a flash.

app/services/scheduling_service.rb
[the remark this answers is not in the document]
emmabeynon: This has been changed to a Time Options helper.

app/services/scheduling_service.rb
emmabeynon: I think it's something to do with users getting confused by what day 12:00am is in but can check with Sonia when she's back.

spec/features/workflow/schedule_spec.rb
emmabeynon: Now changed to `then_i_see_the_edition_has_a_set_scheduled_publishing_date_and_time`

app/controllers/schedule_controller.rb
emmabeynon: Now moved into `ScheduledDatetimeChecker`

app/controllers/schedule_controller.rb
```diff
@@ -0,0 +1,39 @@
+# frozen_string_literal: true
+
+class ScheduleController < ApplicationController
+  def save_scheduled_publishing_datetime
+    Document.transaction do
+      document = Document.with_current_edition.lock.find_by_param(params[:id])
+      @edition = document.current_edition
+
+      checker = Requirements::ScheduledDatetimeChecker.new(permitted_params)
+      date_issues = checker.date_issues
```
benthorner: Could we just have a `pre_submit_issues` method that collects these? We have this pattern in some of the other checkers - this means you can then do `if issues.any?` and `issues.items` :-).

app/services/requirements/scheduled_datetime_checker.rb
```diff
@@ -0,0 +1,92 @@
...
+      @month = is_numeric?(params[:month]) ? params[:month].to_i : params[:month]
+      @year = is_numeric?(params[:year]) ? params[:year].to_i : params[:year]
+      @time = params[:time]
+    end
+
+    def date_issues
+      issues = []
+
+      begin
+        parsed_date
```
benthorner: I'm not sure if we're covering all the cases in these methods, but it might be the whacky behaviour of these classes: in `time_issues`, it looks like we're not expecting the parsing to blow up, while in the `date_issues`, it looks like we're not fussed if it's empty, but that sounds like it should be an issue. Have I missed something?

app/services/requirements/scheduled_datetime_checker.rb
```diff
@@ -0,0 +1,92 @@
+# frozen_string_literal: true
+
+module Requirements
+  class ScheduledDatetimeChecker
+    attr_reader :day, :month, :year, :time
+
+    FUTURE_TIME_PERIOD = { months: 14 }.freeze
+
+    def initialize(params)
+      @day = is_numeric?(params[:day]) ? params[:day].to_i : params[:day]
```
benthorner: Do we need to do this anymore? We only use it in the context of a string, so it doesn't look like we need to convert it to an integer :-).

app/views/documents/show/_scheduling_form.html.erb
```diff
@@ -0,0 +1,52 @@
+<% publish_datetime = @edition.revision.scheduled_publishing_datetime %>
```
benthorner: Minor, but could we do the `||` up here, to simplify the `||`s down below? I guess there's a compromise with the two other checks that involve this variable on its own, but I think it'll help readability if we make the following lines a bit simpler :-).

spec/features/workflow/schedule_spec.rb
```diff
@@ -0,0 +1,69 @@
...
+    visit document_path(@edition.document)
+  end
+
+  def and_i_click_on_schedule_publishing
+    find(".govuk-details__summary-text", text: "Schedule publishing").click
+  end
+
+  def then_i_see_a_default_date_and_time_selected
+    default_date = Time.zone.tomorrow
+    expect(page).to have_selector(
```
benthorner: Another minor one, but I think we can do this with `find_field` and `.value`, which should make the expectation a little easier to read :-).

app/views/documents/show/_scheduling_form.html.erb
```diff
@@ -0,0 +1,52 @@
...
+
+<%= form_tag(save_scheduled_publishing_datetime_path(@edition.document)) do %>
+  <%= render "govuk_publishing_components/components/details", {
+    title: if publish_datetime
+             simple_format("Publish date: \n" + publish_datetime.strftime("%-d %B %Y - %l:%M%P"))
+           else
+             "Schedule publishing"
+           end
+  } do %>
+    <% if flash[:issues] %>
```
benthorner: I think we should be using the `error_summary` component here, as it looks like we're trying to summarise some errors - the checker framework is designed to be compatible with its interface, so it should be easier to slot in here, and gives better results with assistive tech if we do have multiple errors :-).

emmabeynon:
> is there a way to position the three date fields side-by-side?

We have a card in the backlog to produce a component that does this.  The date inputs are temporary (mentioned in PR description).

app/services/requirements/scheduled_datetime_checker.rb
emmabeynon: I had expected Time.parse to work in the same way as Date.parse (i.e. throwing an ArgumentError) but turns out it just returns nil if you pass in an empty string or alphanumerical characters.

app/services/requirements/scheduled_datetime_checker.rb
emmabeynon: Yep, totally missed that!

app/views/documents/show/_scheduling_form.html.erb
emmabeynon: To be honest, the error summary looks pretty horrendous in that small space.  I think it looks better visually to have just the errors listed out on their own.  I'll run it past Sonia when she's back in any case.

app/controllers/schedule_controller.rb
benthorner: Just thought of another advantage of doing this - it would help get rid of the duplication of `return unless date_issues.items.empty? && time_issues.items.empty?` :-).

app/views/documents/show/_scheduling_form.html.erb
benthorner: That makes sense - @alex-ju, do you have any thoughts about how we could improve the screen reader experience here?

alex-ju: My personal view is that we should use the error-summary placed at the top of the page - as we use it on other pages. It will point users to the input that needs to be reviewed so they won't have a problem in finding it in the current page. We may need to open the details element for them thought.. Let's see what are our designer's thoughts on this too.

app/controllers/schedule_controller.rb
```diff
@@ -0,0 +1,33 @@
+# frozen_string_literal: true
+
+class ScheduleController < ApplicationController
+  def save_scheduled_publishing_datetime
+    Document.transaction do
+      document = Document.with_current_edition.lock.find_by_param(params[:id])
+      @edition = document.current_edition
```
kevindew: Don't think we need to set this as a class instance variable as it's not going into any views.

app/controllers/schedule_controller.rb
```diff
@@ -0,0 +1,33 @@
...
+
+      checker = Requirements::ScheduledDatetimeChecker.new(permitted_params)
+      issues = checker.pre_submit_issues
+
+      if issues.any?
+        flash[:issues] = issues.items_for(:scheduled_datetime)
```
kevindew: maybe we want to name this flash more specific to the form?

kevindew: I think we also need to set a flash with data of the fields they filled in so it can populate the form.

app/controllers/schedule_controller.rb
emmabeynon: Done

PR #797, opened 2019-02-19
Inline image modal

benthorner:
https://trello.com/c/jARJ0xxA/642-use-a-modal-to-insert-an-existing-image-as-an-inline-snippet

This adds a 'glue' module to operate a modal for the purpose of inserting inline images. Please see the commits for further details.

### DIFF
--- a/app/assets/javascripts/modules/inline-image-modal.js
+++ b/app/assets/javascripts/modules/inline-image-modal.js
@@ -19,7 +19,7 @@ InlineImageModal.prototype.init = function () {
 
 InlineImageModal.prototype.fetchModalContent = function (url) {
   var controller = new window.AbortController()
-  var headers = { 'X-Requested-With': 'XMLHttpRequest' }
+  var headers = { 'Content-Publisher-Rendering-Context': 'modal' }
   var options = { credentials: 'include', signal: controller.signal, headers: headers }
   setTimeout(function () { controller.abort() }, 5000)
 
--- a/app/assets/javascripts/modules/inline-image-modal.js
+++ b/app/assets/javascripts/modules/inline-image-modal.js
@@ -1,0 +1,73 @@
+function InlineImageModal ($module) {
+  this.$module = $module
+  this.$modal = document.getElementById('modal')
+}
+
+InlineImageModal.prototype.init = function () {
+  if (!this.$module || !this.$modal) {
+    return
+  }
+
+  this.$multiSectionViewer = this.$modal
+    .querySelector('[data-module="multi-section-viewer"]')
+
+  this.$module.addEventListener('click', function (event) {
+    event.preventDefault()
+    this.performAction(this.$module)
+  }.bind(this))
+}
+
+InlineImageModal.prototype.fetchModalContent = function (url) {
+  var controller = new window.AbortController()
+  var headers = { 'X-Requested-With': 'XMLHttpRequest' }
+  var options = { credentials: 'include', signal: controller.signal, headers: headers }
+  setTimeout(function () { controller.abort() }, 5000)
+
+  return window.fetch(url, options)
+    .then(function (response) {
+      if (!response.ok) {
+        return window.Promise.reject('Unable to render the content.')
+      }
+      return response.text()
+    })
+}
+
+InlineImageModal.prototype.performAction = function (item) {
+  var handlers = {
+    'open': function () {
+      this.$modal.open()
+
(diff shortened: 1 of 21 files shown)


app/assets/javascripts/modules/inline-image-modal.js
```diff
@@ -0,0 +1,79 @@
...
+
+  this.$trigger.addEventListener('click', function (event) {
+    event.preventDefault()
+    $module.performAction($module.$trigger)
+  })
+}
+
+InlineImageModal.prototype.fetchModalContent = function (url) {
+  var controller = new window.AbortController()
+  var headers = { 'X-Requested-With': 'XMLHttpRequest' }
```
kevindew: As we chatted about the other day - since this isn't actually an XMLHttpRequest maybe we could go with a header that specifies context?

app/assets/javascripts/modules/inline-image-modal.js
```diff
@@ -0,0 +1,79 @@
...
+  var options = { credentials: 'include', signal: controller.signal, headers: headers }
+  setTimeout(function () { controller.abort() }, 5000)
```
kevindew: Is there intentions of consistency with the prefixing of window to things? I notice this isn't prefixed with window whereas fetch and promise are, and then we've also not prefixed MultiSectionViewer

app/controllers/images_controller.rb
```diff
@@ -8,6 +8,7 @@ class ImagesController < ApplicationController
 
   def index
     @document = Document.with_current_edition.find_by_param(params[:document_id])
+    render layout: (@context = "modal") if request.xhr?
```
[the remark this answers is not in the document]
benthorner: Can do. I was thinking of still using the same header, but with a value of `Modal`. Also, in order to keep the controllers simple, I was thinking of adding a `set_context` method in `ApplicationController`, and then doing a `before_action` close to where it's needed. This is based on the repetition that happens in the next PR, so it can wait for that if you like :-).

app/assets/javascripts/modules/inline-image-modal.js
benthorner: We could go with `X-Context` if you like, and this would be blank by default?

kevindew: Yup referencing context sounds good.

It's probably worth checking the guidance on custom HTTP headers, from a quick search I think we're discouraged from using the X prefix on custom. Might also be prudent to include GOVUK or Content Publisher to make it really clearly custom.

app/assets/javascripts/modules/inline-image-modal.js
benthorner: Good point - I was going from the `MarkdownEditor` component, which does the same thing for a referenced class using `global`, but both are unnecessary by using `window`. The linter doesn't actually complain about `setTimeout`, but does for a bunch of other things, which is why they have the `window.` prefix. Unless it's causing a problem (including the linter complaining about it), I'm inclined to not add the prefix.

app/assets/javascripts/modules/inline-image-modal.js
benthorner: @kevindew, from my quick reading, the `X-` prefix is discouraged except for internal uses that will definitely remain internal, which I think is true here. However, `Content-Publisher-Context` sounds good too, so I think I'll go with that.

benthorner: Done :-). Normally I might avoid putting it in ApplicationController straight away, but I've already done the next PR that demonstrates that it's used multiple times.

benthorner:
@kevindew I've updated this to use the helper method as discussed. I thought it might be a bit weird to include the `ApplicationHelper` in the `ApplicationController`, so I've done it in the `ImagesController` for now.

kevindew:
@benthorner yup does seem a bit weird to include a view helper into a controller. What I'd suggest is the method is in ApplicationController rather ApplicationHelper and then we make it available by running `helper_method :rendering_context`

benthorner:
@kevindew done - I guess it's going to be slightly hard to find, either way :-).

app/controllers/images_controller.rb
```diff
@@ -1,13 +1,16 @@
 # frozen_string_literal: true
 
 class ImagesController < ApplicationController
+  include ApplicationHelper
```
kevindew: Just need to remove this again.

benthorner: Oops - done.

PR #798, opened 2019-02-19
Remove explicit scss imports

alex-ju:
We're currently [importing govuk-frontend/all via govuk_publishing_components](https://github.com/alphagov/govuk_publishing_components/blob/master/app/assets/stylesheets/govuk_publishing_components/_all_components.scss#L9) since [#683](https://github.com/alphagov/govuk_publishing_components/pull/683), so there's no point in continuing to explicitly import only the needed dependencies as they're all available.

### DIFF
--- a/app/assets/stylesheets/components/_autocomplete.scss
+++ b/app/assets/stylesheets/components/_autocomplete.scss
@@ -1,4 +1,3 @@
-@import "govuk-frontend/helpers/typography";
 @import "vendor/accessible-autocomplete/dist/accessible-autocomplete.min";
 
 .autocomplete__option {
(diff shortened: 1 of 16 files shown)


benthorner:
@alex-ju, I notice this came from b2d3f39bf69d557db4cb0788fbbbd11e77cdae26, although this doesn't explain why the change was made. Do you think there's a performance gain from going in the opposite direction?

alex-ju:
@benthorner hard so say... I'm for explicitly defining dependencies and only importing what's needed, but there's no point in that when everything is imported from the gem. We would need some sort of a benchmark app to tests real life scenarios.

PR #800, opened 2019-02-19
Support upload and display of new images

benthorner:
https://trello.com/c/QDEuNPgp/653-allow-users-to-upload-an-inline-image-in-the-context-of-a-modal

This exposes the upload form in the modal context of the images index
page, and overrides the 'Upload' button to post the form data i.e. upload
the image. Depending on the form data, we may need to render the
response HTML to display requirements issues, or follow a redirect.

The API of the FormData class means we can support posting a form
generically, which is handled using a new 'postModalForm' function in
the existing 'glue' module called 'inline-image-modal'. We also follow
any redirects automatically, to avoid unnecessary boilerplate '.fetch's.

In future work we will implement the rest of the upload workflow, so
that the modal loads the crop page after the image is uploaded.

### DIFF
--- a/app/assets/javascripts/modules/inline-image-modal.js
+++ b/app/assets/javascripts/modules/inline-image-modal.js
@@ -65,6 +65,7 @@ InlineImageModal.prototype.performAction = function (item) {
         .then(function (text) {
           this.$multiSectionViewer.showDynamicSection(text)
           this.overrideActions()
+          this.initComponents()
         }.bind(this))
         .catch(function (result) {
           this.$multiSectionViewer.showStaticSection('error')
@@ -80,6 +81,7 @@ InlineImageModal.prototype.performAction = function (item) {
         .then(function (text) {
           this.$multiSectionViewer.showDynamicSection(text)
           this.overrideActions()
+          this.initComponents()
         }.bind(this))
         .catch(function (result) {
           this.$multiSectionViewer.showStaticSection('error')
@@ -93,12 +95,14 @@ InlineImageModal.prototype.performAction = function (item) {
 }
 
 InlineImageModal.prototype.initComponents = function () {
-  // TODO: change govuk-frontend so we can do GOVUKFrontend.initAll(dynamicSection)
+  // TODO: change ErrorSummary so it just focusses when it's initialised
   var $errorSummary = document.querySelector('[data-module="error-summary"]')
-  new GOVUKFrontend.ErrorSummary($errorSummary).init()
 
-  // TODO: change ErrorSummary to listen on the scoped element (not static window)
-  window.dispatchEvent(new window.Event('load'))
+  if (!$errorSummary) {
+    return
+  }
+
+  $errorSummary.focus()
 }
 
 InlineImageModal.prototype.overrideActions = function () {
--- a/app/assets/javascripts/modules/inline-image-modal.js
+++ b/app/assets/javascripts/modules/inline-image-modal.js
@@ -32,6 +32,30 @@ InlineImageModal.prototype.fetchModalContent = function (url) {
     })
 }
 
+InlineImageModal.prototype.postModalForm = function (formId) {
+  var form = document.getElementById(formId)
+  var controller = new window.AbortController()
+  setTimeout(function () { controller.abort() }, 10000)
+
+  var options = {
+    credentials: 'include',
+    signal: controller.signal,
+    headers: { 'Content-Publisher-Rendering-Context': 'modal' },
+    redirect: 'follow',
+    method: 'POST',
+    body: (new window.FormData(form))
+  }
+
+  return window.fetch(form.action, options)
+    .then(function (response) {
+      if (!response.ok) {
+        return window.Promise.reject('Unable to render the content.')
+      }
+
+      return response.text()
+    })
+}
+
 InlineImageModal.prototype.performAction = function (item) {
   var handlers = {
     'open': function () {
@@ -50,12 +74,31 @@ InlineImageModal.prototype.performAction = function (item) {
       var editor = this.$module.closest('[data-module="markdown-editor"]')
       this.$modal.close()
       editor.selectionReplace(item.dataset.modalData)
+    },
+    'upload': function () {
+      this.postModalForm(item.dataset.modalActionForm)
+        .then(function (text) {
+          this.$multiSectionViewer.showDynamicSection(text)
+          this.overrideActions()
+        }.bind(this))
+        .catch(function (result) {
+          this.$multiSectionViewer.showStaticSection('error')
+        }.bind(this))
     }
   }
 
   this.$modal.focusDialog()
   this.$multiSectionViewer.showStaticSection('loading')
   handlers[item.dataset.modalAction].bind(this)()
+}
+
+InlineImageModal.prototype.initComponents = function () {
+  // TODO: change govuk-frontend so we can do GOVUKFrontend.initAll(dynamicSection)
+  var $errorSummary = document.querySelector('[data-module="error-summary"]')
+  new GOVUKFrontend.ErrorSummary($errorSummary).init()
+
+  // TODO: change ErrorSummary to listen on the scoped element (not static window)
+  window.dispatchEvent(new window.Event('load'))
 }
 
 InlineImageModal.prototype.overrideActions = function () {
--- a/app/assets/javascripts/modules/inline-image-modal.js
+++ b/app/assets/javascripts/modules/inline-image-modal.js
@@ -113,6 +113,11 @@ InlineImageModal.prototype.overrideActions = function () {
       event.preventDefault()
       this.performAction(item)
     }.bind(this))
+
+    item.addEventListener('submit', function (event) {
+      event.preventDefault()
+      this.performAction(item)
+    }.bind(this))
   }.bind(this))
 }
 
--- a/app/views/images/index.html.erb
+++ b/app/views/images/index.html.erb
@@ -3,35 +3,37 @@
 
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
+    <div class="app-pane">
+      <% upload_image_heading = capture do %>
+        <h2 class="govuk-heading-m">
+          <%= t("images.index.upload_image") %>
+        </h2>
+      <% end %>
 
-    <% unless rendering_context == 'modal' %>
-      <div class="app-pane">
-        <% upload_image_heading = capture do %>
-          <h2 class="govuk-heading-m">
-            <%= t("images.index.upload_image") %>
-          </h2>
-        <% end %>
+      <%= form_tag(
+        create_image_path(@document),
+        id: "image-upload",
+        multipart: true,
+        data: { gtm: "image-upload-submit" },
+      ) do %>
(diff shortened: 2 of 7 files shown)


app/assets/javascripts/modules/inline-image-modal.js
```diff
@@ -63,6 +100,15 @@ InlineImageModal.prototype.performAction = function (item) {
   handlers[item.dataset.modalAction].bind(this)()
 }
 
+InlineImageModal.prototype.initComponents = function () {
+  // TODO: change govuk-frontend so we can do GOVUKFrontend.initAll(dynamicSection)
+  var $errorSummary = document.querySelector('[data-module="error-summary"]')
+  new window.GOVUKFrontend.ErrorSummary($errorSummary).init()
+
+  // TODO: change ErrorSummary to listen on the scoped element (not static window)
+  window.dispatchEvent(new window.Event('load'))
```
kevindew: Oh gosh - I see what you mean now about the `window.load` thing we were discussing yesterday. This is rather unusual to synthetically call the page load event as there is an anticipation it will only be called once.

Btw that couldn't be changed to be listen to a scope as I think only window has a load event, but it could have a separate function called to initialise.

Anyway from looking at [govuk-frontend/error-summary.js](https://github.com/alphagov/govuk-frontend/blob/master/src/components/error-summary/error-summary.js) I do wonder if it'll be doing some things that are either undesirable or problematic in a modal context. 

The undesirable aspect would be that it adds to the history state which enables back button access. I don't know fully how this works, however I'd anticipate that for us, we could end up with a back button that does nothing once the modal isn't displayed.

The other thing that I imagine either just works fine or is painful is the auto scrolling that the error summary does since the modal is in a position fixed I don't know if it'd scroll the window or the element we want.

benthorner: @kevindew agreed, it's not ideal. I reckon it's actually unnecessary to listen to the event, and the `focus()` can just be called immediately - @alex-ju reckons it's a relic from an older version of this component, and we're going to try and remove the event listener.

The scrolling seems to work fine, especially as we expect the modal to occupy most of the page anyway. I hadn't noticed the 'back' behaviour though, which isn't an issue for the inline image work, as none of the error summaries have any links in their content. I think this is something we could resolve (i.e. turn off) if the component was better isolated from the window, so I'm not too worried about it, although I'll be careful to check this in our local ones.

alex-ju: I proposed a change in `govuk-frontend` to [remove the dependency on `document`](https://github.com/alphagov/govuk-frontend/pull/1215), but I get that that that's there to make sure the component is not being initialised twice.

benthorner:
@kevindew this is now ready, when you get a moment :-).

app/assets/javascripts/modules/inline-image-modal.js
```diff
@@ -32,6 +32,30 @@ InlineImageModal.prototype.fetchModalContent = function (url) {
     })
 }
 
+InlineImageModal.prototype.postModalForm = function (formID) {
+  var form = document.getElementById(formID)
```
kevindew: super minor, but ID => Id

app/assets/javascripts/modules/inline-image-modal.js
```diff
@@ -32,6 +32,30 @@ InlineImageModal.prototype.fetchModalContent = function (url) {
...
+  var controller = new window.AbortController()
+  setTimeout(function () { controller.abort() }, 10000)
+
+  var options = {
+    credentials: 'include',
+    signal: controller.signal,
+    headers: { 'Content-Publisher-Rendering-Context': 'modal' },
+    redirect: 'follow',
+    method: 'POST',
+    body: (new window.FormData(form))
```
kevindew: again minor, but these brackets look superfluous

kevindew: Did we confirm that using FormData object was sufficiently supported cross-browser?

app/assets/javascripts/modules/inline-image-modal.js
```diff
@@ -50,6 +75,17 @@ InlineImageModal.prototype.performAction = function (item) {
       var editor = this.$module.closest('[data-module="markdown-editor"]')
       this.$modal.close()
       editor.selectionReplace(item.dataset.modalData)
+    },
+    'upload': function () {
+      this.postModalForm(item.dataset.modalActionForm)
```
kevindew: Are we vulnerable at this point to dual submissions? I assume that nothing changes in the UI once they submit the form until we receive a response which could be a while for an upload.

Could well be that we have the same problem with the non modal form of this too but feedback seems to be clearer in a browser when you're changing a whole page.

app/assets/javascripts/modules/inline-image-modal.js
kevindew: Rather than having this synthetic load event could we do `$errorSummary.focus()` ?

That feels like it'd less potential consequences and we're already coupling this load behaviour to that particular component.

app/views/images/index.html.erb
```diff
@@ -3,35 +3,37 @@
...
-            text: "Upload",
-          } %>
-        <% end %>
-      </div>
-    <% end %>
+      <%= form_tag(
+        create_image_path(@document),
+        id: "image-upload",
+        multipart: true,
+        data: { gtm: "image-upload-submit" },
```
kevindew: Do you know if GTM will still receive this event? I assume this listens to form submit events but it doesn't look like we listen to any of those.

app/views/images/index.html.erb
```diff
@@ -3,35 +3,37 @@
...
+          name: "image"
+        } %>
+
+        <%= render_govspeak(t("images.index.description_govspeak")) %>
+
+        <%= render "govuk_publishing_components/components/button", {
+          text: "Upload",
+          data_attributes: {
+            "modal-action": "upload",
+            "modal-action-form": "image-upload",
```
kevindew: As we put the action rules on this button are we vulnerable to this not working if someone submits this form without clicking the button.

I'm not sure if we can do it on this form as it only has the input file which is a very browser specific action, but once we have the edit form and pressing enter can submit the form these events wouldn't be caught.

app/assets/javascripts/modules/inline-image-modal.js
benthorner: Oh yeah - it's not Ruby :-).

app/assets/javascripts/modules/inline-image-modal.js
benthorner: So we are actually changing the modal to the loading screen - this happens before any action is invoked, as otherwise it kind of defeats the point of the loading screen ;-). I think there's a separate question, which you also asked, about the behaviour of short-lived loading screens.

I guess they can potentially click the button twice, really quickly, although I can't do this myself - in a very slow browser, perhaps.

app/views/images/index.html.erb
benthorner: @alex-ju is currently working on rewriting our local `modules/` to support re-initialisation, just like components. I think you're referring to `GTMFormListener` here, so it should be fixed by that :-).

app/views/images/index.html.erb
benthorner: Ooh, good thought - that hadn't occurred to me. To my surprise, it works out-of-the-box, thanks to the spec: https://www.w3.org/TR/html52/sec-forms.html#implicit-submission. I realise that's HTML, but it seems to work in IE11 as well, and I'm just checking other browsers.

app/assets/javascripts/modules/inline-image-modal.js
benthorner: Yup 👍 - even made it an A/C on the card.

app/assets/javascripts/modules/inline-image-modal.js
kevindew: Oh ok - I think I might just not be understanding how this works. Is this correct?

- When I click the button to submit the form it prevents default on that event
- It focuses the modal again
- It hides the form and shows the loading view
- it pulls the data from the hidden view and submits it
- that form is only cleared once showDynamic section is called again?

app/assets/javascripts/modules/inline-image-modal.js
benthorner: Done - turns out window events are only used by two components, and we're hoping to get rid of it on this one, so the intent is no longer to 'simulate a window loading'.

app/views/images/index.html.erb
kevindew: I was referring to lack of a submit event existing. If GTM is listening for form submit events then this one will never be called because we preventDefault on the button click event.

GTM likely has all sorts of voodoo though so this may not be remotely a problem.

app/views/images/index.html.erb
kevindew: Oh cool - I had no idea about the implicit submission.

kevindew:
Glad the submit thing wasn't much of a pain. Thinking about it it would be pretty trivial to support form submissions too if we had https://github.com/alphagov/content-publisher/blob/1800d6e00bda22c5b68ae72756b42c20fbbb3e07/app/assets/javascripts/modules/inline-image-modal.js#L110 listen on submit events as well and set data on form rather than buttons.

spec/features/editing_images/upload_image_modal_requirements_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+
+  def and_i_click_to_insert_an_image
+    within(".app-c-markdown-editor") do
+      find("markdown-toolbar details").click
+      click_on "Image"
+    end
+  end
+
+  def and_i_upload_an_invalid_image
+    find('form input[type="file"]').set(Rails.root.join(file_fixture("text-file.txt")))
```
kevindew: I wrote a comment on this before but it doesn't seem to be here so apologies if it's come through twice - I'm assuming I just lost it in the review

I notice we didn't need the Rails.root.join on https://github.com/alphagov/content-publisher/blob/7386e790f9669c4bb3fbd9e0ea7d15fd40b6494c/spec/features/editing_images/upload_image_requirements_spec.rb#L22 so wondering if this is needed or special.

app/assets/javascripts/modules/inline-image-modal.js
benthorner: Yes, that's right - is there still something you're concerned about here?

kevindew: Nope - just taking my old brain a while to understand

app/views/images/index.html.erb
benthorner: You're right - it wouldn't have worked. I've just added a commit that, with the work @alex-ju is doing, will preserve the current behaviour.

spec/features/editing_images/upload_image_modal_requirements_spec.rb
benthorner: Yes, it was strange - I encountered this and forgot to come back to it. I've added another commit to explain, but it seems like absolute paths are required for JS feature specs.

kevindew: Oh that's a pain - thanks for the commit explanation

PR #801, opened 2019-02-19
Send user uid value to Google Tag Manager

kevindew:
Trello: https://trello.com/c/ZhLNktSm/644-add-user-id-tracking-to-content-publisher

This is so we can monitor session behaviour of particular users and see
when they're using different browsers/devices.

According to Andy it is acceptable privacy wise to send an id that has a
meaning in our internal systems but cannot be used out of context to
identify the user. I debated whether it would be better to run this
through a one way cryptographic hash function such as SHA1 but I'm not
sure that offers any actual benefits.

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -5,7 +5,10 @@
   <% end %>
 
   <script>
-    var dataLayer = [{ 'dimension1': '<%= current_user.organisation_slug %>' }];
+    var dataLayer = [{
+      'dimension1': '<%= current_user.organisation_slug %>',
+      'cp_uid': '<%= current_user.uid %>'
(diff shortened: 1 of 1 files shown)


kevindew:
> I see it's not introduced in this PR, but should we be using `&.` here to be on the safe side?

I don't think that is necessary as there are other parts of this file that rely on current_user being set and we have no scenarios considered where we anticipate unauthenticated access.

Similarly I'd say by doing that it looks like we have a clear plan of what to send if/when we did allow unathenticated access, when actually it's just a fallback and instead it might be that we'd want to send very different data to GTM if we had unauthenticated access.

kevindew:
@benthorner thanks 👍

PR #803, opened 2019-02-21
Add toolbar dropdown component

alex-ju:
This PR adds a [toolbar dropdown component](https://content-publisher-revie-pr-803.herokuapp.com/component-guide/toolbar_dropdown) ✨ to be used within the markdown toolbar to group "insert" action buttons (such as images and contacts).

This component does not require JavaScript to be toggled, so the actions will be accessible in a no-js experience.

This implementation is purely functional. Have had a chat with @soniaturcotte and we'll iterate the style of this component.

Another PR will follow this one to use the component in the markdown toolbar to launch a modal dialogue for each of these two "insert" actions.

### Dropdown in markdown toolbar - default
screenshot-content-publisher-revie-pr-803 herokuapp com-2019 02 21-15-52-52

### Dropdown in markdown toolbar - opened
screenshot-content-publisher-revie-pr-803 herokuapp com-2019 02 21-15-53-09

[Trello card](https://trello.com/c/Wf0A8zsq)

### DIFF
--- a/app/views/components/_toolbar_dropdown.html.erb
+++ b/app/views/components/_toolbar_dropdown.html.erb
@@ -1,0 +1,30 @@
+<%
+  id ||= "toolbar-dropdown-#{SecureRandom.hex(4)}"
+  align ||= false
+  items ||= []
+  data_attributes ||= {}
+  data_attributes[:module] = 'toolbar-dropdown'
+  css_classes = %w(app-c-toolbar-dropdown)
+  css_classes << 'app-c-toolbar-dropdown--end' if align == 'right'
+%>
+
+<%= tag.details id: id, class: css_classes, data: data_attributes do %>
+  <%= tag.summary title, class: "app-c-toolbar-dropdown__title" %>
+  <%= tag.div class: "app-c-toolbar-dropdown__container" do %>
+    <% if items.any? %>
+      <%= tag.ul class: "app-c-toolbar-dropdown__list" do %>
+        <% items.each do |item| %>
+          <%= tag.li class: "app-c-toolbar-dropdown__list-item" do %>
+            <%= tag.button item[:text],
+              class:"app-c-toolbar-dropdown__button",
+              type: "submit",
+              name: item[:name],
+              value: item[:value],
+              data: item[:data_attributes]
+            %>
+          <% end %>
+        <% end %>
+      <% end %>
+    <% end %>
+  <% end %>
+<% end %>
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -16,6 +16,7 @@
 //= require whatwg-fetch/dist/fetch.umd.js
 
 //= require components/autocomplete.js
+//= require components/toolbar-dropdown.js
 //= require components/image-cropper.js
 //= require components/input-length-suggester.js
 //= require components/markdown-editor.js
--- a/app/assets/javascripts/components/toolbar-dropdown.js
+++ b/app/assets/javascripts/components/toolbar-dropdown.js
@@ -1,0 +1,41 @@
+window.GOVUK = window.GOVUK || {}
+window.GOVUK.Modules = window.GOVUK.Modules || {};
+
+(function (Modules) {
+  function ToolbarDropdown () { }
+
+  ToolbarDropdown.prototype.start = function ($module) {
+    this.$module = $module[0]
+    this.$title = this.$module.querySelector('.app-c-toolbar-dropdown__title')
+    this.$container = this.$module.querySelector('.app-c-toolbar-dropdown__container')
+    this.$buttons = this.$module.querySelectorAll('.app-c-toolbar-dropdown__button')
+
+    this.$module.addEventListener('blur', ToolbarDropdown.prototype.handleBlur.bind(this), true)
+    this.$buttons.forEach(function ($button) {
+      $button.addEventListener('click', ToolbarDropdown.prototype.handleButtonClick.bind(this), true)
+    }, this)
+  }
+
+  ToolbarDropdown.prototype.handleBlur = function (event) {
+    var target = event.relatedTarget
+    if (!target) {
+      target = document.activeElement
+    }
+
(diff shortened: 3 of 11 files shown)


app/assets/javascripts/components/dropdown.js
```diff
@@ -0,0 +1,44 @@
...
+    this.closeDropdown()
+  }
+}
+
+Dropdown.prototype.handleButtonClick = function (event) {
+  this.closeDropdown()
+}
+
+Dropdown.prototype.closeDropdown = function (event) {
+  if (this.$module.hasAttribute('open')) {
```
benthorner: I'm guessing we need this for when we tab over the dropdown?

app/views/components/_dropdown.html.erb
```diff
@@ -0,0 +1,21 @@
+<%
+  id ||= "image-meta-#{SecureRandom.hex(4)}"
+  items ||= []
+  data_attributes ||= {}
+  data_attributes[:module] = 'dropdown'
+%>
+
+<%= tag.details class: "app-c-dropdown", data: data_attributes do %>
```
benthorner: W3C indicates this doesn't work on IE/Edge. Is that a problem?

app/assets/javascripts/components/dropdown.js
```diff
@@ -0,0 +1,44 @@
+function Dropdown ($module) {
+  if (!$module) $module = document
+  this.$module = $module
+  this.$title = $module.querySelector('.app-c-dropdown__title')
+  this.$container = $module.querySelector('.app-c-dropdown__container')
+  this.$buttons = $module.querySelectorAll('.app-c-dropdown__button')
+}
+
+Dropdown.prototype.handleBlur = function (event) {
```
benthorner: I'm guessing this is partly here to cope with the fact that we override the button click events in the inline-image-module? It took a little while to understand the `relatedTarget` stuff, although I'm still not sure why we can't just do `this.closeDropdown()` on blur, as it checks the `open` state. Can you explain a bit?

app/views/components/_dropdown.html.erb
alex-ju: We have a details polyfill covering for IE/Edge.

app/assets/javascripts/components/dropdown.js
alex-ju: Yep

app/assets/javascripts/components/dropdown.js
alex-ju: Blur can happen for both moving to an element inside or outside the component. With `relatedTarget` we try to capture where the user goes to after opening the dropdown. If he goes inside, we don't want to close the dropdown.

app/assets/javascripts/components/dropdown.js
```diff
@@ -0,0 +1,44 @@
+function Dropdown ($module) {
+  if (!$module) $module = document
```
kevindew: Are we expecting to need this?

app/views/components/_dropdown.html.erb
```diff
@@ -0,0 +1,21 @@
+<%
+  id ||= "image-meta-#{SecureRandom.hex(4)}"
```
kevindew: Looks like this is left over from a copy + paste (image-meta)? It doesn't look like we use this id.

app/views/components/docs/dropdown.yml
```diff
@@ -0,0 +1,44 @@
...
+          data_attributes:
+            module: inline-image-modal
+            modal-action: open
+        - text: Contact
+          data_attributes:
+            module: inline-contact-modal
+            modal-action: open
+
+  within_markdown_editor_toolbar:
+    embed: |
```
kevindew: yay embed 🎉

app/views/components/_dropdown.html.erb
alex-ju: Oh.. seems like that. Will update now.

app/assets/javascripts/components/dropdown.js
alex-ju: Most probably not.

alex-ju:
> My big one question though is with regards to the naming. I wonder if "Dropdown" is quite bold unless we anticipate using this in other scenarios? Do you think it's coupled to usage with the toolbar, in which case maybe it's name should reflect that? I ask specifically as it only accepts button elements so wouldn't be useful in a context where we wanted a dropdown of links.

@kevindew I agree with the naming, I went to wide with it. we can name it `toolbar-dropdown` or something similar to tie it with the current usage and leave the code as is, or we can extend it to support links and we can name it `dropdown-menu` (or something similar).

kevindew:
> @kevindew I agree with the naming, I went to wide with it. we can name it `toolbar-dropdown` or something similar to tie it with the current usage and leave the code as is, or we can extend it to support links and we can name it `dropdown-menu` (or something similar).

FWIW I'd go with toolbar-dropdown and we can always change that if we want to use it wider. Whereas planning for it wider from the start might lead to code we maintain that's never used.

alex-ju:
@kevindew renamed and addressed comments

app/views/components/_toolbar_dropdown.html.erb
```diff
@@ -0,0 +1,21 @@
+<%
+  id ||= "toolbar-dropdown-#{SecureRandom.hex(4)}"
```
kevindew: Thanks for changing this - I don't think we actually use this id though?

alex-ju: We're not, but we're doing this for most of our components. Happy to drop it

kevindew: It's not actually set on the details element so all we're doing is setting a variable.

alex-ju: 🤦‍♂️ now I get what you meant.

benthorner:
@alex-ju can we change this to use the new `Modules` style?

alex-ju:
@benthorner updated

PR #804, opened 2019-02-21
[WIP] Modify when change note guidance appears

1pretz1:
This passes a data_attribute for contextual guidance to the radio button
component, this means that change note guidance will only appear when a
user selects a major change and will disappear if the user selects a
minor change.

JS for contextual guidance adds event listeners to all elements with
data attribute of `data-contextual-guidance` to handle displaying
and hiding a components guidance, this enables the minor / major change
note radio button to benefit from this functionality, instead of before
where only focusing on the textarea would display the guidance.

Dependant on this for govuk_publishing_components:
alphagov/govuk_publishing_components#766

Trello:
https://trello.com/c/pcbHvl7B/636-change-note-guidance

### DIFF
--- a/app/views/documents/edit/_change_notes.html.erb
+++ b/app/views/documents/edit/_change_notes.html.erb
@@ -27,11 +27,13 @@
             text: t("documents.edit.update_type.major_name"),
             conditional: textarea,
             checked: @revision.major?,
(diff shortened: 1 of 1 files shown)


alex-ju:
@1pretz1 this PR is superseded by #823. thanks for the initial work that allows us to pass data attributes to the radios

PR #807, opened 2019-02-22
Add loading spinner component

alex-ju:
This PR adds a [loading spinner component](https://content-publisher-revie-pr-807.herokuapp.com/component-guide/loading_spinner) ✨ which displays an animated graphic to indicate loading activity. In our application it will be used when transitioning from one section to another within the 'multi section viewer' component. It relies on SVG and CSS animation.

### Performance notes
GPU Memory Usage - Google Chrome v 72

screen shot 2019-02-26 at 10 51 50

[Trello card](https://trello.com/c/qouF8txw)

### DIFF
--- a/app/assets/stylesheets/components/_loading-spinner.scss
+++ b/app/assets/stylesheets/components/_loading-spinner.scss
@@ -1,6 +1,100 @@
-$loading-spinner-container-padding: govuk-spacing(8) + govuk-spacing(8);
-
 .app-c-loading-spinner {
+  @include govuk-focusable;
   text-align: center;
-  padding: $loading-spinner-container-padding;
+}
+
+@keyframes app-loading-spinner-opacity-0 {
+  0% { opacity: 0.250075; }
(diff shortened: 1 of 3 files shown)


benthorner:
@alex-ju this might need a bit more testing, but it looks like this component uses a significant amount of CPU. I was resizing my browser and noticed it stuttered a bit, which lead me to take a look `top`. It looks like it requires around 30% CPU time in order to render the animation, but that's a very rough result :-|.

alex-ju:
@benthorner that sounds bad 😞. I'll look into it.

alex-ju:
@benthorner I did a bit of testing and it seems like CSS animation affects SVG and HTML elements in the same way (so no change in performance from this point) - the GPU results in Chrome.
Since the spike in CPU and GPU usage only happens for the duration of the spinner being presented/animated I consider this something to look into and find out from users.

benthorner:
👍

PR #815, opened 2019-02-27
Inline image meta

benthorner:

This adds the metadata screen to the existing upload/edit workflow for inline images. I also came around to restoring the dubious 'wizard' pattern, in order to modify the workflow when an image is being uploaded, so that the user can directly 'Save and insert' it. Finally, this also adds 'Back' links for the crop and metadata screens, and does a bit of initial refactoring on the inline-image module.


benthorner:
@kevindew @alex-ju the dust has settled - this one should be ready to review again :-).

app/assets/javascripts/modules/inline-image-modal.js
```diff
@@ -109,20 +116,37 @@ InlineImageModal.prototype.performAction = function (item) {
...
+      this.postModalForm(item)
+        .then(function (result) {
+          if (result.finished) {
+            var editor = this.$module.closest('[data-module="markdown-editor"]')
+            this.$modal.close()
+            editor.selectionReplace(item.dataset.modalData)
+          } else {
+            this.$multiSectionViewer.showDynamicSection(result.body)
+            this.overrideActions()
+            this.initComponents()
```
kevindew: I wonder if you're getting to a point where it'd be useful to have a single function that does the show, override and init?

app/assets/javascripts/modules/inline-image-modal.js
```diff
@@ -57,103 +58,62 @@ InlineImageModal.prototype.postModalForm = function (form) {
...
         })
     })
 }
 
+InlineImageModal.prototype.renderResponse = function (response) {
+  response
+    .then(function (result) {
+      this.$multiSectionViewer.showDynamicSection(result.body)
+      this.overrideActions()
+      this.initComponents()
```
kevindew: I wonder if you'd want to make these three things a single method so that this logic doesn't have to be repeated in metaInsert

benthorner: Indeed - I wanted to avoid too much indirection in this module, with too many methods. Most of the duplication is gone, but I'll try and address this when I do the refactoring PR to extract out generic functionality from this module.

PR #817, opened 2019-02-27
Users can set content to scheduled

emmabeynon:
For https://trello.com/c/t0mha6An/649-schedule-content-for-publication

This PR allows users to set content to a `scheduled` state in the application.  Enabling scheduled publishing functionality via the Publishing API will be tackled in a separate PR.

Content must be in an editable state in order to be scheduled; it cannot be in a live state.  It must also have a `scheduled_publishing_datetime` set.  There is some existing validation for `scheduled_publishing_datetime`, and this PR introduces a check that this value is not in the past or too close to now when the user goes to schedule the content for publication.  There isn't a clear decision on how to handle this check failing, so for now failure will result in the "Schedule" button not displaying.  However in edge cases where users may access the scheduling confirmation page regardless, we raise an error.

__Schedule button__
screen shot 2019-02-27 at 14 32 12

**Review options**
screen shot 2019-02-27 at 14 34 24

**Scheduling confirmation**
screen shot 2019-02-27 at 14 34 36

**Scheduled status**
screen shot 2019-02-27 at 14 31 28

### DIFF
--- a/app/models/edition.rb
+++ b/app/models/edition.rb
@@ -102,6 +102,13 @@ class Edition < ApplicationRecord
     !live? && !scheduled?
   end
 
+  def schedulable?
+    return false unless editable?
+    return false if scheduled_publishing_datetime.nil?
+
+    scheduled_publishing_datetime > Time.zone.now
+  end
+
   def resume_discarded(live_edition, user)
     revision = live_edition.revision.build_revision_update(
       { change_note: "", update_type: "major" },
--- a/app/models/edition.rb
+++ b/app/models/edition.rb
@@ -99,7 +99,7 @@ class Edition < ApplicationRecord
   end
 
   def editable?
-    !live?
+    !live? && !scheduled?
   end
 
   def resume_discarded(live_edition, user)
--- a/app/models/edition.rb
+++ b/app/models/edition.rb
@@ -64,6 +64,8 @@ class Edition < ApplicationRecord
            :scheduled_publishing_datetime,
            to: :revision
 
+  MINIMUM_SCHEDULING_TIME = { minutes: 15 }.freeze
+
   def self.create_initial(document, user = nil, tags = {})
     revision = Revision.create_initial(document, user, tags)
     status = Status.create!(created_by: user,
@@ -106,7 +108,7 @@ class Edition < ApplicationRecord
     return false unless editable?
     return false if scheduled_publishing_datetime.nil?
 
-    scheduled_publishing_datetime > Time.zone.now
+    scheduled_publishing_datetime > Time.zone.now.advance(MINIMUM_SCHEDULING_TIME)
   end
 
   def resume_discarded(live_edition, user)
--- a/app/models/edition.rb
+++ b/app/models/edition.rb
@@ -61,6 +61,7 @@ class Edition < ApplicationRecord
            :lead_image_revision,
            :image_revisions,
            :image_revisions_without_lead,
+           :scheduled_publishing_datetime,
            to: :revision
 
   def self.create_initial(document, user = nil, tags = {})
--- a/app/views/documents/show/_actions.html.erb
+++ b/app/views/documents/show/_actions.html.erb
@@ -65,7 +65,7 @@
 
       <%= render "govuk_publishing_components/components/button", text: "Preview", href: preview_document_path(@edition.document), secondary: true %>
 
-      <% if @edition.scheduled_publishing_datetime.present? && !@edition.scheduled? %>
+      <% if @edition.schedulable? && current_user.has_permission?(User::PRE_RELEASE_FEATURES_PERMISSION) %>
         <%= link_to "Schedule", scheduling_confirmation_path(@edition.document), class: "govuk-link govuk-link--no-visited-state" %>
       <% else %>
         <%= link_to "Publish", publish_confirmation_path(@edition.document), class: "govuk-link govuk-link--no-visited-state" %>
--- a/app/views/documents/show/_actions.html.erb
+++ b/app/views/documents/show/_actions.html.erb
@@ -78,7 +78,7 @@
       <%= link_to "View published edition on GOV.UK", EditionUrl.new(@edition.document.live_edition).public_url, class: "govuk-link govuk-link--no-visited-state" %>
     <% end %>
 
-    <% if (@edition.submitted_for_review? || @edition.draft?) && current_user.has_permission?(User::PRE_RELEASE_FEATURES_PERMISSION) %>
+    <% if (@edition.editable? || @edition.scheduled?) && current_user.has_permission?(User::PRE_RELEASE_FEATURES_PERMISSION) %>
       <hr class="govuk-section-break govuk-section-break--m govuk-section-break--visible" />
       <%= render "documents/show/scheduling_form" %>
     <% end %>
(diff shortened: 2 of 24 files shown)


spec/features/workflow/schedule_set_datetime_passes_spec.rb
```diff
@@ -0,0 +1,27 @@
+# frozen_string_literal: true
+
+RSpec.feature "Set scheduled publishing date and time passes before scheduling" do
+  scenario do
+    given_there_is_an_edition_with_a_scheduled_publishing_date_and_time_set_before_now
+    when_i_visit_the_summary_page
+    then_i_cannot_see_a_schedule_link
+  end
+
+  def given_there_is_an_edition_with_a_scheduled_publishing_date_and_time_set_before_now
```
kevindew: This method name feels pretty big - could it be "given_there_is_an_edition_with_a_past_scheduled_publishing_datetime" (which admittedly isn't much shorter) 

I thought maybe you could get away with `given_there_is_an_edition_scheduled_for_the_past` but that felt a bit wrong as status isn't scheduled.

spec/features/workflow/schedule_set_datetime_passes_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+    @edition = create(:edition)
+    user = User.first
+    datetime = Time.zone.now - 1
+    revision = @edition.revision.build_revision_update(
+      { scheduled_publishing_datetime: datetime }, user
+    )
+    @edition.assign_revision(revision, user).save!
```
kevindew: We can hopefully do all of this through the factory by: `create(:edition, scheduled_publishing_datetime: Time.zone.now - 1)`

It may need the factory updating it lacks that attribute.

app/views/documents/show/_actions.html.erb
```diff
@@ -53,15 +53,23 @@
...
     <% else %>
       <%= form_tag submit_document_for_2i_path(@edition.document) do %>
         <%= render "govuk_publishing_components/components/button", text: "Submit for 2i review" %>
       <% end %>
 
       <%= render "govuk_publishing_components/components/button", text: "Preview", href: preview_document_path(@edition.document), secondary: true %>
 
-      <%= link_to "Publish", publish_confirmation_path(@edition.document), class: "govuk-link govuk-link--no-visited-state" %>
-
+      <% if @edition.schedulable? && current_user.has_permission?(User::PRE_RELEASE_FEATURES_PERMISSION) %>
```
kevindew: Do we need to be considering here whether it's in that 15 minute window?

spec/features/workflow/schedule_spec.rb
```diff
@@ -66,4 +76,32 @@ def then_i_see_the_edition_has_a_set_scheduled_publishing_date_and_time
...
+  def then_i_see_the_edition_has_been_scheduled
+    expect(page).to have_content(I18n.t!("user_facing_states.scheduled.name"))
+  end
+
+  def and_i_can_no_longer_see_a_schedule_action
+    expect(page).not_to have_link("Schedule")
+  end
+
+  def and_i_can_no_longer_edit_the_content
+    expect(page).not_to have_link("Change")
```
kevindew: I'm not sure if this holds right. As even when the content isn't editable they can change topics and that would mean the page has a "Change" link

spec/services/schedule_service_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+
+    context "when the edition has not been reviewed" do
+      it "sets the scheduling reviewed state to false" do
+        edition = create(:edition)
+        ScheduleService.new(edition).schedule(reviewed: false)
+
+        expect(edition.status.details.reviewed).to be false
+      end
+    end
+  end
```
kevindew: These were all nice and easy to read 👍

spec/features/workflow/schedule_spec.rb
```diff
@@ -10,6 +10,16 @@
...
+
+    when_i_click_on_schedule
+    and_i_select_the_content_has_been_reviewed_option
+    and_click_publish
+    then_i_see_a_confirmation_that_the_edition_has_been_scheduled
+
+    when_i_visit_the_summary_page
+    then_i_see_the_edition_has_been_scheduled
+    and_i_can_no_longer_see_a_schedule_action
+    and_i_can_no_longer_edit_the_content
```
kevindew: This feature test seems to have got rather long. I wonder if it's actually testing two distinct features: setting a scheduled_publishing_datetime and scheduling an edition. What do you think?

spec/features/workflow/schedule_set_datetime_passes_spec.rb
benthorner: Could we split the step in two?

spec/features/workflow/schedule_spec.rb
benthorner: @kevindew we do use this approach already for the `edition_spec`. The reason it works it because topics are optional for a document type - it was unclear what other visual queue we could use to test this with.

kevindew: Oh yeah shoot I forgot about that being conditional.

spec/features/workflow/schedule_set_datetime_passes_spec.rb
emmabeynon: Not sure it makes sense to split it in two when it's setting up the edition to be in the state that we need it to be. Could do `given_there_is_an_edition_with_past_scheduled_publishing_datetime`?

app/views/documents/show/_actions.html.erb
emmabeynon: `#schedulable?` checks for that.

spec/features/workflow/schedule_set_datetime_passes_spec.rb
benthorner: How about `given_there_is_an_edition_to_be_scheduled_in_the_past`?

app/models/edition.rb
```diff
@@ -98,7 +101,14 @@ def self.create_next_edition(preceding_edition, user)
   end
 
   def editable?
-    !live?
+    !live? && !scheduled?
+  end
+
+  def schedulable?
+    return unless editable?
+    return if scheduled_publishing_datetime.nil?
```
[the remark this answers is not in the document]
emmabeynon: Done

spec/features/workflow/schedule_set_datetime_passes_spec.rb
emmabeynon: I'm not sure that makes it clear what exactly is in the past.

spec/features/workflow/schedule_spec.rb
emmabeynon: Are we good to keep this in?

spec/features/workflow/schedule_spec.rb
emmabeynon: Now split into two test files.

spec/features/workflow/schedule_set_datetime_passes_spec.rb
benthorner: I think it's a bit of a compromise with readability. For example, we've got steps like `when_i_go_to_the_tags_page`, which leave it ambiguous as to how that happens. I think it's harder to understand what 'past scheduled publishing datetime' means - one too many adjectives for me ;-).

spec/features/workflow/schedule_set_datetime_passes_spec.rb
emmabeynon: I've updated the factories to handle `scheduled_publishing_datetime` 👍

PR #818, opened 2019-02-27
Support deletion of images in a modal context

benthorner:
https://trello.com/c/MkWywL4g/590-support-inline-images

There's no specific card for this, but it seems to be part of the design 
and it only took 20 minutes :-).

This adds a delete action for all non-lead images in the inline-images
modal. As we only show the lead image for context, this action is not
supported for it. One slight surprise was that we actually do a POST
request for deletions, using a hidden parameter to override the method.


benthorner:
@kevindew I've done the same for these specs, too - look OK to you?

benthorner:
Discussed with @kevindew offline - ready to merge...

PR #822, opened 2019-02-28
Fix flaky Scheduled Datetime Checker tests

emmabeynon:
We discovered some flakiness in the setup of Scheduled Datetime Checker tests
that were attempting to change a Time to be in the past by subtracting 2.
Unfortunately this actually subtracted 2 from the 'Day' value of Time, so in
instances where it was the 2nd of the month the day would be set to 0 which is
invalid. This introduces a more robust way of setting the Time to be in the
past.


alex-ju:
Thanks! I thought I was doing something wrong :)

PR #823, opened 2019-02-28
Update contextual guidance

alex-ju:
This PR refactors the contextual guidance as follows:
- remove relative option on contextual guidance and position the last guidance on the page as relative using CSS selectors.
- fix contextual guidance overflowing horizontally [Trello card](https://trello.com/c/FlW8s7G7)
- add 'hide only' support for contextual guidance triggers based on additional data attribute [Trello card](https://trello.com/c/pcbHvl7B)

### Before
screen shot 2019-03-01 at 15 31 22

### After
screen shot 2019-03-01 at 15 30 35

### DIFF
--- a/app/assets/javascripts/components/contextual-guidance.js
+++ b/app/assets/javascripts/components/contextual-guidance.js
@@ -22,7 +22,7 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
   }
 
   ContextualGuidance.prototype.hideAllGuidance = function () {
-    var guidances = document.querySelectorAll('.app-c-contextual-guidance-wrapper')
+    var guidances = document.querySelectorAll('[data-module="contextual-guidance"]')
 
     guidances.forEach(function (guidance) {
       guidance.style.display = 'none'
--- a/app/assets/javascripts/components/contextual-guidance.js
+++ b/app/assets/javascripts/components/contextual-guidance.js
@@ -18,7 +18,9 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
 
   ContextualGuidance.prototype.handleFocus = function (event) {
     this.hideAllGuidance()
-    this.$module.style.display = 'block'
+    if (!event.target.dataset.contextualGuidanceHideOnly) {
+      this.$module.style.display = 'block'
+    }
   }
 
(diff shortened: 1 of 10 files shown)


benthorner:
This is a nice small change to solve the 'no' problem for the change note, but it feels like we're trying to solve a problem that's caused by the per-element way the contextual guidance is now working. I think it's going to be a bit unintuitive to say 'no' has 'change-note-guidance' when it doesn't.

I think we should consider moving back to the original 'manager' approach, whereby the script is initialised on some outer element and 'manages' the appearance of guidance within. Scoping the module to a specific element means we can still initialise it multiple times, and also has the advantage that we can support multiple, concurrent bits of guidance on the same page.

Let me know if that needs some more explanation :-).

alex-ju:
> This is a nice small change to solve the 'no' problem for the change note, but it feels like we're trying to solve a problem that's caused by the per-element way the contextual guidance is now working. I think it's going to be a bit unintuitive to say 'no' has 'change-note-guidance' when it doesn't.

I see your point, but I don't think having a manager script and using only `data-contextual-guidance` attached to the 'no' input radio element will make it more clear that the no element is actually hiding any guidance on the page. We're basically saying 'no' is meant to hide the 'change-note-guidance', which it does.

> I think we should consider moving back to the original 'manager' approach, whereby the script is initialised on some outer element and 'manages' the appearance of guidance within. Scoping the module to a specific element means we can still initialise it multiple times, and also has the advantage that we can support multiple, concurrent bits of guidance on the same page.

The current script does supports multiple, concurrent bits of guidance, so I'm not very clear of the advantage of introducing a manager script. I would rather try to keep scripts attached to each component instance where possible.

> Let me know if that needs some more explanation :-).

To wrap my response up, I'm happy to introduce a manager script if it proves to be needed in future, but at this point it feels that it introduces code complexity without a benefit in terms of functionality.

benthorner:
I'm not sure I agree with your last sentence: clicking 'no' should hide any guidance on the page e.g. if we went straight from the 'body' textarea to clicking 'no'. This means there's no special relationship between the 'no' input and the 'change-note-guidance', so I think it's confusing to suggest there is.
Since we used to have a manager script, I'm not sure what complexity we're concerned about here?

PR #827, opened 2019-02-28
Fix details components

alex-ju:
This fixes instances of details component after #811 introduced the default spacing.
Tries to deal with excessive padding and markup on the scheduling details element.

### Before
screen shot 2019-03-01 at 10 44 18

### After
screen shot 2019-03-01 at 10 44 38

@emmabeynon: will rebase after #826

### DIFF
--- a/app/views/documents/show/_scheduling_form.html.erb
+++ b/app/views/documents/show/_scheduling_form.html.erb
@@ -4,12 +4,17 @@
 <% submitted_values = flash[:scheduled_publishing_params] %>
 <% selected_time = submitted_values&.fetch("time") ||  current_scheduled_publish_datetime&.strftime("%l:%M%P")&.strip %>
 
+<% scheduling_title = capture do %>
+  <% if current_scheduled_publish_datetime %>
+    Publish date: <br><%= current_scheduled_publish_datetime.strftime("%-d %B %Y - %l:%M%P") %>
+  <% else %>
+    Schedule publishing
+  <% end %>
+<% end %>
+
 <%= render "govuk_publishing_components/components/details", {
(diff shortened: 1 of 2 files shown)


app/views/documents/show/_scheduling_form.html.erb
```diff
@@ -5,12 +5,18 @@
...
+    <% if current_scheduled_publish_datetime %>
+      Publish date: <br><%= current_scheduled_publish_datetime.strftime("%-d %B %Y - %l:%M%P") %>
+    <% else %>
+      Schedule publishing
+    <% end %>
+  <% end %>
+
   <%= render "govuk_publishing_components/components/details", {
-    title: if current_scheduled_publish_datetime
-             simple_format("Publish date: \n" + current_scheduled_publish_datetime.strftime("%-d %B %Y - %l:%M%P"))
```
alex-ju: `simple_format` works fine, but there's no way to remove the wrapper, only to change the element type

app/views/documents/show/_scheduling_form.html.erb
```diff
@@ -5,12 +5,18 @@
 <% selected_time = submitted_values&.fetch("time") ||  current_scheduled_publish_datetime&.strftime("%l:%M%P")&.strip %>
 
 <%= form_tag(save_scheduled_publishing_datetime_path(@edition.document)) do %>
+
+  <% sheduleding_title = capture do %>
```
emmabeynon: Just a small typo here, should be `scheduling_title`

alex-ju: I would say more than small. rushing is not good..
Thanks! Updated

PR #828, opened 2019-02-28
Merge autocomplete and its hinted variant

benthorner:
Previously we had a flag that would change the autocomplete to add hint
text to each entry. The hinted variant is only used for contacts, and
was broken due to a misplaced data attribute.

This merges the hinted variant with the normal non-hinted autocomplete,
which fixes the issue with contacts and removes some duplicated code,
noting that the hinted variant also supports not using a hint.

This also removes support for contextual guidance, since this is not
currently being used with the autocomplete, and in any case was only
implemented for the non-hinted variant.

### DIFF
--- a/app/views/components/_autocomplete.html.erb
+++ b/app/views/components/_autocomplete.html.erb
@@ -4,12 +4,10 @@
   selected_options ||= []
   multiple ||= nil
   size ||= nil
-  data ||= nil
   hint ||= nil
-  data_module ||= "autocomplete"
 %>
 
-<div class="app-c-autocomplete govuk-form-group" data-module="<%= data_module %>">
+<div class="app-c-autocomplete govuk-form-group" data-module="autocomplete">
   <%= render "govuk_publishing_components/components/label", {
     html_for: id
   }.merge(label.symbolize_keys) %>
@@ -27,7 +25,6 @@
       options_for_select(options, selected_options),
       id: id,
       class: "govuk-select",
-      data: data,
       size: size,
       multiple: multiple
     %>
--- a/app/views/tags/edit.html.erb
+++ b/app/views/tags/edit.html.erb
@@ -16,16 +16,6 @@
         tags: @revision.tags,
         contextual_guidance_id: "#{tag_field.id}-guidance" %>
     </div>
-    <% if @document.document_type.guidance_for(tag_field.id) %>
-    <div class="govuk-grid-column-one-third">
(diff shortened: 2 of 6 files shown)


app/views/components/_autocomplete.html.erb
```diff
@@ -4,12 +4,10 @@
   selected_options ||= []
   multiple ||= nil
   size ||= nil
-  data ||= nil
   hint ||= nil
-  data_module ||= "autocomplete"
 %>
 
-<div class="app-c-autocomplete govuk-form-group" data-module="<%= data_module %>">
+<div class="app-c-autocomplete govuk-form-group", data-module="autocomplete">
```
alex-ju:
```suggestion
<div class="app-c-autocomplete govuk-form-group" data-module="autocomplete">
```

app/views/tags/edit.html.erb
```diff
@@ -16,16 +16,6 @@
         tags: @revision.tags,
         contextual_guidance_id: "#{tag_field.id}-guidance" %>
     </div>
-    <% if @document.document_type.guidance_for(tag_field.id) %>
-    <div class="govuk-grid-column-one-third">
-      <%= render "components/contextual_guidance", {
-        id: "#{tag_field.id}-guidance",
-        title: @document.document_type.guidance_for(tag_field.id).title
-      } do %>
-        <p class="govuk-body"><%= @document.document_type.guidance_for(tag_field.id).body %></p>
```
alex-ju: How sure are we that we're not needing the guidance for this (maybe we've lost it along the way)?
I remember I did some work on the guidance component a while back just to make it work with autocomplete.

app/views/components/_autocomplete.html.erb
benthorner: Good spot :-).

app/views/tags/edit.html.erb
benthorner: So I can see it's removed for tags, to be replaced with hint text, while for something like contacts it's just not applicable. It looks like it would be pretty simple to add it in again, if we did need to support it, but I thought this was a good opportunity to clean up. Do you think we should keep it?

alex-ju: Sure, if we're providing the hint text I'm happy to drop it. The less we're using the contextual guidance aside from the edit page the better.

benthorner: 👍

PR #829, opened 2019-02-28
Add check for scheduled datetime being too close to now

emmabeynon:


emmabeynon:
@kevindew I think that's certainly simpler.  My concern was that if those time periods ever changed then you would need to remember to change the translation, otherwise users would be shown the wrong message.  But they're values that are probably unlikely to change, or at least not frequently, so perhaps it's a risk worth taking.

PR #833, opened 2019-02-28
Add a bit of vertical padding for the spinner

benthorner:
This is just a quick fix so that the modal height doesn't collapse too
much in between rendering different sections.

### DIFF
--- a/app/assets/stylesheets/components/_loading-spinner.scss
+++ b/app/assets/stylesheets/components/_loading-spinner.scss
@@ -1,6 +1,7 @@
 .app-c-loading-spinner {
   @include govuk-focusable;
   text-align: center;
+  padding: govuk-spacing(8) * 2;
 }
 
 @keyframes app-loading-spinner-opacity-0 {


app/assets/stylesheets/components/_loading-spinner.scss
```diff
@@ -1,6 +1,7 @@
 .app-c-loading-spinner {
   @include govuk-focusable;
   text-align: center;
+  padding: 200px 0;
```
alex-ju: Can we use the [design system spacing scale](https://design-system.service.gov.uk/styles/spacing/) for a 'more consistent' spacing around the page? `padding: govuk-spacing(8) * 2` or something. I would ask @soniaturcotte about this too.

benthorner: Sorry - still learning to use that. Fixed :-). I'd quite like to get this in and leave any further changes to the product/testing phase - is that OK with you?

alex-ju: Yep

PR #834, opened 2019-02-28
Make re-initialisation for modal more efficient

benthorner:
Previously we were trying to re-initialise all components in the app,
even though we only need to attempt this for the modal content.

### DIFF
--- a/app/assets/javascripts/modules/inline-image-modal.js
+++ b/app/assets/javascripts/modules/inline-image-modal.js
@@ -1,3 +1,5 @@
+/* global $ */
+
 function InlineImageModal ($module) {
   this.$module = $module
   this.$modal = document.getElementById('modal')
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/modules/inline-image-modal.js
```diff
@@ -127,7 +129,7 @@ InlineImageModal.prototype.performAction = function (item) {
 }
 
 InlineImageModal.prototype.initComponents = function () {
-  window.GOVUK.modules.start()
+  window.GOVUK.modules.start($(this.$module))
```
kevindew: minor, but I think given we have a convention for $ on variables the usage of jQuery might be clearer by using the jQuery global variable rather than $

benthorner: @alex-ju what do you think? I was going from the use of `$` in govuk_frontend_toolkit, which I've used in the tests. Can change everywhere if we agree it's preferable :-).

alex-ju: I see @kevindew's point that the use of `$` for illustrating `nodeList` objects is not great as it causes confusion with the jQuery alias, but on the other hand we use it this way all over `govuk_publishing_components` and other projects across GOVUK. I hope to get rid of jQuery from this project soon so I have no strong opinions about how it will looks like until then.

PR #838, opened 2019-03-01
Insert a new line after the inline snippet

benthorner:

This is to ensure certain inline elements render correctly.

### DIFF
--- a/app/assets/javascripts/modules/inline-image-modal.js
+++ b/app/assets/javascripts/modules/inline-image-modal.js
@@ -77,7 +77,7 @@ InlineImageModal.prototype.renderResponse = function (response) {
 
 InlineImageModal.prototype.insertSnippet = function (item) {
   var editor = this.$module.closest('[data-module="markdown-editor"]')
-  editor.selectionReplace(item.dataset.modalData)
+  editor.selectionReplace(item.dataset.modalData + '\n')
 }
 
 InlineImageModal.prototype.performAction = function (item) {


kevindew:
Thanks for doing this. 

Sorry for back-pedalling but I'm wondering now if this is starting to open up a can of worms and my new line suggestion was too simplistic. For example you'd want it to replace an insertion without changes, or prefix a new line if we're already in a paragraph, or do nothing if there's already a blank line.

I had a brief ponder over how this should work and then and a chat with Alex. I notice we already have this behaviour (and tons of logic behind it) on our other block elements like headers and lists. Digging into this we have this: https://github.com/alphagov/markdown-toolbar-element/blob/master/index.js#L405-L434 which probably does all of this.

I'll have a brief look to see if it's easy to use that code, but if not it might be better to just delay this until videos - unless you think I'm overthinking this whole thing?

kevindew:
I got it working quickly in this branch: https://github.com/alphagov/content-publisher/compare/new-line-after-snippet-kevin?expand=1 

Felt a bit wrong to expose the private function but looks like we've just had to do that for insertText as well. I feel like my boolean interface to the function isn't ideal either.

benthorner:
@kevindew that looks OK to me - except I think `prepend` and `append` are the other way round - but I still wonder if this it's really necessary. Unless we're really concerned, can we leave this to research to see if we actually need any intelligent behaviour here?

kevindew:
@benthorner we can do this as part of video work.

One thing I noticed digging into this is it looks like our markdown-toolbar-editor js script has diverged from: https://github.com/alphagov/markdown-toolbar-element are we planning to resolve that?

benthorner:
OK @kevindew. I don't know about the markdown-toolbar - perhaps @alex-ju can comment?

PR #843, opened 2019-03-04
Refactor modal module

benthorner:

This is a small attempt to refactor the inline-image-modal module, so that it's a little 
easier to navigate and to extract some of the more generic functions that don't have 
to be in the same file as the ones that are specific to the inline-image workflow.


app/assets/javascripts/utilities/modal-workflow.js
```diff
@@ -0,0 +1,37 @@
+/* global $ */
```
kevindew: Something I learnt last week is that for many situations there are eslint-env settings for different global variables. 

So this could be defined `/* eslint-env jQuery */` as per: https://eslint.org/docs/user-guide/configuring#specifying-environments

Since it's actually more code and we probably already have this global in a few places it's probably of low value. But I thought I'd pass it in on.

app/assets/javascripts/utilities/modal-fetch.js
```diff
@@ -0,0 +1,47 @@
+function ModalFetch () { }
```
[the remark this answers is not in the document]
benthorner: Yeah - I was going for consistency, but I've made it an object now.

app/assets/javascripts/utilities/modal-workflow.js
benthorner: Agreed - our current approach seems succinct and descriptive enough for now.

benthorner:
@kevindew this should be ready for another look...

PR #844, opened 2019-03-04
Display scheduled publishing date time errors in error summary

emmabeynon:
This moves the scheduled publishing datetime form errors to be displayed in the error summary on top of the document show page.  We also add a href so that errors link to the form, and ensure the form stays open in the event of invalid data being submitted.

**Before**
screen shot 2019-03-04 at 10 29 19

**After**
screen shot 2019-03-04 at 10 14 16

### DIFF
--- a/app/controllers/schedule_controller.rb
+++ b/app/controllers/schedule_controller.rb
@@ -10,7 +10,10 @@ class ScheduleController < ApplicationController
       issues = checker.pre_submit_issues
 
       if issues.any?
-        flash[:scheduled_publishing_datetime_issues] = issues.items_for(:scheduled_datetime)
+        flash["alert_with_items"] = {
+          title: I18n.t!("requirements.scheduled_datetime.title"),
+          items: issues.items,
+        }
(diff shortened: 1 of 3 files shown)


emmabeynon:
@benhazell Content improvements very much welcomed 😄

app/controllers/schedule_controller.rb
```diff
@@ -10,7 +10,11 @@ def save_scheduled_publishing_datetime
       issues = checker.pre_submit_issues
 
       if issues.any?
-        flash[:scheduled_publishing_datetime_issues] = issues.items_for(:scheduled_datetime)
+        href = { scheduled_datetime: "#scheduled_publishing_datetime" }
+        flash["alert_with_items"] = {
+          title: I18n.t!("requirements.scheduled_datetime.title"),
+          items: issues.items_for(:scheduled_datetime, hrefs: href),
```
benthorner: It looks like we want to get all issues here - could we use the `items` method instead?

PR #846, opened 2019-03-04
Improve quality of errors when debugging

benthorner:

It didn't take long to track down the pain-points when debugging with 
the modal, but I'm happy to break this out into another card if we feel 
it deserves some more thorough attention.

This adds a 'console.error' to ensure we don't swallow errors that we
don't hide errors caused by things like syntax issues with JavaScript.
I've also added a couple of 'console.debug' statements to help inspect
server errors when they occur - these won't be shown in production.


benthorner:
@kevindew turns out it doesn't - rather than stop the world, I've used `captureException` as it's something we can foresee.

PR #849, opened 2019-03-04
[DO NOT MERGE] Bh images release

benhazell:
Guidance updates for release of inline images feature - only release when new features are live.
https://trello.com/c/Ub7JHdhg/637-feature-release-master-card

### DIFF
--- a/app/views/publisher_information/beta_capabilities.govspeak.md
+++ b/app/views/publisher_information/beta_capabilities.govspeak.md
@@ -13,7 +13,6 @@ It is possible to remove (unpublish) pages on request.
 
 Some features are not available yet:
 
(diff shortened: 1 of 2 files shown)


benthorner:
@benhazell is it worth hinting that users can do this via a new 'widget' on the markdown toolbar? It might be confusing to some if they are trying to relate with the existing flow for lead images.

benhazell:
Not here - this is just an update to say it is now available. 
We're aiming to make it intuitive to do it without guidance. I need to review the build, but if it isn't it'll be more useful to see people get stuck and improve the design than to provide lots of notes.

kevindew:
This has gone out with the release

PR #852, opened 2019-03-04
Filter by a users organisation by default

kevindew:
Trello: https://trello.com/c/RAspjwep/673-set-default-view-to-filter-by-organisation

This is a draft PR - my first one 🌮 - as there are still design decisions to be made in the trello card and there's a couple of aspects of this PR I want to open to discussion first.

This changes the default behaviour of the `/documents` of Content Publisher to redirect to `/documents?organisation#{current_user.organisation.content_id}` when requested without an organisation parameter.

This is done so that documents are filtered by the current users organisation and so that the URL used to browse documents is not personalised towards a particular session - as per [this trello comment](https://trello.com/c/RAspjwep/673-set-default-view-to-filter-by-organisation#comment-5c790b2694ca0d8b930b48e8).

This has led to this containing a few things that I'm not super duper happy about and I thought I'd open it open for feedback/suggestions.

### Whether to still link to `documents_path` or link to organisation filtered path

It feels a bit annoying to link users to `documents_path` when they'll only be served a redirect we can anticipate but it doesn't feel great to be typing `documents_path(organisation: current_user.organisation_content_id)` all over the place too and that feels a bit fragile. I also wondered if this could mean lost flash messages, but I believe that won't be the case.

Currently I've gone with just linking to `documents_path`. Although setting `documents_path(organisation: current_user.organisation_content_id)` isn't actually much different to most route parameters but it does feel a bit strange since it's an optional rather than a required parameter.

My ideal would have been to set up a proc in the routes.rb file which automatically adds this as part of writing the route using the warden config, but I don't know how to do that. 

Finally another option would be a helper function, but I feel like this would be a bit of a wtf since it's not really that complex and it'd be tricky to make that distinguishable from routes.rb paths yet still be clearly a path.

### Usage of current user in tests

I think this is one of the first features where we're making references to the object of the current user. I've gone with using `User.first` to access them, since I know how gds-sso mock mode works. 

This does feel a bit opaque though. I wonder if a `current_user` helper in tests could be easier. It could also be good to embrace `Warden::Test::Helpers` per https://github.com/plataformatec/devise/wiki/How-To:-Test-with-Capybara although it's probably overkill at this point.

### Coupling of index filtering behaviour to other tests

The other pain point I hit here was that I had to change tests where a user ended up being redirect to the documents page, see https://github.com/alphagov/content-publisher/commit/f999b1ad2632530404ad28fe9afd7e77909e47a8. It's ended up with these tests being rather coupled to this index change which again doesn't feel great. 

I decided against pursing ideas that removed the index filtering in those cases, as they'd no longer be representative as integration tests. One option I considered was specifying the User as the creator edition and having that automatically set the organisation tag, but that's maybe somewhat opaque.

Any thoughts/suggestions?

### DIFF
--- a/app/controllers/documents_controller.rb
+++ b/app/controllers/documents_controller.rb
@@ -2,6 +2,11 @@
 
 class DocumentsController < ApplicationController
   def index
+    if filter_params[:filters].empty? && current_user.has_permission?(User::PRE_RELEASE_FEATURES_PERMISSION)
+      redirect_to documents_path(organisation: current_user.organisation_content_id)
+      return
+    end
+
     filter = EditionFilter.new(filter_params)
     @editions = filter.editions
     @filter_params = filter.filter_params
--- a/app/services/edition_filter.rb
+++ b/app/services/edition_filter.rb
@@ -14,7 +14,7 @@ class EditionFilter
   def initialize(params)
     @filters = params[:filters].to_h.symbolize_keys
     @sort = allowed_sort?(params[:sort]) ? params[:sort] : DEFAULT_SORT
-    @page = params[:page]
+    @page = params.fetch(:page, 1).to_i
     @per_page = params[:per_page]
   end
 
@@ -29,9 +29,9 @@ class EditionFilter
   end
 
   def filter_params
-    filters.select { |_, value| value.present? }.tap do |params|
+    filters.dup.tap do |params|
       params[:sort] = sort if sort != DEFAULT_SORT
-      params[:page] = page if page != "1"
+      params[:page] = page if page > 1
     end
   end
 
--- a/spec/features/finding/index_details_spec.rb
+++ b/spec/features/finding/index_details_spec.rb
@@ -4,36 +4,25 @@ RSpec.feature "Index details" do
   scenario do
     given_there_is_an_edition
     when_i_visit_the_index_page
-    then_i_can_see_the_title
-    and_i_can_see_the_document_type
-    and_i_can_see_the_state
-    and_i_can_see_who_last_edited_it
+    then_i_can_see_the_edition
   end
 
   def given_there_is_an_edition
-    @editor = create(:user)
-    @edition = create(:edition, last_edited_by: @editor)
+    @edition = create(:edition,
+                      last_edited_by: current_user,
+                      created_by: current_user)
   end
 
   def when_i_visit_the_index_page
     visit documents_path
   end
 
-  def then_i_can_see_the_title
+  def then_i_can_see_the_edition
     expect(page).to have_content(@edition.title)
-  end
-
-  def and_i_can_see_the_document_type
     expect(page).to have_content(@edition.document_type.label)
-  end
-
-  def and_i_can_see_the_state
     expect(page).to have_content(I18n.t!("user_facing_states.draft.name"))
-  end
-
-  def and_i_can_see_who_last_edited_it
     expect(page).to have_content(
-      I18n.t!("documents.index.search_results.last_edited_by", user: @editor.name),
+      I18n.t!("documents.index.search_results.last_edited_by", user: current_user.name),
     )
   end
 end
--- a/spec/features/finding/index_filtering_spec.rb
+++ b/spec/features/finding/index_filtering_spec.rb
@@ -4,21 +4,20 @@ RSpec.feature "Index filtering" do
   scenario do
     given_there_are_some_editions
     when_i_visit_the_index_page
-    and_i_filter_by_title
-    then_i_see_just_the_ones_that_match
+    then_i_only_see_my_organisations_editions
 
     when_i_clear_the_filters
-    then_i_see_all_the_editions
+    then_i_see_all_editions
+
+    when_i_clear_the_filters
+    and_i_filter_by_title
+    then_i_see_just_the_ones_that_match
 
     when_i_filter_by_document_type
     then_i_see_just_the_ones_that_match
 
     when_i_clear_the_filters
     and_i_filter_by_state
-    then_i_see_just_the_ones_that_match
-
-    when_i_clear_the_filters
-    and_i_filter_by_primary_organisation
     then_i_see_just_the_ones_that_match
 
     when_i_clear_the_filters
@@ -30,21 +29,36 @@ RSpec.feature "Index filtering" do
   end
 
   def given_there_are_some_editions
-    create(:edition, :published, title: "Totally irrelevant")
-    @primary_organisation = { "content_id" => SecureRandom.uuid, "internal_name" => "Organisation 1" }
-    @organisation = { "content_id" => SecureRandom.uuid, "internal_name" => "Organisation 2" }
+    @primary_organisation = { "content_id" => current_user.organisation_content_id,
+                              "internal_name" => "Organisation 1" }
+    @other_organisation = { "content_id" => SecureRandom.uuid,
+                            "internal_name" => "Organisation 2" }
 
     @relevant_edition = create(:edition,
                                title: "Super relevant",
                                tags: {
                                  primary_publishing_organisation: [@primary_organisation["content_id"]],
-                                 organisations: [@organisation["content_id"]],
+                                 organisations: [
+                                   @primary_organisation["content_id"],
+                                   @other_organisation["content_id"],
+                                 ],
                                })
+    create(:edition,
+           :published,
+           title: "Irrelevant but my organistion",
+           tags: {
+             primary_publishing_organisation: [@primary_organisation["content_id"]],
+             organisations: [@primary_organisation["content_id"]],
+           })
+
+    create(:edition, :published, title: "Not even my organisation")
   end
 
   def when_i_visit_the_index_page
-    stub_publishing_api_has_linkables([@primary_organisation, @organisation],
-                                      document_type: "organisation")
+    stub_publishing_api_has_linkables(
+      [@primary_organisation, @other_organisation],
+      document_type: "organisation",
+    )
 
     visit documents_path
   end
@@ -63,8 +77,17 @@ RSpec.feature "Index filtering" do
     click_on "Clear all filters"
   end
 
-  def then_i_see_all_the_editions
+  def then_i_only_see_my_organisations_editions
     expect(page).to have_content("2 documents")
+  end
+
+  def when_i_filter_by_all_organisations
+    select "", from: "organisation"
+    click_on "Filter"
+  end
+
+  def then_i_see_all_editions
+    expect(page).to have_content("3 documents")
   end
 
   def when_i_filter_by_document_type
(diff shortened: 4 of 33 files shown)


spec/features/finding/index_details_spec.rb
```diff
@@ -2,24 +2,33 @@
 
 RSpec.feature "Index details" do
   scenario do
-    given_there_is_an_edition
+    given_there_is_an_edition_in_my_organisation
     when_i_visit_the_index_page
-    then_i_can_see_the_title
+    then_i_am_redirected_to_my_organisation_filter
```
benthorner: I'm never really sure how much we care about URLs in testing. But it seems like we're missing the behavioural aspect of this: where we don't see an edition that's not in our org.

spec/features/finding/index_ordering_spec.rb
```diff
@@ -12,8 +12,18 @@
   end
 
   def given_there_are_some_editions
-    @most_recent = create(:edition, title: "Most recent", last_edited_at: 1.minute.ago)
-    @least_recent = create(:edition, title: "Least recent", last_edited_at: 2.minutes.ago)
+    @most_recent = create(:edition,
+                          title: "Most recent",
+                          last_edited_at: 1.minute.ago,
+                          tags: {
+                            primary_publishing_organisation: [User.first.organisation_content_id],
```
benthorner: This seems like a common thing to do - should it be a factory default?

spec/features/workflow/delete_draft_after_publishing_spec.rb
```diff
@@ -50,7 +54,7 @@ def then_i_see_the_updated_draft
   end
 
   def then_i_see_the_draft_is_gone
-    expect(page).to have_current_path(documents_path)
+    expect(page).to have_current_path(documents_path(organisation: User.first.organisation_content_id))
```
benthorner: Would it be cleaner to use a regex match here, so that we're not too coupled to our filtering behaviour? I guess there's a slight compromise, as the reason this is here is to try and verify we're on the right page to see something is gone, but then we do have the filtering feature test for that.

spec/features/finding/index_filtering_spec.rb
```diff
@@ -30,9 +29,19 @@
   end
 
   def given_there_are_some_editions
-    create(:edition, :published, title: "Totally irrelevant")
-    @primary_organisation = { "content_id" => SecureRandom.uuid, "internal_name" => "Organisation 1" }
-    @organisation = { "content_id" => SecureRandom.uuid, "internal_name" => "Organisation 2" }
+    @primary_organisation = { "content_id" => User.first.organisation_content_id,
```
benthorner: Comparing this against the format specs (e.g. `news_story_spec`), perhaps it would be a good idea to introduce a 'tag' factory for these things?

app/controllers/documents_controller.rb
```diff
@@ -2,6 +2,11 @@
 
 class DocumentsController < ApplicationController
   def index
+    if filter_params[:filters].empty?
```
benthorner: Would an alternative be to default the param in the `EditionFilter`? My thinking here is that `documents_path` offers a 'default view' for the user, which might be equivalent to a specific search. I understand it might be a bit weird being able to get the same data from two different URLs, but I can't see any practical downside to this, and the concept of the unadorned `documents_path` as a 'default view' or 'dashboard' sounds pretty intuitive to me.

spec/features/finding/index_details_spec.rb
```diff
@@ -2,24 +2,33 @@
...
+                      })
   end
 
   def when_i_visit_the_index_page
     visit documents_path
   end
 
-  def then_i_can_see_the_title
+  def then_i_am_redirected_to_my_organisation_filter
+    expect(page).to have_current_path(documents_path(organisation: User.first.organisation_content_id))
```
benthorner: Would this help?

https://github.com/teamcapybara/capybara/blob/master/lib/capybara/spec/session/has_current_path_spec.rb#L80

kevindew: Yes - great find 👍

kevindew:
@benthorner This has been updated now. We've now got:

- Actual feature behind pre_release flag
- An authentication_helper across our feature tests
- A FactoryBot trait that stops repeating of revision fields in edition and revision factories (bit of a balance of opaqueness vs dry)
- Revisions are automatically tagged to the creators organisation in factories
- Using the ignore_query on feature tests unrelated to index tests.

Quite a lot for such a tiny functional change 😱 

I need to check in with Ben and Sonia that they agree this is the way it should be still.

spec/features/workflow/withdraw_without_managing_editor_permission_spec.rb
```diff
@@ -2,7 +2,7 @@
 
 RSpec.feature "Withdraw without managing editor permission" do
   background do
-    given_i_dont_have_the_managing_editor_permission
+    given_i_am_not_a_managing_editor
```
benthorner: Do we need this step in the first place?

spec/features/workflow/unwithdraw_without_managing_editor_permissions.spec.rb
```diff
@@ -2,12 +2,17 @@
 
 RSpec.feature "Undo a withdraw without managing editor permission" do
   scenario "withdrawn edition" do
+    given_i_am_not_a_managing_editor
```
benthorner: Similarly, do we actually need this step?

spec/features/workflow/show_with_creator_spec.rb
```diff
@@ -10,15 +10,16 @@
 
   def given_there_is_a_document_with_a_creator
     @user = create(:user, name: "Joe Bloggs")
```
benthorner: Could we use `current_user` here instead, just to avoid the coupling with the search filters?

spec/features/finding/index_details_spec.rb
```diff
@@ -2,38 +2,61 @@
...
+    @edition_in_organisation = create(
+      :edition,
+      last_edited_by: @editor,
+      tags: {
+        primary_publishing_organisation: [current_user.organisation_content_id],
+      },
+    )
+  end
+
+  def and_an_an_edition_not_in_my_organisation
```
benthorner: Typo here :-).

spec/features/finding/index_details_spec.rb
```diff
@@ -2,38 +2,61 @@
...
-    given_there_is_an_edition
+    given_there_is_an_edition_in_my_organisation
+    and_an_an_edition_not_in_my_organisation
     when_i_visit_the_index_page
-    then_i_can_see_the_title
-    and_i_can_see_the_document_type
-    and_i_can_see_the_state
-    and_i_can_see_who_last_edited_it
+    then_i_am_redirected_to_my_organisation_filter
+    and_i_can_see_the_edition_title
```
benthorner: Just a minor thing, but since we're changing these steps, could we just combine into `i_can_see_the_edition`?

spec/features/finding/index_details_spec.rb
```diff
@@ -2,38 +2,61 @@
...
+    and_i_can_see_the_edition_document_type
+    and_i_can_see_the_edition_state
+    and_i_can_see_who_last_edited_the_edition
+    and_i_dont_see_the_edition_not_in_my_organisation
```
benthorner: Does this step belong in here? It looks like this feature is more about what each entry looks like, and we have the filtering spec for what we do/don't see.

spec/features/workflow/withdraw_without_managing_editor_permission_spec.rb
kevindew: No, suppose we don't since we don't do anything explicit anymore - will snip.

spec/features/workflow/show_with_creator_spec.rb
kevindew: Sure

spec/features/finding/index_details_spec.rb
kevindew: darn - thanks

spec/features/finding/index_details_spec.rb
kevindew: yeah suppose so, will mean a few assertions.

spec/features/finding/index_details_spec.rb
kevindew: Sure, can strip it out.

kevindew:
@benthorner your comments are addressed

spec/features/workflow/unwithdraw_without_managing_editor_permissions.spec.rb
emmabeynon: I think this is helpful just to be really clear when reading the tests that the user must not be a managing editor.

benthorner:
@kevindew looks good, but I found a bug: if you change the sort order of the table, it unclears the filters :-|.

spec/features/workflow/unwithdraw_without_managing_editor_permissions.spec.rb
kevindew: @emmabeynon yup the nice part is the clarity in the naming, the bummer though of my change was that the code was just `login_as(create(:user))` which lacks the clarity of the method.

Could presumably go down a route of `create(:user, :non_managing_editor)` but then the code for it is growing.

kevindew:
@benthorner thanks for catching.

kevindew:
@benthorner ok this is sorted now. I experimented with putting it into a feature test but it ended up just using chunks of other feature tests so I'm happy unit tests suffice.

app/services/edition_filter.rb
```diff
@@ -29,9 +29,9 @@ def editions
   end
 
   def filter_params
-    filters.select { |_, value| value.present? }.tap do |params|
+    filters.tap do |params|
```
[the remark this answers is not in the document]
kevindew: D'oh that's an oversight - I wasn't meaning to modify the hash. I'll fix this here.

kevindew:
@benthorner ooh what do you mean order changing? bug

The utf8 thing we're pretty safe to ignore as it's for < IE8 https://github.com/rails/rails/issues/30852 or we could disable it on this form since it's a GET and less hidden from the world.

benthorner:
@kevindew it is a little disconcerting that it changes so much. One nice thing about the old sorting behaviour, albeit inconsistent, is that it would trim the query params; it's quite a long URL now :-(.

```
# URL after clicking filter
http://localhost:3000/documents?utf8=%E2%9C%93&title_or_url=&organisation=7b839800-b9bc-4a3b-bf6f-55bce6509fc0&status=draft&document_type=&sort=-last_updated

# URL after sorting the table
http://localhost:3000/documents?document_type=&organisation=7b839800-b9bc-4a3b-bf6f-55bce6509fc0&sort=-last_updated&status=draft&title_or_url=
```

kevindew:
@benthorner Oh I see what you mean, thought it might have been a bug.

Heh that amuses me as I thought I was the only person who cared about URL lengths - as I think back to the many months ago when we first worked on it. Personally I don't believe keeping it nice for pagination and sorting adds much value, since it's grim when you submit the form. If we kept it nice with the form submission then I'd see this as a loss.

Going back to it would mean choosing between two paths, either we make EditionFilter aware that organisation is special and can be an empty string - which felt a bit icky too me - or we let organisation have a special parameter of "all" (or similar) for empty submission and then add extra code to handle that. Both have a tinge of snowflake about them.

PR #853, opened 2019-03-04
Begin replacing model build methods with updaters

benthorner:

To some extent I think I've tried to avoid making the ImagesController 
even longer, but this may have made it less readable and some more 
question methods somewhere might be beneficial. I've outlined an 
idea for doing this below, but I worry that it's just disguising a bunch of 
single-use code through a convenient interface...

Previously we used a tree of 'build_revision_update' methods on each
model, which has lead to some duplication between models and quite a lot
of extra code in the model classes. This makes a start at extracting
this logic into a set of 'updater' classes with some helper methods.

The updaters take a hash of 'fields', which can include columns,
collections and sub-fields for sub-revisions. I've used inheritance to
DRY-up some of this logic, as well as create a clear interface for the
updater classes that derive from it.

Some of the existing code doesn't fit very well into the updater
classes, as it's specific to a single use-case and doesn't make sense to
share centrally. This does mean the ImagesController now has a bit more
code in it than before. A nice compromise that would avoid bloating the
existing controller and the new RevisionUpdater, would be to use an
'include' pattern to add more image-specific methods to RevisionUpdater.

### DIFF
--- a/app/services/versioning/base_updater.rb
+++ b/app/services/versioning/base_updater.rb
@@ -1,0 +1,40 @@
+# frozen_string_literal: true
+
+module Versioning
+  class BaseUpdater
+    attr_reader :revision, :user
+
+    def initialize(revision, user)
+      @revision = revision
+      @user = user
+    end
+
+    def column_names
+      raise "Not implemented"
+    end
+
+    def changed?(field = nil)
+      field ? changes.key?(field) : changes.any?
+    end
+
+    def next_revision
+      changed? ? dup_revision : revision
+    end
+
+    def changes
+      raise "Not implemented"
+    end
+
+  protected
+
+    def changed_columns
+      old_key_vals = column_names.map { |c| [c, revision.public_send(c)] }
+      new_key_vals = column_names.map { |c| [c, dup_revision.public_send(c)] }
+      Hash[new_key_vals - old_key_vals]
+    end
+
+    def dup_revision
+      raise "Not implemented"
+    end
+  end
+end
--- a/spec/services/versioning/revision_updater/image_spec.rb
+++ b/spec/services/versioning/revision_updater/image_spec.rb
@@ -1,0 +1,119 @@
+# frozen_string_literal: true
+
+RSpec.describe Versioning::RevisionUpdater::Image do
+  let(:user) { create :user }
+  let(:image_revision) { create :image_revision }
+
+  describe "#update_image" do
+    it "extends image revisions with a new image" do
+      revision = create :revision, image_revisions: [image_revision]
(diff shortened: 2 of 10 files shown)


benthorner:
@kevindew I've now finished adding tests for this. I've also replaced the `NotImplementedError` with a helpful message, which seems preferable to some anonymous/custom exception; but I'm open to an alternative approach if you have one :-).

spec/services/versioning/revision_updater/image_spec.rb
```diff
@@ -0,0 +1,119 @@
...
+      updater = Versioning::RevisionUpdater.new(revision, user)
+      updater.update_image(new_image)
+
+      next_revision = updater.next_revision
+      expect(next_revision.image_revisions).to match_array [image_revision, new_image]
+    end
+
+    it "updates an existing image with a new revision" do
+      revision = create(:revision, image_revisions: [image_revision])
+      updated_image = create :image_revision, image_id: image_revision.image_id
```
kevindew: Nitpick, but bit jarring to have the two different parenthesis styles here.

kevindew: There seem to be a few of these in this file.

app/services/versioning/base_updater.rb
```diff
@@ -0,0 +1,40 @@
...
+  class BaseUpdater
+    attr_reader :revision, :user
+
+    def initialize(revision, user)
+      @revision = revision
+      @user = user
+    end
+
+    def column_names
+      raise "Implement with e.g. 'revision.class.column_names'"
```
kevindew: Super minor as no-one will see these exceptions but I don't think this quite reads right. As if you substitute e.g. with for example it's like "Implement with for example .." which doesn't feel like correct English. 

I don't mind if you include an example of what they need to implement but I don't think that's needed. As if we can suggest it maybe we should have just implemented it as a default? Just an error saying it's not implemented is likely enough.

spec/services/versioning/revision_updater/image_spec.rb
benthorner: Oh yeah, I've gone without parentheses as they're not necessary. I remember we thought about getting the linter to check for this :-|.

app/services/versioning/base_updater.rb
benthorner: Fair enough - although I think the English was fine :-).

benthorner:
@kevindew I've addressed your comments :-).

benthorner:
@kevindew this is ready to merge again if you want to take another look.

PR #859, opened 2019-03-07
Add Scheduled Publishing worker

emmabeynon:
For https://trello.com/c/rXTLvzSu/685-scheduling-is-sent-to-the-publishing-api

This PR creates a Scheduled Publishing worker using Sidekiq.

### DIFF
--- a/app/workers/scheduled_publishing_worker.rb
+++ b/app/workers/scheduled_publishing_worker.rb
@@ -14,6 +14,11 @@ class ScheduledPublishingWorker
         return
       end
 
+      unless edition.scheduled?
+        logger.warn("Cannot schedule an edition that is not in a scheduled state")
+        return
+      end
+
       user = edition.status.created_by
       reviewed = edition.status.details.reviewed
 
--- a/app/workers/scheduled_publishing_worker.rb
+++ b/app/workers/scheduled_publishing_worker.rb
@@ -1,0 +1,16 @@
+# frozen_string_literal: true
+
+class ScheduledPublishingWorker
+  include Sidekiq::Worker
+
+  def perform(id)
+    Edition.transaction do
+      edition = Edition.lock.find_by(id: id, current: true)
+      user = edition.status.created_by
+      reviewed = edition.status.details.reviewed
+
+      PublishService.new(edition.document)
+                    .publish(user: user, with_review: reviewed)
+    end
+  end
+end
--- a/app/workers/scheduled_publishing_worker.rb
+++ b/app/workers/scheduled_publishing_worker.rb
@@ -19,11 +19,22 @@ class ScheduledPublishingWorker
         return
       end
 
+      if scheduled_publishing_datetime_in_the_future?(edition)
+        logger.warn("Cannot schedule an edition whose scheduled publishing datetime is in the future")
+        return
+      end
+
       user = edition.status.created_by
       reviewed = edition.status.details.reviewed
 
       PublishService.new(edition.document)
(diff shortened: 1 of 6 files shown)


app/workers/scheduled_publishing_worker.rb
```diff
@@ -0,0 +1,33 @@
...
+      unless edition.scheduled?
+        raise "Cannot schedule an edition that is not in a scheduled state"
+      end
+
+      if scheduled_publishing_datetime_in_the_past?(edition)
+        raise "Cannot schedule an edition whose scheduled publishing datetime is in the past"
+      end
+
+      user = edition.status.created_by
+      reviewed = edition.status.details&.reviewed
```
kevindew: I think we can avoid the safe navigation operator here as we always expect them to have scheduled details don't we?

app/workers/scheduled_publishing_worker.rb
```diff
@@ -0,0 +1,33 @@
+# frozen_string_literal: true
+
+class ScheduledPublishingWorker
+  include Sidekiq::Worker
+  sidekiq_options retry: 10
+  sidekiq_retry_in { 30 }
```
kevindew: These look good 👍 Might be worth leaving a comment above them about the maths of how they add up to 5 mins

app/workers/scheduled_publishing_worker.rb
```diff
@@ -0,0 +1,33 @@
...
+
+  def perform(id)
+    Edition.transaction do
+      edition = Edition.lock.find(id)
```
kevindew: Not sure if this is a great suggestion of mine or not - as it may well conflict with part of a suggestion here: https://trello.com/c/VrE9kdK4/744-we-make-unnecessary-references-to-document-in-our-controllers

But to avoid having to worry about the RecordNotFound exception we could do `find_by(id: id)` which would return nil instead of raising.

spec/workers/scheduled_publishing_worker_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+                       :scheduled,
+                       scheduled_publishing_datetime: Time.zone.now.tomorrow)
+      user = edition.status.created_by
+
+      expect_any_instance_of(PublishService)
+        .to receive(:publish)
+        .with(user: user, with_review: edition.status.details.reviewed)
+
+      ScheduledPublishingWorker.new.perform(edition.id)
+    end
```
kevindew: Looks good 👍

app/workers/scheduled_publishing_worker.rb
```diff
@@ -0,0 +1,33 @@
...
+
+      PublishService.new(edition.document)
+                    .publish(user: user, with_review: reviewed)
+    end
+  end
+
+private
+
+  def scheduled_publishing_datetime_in_the_past?(edition)
+    edition.scheduled_publishing_datetime <= Time.zone.now
```
kevindew: super duper minor, but using `Time.current` is generally preferred to `Time.zone.now` since it's not nested. 

There was this a bit ago: https://github.com/alphagov/content-publisher/pull/708

app/workers/scheduled_publishing_worker.rb
emmabeynon: Done 👍

app/workers/scheduled_publishing_worker.rb
emmabeynon: I've changed this to `find_by(id: id)`.  I've kept the lock, is that correct?

PR #860, opened 2019-03-07
Image copy changes

kevindew:
Trello: https://trello.com/c/MkWywL4g/590-support-inline-images

This is DNM as it is pending chats between @benhazell and @soniaturcotte.

This implements some of the changes @benhazell raised in the linked trello card for copy changes - the other changes suggested looked to require further discussion. Though to do these consistently I had to refer back to designs for some copy where there were gaps in Ben's suggestions.

Things that have changed:

Before:

screen shot 2019-03-07 at 12 48 47

After:

screen shot 2019-03-07 at 12 49 44

Before: 

screen shot 2019-03-07 at 12 49 16

After:

screen shot 2019-03-07 at 12 51 02

Before:

screen shot 2019-03-07 at 12 49 24

After:

screen shot 2019-03-07 at 12 51 08

Before:

screen shot 2019-03-07 at 12 52 35

After:

screen shot 2019-03-07 at 12 52 01

### DIFF
--- a/app/views/images/crop.html.erb
+++ b/app/views/images/crop.html.erb
@@ -1,4 +1,7 @@
-<% content_for :title, t("images.crop.title", title: @document.current_edition.title_or_fallback) %>
+<%
+title_key = rendering_context == "modal" ? "images.crop.title_modal" : "images.crop.title"
+content_for :title,t(title_key, title: @document.current_edition.title_or_fallback)
+%>
 
 <% content_for :back_link, render_back_link(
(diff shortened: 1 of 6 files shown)


kevindew:
@benthorner Yeah I pondered that but I think actually our use of "Edit" on Image details is unusual. As we don't have "Edit" for tags or topics, just content. Not something I'm strongly opinionated on though.

kevindew:
Got the ok from Sonia and Ben 👍

PR #868, opened 2019-03-11
Edition methods to load current edition with joins

kevindew:
Trello: https://trello.com/c/VrE9kdK4/744-we-make-unnecessary-references-to-document-in-our-controllers

This is intended to replace the need to have the following methods in
controllers:

```
document = Document.with_current_edition.find_by_param(params[:id])
edition = document.current_edition
```

and

```
Document.transaction do
  document = Document.with_current_edition.lock.find_by_param(params[:id])
  edition = document.current_edition
end
```

with:

```
edition = Edition.find_current(document: params[:document])
```

and

```
end
```

This is intended so that controller code can be more focused on an
edition, as most operations require a current edition.

The scope used for this accepts a hash where a current edition can be
looked up via either an id key or a document key. The id key is an
integer of the edition id. The document key is expected to be a document
param - a string containing content_id and locale. Were needs to change
for more document related lookup this could query differently depending
on the input, e.g. it could lookup for a document if given that or a
param string.

The aspect of this I'm least clear on is what the best naming for these
methods should be. I've gone for succinct, low detail names that hopefully
reflect a users intention.

### DIFF
--- a/app/controllers/documents_controller.rb
+++ b/app/controllers/documents_controller.rb
@@ -9,50 +9,44 @@ class DocumentsController < ApplicationController
   end
 
   def edit
-    @document = Document.with_current_edition.find_by_param(params[:id])
-    @revision = @document.current_edition.revision
+    @edition = Edition.find_current(document: params[:document])
+    @revision = @edition.revision
   end
 
   def show
-    @document = Document.with_current_edition.find_by_param(params[:id])
-    @edition = @document.current_edition
+    @edition = Edition.find_current(document: params[:document])
   end
 
   def confirm_delete_draft
-    document = Document.with_current_edition.find_by_param(params[:id])
-    redirect_to document_path(document), confirmation: "documents/show/delete_draft"
+    edition = Edition.find_current(document: params[:document])
+    redirect_to document_path(edition.document), confirmation: "documents/show/delete_draft"
   end
 
   def destroy
-    Document.transaction do
-      document = Document.with_current_edition.lock.find_by_param(params[:id])
-
+    Edition.find_and_lock_current(document: params[:document]) do |edition|
       begin
-        current_edition = document.current_edition
-        DeleteDraftService.new(document, current_user).delete
+        DeleteDraftService.new(edition.document, current_user).delete
 
         TimelineEntry.create_for_status_change(entry_type: :draft_discarded,
-                                               status: current_edition.status)
+                                               status: edition.status)
 
         redirect_to documents_path
       rescue GdsApi::BaseError => e
         GovukError.notify(e)
-        redirect_to document, alert_with_description: t("documents.show.flashes.delete_draft_error")
+        redirect_to edition.document, alert_with_description: t("documents.show.flashes.delete_draft_error")
       end
     end
   end
 
   def update
-    Document.transaction do # rubocop:disable Metrics/BlockLength
-      @document = Document.with_current_edition.lock.find_by_param(params[:id])
-      current_edition = @document.current_edition
-      current_revision = current_edition.revision
+    Edition.find_and_lock_current(document: params[:document]) do |edition|
+      current_revision = edition.revision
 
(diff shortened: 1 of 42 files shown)


app/controllers/documents_controller.rb
```diff
@@ -61,30 +55,30 @@ def update
           "items" => @issues.items,
         }
 
-        render :edit, status: :unprocessable_entity
-        return
+        render :edit, assign: { edition: edition }, status: :unprocessable_entity
```
benthorner: This looks nice, but why the `next`?

app/models/edition.rb
```diff
@@ -66,6 +66,31 @@ class Edition < ApplicationRecord
...
+    join_tables = %i[document revision status]
+    where(current: true)
+      .joins(join_tables)
+      .includes(join_tables)
+      .find_by!(find_by)
+  end
+
+  def self.find_and_lock_current(*args)
+    transaction do
+      lock.find_current(*args).tap do |edition|
```
benthorner: Very minor, but I think you can do `tap(&block)` here if `&block` is also an arg to the main method.

app/controllers/documents_controller.rb
```diff
@@ -61,30 +55,30 @@ def update
...
-        redirect_to @document
+        redirect_to edition.document
       end
     end
   end
 
   def generate_path
-    document = Document.find_by_param(params[:id])
-    base_path = PathGeneratorService.new.path(document, params[:title])
+    edition = Edition.find_current(params[:document])
```
benthorner: Shouldn't this be `document: ...`?

app/models/edition.rb
```diff
@@ -66,6 +66,31 @@ class Edition < ApplicationRecord
 
   MINIMUM_SCHEDULING_TIME = { minutes: 15 }.freeze
 
+  scope :find_current, ->(id: nil, document: nil) do
```
benthorner: It's strange to see this phrased as a scope, as the name contains a verb and I'm used to scopes reflecting traits of a collection. Would it make more sense for it to be a method?

app/controllers/documents_controller.rb
kevindew: d'oh thanks!

app/controllers/documents_controller.rb
kevindew: That something I fixed based off the linter without much thought. 

It's suggested because return leaves the whole method whereas a next leaves the block, and I took that as a sign that it was a consequence of switching to a block.

But now you've raised that I realise it always used to be in a block so I wonder if there is a special linting rule for transactions and return 😕

app/models/edition.rb
kevindew: That's cool - thanks, I've tried that out. 

Pro's are it can be nice and succinct - now fits on one line.
Con's seem to be that it's upped the magic of it and has probably pushed it past being able to read the method and now how it should be used.

app/models/edition.rb
kevindew: The advantage to this being a scope is that we can chain this like we do with `lock.find_current` and wouldn't work in the same way as `find` or `find_by` that also act as chainable scopes.

app/controllers/documents_controller.rb
kevindew: Ah I see why it triggered, as per: https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Lint/NonLocalExitFromIterator

It meets these criteria: 

> - No value is returned,
> - the block is preceded by a method chain,
> - the block has arguments,

Whereas `transaction` does not meet those.

app/models/edition.rb
benthorner: Fair enough :-).

app/models/edition.rb
benthorner: Also fair enough :-).

PR #870, opened 2019-03-12
Schedule the scheduled publishing worker

emmabeynon:
For https://trello.com/c/rXTLvzSu/685-scheduling-is-sent-to-the-publishing-api

We schedule the Scheduled Publishing Worker to perform a job at the time that
content has been scheduled to publish.


spec/features/workflow/schedule_spec.rb
```diff
@@ -38,6 +38,11 @@ def and_i_click_on_publish
 
   def then_i_see_a_confirmation_that_the_edition_has_been_scheduled
     expect(page).to have_content(I18n.t!("schedule.scheduled.title"))
+    govuk_header_args = { "authenticated_user" => nil, "request_id" => nil }
```
kevindew: Since this data is all tied into govuk_sidekiq gem could this be just something like `an_instance_of(Hash)` so we don't have to worry about the content of the hash?

emmabeynon: Done 👍

PR #871, opened 2019-03-12
Stop scheduling feature

1pretz1:
Users can cancel their scheduled content so Content Publisher no longer sets its state to `scheduled`.  Content reverts back to its pre-scheduling status, factoring in whether it was flagged has having been reviewed or not when it was scheduled.

https://trello.com/c/iwxNCu9n/692-allow-users-to-revert-scheduled-status

### DIFF
--- a/app/views/documents/show/_actions.html.erb
+++ b/app/views/documents/show/_actions.html.erb
@@ -57,6 +57,10 @@
       <%= form_tag create_preview_path(@edition.document) do %>
         <%= render "govuk_publishing_components/components/button", text: "Preview" %>
       <% end %>
+      <%= form_tag unschedule_path(@edition.document) do %>
+        <%= render "govuk_publishing_components/components/button", text: "Stop scheduled publishing", secondary: true
+      %>
+      <% end %>
       <%= delete_draft_link %>
     <% else %>
       <%= form_tag submit_document_for_2i_path(@edition.document) do %>
--- a/spec/features/workflow/unschedule_spec.rb
+++ b/spec/features/workflow/unschedule_spec.rb
@@ -1,0 +1,31 @@
+# frozen_string_literal: true
+
+RSpec.feature "Unschedule an edition" do
+  scenario do
+    given_there_is_a_scheduled_document
+    when_i_visit_the_summary_page
+    and_i_click_stop_scheduled_publishing
+    then_the_document_is_no_longer_scheduled
+  end
+
+  def given_there_is_a_scheduled_document
+    schedule = create(:scheduling, reviewed: true)
+    @edition = create(:edition, :scheduled, scheduling: schedule)
+  end
+
+  def when_i_visit_the_summary_page
+    visit document_path(@edition.document)
+  end
+
+  def and_i_click_stop_scheduled_publishing
+    click_on "Stop scheduled publishing"
+  end
+
+  def then_the_document_is_no_longer_scheduled
+    expect(page).to have_content(I18n.t!("user_facing_states.submitted_for_review.name"))
+
+    within first(".app-timeline-entry") do
+      expect(page).to have_content I18n.t!("documents.history.entry_types.unscheduled")
+    end
+  end
+end
--- a/app/controllers/unschedule_controller.rb
+++ b/app/controllers/unschedule_controller.rb
@@ -1,0 +1,35 @@
+# frozen_string_literal: true
+
+class UnscheduleController < ApplicationController
+  def unschedule
+    Document.transaction do
+      document = Document.with_current_edition.lock!.find_by_param(params[:id])
+      edition = document.current_edition
+      schedule = edition.status.details
+
+      remove_scheduled_publishing_datetime(edition)
+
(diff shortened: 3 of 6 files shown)


app/views/documents/show/_actions.html.erb
```diff
@@ -57,7 +57,12 @@
       <%= form_tag create_preview_path(@edition.document) do %>
         <%= render "govuk_publishing_components/components/button", text: "Preview" %>
       <% end %>
+      <%= form_tag unschedule_path(@edition.document) do %>
+        <%= render "govuk_publishing_components/components/button", text: "Stop scheduled publishing", secondary: true
+      %>
+      <% end %>
       <%= delete_draft_link %>
+
```
emmabeynon: I think we can remove this extra line.

spec/features/workflow/unschedule_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+    given_there_is_a_scheduled_document
+    when_i_visit_the_summary_page
+    and_i_click_stop_scheduled_publishing
+    then_the_document_is_no_longer_scheduled_and_the_status_reverted
+    and_the_scheduled_datetime_is_still_persisted
+  end
+
+  def given_there_is_a_scheduled_document
+    schedule = create(:scheduling)
+    @datetime = Time.current.tomorrow.at_noon
```
emmabeynon: We can get rid of `.at_noon` here.  `.tomorrow` should be enough to indicate it's in the future.

spec/features/workflow/unschedule_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+    click_on "Stop scheduled publishing"
+  end
+
+  def then_the_document_is_no_longer_scheduled_and_the_status_reverted
+    expect(@edition.status.details.pre_scheduled_status.draft?).to eq(true)
+    expect(@edition.scheduled?).to eq(true)
+
+    @edition.reload
+
+    expect(@edition.draft?).to eq(true)
```
emmabeynon: Given these are feature tests, we should be ok to see from the UI that the edition has reverted to its pre-scheduled status.  Something like `expect(page).to have_content(I18n.t!("user_facing_states.submitted_for_review.name"))` and that the timeline entry appears (see https://github.com/alphagov/content-publisher/blob/082d1bcf33523b8dce9ed458f353f175bb6bce16/spec/features/workflow/clear_scheduled_publishing_datetime_spec.rb#L59 for an example) should be enough here.

app/controllers/schedule_controller.rb
```diff
@@ -74,6 +74,26 @@ def scheduled
     @edition = document.current_edition
   end
 
+  def unschedule
```
emmabeynon: I think this might benefit from being in its own controller, similar to how we have Withdraw vs. Unwithdraw controllers, as the the actions here are quite separate from each other.

spec/features/workflow/unschedule_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+    expect(@edition.timeline_entries.last.entry_type).to eq("unscheduled")
+  end
+
+  def and_the_scheduled_datetime_is_still_persisted
+    expect(@edition.scheduled_publishing_datetime).to eq(@datetime)
```
emmabeynon: If Sonia decides we should still show the set scheduled_publishing_Datetime in the form, we should check that it's still displayed in the UI rather than in the db.

spec/features/workflow/unschedule_spec.rb
1pretz1: The `at_noon` method is an arbitrary choice which is needed due to if we just did `Time.current.tomorrow` ruby saves the time in the `@datetime` variable to a higher precision then postgres does when the `scheduled_publishing_datetime` is persisted (factory is built).

This caused the test on line 39 to fail when the dates were compared between the record and the saved variable so specifying an exact time was needed.

app/views/documents/show/_actions.html.erb
1pretz1: Yep good spot

spec/features/workflow/unschedule_spec.rb
1pretz1: Good point, after a chat with Sonia and how the new design works we decided it would be best not to save the time after cancelling. 

We could always amend this if it was seen to be a problem later on.

app/controllers/schedule_controller.rb
1pretz1: Yep i agree with you now i've thought about it, thanks!

spec/features/workflow/unschedule_spec.rb
1pretz1: Sure, i originally only did db level checks as i didn't know how drastically the design would change and so how much the frontend would change, but now we know i can make those changes :)

1pretz1:
@emmabeynon githubs been a lil slow to update my new changes but i've addressed all your comments!

PR #873, opened 2019-03-13
Create publish intent request

1pretz1:
When a user schedules content to be published, we need to create a
publish intent in the Publishing API so that Content Store will run down
the TTL and the document's cache will expire as the new content is
published.

Trello:
https://trello.com/c/fXtvPpIO/696-create-publish-intent

### DIFF
--- a/app/services/schedule_service.rb
+++ b/app/services/schedule_service.rb
@@ -11,6 +11,8 @@ class ScheduleService
     scheduling = Scheduling.new(pre_scheduled_status: edition.status, reviewed: reviewed)
     set_edition_status(scheduling, user)
     create_timeline_entry(scheduling)
+
+    create_publish_intent(edition)
   end
 
 private
@@ -26,5 +28,10 @@ private
       status: edition.status,
       details: scheduling,
     )
+  end
+
+  def create_publish_intent(edition)
+    payload = PublishingApiPayload.new(edition).intent_payload
+    GdsApi.publishing_api.put_intent(edition.base_path, payload)
   end
 end
--- a/spec/features/workflow/schedule_publishing_api_down_spec.rb
+++ b/spec/features/workflow/schedule_publishing_api_down_spec.rb
@@ -1,0 +1,40 @@
+# frozen_string_literal: true
+
+RSpec.feature "Schedule a document when Publishing API is down" do
+  scenario do
+    given_there_is_an_edition_with_set_scheduled_publishing_datetime
+    and_the_publishing_api_is_down
+    when_i_visit_the_summary_page
+    and_i_try_to_schedule_the_edition
+    then_i_see_an_error_message
+    and_the_document_has_not_been_scheduled
+  end
+
+  def given_there_is_an_edition_with_set_scheduled_publishing_datetime
+    @datetime = Time.current.tomorrow
+    @edition = create(:edition, scheduled_publishing_datetime: @datetime)
+  end
+
+  def and_the_publishing_api_is_down
+    publishing_api_isnt_available
+  end
+
+  def when_i_visit_the_summary_page
+    visit document_path(@edition.document)
+  end
+
+  def and_i_try_to_schedule_the_edition
+    click_on "Schedule"
+    choose I18n.t!("schedule.confirmation.review_status.reviewed")
+    click_on "Publish"
+  end
+
+  def then_i_see_an_error_message
+    expect(page).to have_content(I18n.t!("documents.show.flashes.schedule_error.title"))
+  end
+
+  def and_the_document_has_not_been_scheduled
+    @edition.reload
+    expect(@edition.state).to_not eq("scheduled")
+  end
+end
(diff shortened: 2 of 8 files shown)


spec/services/schedule_service_spec.rb
```diff
@@ -3,13 +3,17 @@
 RSpec.describe ScheduleService do
   describe "#schedule" do
     it "sets an edition's state to 'scheduled'" do
+      stub_default_publishing_api_put_intent
```
kevindew: Should we consider doing this in a before_each if we're writing it in most methods?

spec/factories/document_type_factory.rb
```diff
@@ -9,7 +9,13 @@
     contents { [] }
     tags { [] }
     guidance { [] }
-    publishing_metadata { DocumentType::PublishingMetadata.new }
+    publishing_metadata {
+      DocumentType::PublishingMetadata.new(
+        "schema_name": "news_article",
+        "rendering_app": "government-frontend",
```
kevindew: It looks like these are really specific values towards news, whereas the rest of this factory is rather generic. 

It'd be more fitting with the existing patterns of this factory if these were randomly generated.

Though I actually think you can avoid making a change here - I'll reference that in my next comment.

spec/services/schedule_service_spec.rb
```diff
@@ -18,14 +22,38 @@
...
+
       edition = create(:edition)
       ScheduleService.new(edition).schedule
 
       expect(edition.timeline_entries.first.entry_type).to eq("scheduled")
     end
 
+    it "makes a request to publishing-api to create / update a publishing intent" do
+      publish_time = Time.current.tomorrow.at_noon
+      edition = create(:edition, scheduled_publishing_datetime: publish_time)
```
kevindew: Since this seems to be the only class where we're concerned with the `document_type.publishing_metadata.rendering_app` we might want to create our document type in this test rather than change the factory.

Then we could also avoid the big chain of `edition.document_type.publishing_metadata.rendering_app`

spec/spec_helper.rb
```diff
@@ -38,6 +39,7 @@
   config.infer_spec_type_from_file_location!
   config.use_transactional_fixtures = true
   config.include FactoryBot::Syntax::Methods
+  config.include GdsApi::TestHelpers::PublishingApi
```
kevindew: Oh boo. Is this only in v1 test helpers? that's a pain.

kevindew:
Oh - just thought of something. Since this introduces communication with the Publishing API the controller should rescue from API errors.

spec/services/schedule_service_spec.rb
1pretz1: Yep that makes sense! Originally I thought that because I was making a unique stub in one of the tests i'd get some sort of conflict, but as it turns out the unique stub overrides the before(:each)'s default stub so alls good :)

spec/factories/document_type_factory.rb
1pretz1: Sure, IIve removed the bespoke behaviour and created the specific document type i needed in the test itself

spec/services/schedule_service_spec.rb
1pretz1: Yep, see the above comment.

spec/spec_helper.rb
1pretz1: Yep, unfortunately! I mean I could include it directly in the test file itself but as I still need to do the destroying intent workflow when someone cancels a scheduling it makes sense to have it here as I will need the helper methods for that piece of work too.

1pretz1:
@kevindew I've added the rescue and a publishing_api_down spec to test this workflow.

app/services/schedule_service.rb
```diff
@@ -27,4 +29,21 @@ def create_timeline_entry(scheduling)
...
+  end
+
+  def intent_payload
+    rendering_app = edition.document_type.publishing_metadata.rendering_app
+    publish_time = edition.scheduled_publishing_datetime
+    base_path = edition.base_path
+
+    {
+      publish_time: publish_time,
+      publishing_app: "content-publisher",
```
benthorner: Although I can't see this ever being a problem, could we use `PublishingApiPayload::PUBLISHING_APP` here? In fact, would this whole method fit better in that class?

spec/features/workflow/schedule_publishing_api_down_spec.rb
```diff
@@ -0,0 +1,48 @@
+# frozen_string_literal: true
+
+RSpec.feature "Schedule a document when Publishing API is down" do
+  scenario do
+    given_there_is_an_edition_with_set_scheduled_publishing_datetime
+    and_the_publishing_api_is_down
+    when_i_visit_the_summary_page
```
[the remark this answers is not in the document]
1pretz1: Yep sure!

spec/features/workflow/schedule_spec.rb
```diff
@@ -18,6 +18,8 @@
   def given_there_is_an_edition_with_set_scheduled_publishing_datetime
     @datetime = Time.current.tomorrow
     @edition = create(:edition, scheduled_publishing_datetime: @datetime)
+
+    stub_default_publishing_api_put_intent
```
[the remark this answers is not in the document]
1pretz1: Done!

app/services/schedule_service.rb
1pretz1: Good spot, `intent_payload` fits nicely in PublishingApiPayload!

PR #875, opened 2019-03-13
Remove redundant withdrawal rake task

1pretz1:
As we can now do this through the UI there is no need to have this rake
task anymore.


benthorner:
Can we also remove that unnecessary nested transaction now?

kevindew:
Thanks @1pretz1 - there's also some documentation on this that should be snipped too: https://github.com/alphagov/content-publisher/blob/master/docs/withdrawing-and-removing-documents.md

(you'll soon regret ever starting this)

1pretz1:
@benthorner Yep we can, i'll address this in another PR however as we discussed offline as there is more than just this example that needs to be changed!

1pretz1:
@kevindew I've updated the guidance

docs/removing-documents.md
```diff
@@ -15,18 +15,7 @@ Environment variables are being used to pass parameters to the rake tasks.
 
 ## Withdrawing documents
```
kevindew: I'd just remove this section since the title no-longer references Withdrawing and we then end up with a h1 of Removing Documents followed by a h2 of Removing Documents.

1pretz1: Done

docs/removing-documents.md
```diff
@@ -1,33 +1,15 @@
...
-
-Optional parameters:
-
-- LOCALE (set to "en" by default)
-
-```
-rake unpublish:withdraw['a-content-id'] NOTE='A note'
-```
-
 ## Removing documents
```
kevindew: We should now look at changing this header to reference that it'll be gone as we still have a 
```
# Removing Documents
## Removing Documents
```
problem

1pretz1: Ah yes, doh!

docs/removing-documents.md
```diff
@@ -1,33 +1,15 @@
-# Withdrawing and removing documents
+# Remove / remove and redirect documents
```
benthorner: Is 'Remove' different to 'remove' 😆?

docs/removing-documents.md
```diff
@@ -1,33 +1,15 @@
...
 
 We need to allow users to unpublish content from GOV.UK. In Content Publisher we are replacing
 unpublish terminology with "withdrawn" and "removed".
```
benthorner: I think we need a bit more explanation about withdrawn, if we're mentioning it here. But I'm not really sure what the point of this file is - do we need it?

docs/removing-documents.md
1pretz1: Well no if you put it like that 🤣But as there are two different actions, either `Remove` or `Remove and redirect` this title tries to convey the content of the guidance.

docs/removing-documents.md
1pretz1: Not sure we do, but as we can't remove a document through the UI it's maybe best we keep it for guidance for people dealing with support requests maybe?

1pretz1: I'll just remove the mention of 'withdrawn' as it isn't relevant to the rest of the doc

1pretz1:
@benthorner updated as per your comments.

docs/removing-documents.md
benthorner: Oh yeah, I'm afraid I missed that. Would it be OK to go with 'removing'?

PR #876, opened 2019-03-13
Check govspeak is valid

kevindew:
Trello: https://trello.com/c/SdMUjUmj/722-resolve-xss-issues-raised-by-pen-testers

Marked as DNM until final content decisions.

This adds validation for the places where we accept govspeak input. It is used as a means to tell them that the content they have entered can't be accepted. Writing the validation message is rather grim as there are quite a few reasons the `Govspeak#valid?` method can fail. I felt the simplest way
to resolve this was just to tell users to "Remove HTML" - this hasn't been content designed though and may need revising if there are situations where users are actively advised to enter HTML.

### DIFF
--- a/app/controllers/withdraw_controller.rb
+++ b/app/controllers/withdraw_controller.rb
@@ -32,7 +32,9 @@ class WithdrawController < ApplicationController
           "items" => issues.items,
         }
 
-        render :new, assigns: { edition: edition }, status: :unprocessable_entity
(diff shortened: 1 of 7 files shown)


kevindew:
Thanks @emmabeynon - I'm going to get some content design help from @benhazell before merging.

kevindew:
thanks @emmabeynon 👍

PR #877, opened 2019-03-14
Edition loading controllers

kevindew:
This builds on top of the work of https://github.com/alphagov/content-publisher/pull/868 to add these methods into all of our controllers.

There aren't intended to be any functionality changes here. Small changes are: use of assigns in some places where instance variables are confusing, and using `edition.revision` in some places where we don't need to set it to a variable.

The biggest change is putting all the routes in a scope in: https://github.com/alphagov/content-publisher/blob/edition-loading-controllers/config/routes.rb

### DIFF
--- a/app/controllers/editions_controller.rb
+++ b/app/controllers/editions_controller.rb
@@ -2,31 +2,27 @@
 
 class EditionsController < ApplicationController
   def create
-    Document.transaction do
-      document = Document.with_current_edition.lock.find_by_param(params[:document_id])
-
-      current_edition = document.current_edition
-
-      unless current_edition.live?
+    Edition.find_and_lock_current(document: params[:document]) do |edition|
+      unless edition.live?
         # FIXME: this shouldn't be an exception but we've not worked out the
         # right response - maybe bad request or a redirect with flash?
         raise "Can't create a new edition when the current edition is a draft"
       end
 
-      current_edition.update!(current: false)
+      edition.update!(current: false)
 
       next_edition = Edition.find_by(
-        document: document,
-        number: current_edition.number + 1,
+        document: edition.document,
+        number: edition.number + 1,
       )
 
       if next_edition
-        next_edition.resume_discarded(current_edition, current_user)
+        next_edition.resume_discarded(edition, current_user)
 
         TimelineEntry.create_for_status_change(entry_type: :draft_reset,
                                                status: next_edition.status)
       else
-        next_edition = Edition.create_next_edition(current_edition, current_user)
+        next_edition = Edition.create_next_edition(edition, current_user)
 
         TimelineEntry.create_for_status_change(entry_type: :new_edition,
                                                status: next_edition.status)
@@ -34,7 +30,7 @@ class EditionsController < ApplicationController
 
       PreviewService.new(next_edition).try_create_preview
 
-      redirect_to edit_document_path(document)
+      redirect_to edit_document_path(next_edition.document)
     end
   end
 end
--- a/app/controllers/images_controller.rb
+++ b/app/controllers/images_controller.rb
@@ -7,14 +7,12 @@ class ImagesController < ApplicationController
   end
 
   def index
-    @document = Document.with_current_edition.find_by_param(params[:document_id])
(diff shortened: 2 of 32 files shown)


app/controllers/images_controller.rb
```diff
@@ -145,62 +131,58 @@ def update
...
-      document, image_revision = find_locked_document_and_image_revision(
-        params[:document_id],
-        params[:image_id],
-      )
-
-      current_edition = document.current_edition
-      current_revision = current_edition.revision
+    Edition.find_and_lock_current(document: params[:document]) do |edition|
+      image_revision = edition.image_revisions.find_by!(image_id: params[:image_id])
+      current_revision = edition.revision
```
benthorner: Minor, but it looks like you've been aiming to say `revision` instead of `current_revision` :-).

app/controllers/internal_notes_controller.rb
```diff
@@ -2,34 +2,34 @@
...
 class InternalNotesController < ApplicationController
   def create
-    Document.transaction do
-      document = Document.with_current_edition.lock!.find_by_param(params[:id])
+    Edition.find_and_lock_current(document: params[:document]) do |edition|
       note = params.fetch(:internal_note)
 
       if note&.chomp.blank?
-        return redirect_to_document_history(document)
+        redirect_to_document_history(edition)
```
benthorner: I like that we're not returning our redirects now :-)

app/controllers/internal_notes_controller.rb
```diff
@@ -2,34 +2,34 @@
...
+      redirect_to_document_history(edition)
     end
   end
 
 private
 
-  def redirect_to_document_history(document)
-    redirect_to("#{document_path(document)}#document-history")
+  def redirect_to_document_history(edition)
+    redirect_to("#{document_path(edition.document)}#document-history")
```
benthorner: Minor, but could we do `redirdct_to(document_path(edition.document), anchor: '...')`?

app/controllers/preview_controller.rb
```diff
@@ -2,24 +2,26 @@
 
 class PreviewController < ApplicationController
   def create
-    Document.transaction do
-      @document = Document.with_current_edition.lock.find_by_param(params[:id])
+    Edition.find_and_lock_current(document: params[:document]) do |edition|
```
benthorner: I'm wondering if we really need a lock here?

app/controllers/editions_controller.rb
```diff
@@ -2,39 +2,35 @@
 
 class EditionsController < ApplicationController
   def create
-    Document.transaction do
-      document = Document.with_current_edition.lock.find_by_param(params[:document_id])
-
-      current_edition = document.current_edition
-
-      unless current_edition.live?
+    Edition.find_and_lock_current(document: params[:document]) do |edition|
```
benthorner: Just to confirm, we don't need to lock on the document here, as we'll be updating the `current` property of the edition we're locking on?

app/controllers/images_controller.rb
kevindew: I'm going to go back to current_revision in the other file here for consistency, as the consequences of changing that for images is more complex.

This should be a low running concerns as your updater work clears it out.

app/controllers/internal_notes_controller.rb
kevindew: Aces, I didn't know that was an option.

app/controllers/preview_controller.rb
kevindew: We do as we modify the edition model, we set revision_synced. We also potentially modify some asset models.

app/controllers/editions_controller.rb
kevindew: We lock on the document and the edition with `find_and_lock_current` method so both bases are covered, but I think you're right that we'd get away with just a lock on edition here anyway.

app/controllers/preview_controller.rb
benthorner: Oh yes - I forgot about all the coordination the preview service does!

PR #881, opened 2019-03-14
Add Notify integration with Action Mailer

emmabeynon:
For https://trello.com/c/NUdRomQV/714-set-up-content-publisher-to-use-notify

This sets up this app to use GOV.UK Notify and creates a `NotifyDeliveryMethod` for future use as a delivery method by Action Mailer.  We also add a rake task so that we can test sending emails ahead of future mailers being created.


config/notify.yml
```diff
@@ -0,0 +1,10 @@
+<% fake_api_key = "fake_api_key-00000000-0000-0000-0000-000000000000-00000000-0000-0000-0000-000000000000" %>
```
kevindew: Does notify complain if the key is not in this format? just curious if we can get away with something simpler.

config/initializers/mailer.rb
```diff
@@ -0,0 +1,2 @@
+require Rails.root.join("lib", "notify_delivery_method")
```
kevindew: We can just use `require "notify_delivery_method"` I believe here, since lib is in the load path.

lib/notify_delivery_method.rb
```diff
@@ -0,0 +1,44 @@
+# frozen_string_literal: true
+
+require "notifications/client"
+
+class NotifyDeliveryMethod
+  def initialize(mail); end
```
kevindew: Does this need to exist if it doesn't do anything?

lib/notify_delivery_method.rb
```diff
@@ -0,0 +1,44 @@
...
+  def notify
+    @notify ||= notify_environment_config.symbolize_keys.freeze
+  end
+
+  def notify_config_path
+    Rails.root.join("config", "notify.yml")
+  end
+
+  def notify_environment_config
+    YAML.safe_load(ERB.new(File.read(notify_config_path)).result).fetch(Rails.env)
```
kevindew: Might be worth breaking this up into multiple lines?

Could even be worth moving this and notify_config_path into the same method since nothing else calls them and they're short.

lib/notify_delivery_method.rb
```diff
@@ -0,0 +1,44 @@
...
+  def initialize(mail); end
+
+  def deliver!(mail)
+    client.send_email(payload(mail))
+  end
+
+private
+
+  def client
+    @client ||= Notifications::Client.new(notify_api_key)
```
kevindew: We can probably access the api_key from notify here and save a few lines of code since it's similar in length.

config/notify.yml
```diff
@@ -0,0 +1,10 @@
...
+
+development:
+  api_key: <%= ENV.fetch('GOVUK_NOTIFY_API_KEY', fake_api_key) %>
+
+test:
+  api_key: <%= ENV.fetch('GOVUK_NOTIFY_API_KEY', fake_api_key) %>
+
+production:
+  api_key: <%= ENV.fetch('GOVUK_NOTIFY_API_KEY', nil) %>
```
kevindew: I just had a thought - and I'm sorry it's inconsistent with our earlier conversation 🤕 - but I think it'd actually be consistent with our app if we moved this to our [secrets file](https://github.com/alphagov/content-publisher/blob/master/config/secrets.yml) where we already have a precedent.

Doing this would mean we could replace `YAML.safe_load(ERB.new(File.read(notify_config_path)).result).fetch(Rails.env)` with `Rails.application.secrets.notify_api_key`

lib/notify_delivery_method.rb
emmabeynon: Ah yes - sorry this was a hangover from trying to figure out how to get it to work with the mailer.

config/notify.yml
emmabeynon: Yes unfortunately Notify complains about an invalid API key if it's not in a format that's something along the lines of "account_name-long-uuid".

emmabeynon:
@kevindew Thanks for the suggestions - `NotifyDeliveryMethod` is looking a lot cleaner now and the config is much simpler to access.  Will work on the rake task next 👍

lib/tasks/notify.rake
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+namespace :notify do
+  desc "Send an email notification"
+  task :send_email, [:email_address] => :environment do |_, args|
+    raise "Missing email address" unless args.email_address
+
+    headers = {
+      to: args.email_address,
+      template_id: ENV["TEMPLATE_ID"] || "759acac6-da53-4a19-b591-b7538c7c39de",
```
kevindew: Not a concern for now but do we need to set this key up to be an environment variable (ignoring the one as an argument) so we can have different ones in integration/staging/prod

lib/tasks/notify.rake
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+namespace :notify do
+  desc "Send an email notification"
+  task :send_email, [:email_address] => :environment do |_, args|
+    raise "Missing email address" unless args.email_address
+
+    headers = {
```
kevindew: minor, but I don't think these are headers and more parameters.

lib/tasks/notify.rake
```diff
@@ -0,0 +1,18 @@
...
+      to: args.email_address,
+      template_id: ENV["TEMPLATE_ID"] || "759acac6-da53-4a19-b591-b7538c7c39de",
+      body: ENV["BODY"] || "This is a test email notification",
+      subject: ENV["SUBJECT"] || "Test email notification",
+    }
+
+    message = Mail::Message.new(headers)
+    NotifyDeliveryMethod.new.deliver!(message)
```
kevindew: Are we able to use action mailer for this? Just as it stands we won't be using the action mailer configuration that's been wrote.

I figure we can just do this with something like `ActionMailer::Base.mail(headers).deliver_now` - although there could be a fatal flaw in my plan.

spec/lib/notify_delivery_method_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+
+      client = instance_double("Notifications::Client")
+      allow(Notifications::Client).to receive(:new).and_return(client)
+
+      expect(client).to receive(:send_email)
+        .with(email_address: headers[:to],
+              template_id: headers[:template_id],
+              personalisation: personalisation)
+
+      described_class.new.deliver!(message)
```
kevindew: I think this is our first usage of described_class. It'd be more consistent, for this codebase, to go with `NotifyDeliveryMethod`.

Aside from that though this test looks good 👍

spec/tasks/notify_rake_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+      allow(Notifications::Client).to receive(:new).and_return(client)
+
+      expect(client).to receive(:send_email)
+        .with(email_address: email_address,
+              template_id: template_id,
+              personalisation: { body: body, subject: subject })
+
+      ClimateControl.modify(TEMPLATE_ID: template_id, BODY: body, SUBJECT: subject) do
+        Rake::Task["notify:send_email"].invoke(email_address)
+      end
```
kevindew: If we're able to go with my suggestion of using actionmailer then I think we could benefit from using action mailer in the test too. 

From [googling around](https://www.lucascaton.com.br/2010/10/25/how-to-test-mailers-in-rails-with-rspec/#) it looks like there's an array set on ActionMail::Base.deliveries that can be checked for what has been mailed. That array [seems to contain Mail objects](https://github.com/mikel/mail/blob/master/lib/mail/network/delivery_methods/test_mailer.rb#L41) so could asset that it appended to that array a mail object with the parameters you'd expect.

I think doing that would abstract a couple of the tests here from being very similar to the notify one and also be an interesting thing to learn about how we should test emails our sent functionally.

config/notify.yml
```diff
@@ -0,0 +1,6 @@
+
+development:
+
+test:
+
+production:
```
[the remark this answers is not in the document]
emmabeynon: Oh whoops, didn't realise I'd added that 😅

lib/tasks/notify.rake
emmabeynon: I've used `headers` here as it's the terminology that's used in Action Mailer https://guides.rubyonrails.org/action_mailer_basics.html#complete-list-of-action-mailer-methods

lib/tasks/notify.rake
emmabeynon: Sure, once we've got the actual template ids then I can delete this one.

lib/tasks/notify.rake
kevindew: Ah yes I see what you mean. I don't think that is referring to the params for mail though - looks like that it's a different hash in a mailer for the email headers. The source code describes it better: https://github.com/rails/rails/blob/1cdc6c6576afe9a9fa67a6b313080a7dc62188d9/actionmailer/lib/action_mailer/base.rb#L50-L55

lib/tasks/notify.rake
emmabeynon: I'm getting `OpenSSL::SSL::SSLError: hostname "localhost" does not match the server certificate` when trying that 😕

emmabeynon: The documentation does suggest that a mailer model needs to be created to use Action Mailer https://apidock.com/rails/ActionMailer/Base

kevindew: Hmm do you have a stack trace of that? do you know if it's coming from notify code or not even making it to that? (could be falling back to a different delivery method maybe)

PR #883, opened 2019-03-15
No js insert image new tab

benthorner:
https://trello.com/c/RxrZShDD/694-open-a-new-tab-when-inserting-inline-images-without-javascript

    This changes the toolbar-dropdown component to work with links by
    default, so that we can use a target attribute to open a new tab for
    inserting images for users without JS. Moving the button-specific
    options into a namespace should make them easy to remove, as the need
    for them is coupled to the current way we insert contacts.

Note that the toolbar-dropdown component doesn't work in Safari, in case you end up testing this in that browser :-).

### DIFF
--- a/app/assets/stylesheets/components/_toolbar-dropdown.scss
+++ b/app/assets/stylesheets/components/_toolbar-dropdown.scss
@@ -62,6 +62,33 @@ $dropdown-arrow-spacing: $dropdown-arrow-size / 2;
   padding: 0;
 }
 
+.app-c-toolbar-dropdown__link {
+  @include govuk-font($size: 16);
+  @include govuk-focusable;
+
+  appearance: none;
+  -webkit-appearance: none;
+
+  background: govuk-colour("grey-4");
+  color: $govuk-text-colour;
+  padding: govuk-spacing(3);
+  text-decoration: none;
+  display: block;
+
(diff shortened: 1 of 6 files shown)


app/assets/stylesheets/components/_toolbar-dropdown.scss
```diff
@@ -62,6 +62,33 @@ $dropdown-arrow-spacing: $dropdown-arrow-size / 2;
   padding: 0;
 }
 
+.app-c-toolbar-dropdown__link {
```
kevindew: As this is almost the same as `.app-c-toolbar-dropdown__button` could they both use the same css and save having to duplicate it?

kevindew: Further to this, I wonder if they could both be the same, more generic, class and then we'd not need to duplicate the event handlers.

benthorner: I did go with that originally, but I think this is a case of 'similar but different'. The 'link' class doesn't require a few of the styles for buttons, plus it adds one of its own.

kevindew: okey doke

PR #889, opened 2019-03-19
Add template_id to Notify settings

emmabeynon:
https://trello.com/c/NUdRomQV/714-set-up-content-publisher-to-use-notify

Following convention of accessing delivery method config via Action Mailer settings, we add Notify template id to Notify settings. This removes the need for us to include a template id in the Mail object and pass a template id env to the `notify:send_email` rake task.

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -40,7 +40,10 @@ module ContentPublisher
     end
 
     config.after_initialize do
-      config.action_mailer.notify_settings = { api_key: Rails.application.secrets.notify_api_key }
+      config.action_mailer.notify_settings = {
+        api_key: Rails.application.secrets.notify_api_key,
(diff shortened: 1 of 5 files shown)


config/application.rb
```diff
@@ -40,7 +40,10 @@ class Application < Rails::Application
     end
 
     config.after_initialize do
-      config.action_mailer.notify_settings = { api_key: Rails.application.secrets.notify_api_key }
+      config.action_mailer.notify_settings = {
+        api_key: Rails.application.secrets.notify_api_key,
+        template_id: Rails.application.secrets.notify_template_id,
```
kevindew: I'd just look up the env var here as the template id is not actually a secret.

emmabeynon: Now changed 👍

PR #892, opened 2019-03-19
Add Mail Recipient Interceptor

emmabeynon:
For https://trello.com/c/eNpUY7Ek/744-add-email-intercept

We don't want users to receive notifications in integration and staging
environments as they could get confused with production notifications. While
Notify allows us to specify certain email addresses to only receive
notifications in integration and staging, if a notification is sent to a
non-allowed email address (for example in future when publishers receive
notifications that their content has been published) then Notify will return a
`BadRequestError` which is not ideal.

To avoid this, we create a Mail Recipient Interceptor to intercept emails that
are sent in integration and staging and route them to a team email address
instead of the original intended recipient.  We prefix the intended recipient to the
body of the email so that we can easily know who was intended to receive the 
intercepted emails.

### DIFF
--- a/config/secrets.yml
+++ b/config/secrets.yml
@@ -5,10 +5,8 @@ development:
 test:
   secret_key_base: secret
   jwt_auth_secret: secret
-  notify_non_production_allow_list: ["allowed@example.com"]
 
 production:
   secret_key_base: <%= ENV['SECRET_KEY_BASE'] %>
   jwt_auth_secret: <%= ENV['JWT_AUTH_SECRET'] %>
   notify_api_key: <%= ENV['GOVUK_NOTIFY_API_KEY'] %>
-  notify_non_production_allow_list: <%= ENV['GOVUK_NOTIFY_ALLOW_LIST'] %>
--- a/config/secrets.yml
+++ b/config/secrets.yml
@@ -5,8 +5,10 @@ development:
 test:
   secret_key_base: secret
   jwt_auth_secret: secret
+  notify_non_production_allow_list: ["allowed@example.com"]
 
 production:
   secret_key_base: <%= ENV['SECRET_KEY_BASE'] %>
   jwt_auth_secret: <%= ENV['JWT_AUTH_SECRET'] %>
   notify_api_key: <%= ENV['GOVUK_NOTIFY_API_KEY'] %>
+  notify_non_production_allow_list: <%= ENV['GOVUK_NOTIFY_ALLOW_LIST'] %>
--- a/spec/lib/mail_recipient_interceptor_spec.rb
+++ b/spec/lib/mail_recipient_interceptor_spec.rb
@@ -8,7 +8,7 @@ RSpec.describe MailRecipientInterceptor do
     let(:mail) { Mail.new(to: original_recipient) }
 
     before do
-      ClimateControl.modify(GOVUK_NOTIFY_ALLOW_LIST: "allowed@example.com") do
+      ClimateControl.modify(EMAIL_ADDRESS_OVERRIDE: "allowed@example.com") do
         MailRecipientInterceptor.delivering_email(mail)
       end
     end
--- a/spec/lib/mail_recipient_interceptor_spec.rb
+++ b/spec/lib/mail_recipient_interceptor_spec.rb
@@ -4,14 +4,21 @@ require "mail_recipient_interceptor"
 
 RSpec.describe MailRecipientInterceptor do
   describe "#delivering_email" do
-    it "intercepts emails and sends to an allow-listed email address" do
-      mail = Mail.new(to: "not-allowed@example.com")
+    let(:original_recipient) { "not-allowed@example.com" }
+    let(:mail) { Mail.new(to: original_recipient) }
 
+    before do
       ClimateControl.modify(GOVUK_NOTIFY_ALLOW_LIST: "allowed@example.com") do
         MailRecipientInterceptor.delivering_email(mail)
       end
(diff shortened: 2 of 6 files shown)


config/secrets.yml
```diff
@@ -5,8 +5,10 @@ development:
 test:
   secret_key_base: secret
   jwt_auth_secret: secret
+  notify_non_production_allow_list: ["allowed@example.com"]
 
 production:
   secret_key_base: <%= ENV['SECRET_KEY_BASE'] %>
   jwt_auth_secret: <%= ENV['JWT_AUTH_SECRET'] %>
   notify_api_key: <%= ENV['GOVUK_NOTIFY_API_KEY'] %>
+  notify_non_production_allow_list: <%= ENV['GOVUK_NOTIFY_ALLOW_LIST'] %>
```
kevindew: I don't think this is a secret so probably belongs in config/application.rb

kevindew: Also note that that this would be a string rather than an array unlike the one in test. I think it'd be better to treat it as a string though as I don't think we're expecting a need for multiple recipients.

lib/non_production_interceptor.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+class NonProductionInterceptor
+  def self.delivering_email(message)
+    email = Rails.application.secrets.notify_non_production_allow_list
+    body_prefix = "Intended recipient: #{message.to}\n"
```
kevindew: Do we expect message.to to be a string or an array? 

I ask as it looks like we set it to an array in the next statement but seem to be treating it here as a string.

lib/non_production_interceptor.rb
```diff
@@ -0,0 +1,11 @@
...
+
+    message.to = email
+    message.body = body_prefix + message.body.raw_source
```
kevindew: Looks good 👍

spec/lib/non_production_interceptor_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+
+    before { NonProductionInterceptor.delivering_email(mail) }
+
+    it "intercepts emails and sends to an allow-listed email address" do
+      expect(mail.to).to include("allowed@example.com")
+    end
+
+    it "intercepts emails and prefixes the original recipient to the body" do
+      expect(mail.body.raw_source).to include(original_recipient)
+    end
```
kevindew: This test looks good 👍

lib/non_production_interceptor.rb
emmabeynon: You can set `mail.to` to a string or an array, and when you call it will return an array https://www.rubydoc.info/github/mikel/mail/Mail/Message#to-instance_method

emmabeynon: I'll make a note of this in the commit message as it's not very clear.

kevindew: Sounds good - do we want to do a conversion of the array to a string for the interpolation?

config/initializers/non_production_interceptor.rb
```diff
@@ -0,0 +1,6 @@
+app_domain = ENV.fetch("GOVUK_APP_DOMAIN", "")
+non_production_environment = %w(integration staging).detect { |s| app_domain.include?(s) }
+
+if non_production_environment
+  ActionMailer::Base.register_interceptor(NonProductionInterceptor)
```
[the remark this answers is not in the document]
emmabeynon: I've gone for `MailRecipientInterceptor` - I think this make sense because it's the recipient we're interested in.  We're not mentioning override anywhere so I think we can get away with not including that concept in the name.

config/secrets.yml
emmabeynon: I've removed this and now referencing the env var directly in MailRecipientInterceptor.  I'll make sure the email is added as a string in Puppet 👍

lib/non_production_interceptor.rb
emmabeynon: I've called join on this so that it can handle multiple email addresses, should we decide to add more to the allow list at a later date.

config/initializers/non_production_interceptor.rb
```diff
@@ -0,0 +1,6 @@
+app_domain = ENV.fetch("GOVUK_APP_DOMAIN", "")
+non_production_environment = %w(integration staging).detect { |s| app_domain.include?(s) }
```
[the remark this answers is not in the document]
emmabeynon: Done 👍

spec/lib/mail_recipient_interceptor_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+
+require "mail_recipient_interceptor"
+
+RSpec.describe MailRecipientInterceptor do
+  describe "#delivering_email" do
+    let(:original_recipient) { "not-allowed@example.com" }
+    let(:mail) { Mail.new(to: original_recipient) }
+
+    before do
+      ClimateControl.modify(GOVUK_NOTIFY_ALLOW_LIST: "allowed@example.com") do
```
benthorner: Would it be possible to use multiple email addresses here? I'm not sure how this env var should look to achieve the 'listiness' :-).

benthorner: Discussed with @emmabeynon - since the NotifyDeliveryMethod payload only supports a single email, we're going to remove the '_LIST' aspect of this PR, since we only anticipate needing one email address for the time being.

PR #895, opened 2019-03-20
Migrate to updaters

benthorner:

This completes the work to move all revision updates to the Updaters pattern, which means we can now delete the old model code that used to do this 🎉 .


1pretz1:
Great work @benthorner!
Screen Shot 2019-03-26 at 11 43 14

All looks good to me apart from you have a few conflicts :)

benthorner:
@1pretz1 this should be ready now :-)

PR #905, opened 2019-03-25
Refactor and fix transactions

1pretz1:
In a few places within the app we seem to use nested transactions which aren't actually necessary as long as the rescue block is put in the correct place.

## Before
Previously there was a nested transaction where the parent transaction didn't perform any database operations and had a confusing rollback procedure.

If we just removed the nested transaction and did nothing else then
this would not work as expected. The reason for this is that when a
GdsApi exception is thrown the exception would be rescued before the
transaction had a chance to roll back, meaning records could be
updated / created even if a GdsApi error is thrown.

## Now 
The rescue is outside the only transaction, this means when a GdsApi exception
is thrown the transaction handles all the roll backs it needs to do,
uninterrupted. Then when this is complete the exception is bubbled up to
the containing method which is then rescued successfully.

Having a single transaction is better due to it being easier to understand.

### DIFF
--- a/app/controllers/unwithdraw_controller.rb
+++ b/app/controllers/unwithdraw_controller.rb
@@ -13,14 +13,12 @@ class UnwithdrawController < ApplicationController
 
   def unwithdraw
     Edition.find_and_lock_current(document: params[:document]) do |edition|
-      begin
-        UnwithdrawService.new.call(edition, current_user)
-        redirect_to edition.document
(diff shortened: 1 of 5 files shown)


benthorner:
Thanks for making this clearer - I think we need to give some more explanation about the behaviour before and after, as it's quite subtle :-).

1pretz1:
Yep sure, i'll update the comment / commit message to give more detail

PR #908, opened 2019-03-26
Use ActiveJob as a wrapper over Sidekiq

benthorner:

Previously we used Sidekiq workers directly. This modifies the existing
code to instead implement the ActiveJob framework, using Sidekiq as the
queueing adapter. Since we're going to use ActiveJob for sending emails
it makes sense to use a consistent framework elsewhere if possible.

In the past ActiveJob has been a poor substitute for Sidekiq when it
came to handling failure and retries. But recently ActiveJob has
improved to the extent that it's a viable drop-in replacement for
Sidekiq, provided we disable Sidekiq's native retry feature!

Moving to ActiveJob has simplified the failure handling compared to
Sidekiq, as the default behaviour of ActiveJob is not to retry; this
makes it much simpler to abort a job, with the compromise that we now
have to be specific about the exceptions we *do* want to retry_on.

https://github.com/rails/rails/blob/5-1-stable/activejob/CHANGELOG.md


app/workers/scheduled_publishing_job.rb
```diff
@@ -16,21 +15,19 @@ def perform(id)
       PublishService.new(edition.document)
                     .publish(user: user, with_review: reviewed)
     end
-  rescue ActiveRecord::RecordNotFound
```
emmabeynon: Just wondering why we're no longer raise a specific error here?

app/workers/scheduled_publishing_job.rb
```diff
@@ -1,10 +1,9 @@
 # frozen_string_literal: true
 
-class ScheduledPublishingWorker
-  include Sidekiq::Worker
-  # We want to retry for up to 5 minutes. 10 retries x 30s intervals = 5 minutes.
-  sidekiq_options retry: 10
-  sidekiq_retry_in { 30 }
+class ScheduledPublishingJob < ApplicationJob
+  retry_on(GdsApi::BaseError,
```
emmabeynon: This doesn't seem to match up with the number of retries in the previous implementation?

app/workers/scheduled_publishing_job.rb
benthorner: Good question :-). Previously we were rescuing this error in order to abort immediately, since it's impossible to send an email for an edition that does not exist e.g. if it's been wiped on integration. Because we're using ActiveJob, which aborts by default (vs. Sidekiq), we can just let the exception bubble-up, rather than having to instruct Sidekiq to abort.

app/workers/scheduled_publishing_job.rb
benthorner: I'm happy to keep the original behaviour, but I thought it would be a good to use the more sophisticated retry mechanism provided by ActiveJob. These settings mean we will retry for the same amount of time, but with fewer attempts. The thinking here is similar to protocols like TCP/WiFi, where the idea is to give the remote increasing amounts of time to recover, rather than pummelling it constantly when we know it's already in trouble.

emmabeynon: Ok, 3 attempts seems a little low to me at first glance but the initial "retry every 30 seconds for up to 5 minutes" was just a first pass at a limit anyway.  I'd find it helpful if there was a comment either in the code or in the commit message to illustrate why we're changing to use `:exponentially_longer` and what that * 3 retry attempts means in terms of time so future devs can more quickly make a decision about how to change that.

app/workers/scheduled_publishing_job.rb
emmabeynon: Ok that makes sense.  I wonder if it would still be helpful from a debugging perspective to log that the edition was not found in a scheduled publishing context?

benthorner: It does now :-).

app/workers/scheduled_publishing_job.rb
benthorner: Turns out I forgot to increase it to 5 anyway, which does give us a 5 minute period. Added a doc too!

benthorner:
@kevindew this should be ready to review again :-).

app/jobs/scheduled_publishing_job.rb
```diff
@@ -0,0 +1,38 @@
...
+      PublishService.new(edition.document)
+                    .publish(user: user, with_review: reviewed)
+    end
+  end
+
+private
+
+  def no_longer_schedulable?(edition)
+    unless edition.scheduled?
+      Rails.logger.warn("Cannot publish an edition that is not scheduled")
```
kevindew: Small thing that I missed last time, it'd probably be useful if this and the other log included the edition id for them to be something we can identify from a log.

kevindew:
Oh and worth keeping notes on things we've lost and gained by the switch to ActiveJob for the [ADR](https://trello.com/c/XBbVir3W/839-write-an-adr-documenting-tech-product-decisions-around-notifications)

If I remember right we identified that we lose the retry set from the sidekiq dashboard. Do you know if we still get failed jobs? Wondering as that data is sent to statsd and used in some grafana dashboards, but I don't think it's a part of any particular monitoring.

app/jobs/scheduled_publishing_job.rb
benthorner: Sure - I've amended with the ID :-).

kevindew: Awesome 👍

benthorner:
@kevindew I think we saw that jobs went straight to 'dead' state, as the 'failed' state is associated with retries, which we've turned off for Sidekiq. I've added a [checklist on the ADR card](https://trello.com/c/XBbVir3W/839-write-an-adr-documenting-tech-product-decisions-around-notifications) to try and capture everything we intend to document.

PR #909, opened 2019-03-26
Publish notification 2

benthorner:

This adds the ability to notify the editors of an edition by email when their document is published, with or without review. Editors are defined to be any user who has modified or change the state of an edition. Please see the commit for further details :-).

### DIFF
--- a/app/controllers/publish_controller.rb
+++ b/app/controllers/publish_controller.rb
@@ -18,7 +18,12 @@ class PublishController < ApplicationController
   end
 
   def publish
-    Edition.find_and_lock_current(document: params[:document]) do |edition| # rubocop:disable Metrics/BlockLength
+    edition = Edition.find_current(document: params[:document])
+    live_edition = nil
+
+    Edition.transaction do
+      edition.lock!
+
       if params[:review_status].blank?
         flash.now["alert_with_items"] = {
           "title" => I18n.t!("publish.confirmation.flashes.requirements"),
@@ -28,12 +33,12 @@ class PublishController < ApplicationController
         render :confirmation,
                assigns: { issues: review_status_issues, edition: edition },
                status: :unprocessable_entity
-        next
+        return
       end
 
       if edition.live?
         redirect_to published_path(edition.document)
-        next
+        return
       end
 
       with_review = params[:review_status] == "reviewed"
@@ -43,16 +48,17 @@ class PublishController < ApplicationController
                                      .publish(user: current_user, with_review: with_review)
       rescue GdsApi::BaseError
         redirect_to edition.document, alert_with_description: t("documents.show.flashes.publish_error")
-        next
+        return
       end
 
       TimelineEntry.create_for_status_change(
         entry_type: with_review ? :published : :published_without_review,
         status: live_edition.status,
       )
-
-      redirect_to published_path(edition.document)
     end
+
+    send_notifications(live_edition)
+    redirect_to published_path(edition.document)
   end
 
   def published
@@ -65,5 +71,11 @@ private
     @review_status_issues ||= Requirements::CheckerIssues.new([
       Requirements::Issue.new(:review_status, :not_selected),
     ])
+  end
+
+  def send_notifications(edition)
+    edition.editors.each do |user|
+      PublishMailer.publish_email(edition.status, user).deliver_later
+    end
   end
 end
--- a/app/controllers/publish_controller.rb
+++ b/app/controllers/publish_controller.rb
@@ -18,11 +18,10 @@ class PublishController < ApplicationController
   end
 
   def publish
-    edition = Edition.find_current(document: params[:document])
-    live_edition = nil
+    edition = nil
 
-    Edition.transaction do
-      edition.lock!
+    Edition.transaction do # rubocop:disable Metrics/BlockLength
+      edition = Edition.lock.find_current(document: params[:document])
 
       if params[:review_status].blank?
         flash.now["alert_with_items"] = {
@@ -44,8 +43,8 @@ class PublishController < ApplicationController
       with_review = params[:review_status] == "reviewed"
 
       begin
-        live_edition = PublishService.new(edition.document)
-                                     .publish(user: current_user, with_review: with_review)
+        PublishService.new(edition)
+                      .publish(user: current_user, with_review: with_review)
       rescue GdsApi::BaseError
         redirect_to edition.document, alert_with_description: t("documents.show.flashes.publish_error")
         return
@@ -53,11 +52,11 @@ class PublishController < ApplicationController
 
       TimelineEntry.create_for_status_change(
         entry_type: with_review ? :published : :published_without_review,
-        status: live_edition.status,
+        status: edition.status,
       )
     end
 
-    send_notifications(live_edition)
+    send_notifications(edition)
     redirect_to published_path(edition.document)
   end
 
--- a/app/controllers/publish_controller.rb
+++ b/app/controllers/publish_controller.rb
@@ -75,7 +75,7 @@ private
 
   def send_notifications(edition)
     edition.editors.each do |user|
-      PublishMailer.publish_email(edition.status, user).deliver_later
+      PublishMailer.publish_email(edition, user).deliver_later
     end
   end
 end
--- a/app/models/edition.rb
+++ b/app/models/edition.rb
@@ -42,6 +42,8 @@ class Edition < ApplicationRecord
 
   has_and_belongs_to_many :revisions
 
+  has_many :statuses
+
   has_many :internal_notes
 
   delegate :content_id, :locale, :document_type, :topics, :document_topics, to: :document
@@ -178,5 +180,10 @@ class Edition < ApplicationRecord
                       last_edited_at: Time.current)
 
     self
+  end
+
+  def editors
+    user_ids = statuses.pluck(:created_by_id) + revisions.pluck(:created_by_id)
+    User.where(id: user_ids.uniq)
   end
 end
--- a/app/services/publish_service.rb
+++ b/app/services/publish_service.rb
@@ -1,16 +1,18 @@
 # frozen_string_literal: true
 
 class PublishService
-  attr_reader :document
-  delegate :current_edition, :live_edition, to: :document
+  attr_reader :edition
+  delegate :document, to: :edition
 
-  def initialize(document)
-    @document = document
+  def initialize(edition)
+    @edition = edition
   end
 
   def publish(user:, with_review:)
+    live_edition = document.live_edition
+
     publish_new_images
-    retire_old_images
+    retire_old_images(live_edition)
 
     GdsApi.publishing_api_v2.publish(
       document.content_id,
@@ -18,11 +20,11 @@ class PublishService
       locale: document.locale,
     )
 
-    supersede_live_edition(user)
+    supersede_live_edition(live_edition, user)
     set_new_live_edition(user, with_review)
     set_first_published_at
 
-    current_edition
+    document.reload
   rescue GdsApi::BaseError => e
     GovukError.notify(e)
     raise
@@ -30,7 +32,7 @@ class PublishService
 
 private
 
-  def supersede_live_edition(user)
+  def supersede_live_edition(live_edition, user)
     return unless live_edition
 
     live_edition.assign_status(:superseded, user, update_last_edited: false)
@@ -40,9 +42,9 @@ private
 
   def set_new_live_edition(user, with_review)
     status = with_review ? :published : :published_but_needs_2i
-    current_edition.assign_status(status, user)
-    current_edition.live = true
-    current_edition.save!
+    edition.assign_status(status, user)
+    edition.live = true
+    edition.save!
   end
 
   def set_first_published_at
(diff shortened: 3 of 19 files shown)


spec/features/workflow/publish_spec.rb
```diff
@@ -21,10 +53,12 @@ def when_i_visit_the_summary_page
   end
 
   def and_i_publish_the_edition
-    click_on "Publish"
-    choose I18n.t!("publish.confirmation.has_been_reviewed")
-    @content_request = stub_publishing_api_publish(@edition.content_id, update_type: nil, locale: @edition.locale)
-    click_on "Confirm publish"
+    travel_to(@publish_date = Time.current + 1.day) do
```
emmabeynon: What's the reason for needing to travel forward in time in order to publish?

spec/features/workflow/publish_spec.rb
```diff
@@ -41,4 +75,59 @@ def and_the_content_is_shown_as_published
...
+    expect(message.subject).to eq(I18n.t("publish_mailer.publish_email.subject.update",
+                                         title: @edition.title))
+
+    expect(message.body).to include(I18n.t("publish_mailer.publish_email.details.update",
+                                           time: publish_time,
+                                           date: publish_date,
+                                           user: publish_user))
+  end
+
+  def then_i_receive_an_email_about_the_minor_change
```
emmabeynon: As this is very similar to `#then_i_receive_an_email_about_the_major_change` I wonder if it's enough here just to test that the email includes the minor update text?

spec/features/workflow/publish_spec.rb
benthorner: This is to ensure the date is distinct from other timestamps in the edition, but I've simplified it slightly by modifying `created_at` when the edition is created, so we're not adding a magic number when we freeze time.

spec/features/workflow/publish_spec.rb
benthorner: Sure - changed :-).

app/models/edition.rb
```diff
@@ -41,6 +41,7 @@ class Edition < ApplicationRecord
   has_many :timeline_entries
 
   has_and_belongs_to_many :revisions
+  has_many :statuses
```
kevindew: Minor but other associations have a new line preceding them.

app/models/edition.rb
```diff
@@ -179,4 +180,8 @@ def assign_revision(revision, user)
 
     self
   end
+
+  def editors
+    (revisions + statuses).map(&:created_by).uniq
```
kevindew: This would cause one query for every revision and status so I think we should refactor it as that could be a ton of queries.

I did some experimenting to see if it could go into a `has_many` but I hit issues with that. This was as far as I got:

```
  has_many :editors,
           -> (edition) {
            user_ids = edition.statuses.pluck(:created_by_id) + edition.revisions.pluck(:created_by_id)
            where(id: user_ids).distinct
           },
           class_name: "User"
```

But I couldn't kill a scope of where `edition_id = edition.id` so that wasn't going anywhere. I think the right way for a has_many would be denormalising it anyway.

So could use that technique in this method to reduce the queries. 

I do feel like we might possibly hit a reason that we do want to denormalise this at some point though. As it feels a bit delicate that this pulls data off immutable data and thus couldn't be altered.

config/sidekiq.yml
```diff
@@ -5,3 +5,6 @@
 :logfile: ./log/sidekiq.json.log
 development:
   :logfile: null
+:queues:
+  - default
+  - mailers
```
kevindew: might be premature to have a second queue?

app/services/publish_service.rb
```diff
@@ -1,11 +1,12 @@
...
-  attr_reader :document
-  delegate :current_edition, :live_edition, to: :document
+  attr_reader :document, :current_edition
+  delegate :live_edition, to: :document
 
-  def initialize(document)
-    @document = document
+  def initialize(edition)
+    @current_edition = edition
+    @document = edition.document
```
kevindew: I feel like the changes here have impacted this class negatively. It seems a bit strange we have instance variables for document and current_edition but not live edition.

Was there a particular reason this was changed?

If so I wonder if it'd be cleaner if we had an instance variable of edition and delegated document to that.

app/views/publish_mailer/publish_email.text.erb
```diff
@@ -0,0 +1,36 @@
+<%= @edition.document_type.label %> ‘<%= @edition.title %>’
+
+<% if @edition.number > 1 %>
+  <%= t("publish_mailer.publish_email.details.update",
+    time: @status.created_at.strftime("%l:%M%P"),
+    date: @status.created_at.strftime("%d %B %Y"),
+    user: @status.created_by.name,
+  ) %>
```
kevindew: It looks a bit like this would indent the text being echoed here. Does that cause any oddities with either the HTML or text view in the Notify email?

It might need to use the erb whitespace suppression features.

app/views/publish_mailer/publish_email.text.erb
```diff
@@ -0,0 +1,36 @@
...
+    <%= t("publish_mailer.publish_email.minor_update") %>
+  <% end %>
+<% end %>
+
+<% if @edition.published_but_needs_2i? %>
+  <%= t("publish_mailer.publish_email.2i_warning") %>
+<% end %>
+
+<%= t("publish_mailer.publish_email.edit_in_app") %>
+<%= document_url(@edition.document) %>
```
kevindew: I'm not sure we gain a lot by this using translation phrases since it's a text file and would presumably be easier to edit here than through translation files. I'm imagining if a content designer was to come to edit this copy it'd be quite a complicated process where it's quite hard to determine the email copy due to the translations.

I'm thinking if we were to support multiple languages we'd probably have separate text view files for each language rather than phrases, since the logic is all around text rather than being wrapped in components/HTML.

Apologies if we already have a precedent and I missed this the first time around, what are your thoughts?

spec/features/workflow/publish_without_review_spec.rb
```diff
@@ -1,11 +1,14 @@
 # frozen_string_literal: true
 
-RSpec.feature "Publish without review" do
+RSpec.feature "Publish without review", inline: true do
```
kevindew: It wasn't until I reached the spec file that I realised inline true was for sidekiq. Perhaps call it sidekiq_inline? Could also be that we want it to default to sidekiq inline for all feature tests as I can't imagine we'd want it it not be inline for feature tests.

app/services/publish_service.rb
benthorner: So this is because of the ambiguous `inverse_of` between current/live editions and their document. We're passing `edition.document` to the service, but this doesn't prevent a DB query when the service tries to do `(edition.)document.current_edition`. The end result is that the `edition` variable in the controller needs to be reloaded.

It looks like this is actually 'fixed' automatically by the serialisation/deserialisation that happens when we pass the edition to the mailer and invoke `deliver_later`. So we could just roll with that, but it's not very intuitive. What might be better is to run the controller by finding/locking on the document and using `with_current_edition`.

How does that sound?

app/views/publish_mailer/publish_email.text.erb
benthorner: Good thought - I'd neglected to check the text variant. Turns out it's mostly OK, apart from a couple of excess newlines. I've added a commit to sort this out a bit, but it's still not perfect (there will be a blank line if there's no 2i warning), and I wonder if it's better to just not worry too much and write templates in the normal HTML style?

app/views/publish_mailer/publish_email.text.erb
benthorner: This one is a bit unusual, as normally we're passing translations as arguments to a component, whereas here they're just inline. But this is covered in the [ADR](https://github.com/alphagov/content-publisher/blob/master/docs/adr/0004-editing-microcopy.md#decision) under 'All other strings', so I've stuck with that. 

The only place I'm aware we're not consistent with this is the [scheduling form](https://github.com/alphagov/content-publisher/blob/master/app/views/documents/show/_scheduling_form.html.erb). I think we've avoided steering content designers towards ERB files thus far - it starts to make it hard to find where a particular string is. 

Possibly this is just another occurrence of where the scope of the original card was too big, but on the other hand, it didn't bother me much putting them in.

app/models/edition.rb
benthorner: The scope looks pretty good to me, as it bounds the number of queries. However, I'm assuming it will be advantageous to denormalise the editors in order to filter by them on the index page. Would a workable compromise be to use the scope for now, but maybe do a separate PR where we look at making editors a persisted concept?

spec/features/workflow/publish_without_review_spec.rb
benthorner: I'm used to having Sidekiq inline in all tests, but we've actually got a use case for that: the `schedule_spec`. I'd like to keep Sidekiq out of the tests if possible, so how about one of: `background` or `async`?

config/sidekiq.yml
benthorner: Ah, so this was just the ActionMailer default and I couldn't think of a good reason to override it. I'm happy to, if you think it would cause confusion :-).

app/models/edition.rb
kevindew: Yup that sounds good, so we'll plan to denormalise as part of index?

That scope I wrote doesn't work, it'd have to be in your method by the way.

config/sidekiq.yml
kevindew: nah - seems fine if it's a default.

app/services/publish_service.rb
kevindew: Oh darn, that makes sense. Yeah that suggestion sounds good - looks like we're only really dealing with a document in the controller anyway.

app/views/publish_mailer/publish_email.text.erb
kevindew: I think we should try make best efforts with it as it'll be so hard to quantify whether or not people are seeing the text ones and we'd want them to not look broken.

Thanks for sorting those bits out.

It could be easier to have a grip on the whitespace by writing them in a less HTML style e.g.

```
<%= 
if @edition.number > 1
  if @edition.major?
    t("publish_mailer.publish_email.change_note") + "\n" + @edition.change_note
 else
    t("publish_mailer.publish_email.minor_update")
end
%>

<%= t("publish_mailer.publish_email.2i_warning") if @edition.published_but_needs_2i? %>
```

app/views/publish_mailer/publish_email.text.erb
kevindew: Yeah, I'd agree this one is unusual. To me it fits more into the realm of the guidance files where we're already sending content designers to erb files, e.g: https://github.com/alphagov/content-publisher/blame/master/app/views/new_document/guidance.govspeak.md but can also understand how it fits into the definition from the ADR. 

For the most part I'm expecting content designers have to search for a phrase whatever since they don't know the view structure.

spec/features/workflow/publish_without_review_spec.rb
kevindew: Oh yes - I remember that one. Darn, I'd forgotten that test. We can set that one inline but from my googling it'll behave a bit weird and run straight away: https://github.com/mperham/sidekiq/issues/2255

Maybe it's best to set all feature tests to run inline and for that (and any other ones) to have a before block of `before { Sidekiq::Testing.fake! }`.

I don't think any option that is provided just for sidekiq can be anything but opaque unless it mentions sidekiq in it.

app/models/edition.rb
benthorner: Ah, I see what you mean now - that's probably for the best :-). I've added a note to the docs section of the card about searching for editors.

spec/features/workflow/publish_without_review_spec.rb
benthorner: Yeah, I think I tried that already, and I the fact it behaves weird with scheduling wasn't surprising, really. I've changed to run everything inline except that one spec.

app/services/publish_service.rb
benthorner: Turns out that doesn't work either, as the PublishService is modifying the attributes on the individual editions, rather than re-assigning stuff on the document. I had a go at changing the service, but the assignment for `live_edition` doesn't work, as re-assigning it tries to set `document_id` to nil on the previous live edition, which breaks a foreign key constraint.

So now we have something else, which actually points to some kind of global `context` variable :-0.

app/controllers/publish_controller.rb
```diff
@@ -18,41 +18,54 @@ def confirmation
   end
 
   def publish
-    Edition.find_and_lock_current(document: params[:document]) do |edition|
+    edition = Edition.find_current(document: params[:document])
+
+    Edition.transaction do
+      edition.lock!
```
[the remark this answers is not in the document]
benthorner: Are you sure? My understanding was that we lock on the edition - we're just doing `lock` in the class method of the Edition model - and this has the side-effect of locking on all associated rows, or at least that's what [this blog](http://shiroyasha.io/selecting-for-share-and-update-in-postgresql.html) says.

app/views/publish_mailer/publish_email.text.erb
benthorner: I had a go at that one, but unfortunately it recreates the problem of an extra blank line if there's no 2i warning - it's tempting to have a second view to cope with this case, but I think they're too similar to justify that.

app/views/publish_mailer/publish_email.text.erb
benthorner: The ADR does also have a specific rule for "big guidance", so I'm reluctant to lump this in with that - it was more about the nature of the content, as opposed to the format it's in.

I wouldn't say this has caused me any pain, but if it's something we'd like to revisit, perhaps the best approach is to open an issue to look at it again - having rules to follow does help simplify this kind of scenario, but I remember we never really satisfied everyone.

app/services/publish_service.rb
benthorner: Refactored again as discussed :-).

benthorner:
@kevindew this should be ready for review again...

app/services/publish_service.rb
```diff
@@ -52,42 +54,32 @@ def set_first_published_at
...
     return unless live_edition
 
-    image_revisions = current_edition.image_revisions
-    current_file_revisions = image_revisions.map(&:file_revision)
-    current_revisions_by_image_id = image_revisions.group_by(&:image_id)
-
-    live_edition.image_revisions.each do |revision|
-      next if current_file_revisions.include?(revision.file_revision)
+    live_edition.image_revisions.each do |old_revision|
+      new_revision = edition.image(old_revision.image_id)
```
kevindew: Were you planning to do something different with this? I'm assuming it's slipped through as I can't see this method defined.

app/services/publish_service.rb
```diff
@@ -52,42 +54,32 @@ def set_first_published_at
...
+      next unless asset.draft?
 
-        if asset.draft?
-          AssetManagerService.new.publish(asset)
-          asset.live!
-        end
-      end
+      AssetManagerService.new.publish(asset)
+      asset.live!
     end
```
kevindew: This is a lot nicer 👍

app/services/publish_service.rb
benthorner: Oops - sorry about that. Fixed.

PR #913, opened 2019-03-26
Use date-input component for scheduling date

alex-ju:
- Replace the multiple inputs for inserting date with the date-input component
- Adjust whitespacing in the details element within the side area to fit the default date-input fields in one line

[Trello card](https://trello.com/c/0ZphzvRd)

### DIFF
--- a/app/assets/stylesheets/objects/_side.scss
+++ b/app/assets/stylesheets/objects/_side.scss
@@ -21,4 +21,8 @@
   .govuk-section-break {
     clear: both;
   }
+
+  .govuk-details__text {
+    padding: govuk-spacing(3);
+  }
 }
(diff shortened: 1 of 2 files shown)


app/assets/stylesheets/application.scss
```diff
@@ -1,5 +1,7 @@
 @import "govuk_publishing_components/all_components";
 
+@include _govuk-font-face-nta;
```
benthorner: It would be helpful to see an example of something that was broken that this fixes. Also, I'm wondering how anyone else would know they needed to add this in another app - that might just be my lack of experience with govuk-frontend, though!

alex-ju: The example is https://content-publisher.integration.publishing.service.gov.uk/ using the fallback font. It probably makes more sense for me to extract this into a separate PR.

benthorner: 👍

alex-ju: Moved it to https://github.com/alphagov/content-publisher/pull/917

PR #914, opened 2019-03-26
A different take on the Interactor pattern

benthorner:
https://trello.com/c/vf6xQW7w/739-prototype-images-controller-with-interactors-pattern

The Interactor pattern is quite simple, but trying to use it leaves two
issues unresolved: the verbose passing of variables, either with context
or between methods; and the structure of the actions we're trying to
encapsulate, most which have a somewhat hidden lock/release mechanism.

Re-implementing the Interactor pattern locally gives the opportunity to
address some of these issues, but also to go completely off-the-rails!
This is really another example of what the end-result could look like,
which has some good and bad attributes over vanilla Interactors.

### DIFF
--- a/app/beneractors/images/create_actor.rb
+++ b/app/beneractors/images/create_actor.rb
@@ -1,0 +1,31 @@
+# frozen_string_literal: true
+
+require "beneractors/beneractor"
+
+module Images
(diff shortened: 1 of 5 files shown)


benthorner:
Closing as I think we've got all the comments we want from this now :-).

PR #915, opened 2019-03-27
Fix and improve radio buttons on confirmation

1pretz1:
Currently when a user clicks 'Publish' on the confirmation screen without selecting whether the document has been approved or not we default the document to 'published without review', this is a bad user experience. 

We should ensure that a user confirms a review status when publishing / scheduling a document.

### DIFF
--- a/app/controllers/schedule_controller.rb
+++ b/app/controllers/schedule_controller.rb
@@ -43,6 +43,15 @@ class ScheduleController < ApplicationController
 
   def schedule
     Edition.find_and_lock_current(document: params[:document]) do |edition|
+      if params[:review_status].nil?
+        flash["alert_with_items"] = {
+          "title" => t("schedule.confirmation.radio_not_selected.title"),
+          "items" => [{ "text": t("schedule.confirmation.radio_not_selected.description_govspeak") }],
+        }
+        redirect_to scheduling_confirmation_path(edition.document)
+        next
+      end
+
       if edition.scheduled_publishing_datetime.blank?
         # FIXME: this shouldn't be an exception but we've not worked out the
         # right response - maybe bad request or a redirect with flash?
@@ -52,7 +61,7 @@ class ScheduleController < ApplicationController
       datetime = edition.scheduled_publishing_datetime
       ScheduledPublishingWorker.perform_at(datetime, edition.id)
 
-      reviewed = review_params == "reviewed"
+      reviewed = params[:review_status] == "reviewed"
       ScheduleService.new(edition).schedule(user: current_user, reviewed: reviewed)
 
       redirect_to scheduled_path(edition.document)
@@ -70,10 +79,6 @@ private
 
   def permitted_params
     params.require(:scheduled).permit(:year, :month, :day, :time)
-  end
-
-  def review_params
-    params.require(:review_status)
   end
 
   def set_scheduled_publishing_datetime(edition, datetime = nil)
--- a/config/locales/en/schedule/confirmation.yml
+++ b/config/locales/en/schedule/confirmation.yml
@@ -6,3 +6,6 @@ en:
       review_status:
         reviewed: This content has been reviewed and approved for publication.
         not_reviewed: This content needs to be published urgently but should be reviewed as soon as possible.
+      radio_not_selected:
+        title: You must
+        description_govspeak: Select a publishing option
--- a/app/controllers/publish_controller.rb
+++ b/app/controllers/publish_controller.rb
@@ -19,19 +19,27 @@ class PublishController < ApplicationController
 
   def publish
     Edition.find_and_lock_current(document: params[:document]) do |edition|
+      if params[:review_status].nil?
+        flash["alert_with_items"] = {
+          "title" => t("publish.confirmation.radio_not_selected.title"),
+          "items" => [{ 'text': t("publish.confirmation.radio_not_selected.description_govspeak") }],
+        }
+        redirect_to publish_confirmation_path(edition.document)
+        next
+      end
+
       if edition.live?
         redirect_to published_path(edition.document)
         next
       end
 
-      with_review = params[:review_state] == "reviewed"
+      with_review = params[:review_status] == "reviewed"
 
       begin
         live_edition = PublishService.new(edition.document)
                                      .publish(user: current_user, with_review: with_review)
       rescue GdsApi::BaseError
-        redirect_to edition.document,
-                    alert_with_description: t("documents.show.flashes.publish_error")
+        redirect_to edition.document, alert_with_description: t("documents.show.flashes.publish_error")
         next
       end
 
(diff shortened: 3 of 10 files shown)


app/controllers/schedule_controller.rb
```diff
@@ -43,6 +43,12 @@ def confirmation
 
   def schedule
     Edition.find_and_lock_current(document: params[:document]) do |edition|
+      if params[:review_status].nil?
```
benthorner: Could we synchronise the naming (`status` vs. `state`) here?

config/locales/en/schedule/confirmation.yml
```diff
@@ -6,3 +6,6 @@ en:
       review_status:
         reviewed: This content has been reviewed and approved for publication.
         not_reviewed: This content needs to be published urgently but should be reviewed as soon as possible.
+      radio_not_selected:
+        title: No radio button has been selected
```
benthorner: I think the card had some different copy to use - could you double check?

spec/features/workflow/publish_without_selection_spec.rb
```diff
@@ -0,0 +1,31 @@
+# frozen_string_literal: true
+
+RSpec.feature "Attempting to publish without confirming a review status" do
+  scenario do
+    given_there_is_an_edition_in_draft
+    when_i_visit_the_summary_page
+    and_try_and_publish_without_selecting_a_review_status
```
benthorner: Also good to synchronise the terminology here (`status` vs. `state`).

spec/features/workflow/publish_without_selection_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+    @edition = create(:edition, :publishable)
+  end
+
+  def when_i_visit_the_summary_page
+    visit document_path(@edition.document)
+  end
+
+  def and_try_and_publish_without_selecting_a_review_status
+    click_on "Publish"
+    stub_publishing_api_publish(@edition.content_id, update_type: nil, locale: @edition.locale)
```
benthorner: I don't think we need to have this stub here, as hopefully we won't be doing anything :-).

spec/features/workflow/schedule_without_selection_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+    @edition = create(:edition, scheduled_publishing_datetime: Time.current.tomorrow)
+  end
+
+  def when_i_visit_the_summary_page
+    visit document_path(@edition.document)
+  end
+
+  def and_try_and_schedule_without_selecting_a_review_status
+    click_on "Schedule"
+    stub_default_publishing_api_put_intent
```
benthorner: Hopefully we shouldn't need a stub here either :-).

spec/features/workflow/schedule_without_selection_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+
+RSpec.feature "Attempting to schedule without confirming a review status" do
+  scenario do
+    given_there_is_an_edition_in_draft
+    when_i_visit_the_summary_page
+    and_try_and_schedule_without_selecting_a_review_status
+    then_an_error_is_displayed
+  end
+
+  def given_there_is_an_edition_in_draft
```
benthorner: To be more accurate, could we go with something like `ready_to_schedule`?

spec/features/workflow/publish_without_selection_spec.rb
```diff
@@ -0,0 +1,31 @@
+# frozen_string_literal: true
+
+RSpec.feature "Attempting to publish without confirming a review status" do
```
emmabeynon: I would remove "Attempting to" here.  I'd also rename the file to be a bit clearer about what we're testing.  Something like `publish_without_selecting_review_status_spec.rb` perhaps?

spec/features/workflow/schedule_without_selection_spec.rb
```diff
@@ -0,0 +1,31 @@
+# frozen_string_literal: true
+
+RSpec.feature "Attempting to schedule without confirming a review status" do
```
emmabeynon: Similar comment, I would remove "Attempting to" and rename the file to be clearer.

spec/features/workflow/publish_without_selection_spec.rb
benthorner: @emmabeynon we had a long think about this yesterday, and were trying to find a naming pattern that could also be used for the `choose_document_spec`, which does exactly the same kind of check. Taking some inspiration from @1pretz1's choice of locale, how about `publish_status_not_selected_spec.rb`, which could work as `create_document_format_not_selected_spec.rb`?

emmabeynon: Sounds good, perhapa `publish_review_status_not_selected_spec.rb` just to be really clear?  `create_document_format_not_selected_spec.rb` works for me 👍

benthorner: Fine with me :-).

1pretz1: Yep i think Emma's last suggestion is a good file name to use, i'll make the changes! And i'll also open a separate pr for renaming the `choose_document_spec` to follow suite

spec/features/workflow/publish_without_selection_spec.rb
1pretz1: Yep nice spot!

spec/features/workflow/publish_without_selection_spec.rb
1pretz1: Yep you're correct!

app/controllers/schedule_controller.rb
1pretz1: Yep, i've gone with `review_status` for both scheduling and publishing controllers

config/locales/en/schedule/confirmation.yml
1pretz1: Yep didn't realise content had already been agreed upon 👍

spec/features/workflow/schedule_without_selection_spec.rb
1pretz1: Yep sure

spec/features/workflow/schedule_without_selection_spec.rb
1pretz1: Yep ⭐️

spec/features/workflow/schedule_without_selection_spec.rb
1pretz1: True true true

benthorner:
I've noticed the styling of the error message is different to e.g. requirements issues for images. We should try and make it so errors appear in the same way - could you look into it?

1pretz1:
@benthorner Yep thats because i was using the error alert component, where as image requirements uses the error summary component. 

Even though the error alert component is used in the application i've made the change as suggested to keep this alert consistent with the other 'requirement' type flashes, although it does certainly look uglier in the code :(

PR #921, opened 2019-03-28
Standardise alerting for not selecting an option

1pretz1:
https://trello.com/c/LhoqJDhC/733-revisit-the-radio-buttons-for-the-publish-confirmation-screen

Standardise alerting for not selecting an option

This updates the flash alerts to follow the alert pattern started here -
https://github.com/alphagov/content-publisher/pull/915.

We have identified a pattern of errors when a user doesn't select an
option and tries to continue to the next page. This pr makes the
alerts for this kind of action and the naming of the specs consistent within the app.


kevindew:
@benthorner do we have a plan for how this will be wrapped up with Peter away? not sure if I missed this at stand up or not. I can probably take it over though.

As for your comment I think we should drop the govspeak as you suggest, but I don't think we should go down the route of setting the items array as this would be pretty weird for a translation file. I can see you suggest that for simplifying the code but that would lead to translation file being used as a data structure rather than a key value lookup like it's normally used.

benthorner:
@kevindew yes, it was only an observation - the actual blocker is the `_govspeak` label.

benthorner:
I don't think @1pretz1 handed-over to anyone, but I can finish this off if you like?

kevindew:
@benthorner If you wouldn't mind that'd be really helpful 👍

benthorner:
@kevindew this should be ready now - I've added another commit to fix the same problem for the publish/schedule locales, where I didn't notice this in the original PR :-|.

kevindew:
> @kevindew this should be ready now - I've added another commit to fix the same problem for the publish/schedule locales, where I didn't notice this in the original PR :-|.
> 
> Regarding the locale approach, we could alternatively add another flash type, along the lines of `alert_with_item`, or maybe make it intelligently deal with a hash. What do you think?

Yeah a different approach sounds like a good idea. As we already have a bunch of [different alert sections on the layout](https://github.com/alphagov/content-publisher/blob/master/app/views/layouts/application.html.erb#L55-L90) I'd be keen to not add another, so I think if we were to convert non hashes to a hash in the view that'd be good (which is what I assume you mean by intelligently deal with the hash)

benthorner:
@kevindew all done - there is actually a precedent for using the YAML data structure to simplify setting the flash messages for `alert_with_description`, which is a technique I've copied here. It was very tempting to add the array data structure, but the fact that this makes the underlying translations inaccessible, which was immediately an issue in the tests, was enough to do it the code-y way instead.

PR #925, opened 2019-03-29
Avoid DateTime and multiple parsing of Date/Time

benthorner:
https://ci.integration.publishing.service.gov.uk/job/content-publisher/job/use-active-job/5/console

    Previously we did two rounds of parsing of the date and time when
    setting the schedule for publishing a document, as follows:
    
      parsed_date = Date.parse('2019-04-01')
      => '2019-04-01'
    
      parsed_time = Time.parse('09:00am')
      => '2019-03-29 09:00:00 +0000'
    
      DateTime.parse("#{parsed_date} #{parsed_time}").in_time_zone
      DateTime.parse('2019-04-01 2019-03-29 09:00:00 +0000').in_time_zone
      => '2019-04-01 10:00:00 +01:00'
    
    Double-parsing the time separately from the date caused us to inject a
    UTC override '+0000' into what would otherwise have been a BST date.
    This fixes the requirements checker to avoid double-parsing of the time.
    
    In order to help avoid future timezone issues, this also adds a cop to
    encourage the use of Time, instead of DateTime. The existance of both
    classes is apparently historical - Time functions equivalently with the
    added benefit that we have a RuboCop to ensure we use it with 'zone'.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -48,3 +48,6 @@ Rails/DynamicFindBy:
 
 Rails/SaveBang:
   Enabled: true
+
+Style/DateTime:
+  Enabled: true
--- a/app/services/requirements/scheduled_datetime_checker.rb
+++ b/app/services/requirements/scheduled_datetime_checker.rb
@@ -14,33 +14,24 @@ module Requirements
       @time = params[:time]
     end
 
-    def date_issues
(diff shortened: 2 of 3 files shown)


benthorner:
I'm sure we'll reap the fruits of this labour next year ;-).

PR #930, opened 2019-04-02
Image crop params

kevindew:
This fixes the warning about unpermitted params, and the longstanding fixme.

It also switches to round a crop number rather than floor it. See commits for further details.

### DIFF
--- a/app/controllers/images_controller.rb
+++ b/app/controllers/images_controller.rb
@@ -160,8 +160,10 @@ private
 
   def update_crop_params
     image_aspect_ratio = Image::HEIGHT.to_f / Image::WIDTH
-    crop_height = params[:crop_width].to_i * image_aspect_ratio
-    # FIXME: this will raise a warning because of unpermitted paramaters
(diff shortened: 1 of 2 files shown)


app/controllers/images_controller.rb
```diff
@@ -160,8 +160,10 @@ def update_params
 
   def update_crop_params
     image_aspect_ratio = Image::HEIGHT.to_f / Image::WIDTH
-    crop_height = params[:crop_width].to_i * image_aspect_ratio
-    # FIXME: this will raise a warning because of unpermitted paramaters
-    params.permit(:crop_x, :crop_y, :crop_width).merge(crop_height: crop_height.to_i)
+
+    params
+      .require(:image_crop)
```
benthorner: Just a minor thing, but could this just be `image`, as on the edit page - it's just a bit redundant to have 'crop' twice :-).

kevindew: Sure, I think image_revision would be consistent.

kevindew:
@benthorner ok updated

PR #931, opened 2019-04-02
Fix debug link

kevindew:
It's now infrequent for us to reference document in views so this debug
link vanished in most places.

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -114,9 +114,9 @@
       }
     end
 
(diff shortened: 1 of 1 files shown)


benthorner:
I wonder if we could just have this on the document summary - ideally outside/away from all the existing component - as I think we only need one way of getting to this page, rather than it being on every page.

kevindew:
Closing in favour of https://github.com/alphagov/content-publisher/pull/932

PR #935, opened 2019-04-03
Video embed

benthorner:
https://trello.com/c/4kQa7EZO/762-allow-users-embed-videos

This adds an embed video workflow, which is only accessible in the context of a modal. Please see the commits for further info. Future things to look into:

   - Reduce duplication of the render logic between the two modal modules
   - Make the modal appear at two thirds with for this workflow

### DIFF
--- a/app/controllers/video_embed_controller.rb
+++ b/app/controllers/video_embed_controller.rb
@@ -1,0 +1,21 @@
+# frozen_string_literal: true
+
+class VideoEmbedController < ApplicationController
+  def new
+    if rendering_context != "modal"
+      head :bad_request
+      return
+    end
+
+    render :new, layout: rendering_context
+  end
+
+  def create
+    if rendering_context != "modal"
+      head :bad_request
+      return
+    end
+
+    render inline: "[#{params[:title]}](#{params[:url]})"
+  end
+end
--- a/app/controllers/video_embed_controller.rb
+++ b/app/controllers/video_embed_controller.rb
@@ -1,6 +1,8 @@
 # frozen_string_literal: true
 
 class VideoEmbedController < ApplicationController
+  include ActionView::Helpers::SanitizeHelper
+
   def new
     if rendering_context != "modal"
       head :bad_request
@@ -16,6 +18,26 @@ class VideoEmbedController < ApplicationController
       return
     end
 
-    render inline: "[#{params[:title]}](#{params[:url]})"
+    title = strip_tags(params[:title])
+    url = strip_tags(params[:url])
+
+    issues = Requirements::VideoEmbedChecker.new
+      .pre_embed_issues(title: title, url: url)
+
+    if issues.any?
+      flash.now["alert_with_items"] = {
+        "title" => t("video_embed.new.flashes.requirements"),
+        "items" => issues.items,
+      }
+
+      render :new,
+        assigns: { issues: issues },
+        layout: rendering_context,
+        status: :unprocessable_entity
+
+      return
+    end
+
+    render inline: "[#{title}](#{url})"
   end
 end
--- a/app/services/requirements/video_embed_checker.rb
+++ b/app/services/requirements/video_embed_checker.rb
@@ -1,0 +1,34 @@
+# frozen_string_literal: true
+
+module Requirements
+  class VideoEmbedChecker
+    YOUTUBE_HOST = "youtube.com"
+    YOUTU_HOST = "youtu.be"
+
+    def pre_embed_issues(title: nil, url: nil)
+      issues = []
+
+      if title.blank?
+        issues << Issue.new(:video_embed_title, :blank)
+      end
+
+      if url.blank?
+        issues << Issue.new(:video_embed_url, :blank)
+      elsif !youtube_url?(url)
+        issues << Issue.new(:video_embed_url, :non_youtube)
+      end
+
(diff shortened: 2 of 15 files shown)


app/controllers/video_embed_controller.rb
```diff
@@ -0,0 +1,37 @@
+# frozen_string_literal: true
+
+class VideoEmbedController < ApplicationController
+  def new
+    if rendering_context != "modal"
+      raise "Video embed only supported in modal context"
```
kevindew: Do you think there are ways we can avoid a raise here and/or what this should return ideally?

I know we've got precedent for raising when edition isn't in an expected state, but I'm not sure if we have one yet for non-modal content and would be good, if possible, to avoid having too many internal server error responses that are for client error.

I'm thinking it'd probably be either a 400 bad request or 412 precondition failed. What do you think?

app/services/requirements/video_embed_checker.rb
```diff
@@ -0,0 +1,23 @@
...
+    def pre_embed_issues(title, url)
+      issues = []
+
+      if title.blank?
+        issues << Issue.new(:video_embed_title, :blank)
+      end
+
+      if url.blank?
+        issues << Issue.new(:video_embed_url, :blank)
+      elsif !url.start_with?(YOUTUBE_PREFIX)
```
kevindew: I see this matches the hint text which is actually a rather simplified approach to what is accepted as video input.

We accept both youtube.com and youtu.be urls. And youtube.com urls require a query string parameter of v. 

It doesn't seem too problematic if the JS is more forgiving than this check, so I think it's probably fine to drop youtu.be URLs (pending us getting feedback) but I think it should match in strictness to check the query string so that it will accept the same.

app/controllers/video_embed_controller.rb
benthorner: Sure - I think I'm still getting used to us moving away from exceptions by default. [This article](https://airbrake.io/blog/http-errors/400-bad-request) was also helpful to understand the decision making process.

As for the specific code, I think 412 is [reserved for specific headers](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/412), but there seems to be [some debate around whether 400 is appropriate](https://stackoverflow.com/questions/19671317/400-bad-request-http-error-code-meaning). 

On the whole, a [422 response code](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/422) seems the most accurate for this situation. What do you think?

app/services/requirements/video_embed_checker.rb
benthorner: You caught me before I had the chance to ask :-). This aspect was missing from the card, and the way the hint text is framed makes it ambiguous as a feature.

My suggestion is actually a bit different to yours: we should support youtu.be URLs as they're produced by YouTube, so it's reasonable a user might try to put one in, and the hint text should be changed to something more generic like 'be a valid YouTube URL'.

When it comes to validation, we could elongate the prefix with `?v=`, but I don't see the value in making the check more complex than simply checking one or more prefixes, as it's unlikely users will be typing the URL manually. How does that sound?

app/controllers/video_embed_controller.rb
kevindew: Ah shoot I should have read up more on precondition failed before suggesting - that probably explains why we never see it anywhere, despite the name feeling most appropriate.

So 422 is all wrapped in validations, particularly in Rails land, so I think we should avoid them - particularly as that'd make it non-distinct from issues. I think the closest precedent we already have in rails is with strong parameters which return a 400 response, for example if you do `params.require(:item)` and item isn't a parameter that would be 400.

app/services/requirements/video_embed_checker.rb
kevindew: I think the hint text is quite good in how explicit and succinct it is, so that could be tricky to tweak for generic.

I don't mind if we support youtu.be as well though.

Yeah I think it's ok to include the query string as part of the prefix check. I did have hesitations as, of course, the v could be anywhere in the query string - but YouTube seems to make a concerted effort to have that as the first param so seems unlikely to not be that.

app/controllers/video_embed_controller.rb
benthorner: Ah, params makes a good precedent - updated.

app/services/requirements/video_embed_checker.rb
benthorner: OK. Had a chat with @benhazell and he agrees the content is fine as-is. I've come round to making the check more complex :-).

benthorner:
@kevindew this should be ready for another look now.

app/services/requirements/video_embed_checker.rb
```diff
@@ -0,0 +1,34 @@
+# frozen_string_literal: true
+
+module Requirements
+  class VideoEmbedChecker
+    YOUTUBE_HOST = "youtube.com"
+    YOUTU_HOST = "youtu.be"
+
+    def pre_embed_issues(title, url)
```
kevindew: Just a minor thing, but I was a bit confused reading the tests and wondered if these would be better as keyword arguments?

benthorner: Sure - I was on the fence at the time. How does it look now?

PR #936, opened 2019-04-03
This fixes a missing newline after the update type

benthorner:

GOV.UK Notify was also styling one of the lines strangely, which seems
to be related to the indentation from ERB. This removes all node
indentationt to make the output more predictable.

### DIFF
--- a/app/views/publish_mailer/publish_email.text.erb
+++ b/app/views/publish_mailer/publish_email.text.erb
@@ -1,17 +1,17 @@
 <%= @edition.document_type.label %> ‘<%= @edition.title %>’
 
 <% if @edition.number > 1 -%>
-  <%= t("publish_mailer.publish_email.details.update",
-    time: @status.created_at.strftime("%l:%M%P").strip,
-    date: @status.created_at.strftime("%d %B %Y"),
-    user: @status.created_by.name,
-  ) %>
+<%= t("publish_mailer.publish_email.details.update",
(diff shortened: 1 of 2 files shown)


app/views/publish_mailer/publish_email.text.erb
```diff
@@ -1,17 +1,17 @@
...
-  <%= t("publish_mailer.publish_email.details.update",
-    time: @status.created_at.strftime("%l:%M%P").strip,
-    date: @status.created_at.strftime("%d %B %Y"),
-    user: @status.created_by.name,
-  ) %>
+<%= t("publish_mailer.publish_email.details.update",
+  time: @status.created_at.strftime("%l:%M%P").strip,
+  date: @status.created_at.strftime("%d %B %Y"),
+  user: @status.created_by.name,
+) %>
```
erino: Would using `-%>` have worked? I know it strips whitespace, but I can't remember where it strips it from

benthorner: So by default ERB adds a \r\n after each ruby statement, which is removed by the '-' syntax on the end of the tag. Although for non-printing tags you can also use '-' on the opening tag to remove indentation, there doesn't seem to be an equivalent for a print '<%=' tag :-(.

erino: 👍

PR #941, opened 2019-04-04
Show markdown component errors on edit document page

kevindew:
Trello: https://trello.com/c/SdMUjUmj/722-resolve-xss-issues-raised-by-pen-testers

This PR does a few things:

- It fixes problems that was causing the markdown to not render in component guide, this whole component is a bit dubious though. See commit for more info.
- Adds error_items support to the markdown component
- Wires errors into markdown component on edit document page

Screenshot:

screencapture-localhost-3221-documents-49ea88bd-1e07-4d65-bf08-00b361f1efe9-en-2019-04-04-16_56_30


kevindew:
> Looks good :-). I agree this is hardly a component, and I remember we discussed one possibility of separating the editor and preview aspects through the tabs component. Do you think this is something for the Wall of Pain?

Yup, sounds good for the wall of pain. Is likely only going to get worse.

PR #945, opened 2019-04-05
Show images wide, video embed narrow

benthorner:
https://trello.com/c/4kQa7EZO/762-allow-users-embed-videos


1pretz1:
Sorry I may be behind as i've been away for a week, but what was the thought around altering the size of the modal depending on whether it's a video or image embed?

benthorner:
So this is mostly an aesthetic thing, which is documented as part of the card e.g. https://trello.com/c/4kQa7EZO/762-allow-users-embed-videos.

There is a functional element to it. In the case of images, we want it to be wider to accommodate cropping, whereas for video it's not ideal, as it means we have really long input fields.

PR #948, opened 2019-04-08
Add ADR for background jobs and notifications

benthorner:
https://trello.com/c/xraOMbHg/757-write-an-adr-documenting-tech-product-decisions-around-notifications

This adds an ADR for sending notifications, which is coupled to the way
we handle background jobs, so it seems sensible to discuss them both in
the same document.

### DIFF
--- a/docs/adr/000x-sending-email-notifications.md
+++ b/docs/adr/000x-sending-email-notifications.md
@@ -1,0 +1,103 @@
+# X. Sending email notifiations
+
+[notify]: https://www.notifications.service.gov.uk/
+[notify-features]: https://www.notifications.service.gov.uk/features
+[notify-allowed-list]: https://docs.notifications.service.gov.uk/ruby.html#team-and-whitelist
+[notify-gem]: https://docs.notifications.service.gov.uk/ruby.html#ruby-client-documentation
+[notify-gem-send-email]: https://docs.notifications.service.gov.uk/ruby.html#send-an-email
+[action-mailer]: https://guides.rubyonrails.org/action_mailer_basics.html
+[action-mailer-interceptor]: https://guides.rubyonrails.org/action_mailer_basics.html#intercepting-emails
+[active-job]: https://edgeguides.rubyonrails.org/active_job_basics.html
+[active-job-retry]: https://edgeguides.rubyonrails.org/active_job_basics.html#retrying-or-discarding-failed-jobs
+[active-job-sidekiq]: https://github.com/mperham/sidekiq/wiki/Active-Job
+[active-job-sidekiq-retry]: https://github.com/mperham/sidekiq/wiki/Active-Job#customizing-error-handling
+[active-job-improvements]: https://github.com/rails/rails/blob/5-1-stable/activejob/CHANGELOG.md
+[sidekiq]: https://github.com/mperham/sidekiq
+[sidekiq-govuk]: https://github.com/alphagov/govuk_sidekiq
+[sidekiq-grafana]: https://grafana.production.govuk.digital/dashboard/file/sidekiq.json?refresh=1m&orgId=1
+[sidekiq-sentry]: https://docs.sentry.io/clients/ruby/#reporting-failures
+[raven]: https://github.com/getsentry/raven-ruby
+[raven-active-job]: https://github.com/getsentry/raven-ruby/blob/master/lib/raven/integrations/rails/active_job.rb
+[consequences-sidekiq-worker]: https://github.com/alphagov/content-publisher/commit/88e27840f7a7f812d83c5878503cc6aece01127b
+[consequences-sidekiq-abort]: https://github.com/alphagov/content-publisher/commit/ce13ccd34e967e091526adb20938b815faf5a912
+
+Date: 2019-XX-XX
+
+## Context
+
+In order to send email notifications about user actions, such as publishing a document, we need to:
+
+   - Work out who should receive the email
+   - Generate the content for the email (subject and body)
+   - Send the email to the recipient(s)
+
+We don't anticipate sending email to be a critical step of any user action, but it is still a step that is prone to error and delay, due to interaction with external systems. We should ensure sending emails is robust against transient failure and that emails are only sent when we are confident they are accurate i.e. after DB updates and critical API calls.
+
+We should have a means of sending notifications in non-production environments, which allows us to see what notifications would be sent in production, but prevents the emails from being sent to the intended recipient(s). It should also be possible to see the notifications we have sent in production in order to debug issues if users report they are not receiving notifications, or that the content of the notifications is incorrect.
+
+## Decision
+
+We will use [GOV.UK Notify][notify] in combination with [Action Mailer][action-mailer], [Active Job][active-job] and [Sidekiq][sidekiq].
+
+### GOV.UK Notify
+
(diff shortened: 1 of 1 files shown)


docs/adr/000x-jobs-and-notifications.md
```diff
@@ -0,0 +1,110 @@
...
+Notify will reject an email when a whitelist is enabled and the recipient is not in the list. In order to retain end-to-end email functionality but avoid emailing users outside of production, we will use an [Action Mailer Interceptor][action-mailer-interceptor] to redirect emails to a Google Group in integration and staging. The Google Group(s) will act as a dumping ground for viewing the emails that have been generated, which enables us to check they appear correctly in text and HTML forms.
+
+### Active Job
+
+We will use [Active Job][active-job] in combination with [Action Mailer][action-mailer] to send emails asynchronously.
+
+   - This means any delay in calling the Notify APIs does not impact on the user request
+   - We can use the [retry feature][active-job-retry] of Active Job to cope with transient failures from Notify
+
+Using Active Job also allows us to switch to a different queueing provider in future. Initially, we will use [Sidekiq][sidekiq], which is one of the [in-built queueing adapters][active-job-sidekiq] provided with Active Job.
```
kevindew: It'd be worth stating here why we're using Active Job instead of Sidekiq (which I believe is because of the Action Mailer interface and improvements in Active Job retry mechanisms). Since GOV.UK has a convention of Sidekiq over Active Job this is worth justifying more as it's likely a WTF.

docs/adr/000x-jobs-and-notifications.md
```diff
@@ -0,0 +1,110 @@
...
+
+In order to avoid [duplicate retry behaviour][active-job-sidekiq-retry], we will disable Sidekiq retries and rely solely on the retry behaviour of Active Job. When a failure occurs, the job is marked as 'processed' by Sidekiq and any retry manifests as a newly enqueued job; when the retries are exhausted, the exception will propagate to Sidekiq and the job will be marked as 'dead'. Normally, the job would be marked as 'failed' by Sidekiq and requeued; this means we lose visibility on the number of retries.
+
+We already use [Raven] to report errors to Sentry, which automatically [integrates with Sidekiq][sidekiq-sentry]. Note that Raven also [integrates with Active Job][raven-active-job], but this is disabled when Sidekiq is present. Active Job catches exceptions as part of its retry behaviour, so an error will only get reported to Sentry when the exception is not handled by Active Job, or the retries for an exception we do handle are exhausted. We already [export Sidekiq metrics to Grafana][sidekiq-grafana].
+
+Using Active Job also simplifies the development of background jobs, as only knowledge of Active Job is required.
```
kevindew: This feels a bit out of place in this section.

docs/adr/000x-jobs-and-notifications.md
```diff
@@ -0,0 +1,110 @@
...
+
+### Redis
+
+We currently use a non-resilient Redis machine to distribute jobs between concurrent Sidekiq workers. Although we [may improve the resilience][sidekiq-pro] in future, we should assume enqueueing jobs is prone to failure.
```
kevindew: There seems to be a bit of a conflict here with what's being described. Sidekiq pro wouldn't make redis machines resilient, just would change the interface of plucking jobs.

I think in this section we've probably got 3 concerns:

1. We're relying on GOV.UK infrastructure for Redis resilience and the application assumes it is operable, in a similar manner to our interactions with PostgreSQL.
2. Email notifications don't have a data source aside from Redis and we're aware the method Sidekiq uses to pull the jobs is not resilient to certain types of failure, we think that's ok - Sidekiq pro could resolve this.

docs/adr/000x-jobs-and-notifications.md
```diff
@@ -0,0 +1,110 @@
...
+
+In order to manage this, we will ensure the code to send notifications is placed outside of any DB transactions. Since it's unclear what the exact failure modes are for Redis/Sidekiq, the response to the user may still be impacted.
```
kevindew: I'd suggest this is only a side-effect of the code being outside the transaction rather than being part of any design for resiliency - since the code expects Redis to be up.

docs/adr/000x-jobs-and-notifications.md
```diff
@@ -0,0 +1,110 @@
...
+
+## Status
+
+Accepted
+
+## Consequences
+
+Previously we used Sidekiq to implement a [background worker for scheduling][consequences-sidekiq-worker]. We will rewrite this to use the Active Job framework. Since Active Job does not retry jobs, we no longer need to [add middleware to facilitate aborting the job][consequences-sidekiq-abort].
```
kevindew: "does not retry jobs" => "does not automatically retry jobs"

docs/adr/000x-jobs-and-notifications.md
```diff
@@ -0,0 +1,110 @@
...
+
+Using the retry feature of Active Job instead of Sidekiq means we will lose visibility of the number of failures in the Sidekiq dashboard and in the metrics exported to Grafana. However, we will retain visibility of jobs that have totally failed.
```
kevindew: Do you mean number of retries rather than number of failures?

Data is exported to Graphite rather than Grafana.

docs/adr/000x-jobs-and-notifications.md
```diff
@@ -0,0 +1,110 @@
+# X. Jobs and Notifications
```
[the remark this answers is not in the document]
benthorner: Fair enough - it felt like a mixture, but you're right that notifications is the primary concern.

docs/adr/000x-jobs-and-notifications.md
```diff
@@ -0,0 +1,110 @@
...
+
+## Context
+
+In order to send email notifications about user actions, such as publishing a document, we need to:
+
+   - Work out who should receive the email
+   - Generate the content for the email (subject and body)
+   - Send the email to the recipient(s)
+
+The last step is non-critical but prone to error and delay. We should ensure sending emails is robust against transient failure and that emails are only sent when we are confident they are accurate i.e. after DB updates and critical API calls.
```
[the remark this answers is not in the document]
benthorner: I see what you mean - reworded :-).

docs/adr/000x-jobs-and-notifications.md
```diff
@@ -0,0 +1,110 @@
...
+We will use [GOV.UK Notify][notify] in combination with [Action Mailer][action-mailer], [Active Job][active-job] and [Sidekiq][sidekiq].
+
+### GOV.UK Notify
+
+We will use [GOV.UK Notify][notify] to handle the low-level concerns of sending emails:
+
+   - This avoids having to setup our own mechanism for sending email, including logic to handle low-level errors and retries, which are [handled by Notify by default][notify-features].
+   - Notify provide a [Ruby gem][notify-gem] that makes it easy to integrate the service into our app, without specific knowledge of the underlying APIs
+   - We can use the [Notify dashboard][notify-features] to see emails we have sent, logs for individual emails, as well as aggregate stats over all emails
+   - Notify provides the facility to restrict the recipients of emails to a [team or whitelist][notify-whitelist], which we can use to restrict notifications outside of production.
```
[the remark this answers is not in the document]
benthorner: 👍

benthorner:
@kevindew I've made changes for all your comments, which should hopefully resolve most of them - should be good for another review now :-).

PR #952, opened 2019-04-09
Link Content Data page

emmabeynon:
For https://trello.com/c/y1CtmFT9/763-link-from-our-app-to-content-data-page

This adds a link to the data page in the Content Data app for a particular piece of content.  The conditions under which we display a link are:
- The user must be part of an organisation that is a Content Data beta partner (otherwise they won't be able to access the page), AND
- Content was published before yesterday, OR
- Content was published yesterday AND it is after 9am (the Content Data ETL runs at 7am and finishes around 7.30am.  I've agreed with Sean in Data Informed Content that 9am gives enough buffer time in case the ETL is delayed).

As part of this PR we have also changed the 'View published content on GOV.UK' link text to 'View on GOV.UK'.  The link to Content Data appears below this, and in order to stop the 'View published content...' link from wrapping we've made it shorter.

Screen Shot 2019-04-09 at 15 43 33

### DIFF
--- a/app/services/content_data_url.rb
+++ b/app/services/content_data_url.rb
@@ -18,12 +18,35 @@ class ContentDataUrl
   end
 
   def displayable?(user)
-    has_content_data_access?(user)
+    has_content_data_access?(user) && expected_in_content_data?
   end
 
 private
 
   def has_content_data_access?(user)
     CONTENT_DATA_BETA_PARTNERS.include?(user.organisation_content_id)
+  end
+
+  def expected_in_content_data?
+    # Content Data ETL runs around 7am each day and pulls in data from the
+    # previous day. If content was first published yesterday we only want to
+    # show a link to the content data page after the ETL has run otherwise users
+    # will get a 404 page. Data Informed Content are looking at changing this so
+    # this check should only be temporary.
+    first_published_before_yesterday? ||
+      (first_published_yesterday? && content_data_etl_end_time_elapsed?)
+  end
+
+  def first_published_yesterday?
+    document.first_published_at.to_date == Date.yesterday
+  end
+
+  def first_published_before_yesterday?
+    document.first_published_at.to_date < Date.yesterday
+  end
+
+  def content_data_etl_end_time_elapsed?
+    # We assume that the Content Data ETL will have finished before 9am.
+    Time.current.hour >= 9
   end
 end
--- a/app/services/content_data_url.rb
+++ b/app/services/content_data_url.rb
@@ -1,0 +1,14 @@
+# frozen_string_literal: true
+
+class ContentDataUrl
+  attr_reader :document
+
+  def initialize(document)
+    @document = document
+  end
+
+  def url
+    content_data_root = Plek.new.external_url_for("content-data-admin")
+    content_data_root + "/metrics" + document.live_edition.base_path
+  end
+end
--- a/app/services/content_data_url.rb
+++ b/app/services/content_data_url.rb
@@ -1,6 +1,11 @@
 # frozen_string_literal: true
 
 class ContentDataUrl
+  # department-for-work-pensions, department-of-health-and-social-care, government-digital-service
+  CONTENT_DATA_BETA_PARTNERS = %w(b548a09f-8b35-4104-89f4-f1a40bf3136d
+                                  7cd6bf12-bbe9-4118-8523-f927b0442156
+                                  af07d5a5-df63-4ddc-9383-6a666845ebe9).freeze
+
   attr_reader :document
 
   def initialize(document)
@@ -10,5 +15,15 @@ class ContentDataUrl
   def url
     content_data_root = Plek.new.external_url_for("content-data-admin")
     content_data_root + "/metrics" + document.live_edition.base_path
+  end
+
+  def displayable?(user)
+    has_content_data_access?(user)
+  end
+
+private
+
+  def has_content_data_access?(user)
+    CONTENT_DATA_BETA_PARTNERS.include?(user.organisation_content_id)
   end
 end
--- a/app/views/documents/show/_actions.html.erb
+++ b/app/views/documents/show/_actions.html.erb
@@ -88,9 +88,12 @@
         <li>
           <%= link_to "View published edition on GOV.UK", EditionUrl.new(@edition.document.live_edition).public_url, class: "govuk-link govuk-link--no-visited-state" %>
         </li>
-        <li>
-          <%= link_to "View data about this page", ContentDataUrl.new(@edition.document).url, target: "_blank", class: "govuk-link govuk-link--no-visited-state" %>
-        </li>
+        <% content_data_url = ContentDataUrl.new(@edition.document) %>
+        <% if content_data_url.displayable?(current_user) %>
+          <li>
+            <%= link_to "View data about this page", content_data_url.url, target: "_blank", class: "govuk-link govuk-link--no-visited-state" %>
+          </li>
+        <% end %>
       </ul>
     <% end %>
   </div>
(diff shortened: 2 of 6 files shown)


app/views/documents/show/_actions.html.erb
```diff
@@ -84,7 +84,17 @@
 
     <% if @edition.document.live_edition %>
       <hr class="govuk-section-break govuk-section-break--m govuk-section-break--visible" />
-      <%= link_to "View published edition on GOV.UK", EditionUrl.new(@edition.document.live_edition).public_url, class: "govuk-link govuk-link--no-visited-state" %>
+      <ul class="govuk-list">
```
alex-ju: We can add a `govuk-!-margin-0` class to this list if we want to remove the bottom margin.

spec/features/content_data/content_data_non_beta_partner_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+  end
+
+  def when_i_visit_the_summary_page
+    visit document_path(@edition.document)
+  end
+
+  def then_i_dont_see_a_link_to_the_content_data_page_for_the_document
+    expect(page).not_to have_link("View data about this page")
+  end
+end
```
kevindew: Since the logic is encapsulated in the ContentDataUrl class I think we probably don't need this test and the recently published one (sorry - I hope they didn't take long to write)

spec/services/content_data_url_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+    it "returns true if the edition was published yesterday and it is currently 9am" do
+      edition.status.update!(created_at: Time.current.yesterday)
+      nine_am_today = Time.current.change(hour: 9)
+
+      travel_to(nine_am_today) do
+        expect(ContentDataUrl.new(edition).displayable?(beta_partner)).to be true
+      end
+    end
+  end
+end
```
kevindew: These tests look good 👍

app/views/documents/show/_actions.html.erb
```diff
@@ -84,7 +84,17 @@
...
+        <li>
+          <%= link_to "View on GOV.UK", EditionUrl.new(@edition.document.live_edition).public_url, class: "govuk-link govuk-link--no-visited-state" %>
+        </li>
+        <% content_data_url = ContentDataUrl.new(@edition) %>
```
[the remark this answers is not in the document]
emmabeynon: It makes sense to me that users would want to be able to access the data while subsequent editions are worked on, so I'll change this to use document instead.

app/services/content_data_url.rb
```diff
@@ -0,0 +1,49 @@
+# frozen_string_literal: true
+
+class ContentDataUrl
+  CONTENT_DATA_BETA_PARTNERS = %w(department-for-work-pensions
+                                  department-of-health-and-social-care
+                                  government-digital-service).freeze
```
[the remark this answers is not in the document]
emmabeynon: Sure, that seems reasonable.  I'll add comments for each one so that it's more readable.

emmabeynon:
> Do you know whats visible in Content Data in the time from a page being published and before they've got data? @kevindew At the moment the page will 404.  Sean mentioned that they want to look into providing a better solution than that, but that's the current situation.

app/views/documents/show/_actions.html.erb
emmabeynon: Done 👍

app/services/content_data_url.rb
```diff
@@ -0,0 +1,49 @@
...
+
+  attr_reader :edition
+
+  def initialize(edition)
+    @edition = edition
+  end
+
+  def data_page
```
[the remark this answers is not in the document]
emmabeynon: Done 👍

app/views/documents/show/_actions.html.erb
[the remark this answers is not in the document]
emmabeynon: Now using document instead of edition.

app/services/content_data_url.rb
```diff
@@ -0,0 +1,49 @@
...
+    content_data_etl_end_time_elapsed?
+  end
+
+private
+
+  def has_content_data_access?(user)
+    CONTENT_DATA_BETA_PARTNERS.include?(user.organisation_slug)
+  end
+
+  def published_today?
```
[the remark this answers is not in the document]
emmabeynon: We're no longer using this method.

app/services/content_data_url.rb
```diff
@@ -0,0 +1,49 @@
...
+    content_data_root + "/metrics" + edition.base_path
+  end
+
+  def displayable?(user)
+    return false unless data_page
+    return false unless has_content_data_access?(user)
+    return false if published_today?
+    return true unless published_yesterday?
+
+    content_data_etl_end_time_elapsed?
```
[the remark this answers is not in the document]
emmabeynon: I've rewritten these checks and ditched checking explicitly if the content was published today.  Hopefully this is a bit easier to read and understand now 😸

spec/features/content_data/content_data_non_beta_partner_spec.rb
emmabeynon: I've deleted both of these files.

app/services/content_data_url.rb
```diff
@@ -0,0 +1,46 @@
...
+
+  attr_reader :document
+
+  def initialize(document)
+    @document = document
+  end
+
+  def url
+    content_data_root = Plek.new.external_url_for("content-data-admin")
+    content_data_root + "/metrics" + document.current_edition.base_path
```
kevindew: I think this should be live_edition?

app/services/content_data_url.rb
```diff
@@ -0,0 +1,46 @@
...
+  def published_yesterday?
+    document.first_published_at.to_date == Date.yesterday
+  end
+
+  def published_before_yesterday?
+    document.first_published_at.to_date < Date.yesterday
+  end
+
+  def content_data_etl_end_time_elapsed?
+    Time.current >= Time.current.change(hour: 9)
```
kevindew: Minor, but this could be `Time.current.hour >= 9`

I think it'd be useful to have a small comment that explains the oddness of this 9am and the 404 as I think it's a high enough WTF that you'd want it documenting beyond a commit message. Also worth stating we expect them to fix this so it can be removed.

spec/factories/edition_factory.rb
```diff
@@ -102,6 +106,10 @@
           revision_at_creation: edition.revision,
           withdrawal: evaluator.withdrawal,
         )
+        unless edition.document.first_published_at
+          edition.document.first_published_at = Time.current.yesterday
+          edition.document.save!
+        end
```
kevindew: I'm keen to see if we can do the factories in a different way to avoid the use of a save method (as this will mean that if we build an edition it will persist a document but none of the other models). It also wouldn't work correctly if the status isn't persisted.

I had a quick go myself at trying to put something together that's consistent with our other transient attributes to see if we could get around this requirement: https://github.com/alphagov/content-publisher/compare/first_published_at-factory?expand=1

app/services/content_data_url.rb
emmabeynon: Yep, changed.

app/services/content_data_url.rb
emmabeynon: Added a comment 👍

spec/factories/edition_factory.rb
emmabeynon: This is great - I had tried a similar approach but missed out adding `first_published_at` in the transient block, so it didn't work.  Updated 👍

emmabeynon:
@kevindew Your comments should be addressed now, I'm checking in with Tobit on the pre-release feature flag but obviously will be easy to remove if needed.

PR #954, opened 2019-04-10
Add paste html to govspeak functionality

emmabeynon:
For https://trello.com/c/3JX1Pc9f/770-embed-paste-html-to-govspeak-functionality

This embeds the [paste-html-to-govspeak](https://www.npmjs.com/package/paste-html-to-govspeak) npm package in the body field on the edit document page.  We put this behind a pre-release feature flag for now; note that when we set this as a data attribute on the text area component the boolean value comes through as a string, which is why we check if this is equal to 'true' before adding a paste event listener.

### DIFF
--- a/app/views/components/_markdown_editor.html.erb
+++ b/app/views/components/_markdown_editor.html.erb
@@ -6,11 +6,15 @@
   govspeak_path ||= nil
   edition ||= nil
   error_items ||= []
+  paste_html_to_govspeak ||= false
 
   component_classes = %w[app-c-markdown-editor govuk-form-group]
   component_classes << "govuk-form-group--error" if error_items.any?
 
   insert_items = []
+
+  textarea[:data] = textarea.fetch(:data, {})
+                            .merge({ "paste-html-to-govspeak": paste_html_to_govspeak })
 
   if edition
     insert_items = [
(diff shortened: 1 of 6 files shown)


kevindew:
@emmabeynon this is looking good, I've got sidetracked reviewing it because I looked into the deployment fail which reveals NPM problems and I'm falling into a rabbit hole.

app/views/components/_markdown_editor.html.erb
```diff
@@ -6,12 +6,19 @@
...
   component_classes = %w[app-c-markdown-editor govuk-form-group]
   component_classes << "govuk-form-group--error" if error_items.any?
 
   insert_items = []
 
+  if textarea[:data]
+    textarea[:data]["paste-html-to-govspeak"] = paste_html_to_govspeak
+  else
+    textarea[:data] = { "paste-html-to-govspeak": paste_html_to_govspeak }
+  end
```
kevindew: You could make this a bit more succinct with something like:

`textarea[:data] = textarea.fetch(:data, {}).merge({ "paste-html-to-govspeak": paste_html_to_govspeak })` and avoid the conditional.

emmabeynon: That looks much better - now updated 👍

kevindew:
@emmabeynon v0.2.0 should be out now

emmabeynon:
@kevindew Now updated to v0.2.0 👍

PR #957, opened 2019-04-11
Use autocomplete as dropdown for time input

alex-ju:
This PR updates the autocomplete to support a custom configuration ([without-narrowing-results](https://content-publisher-revie-pr-957.herokuapp.com/component-guide/autocomplete/autocomplete_without_narrowing_results/preview)) in which essentially acts as a dropdown list which also allows custom input from users.

This custom configuration is used in [time select in the scheduling form](https://content-publisher-revie-pr-957.herokuapp.com/documents/d7e73a7d-17b2-4488-b41c-c6b3b5758dd3:en) as shown below.

### Before
Screen Shot 2019-04-11 at 17 39 26

### After
Screen Shot 2019-04-11 at 17 38 37

[Trello card](https://trello.com/c/QrEdhtTO)

### DIFF
--- a/app/assets/javascripts/components/autocomplete.js
+++ b/app/assets/javascripts/components/autocomplete.js
@@ -11,6 +11,8 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
 
     if (type === 'with-hint-on-options') {
       this.initAutoCompleteWithHintOnOptions()
+    } else if (type === 'without-narrowing-results') {
+      this.initAutoCompleteWithoutNarrowingResults()
     } else {
       this.initAutoComplete()
     }
@@ -67,6 +69,54 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
         suggestion: function (result) {
           return result && result.text + '<span class="autocomplete__option-hint">' + result.hint + '</span>'
         }
+      },
+      onConfirm: function (result) {
+        var value = result && result.value
+        var options = [].filter.call($select.options, function (option) {
+          return option.value === value
+        })
+
+        if (options.length) {
+          options[0].selected = true
+        }
+      }
+    })
+
+    $select.style.display = 'none'
+    $select.id = $select.id + '-select'
+  }
+
+  Autocomplete.prototype.initAutoCompleteWithoutNarrowingResults = function () {
+    // Read options and associated data attributes and feed that as results for inputValueTemplate
+    var $select = this.$module.querySelector('select')
+
+    if (!$select) {
+      return
+    }
+
+    var $options = $select.querySelectorAll('option')
+    var $selectedOption = $select.querySelector('option[selected]')
+    if (!$selectedOption) {
+      $selectedOption = $options[0]
+    }
+    var defaultValue = $selectedOption.textContent
+
+    new window.accessibleAutocomplete({ // eslint-disable-line no-new, new-cap
+      id: $select.id,
+      name: $select.name,
+      element: this.$module,
+      showAllValues: true,
+      defaultValue: defaultValue,
+      autoselect: false,
+      dropdownArrow: function (config) {
+        return '<svg class="' + config.className + '" style="top: 8px;" viewBox="0 0 512 512" ><path d="M256,298.3L256,298.3L256,298.3l174.2-167.2c4.3-4.2,11.4-4.1,15.8,0.2l30.6,29.9c4.4,4.3,4.5,11.3,0.2,15.5L264.1,380.9  c-2.2,2.2-5.2,3.2-8.1,3c-3,0.1-5.9-0.9-8.1-3L35.2,176.7c-4.3-4.2-4.2-11.2,0.2-15.5L66,131.3c4.4-4.3,11.5-4.4,15.8-0.2L256,298.3  z"/></svg>'
+      },
+      source: function (query, syncResults) {
+        var results = []
+        $options.forEach(function ($el) {
+          results.push($el.textContent)
+        })
+        syncResults(results)
       },
       onConfirm: function (result) {
         var value = result && result.value
--- a/app/assets/javascripts/components/autocomplete.js
+++ b/app/assets/javascripts/components/autocomplete.js
@@ -120,7 +120,7 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
       },
       onConfirm: function (result) {
         var value = result && result.value
-        var options = [].filter.call($select.options, function (option) {
+        var options = [].filter.call($options, function (option) {
           return option.value === value
         })
 
@@ -130,8 +130,7 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
       }
     })
 
-    $select.style.display = 'none'
-    $select.id = $select.id + '-select'
+    $select.parentNode.removeChild($select)
   }
 
   Modules.Autocomplete = Autocomplete
--- a/app/views/components/docs/autocomplete.yml
+++ b/app/views/components/docs/autocomplete.yml
@@ -105,3 +105,21 @@ examples:
           - United Kingdom
           - uk
           - data-hint: London
+  autocomplete_without_narrowing_results:
+    data:
+      id: autocomplete-without-narrowing-results
+      name: autocomplete-without-narrowing-results
+      type: without-narrowing-results
+      label:
+        text: Select your country
+      data_module: autocomplete
+      options:
+        -
+          - France
+          - fr
+        -
+          - United Arab Emirates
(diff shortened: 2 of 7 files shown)


kevindew:
The code here looks good to me.

I did wonder if we definitely wanted to remove the non narrowing behaviour, this isn't something that's part of data lists for instance? 

Do we intend to have just a single field for time or for different hours and minutes fields?

alex-ju:
> I did wonder if we definitely wanted to remove the non narrowing behaviour, this isn't something that's part of data lists for instance?

We did some usability tests with this feature both on and off and there is a usability concern if we preserve the original behaviour of the autocomplete. By default we set a value for the time (currently 9:00AM) and when a value is set, clicking on the filed or dropdown icons will not reveal any results (as it narrows down to only the selected one). This way the user would need to remove the date and refill the input or select another option which increases the interaction steps and potentially makes it more confusing for the users.

> Do we intend to have just a single field for time or for different hours and minutes fields?

As per design, the time will be only one field as it makes it easier for users to select/fill in a single field compared to 2 fields and a dropdown, or 3 dropdowns, which will also overload the UI.

soniaturcotte:
Where can I review this ?

alex-ju:
@soniaturcotte: I've fixed the review deploy. You can now review the component [in solitude](https://content-publisher-revie-pr-957.herokuapp.com/component-guide/autocomplete/autocomplete_without_narrowing_results/preview) or [in context](https://content-publisher-revie-pr-957.herokuapp.com/documents/d7e73a7d-17b2-4488-b41c-c6b3b5758dd3:en).

app/views/components/docs/autocomplete.yml
```diff
@@ -105,3 +105,21 @@ examples:
           - United Kingdom
           - uk
           - data-hint: London
+  autocomplete_without_narrowing_results:
+    data:
+      id: autocomplete-withour-narrowing-results
+      name: autocomplete-withour-narrowing-results
```
kevindew: s/withour/without/

app/assets/javascripts/components/autocomplete.js
```diff
@@ -84,5 +86,53 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
...
+    if (!$select) {
+      return
+    }
+
+    var $options = $select.querySelectorAll('option')
+    var $selectedOption = $select.querySelector('option[selected]')
+    if (!$selectedOption) {
+      $selectedOption = $options[0]
+    }
+    var defaultValue = $selectedOption.textContent
```
kevindew: As we're hoping they can type in their own input should this be dealing with an input field rather than a select box? 

I realise this complicates matters slightly as the option is just "without narrowing results" which doesn't make any implications about input or select.

app/assets/javascripts/components/autocomplete.js
```diff
@@ -84,5 +86,53 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
...
+    new window.accessibleAutocomplete({ // eslint-disable-line no-new, new-cap
+      id: $select.id,
+      name: $select.name,
+      element: this.$module,
+      showAllValues: true,
+      defaultValue: defaultValue,
+      autoselect: false,
+      dropdownArrow: function (config) {
+        return '<svg class="' + config.className + '" style="top: 8px;" viewBox="0 0 512 512" ><path d="M256,298.3L256,298.3L256,298.3l174.2-167.2c4.3-4.2,11.4-4.1,15.8,0.2l30.6,29.9c4.4,4.3,4.5,11.3,0.2,15.5L264.1,380.9  c-2.2,2.2-5.2,3.2-8.1,3c-3,0.1-5.9-0.9-8.1-3L35.2,176.7c-4.3-4.2-4.2-11.2,0.2-15.5L66,131.3c4.4-4.3,11.5-4.4,15.8-0.2L256,298.3  z"/></svg>'
+      },
```
kevindew: Is there anyway we can think about moving this stuff into the library itself? I imagine it already has some styling and it feels like we're basically providing a default here.

kevindew:
> > I did wonder if we definitely wanted to remove the non narrowing behaviour, this isn't something that's part of data lists for instance?
> 
> We did some usability tests with this feature both on and off and there is a usability concern if we preserve the original behaviour of the autocomplete. By default we set a value for the time (currently 9:00AM) and when a value is set, clicking on the filed or dropdown icons will not reveal any results (as it narrows down to only the selected one). This way the user would need to remove the date and refill the input or select another option which increases the interaction steps and potentially makes it more confusing for the users.

Ah darn I can see that initial step being a pain, I also think the way we've got it now is a bit of a pain as if you're typing your own value and it's not in the options it looks a bit like it's invalid 🤔 

Maybe there's a middle option that gives you a dropdown if you've not changed the content of the input since page load - that could be too complex though.

> > Do we intend to have just a single field for time or for different hours and minutes fields?
> 
> As per design, the time will be only one field as it makes it easier for users to select/fill in a single field compared to 2 fields and a dropdown, or 3 dropdowns, which will also overload the UI.

Sure sounds good - lets never consider seconds as part of scheduling :wink:

app/assets/javascripts/components/autocomplete.js
alex-ju: Yes, I'd really like to do that. I'm afraid it will take ages to get merged, but I'll take my chances.

alex-ju:
@soniaturcotte I did cross-browser and assistive technology testing and they all look/behave as expected. May have to repeat that after updating the code.

app/assets/javascripts/components/autocomplete.js
alex-ju: It will be quite tricky since we won't have a way to get the values to populate the dropdown and the autocomplete generates its own input.

kevindew: Sure, might be worth checking with @soniaturcotte then. As far as I was aware, I thought it was essential that a user could type in their own time outside the options?

alex-ju: Right, in the current format the users won't be able to enter a custom time if they don't have JavaScript enabled.

kevindew: I don't think they can even if they have JavaScript as the select box is what is submitted as the value rather than the input.

alex-ju: In this case the input is what's submitted, not the select. The problem we were seeing was not playing back the user their custom input (as we only got the default items in the list). I've extended the code (with my rusty ruby skills) to include the custom input within the list. Let me know if this makes sense or there's a better approach to it.

app/assets/javascripts/components/autocomplete.js
```diff
@@ -84,5 +86,53 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
...
+        })
+
+        if (options.length) {
+          options[0].selected = true
+        }
+      }
+    })
+
+    $select.style.display = 'none'
+    $select.id = $select.id + '-select'
```
kevindew: As discussed IRL, maybe we could just remove the select element here? then we wouldn't have any concerns about form name collisions

alex-ju: Updated as discussed to avoid confusion.

app/assets/javascripts/components/autocomplete.js
```diff
@@ -84,5 +86,52 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
...
+      source: function (query, syncResults) {
+        var results = []
+        $options.forEach(function ($el) {
+          results.push($el.textContent)
+        })
+        syncResults(results)
+      },
+      onConfirm: function (result) {
+        var value = result && result.value
+        var options = [].filter.call($select.options, function (option) {
```
kevindew: should this be $options?

alex-ju: it's an array of items not node elements, that's why I didn't used the $ sign. I share your view on this prefix, for me they're all objects so I will raise it with the other frontend devs and we may get rid of it.

kevindew: Sorry I actually meant whether $select.options should reference the $options variable rather than the $select.options (as per the method above)

Thought the other part is a good point too 😉

alex-ju: We're removing the `$select` element from the DOM, but the object is still in the memory and good to be referenced. I updated to reference the `$options` to avoid any doubt.

benthorner:
This took me a little while to understand - just to make sure I'm getting it, would it also be correct to say 'autocomplete-without-typeahead'?

PR #963, opened 2019-04-16
Create File Attachment models

emmabeynon:
For https://trello.com/c/3c6nomPo/782-mvp-of-inline-attachment-flow-add-and-insert

In this PR we create the various models needed to store File Attachments.  This is based strongly on the current modelling for Images.

### DIFF
--- a/app/models/file_attachment/asset.rb
+++ b/app/models/file_attachment/asset.rb
@@ -1,0 +1,19 @@
+# frozen_string_literal: true
+
+# A FileAttachment::Asset is used to interface with Asset Manager to represent a
+# file uploaded as part of a FileAttachment::Revision. It is used to track the
+# state of the media on Asset Manager and its URL.
+#
+# This is a mutable model that is mutated when state changes on Asset Manager
+class FileAttachment::Asset < ApplicationRecord
+  belongs_to :file_revision,
+             class_name: "FileAttachment::FileRevision"
+
+  enum state: { absent: "absent",
+                draft: "draft",
+                live: "live" }
+
+  enum variant: { file: "file", thumbnail: "thumbnail" }
+
+  delegate :filename, to: :file_revision
+end
--- a/app/models/revision.rb
+++ b/app/models/revision.rb
@@ -36,6 +36,12 @@ class Revision < ApplicationRecord
                           association_foreign_key: "image_revision_id",
                           join_table: "revisions_image_revisions"
 
+  has_and_belongs_to_many :file_attachment_revisions,
+                          -> { order("file_attachment_revisions.file_attachment_id ASC") },
+                          class_name: "FileAttachment::Revision",
+                          association_foreign_key: "file_attachment_revision_id",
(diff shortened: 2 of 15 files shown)


app/models/file_attachment/asset.rb
```diff
@@ -0,0 +1,22 @@
...
+# A FileAttachment::Asset is used to interface with Asset Manager to represent a
+# file uploaded as part of a FileAttachment::Revision. It is used to track the
+# state of the media on Asset Manager and its URL.
+#
+# This is a mutable model that is mutated when state changes on Asset Manager
+class FileAttachment::Asset < ApplicationRecord
+  belongs_to :file_revision,
+             class_name: "FileAttachment::FileRevision"
+
+  belongs_to :superseded_by,
```
kevindew: We'll only need this if we support replacing files, shall we assume we support replacing or add it in later?

app/models/revision.rb
```diff
@@ -36,6 +36,12 @@ class Revision < ApplicationRecord
                           association_foreign_key: "image_revision_id",
                           join_table: "revisions_image_revisions"
 
+  has_and_belongs_to_many :file_attachment_revisions,
```
kevindew: The use of this join table feels like a decision we should document at some point (maybe a section in dev doc for things to be part of an ADR)

It seems pragmatic now that we store an association of file_attachments to revisions, but this doesn't feel long term friendly to a world where we may also have html_attachments or if we need to do anything like ordering (as per publications) so probably flagged as something that is fine for now but we can anticipate issues.

db/migrate/20190416075600_create_file_attachment_asset.rb
```diff
@@ -0,0 +1,27 @@
...
+                                  on_delete: :restrict },
+                   index: true,
+                   null: false
+
+      t.references :superseded_by,
+                   foreign_key: { to_table: :file_attachment_assets,
+                                on_delete: :nullify },
+                   null: true
+
+      t.string :variant, null: false
```
kevindew: We may need to work out what we do with variant in this file attachment context.

With images we use this for all the various file sizes, so it can have quite a variety of values. Whereas for attachments the only scenario where we may have two would be a pdf with a thumbnail and we've have assets for each of those.

I guess we'd maybe want variant to be an enum in that case?

app/models/revision.rb
emmabeynon: This has been documented here https://docs.google.com/document/d/1ZeWCxQ71NU_8QvVT3Y5ukl5Xei-U3DLWsc7-e5LKPB4/edit# so will resolve.

db/migrate/20190416075600_create_file_attachment_asset.rb
emmabeynon: Added a variant enum to the model 👍

app/models/file_attachment/asset.rb
emmabeynon: I've removed `superseded_by` from the model and the migration.

PR #965, opened 2019-04-16
Remove feature flag from paste-html-to-govspeak

emmabeynon:
For https://trello.com/c/3JX1Pc9f/770-embed-paste-html-to-govspeak-functionality

### DIFF
--- a/app/views/documents/fields/_govspeak_input.html.erb
+++ b/app/views/documents/fields/_govspeak_input.html.erb
@@ -21,7 +21,7 @@
       govspeak_path: govspeak_preview_path(edition.document),
       edition: edition,
       error_items: issues&.items_for(field.id.to_sym),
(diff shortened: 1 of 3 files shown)


app/views/documents/fields/_govspeak_input.html.erb
```diff
@@ -20,8 +20,7 @@
       contextual_guidance: "document-contents-guidance",
       govspeak_path: govspeak_preview_path(edition.document),
       edition: edition,
-      error_items: issues&.items_for(field.id.to_sym),
-      paste_html_to_govspeak: current_user.has_permission?(User::PRE_RELEASE_FEATURES_PERMISSION)
```
kevindew: I think we should just change this to true for now rather than remove the conditional aspects. In the early days of this it would give us a bit more flexibility to turn it off and on and would leave both iterations in the component-guide.

emmabeynon: I've changed this to true now.

PR #968, opened 2019-04-16
Add attachment meta component

alex-ju:
Add an ["attachment meta" component](https://content-publisher-revie-pr-968.herokuapp.com/component-guide/attachment_meta) in a similar fashion with the ["image meta" component](https://content-publisher-revie-pr-968.herokuapp.com/component-guide/image_meta)

[Trello card](https://trello.com/c/z1aiStcD)

### DIFF
--- a/app/views/components/_attachment_meta.html.erb
+++ b/app/views/components/_attachment_meta.html.erb
@@ -1,0 +1,41 @@
+<%
+  id ||= nil
+  title ||= nil
+  thumbnail_src ||= nil
+  actions ||= []
+  metadata_items ||= []
+%>
+
+<%= tag.div class: "app-c-attachment-meta", id: id do %>
+  <%= tag.div class: "app-c-attachment-meta__thumbnail-container" do %>
+
+    <%= tag.img class: "app-c-attachment-meta__thumbnail-image",
+      src: thumbnail_src,
+      alt: title
+    %>
+
+  <% end %>
+
+  <%= tag.div class: "app-c-attachment-meta__metadata-container" do %>
+
+    <%= tag.p class: "app-c-attachment-meta__title" do %>
+      <%= link_to title, url, class: 'govuk-link' %>
+    <% end %>
+
+    <% if metadata_items.any? %>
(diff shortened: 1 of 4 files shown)


app/views/components/_attachment_meta.html.erb
```diff
@@ -0,0 +1,41 @@
+<%
+  id ||= "attachment-meta-#{SecureRandom.hex(4)}"
```
kevindew: Do we use this for anything?

app/views/components/docs/attachment_meta.yml
```diff
@@ -0,0 +1,33 @@
...
+
+  * have alt text
+examples:
+  default:
+    data:
+      url: https://www.gov.uk/government/publications/minimum-wage-jobs-in-ashfield-local-authority-since-2009
+      thumbnail_src: https://assets.digital.cabinet-office.gov.uk/government/assets/pub-cover-doc-afe3b0a8a9511beeca890340170aee8b5649413f948e512c9b8ce432d8513d32.png
+      title: Document attachment
+      metadata_items:
+       - file_format: MS Word Document
```
kevindew: We're going to hit something tricky with this and govspeak.

At the moment the information on the various texts to display here is stored in govspeak: https://github.com/alphagov/govspeak/blob/master/lib/govspeak/presenters/attachment_presenter.rb#L203-L240 (presumably Whitehall has a similar thing).

I don't really know what the right answer is for where to put this is.

This is a bit outside the scope of your card, just mentioning it as I see a problem looming which may make it harder to use this component 🤔

app/views/components/_attachment_meta.html.erb
alex-ju: Nope, not much. I thought would be good to have some consistency about the attributes that can be passed to a component.

app/views/components/docs/attachment_meta.yml
benthorner: Could we render the attachment using Govspeak and append the actions separately?

PR #971, opened 2019-04-17
Frontend cleanup

alex-ju:
This PR does the following in an attempt to code consistency:
- Update templates to only generate unique `id`s when mandatory and make use of rails view helpers to simplify the logic and preserve markup formatting
- Rename `dropdown-spec` to `toolbar-dropdown-spec` as it was missed when renaming the component
- Declare `eslint-env` consistently in spec files
- Fixed a wrongly formatted YAML line which is currently breaking the component-guide

### DIFF
--- a/app/views/components/_warning_summary.html.erb
+++ b/app/views/components/_warning_summary.html.erb
@@ -2,30 +2,27 @@
   title ||= false
   description ||= false
   items ||= []
-  id ||= "app-c-warning-summary-#{SecureRandom.hex(4)}"
+  id ||= nil
 %>
 
-<div class="app-c-warning-summary" id="<%= id %>">
-  <h2 class="app-c-warning-summary__title">
-    <%= title %>
-  </h2>
+<%= tag.div class: "app-c-warning-summary", id: id do %>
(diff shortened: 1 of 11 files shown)


app/views/components/_warning_summary.html.erb
```diff
@@ -2,30 +2,27 @@
   title ||= false
   description ||= false
   items ||= []
-  id ||= "app-c-warning-summary-#{SecureRandom.hex(4)}"
+  id ||= id
```
kevindew: is this meant to be nil?

alex-ju: 🤦‍♂️ yep, good spot.

alex-ju: Updated

kevindew: Looks still here - did you push?

alex-ju: Yep, put the it seems like I got the same error in two files, now is fixed in both.

benthorner:
Nice one @alex-ju !

PR #973, opened 2019-04-17
Replace vendored files with their equivalent via npm/yarn

alex-ju:
This PR removes the assets following vendored packages:
- cropperjs
- accessible-autocomplete
- miller-columns-element
- markdown-toolbar-element

Ideally we would publish the two remaining packages (miller-columns-element and markdown-toolbar-element) and merge the multiselect work into the accessible autocomplete (although we're still iterating on this component) then grab them from npm instead of github.


kevindew:
This looks ace, I didn't realise these were on NPM? Did you do that legwork too?

alex-ju:
I'm going to close this as we're gradually doing these changes component by component.
miller-columns-element will be published soon, I'll raise one soon for cropperjs and hopefully we'll find a way for the accessible-autocomplete too.

PR #977, opened 2019-04-18
Upload attachments

emmabeynon:
For https://trello.com/c/3c6nomPo/782-mvp-of-inline-attachment-flow-add-and-insert

This creates a basic workflow for users to upload attachments.  Once an attachment has been uploaded users will see a markdown snippet for both attachments and link attachments that they can copy and paste in to the body of an edition.

**Insert attachment link**

body_insert_attachment

**Attachment upload form**

upload_file_attachment_form


1pretz1:
Just noticed in the comment you refer to the general trello doing board, I think you meant to reference - https://trello.com/c/3c6nomPo/782-mvp-of-inline-attachment-flow-add-and-insert?

app/views/attachments/index.html.erb
```diff
@@ -29,3 +29,18 @@
     </div>
   </div>
 </div>
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <% attachments = @edition.file_attachment_revisions %>
+
+    <% if attachments.any? %>
```
1pretz1: Missing indentation for line 37 under the if block

spec/features/editing_attachments/upload_attachment_spec.rb
```diff
@@ -29,5 +30,9 @@ def and_i_upload_an_attachment
     find('form input[type="file"]').set(Rails.root.join(file_fixture(@attachment_filename)))
     fill_in "title", with: "A title"
     click_on "Upload"
+
+  def then_i_can_see_the_attachment_markdown_on_the_page
```
1pretz1: Maybe this should be renamed to - `then_i_can_see_the_attachment_markdown`?

spec/factories/file_attachment_file_revision_factory.rb
```diff
@@ -8,6 +8,7 @@
 
     transient do
       fixture { "text-file.txt" }
+      assets { nil }
```
1pretz1: FileAttachment::FileRevision does not have many assets, it only `has_one` file_asset so this particular transient attribute is incorrect.

spec/factories/file_attachment_revision_factory.rb
```diff
@@ -6,6 +6,7 @@
     association :created_by, factory: :user
 
     transient do
+      assets { nil }
```
1pretz1: See above comment ^

app/controllers/attachments_controller.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+class AttachmentsController < ApplicationController
```
1pretz1: Should the name of this controller be `FileAttachmentController` as we're using FileAttachment for the model.

app/services/preview_service.rb
```diff
@@ -44,15 +44,21 @@ def upload_assets(edition)
     edition.image_revisions.each do |image_revision|
       image_revision.ensure_assets
 
-      image_revision.assets.each { |asset| upload_image(edition, asset) }
+      image_revision.assets.each { |asset| upload_asset(edition, asset) }
+    end
+
+    edition.file_attachment_revisions.each do |file_attachment_revision|
+      file_attachment_revision.ensure_assets
```
1pretz1: Just wondering why we are calling `ensure_assets` from here and from the file_attachment_upload_service?

app/services/versioning/revision_updater/file_attachment.rb
```diff
@@ -0,0 +1,20 @@
...
+    module FileAttachment
+      def update_file_attachment(attachment_revision)
+        revisions = other_file_attachments(attachment_revision) + [attachment_revision]
+        assign(file_attachment_revisions: revisions)
+      end
+
+    private
+
+      def other_file_attachments(attachment_revision)
+        revision.file_attachment_revisions.reject do |revision|
```
1pretz1: Revision being shadowed on the end of the line is confusing, maybe the block parameter should be made into an acronym `far`, this is what's currently done in the image mixin for revision updater - `revision_updater/image.rb`.

1pretz1:
We've noticed that the Govspeak preview work has been started and not completed, and wondered if it it should be a separate pr with the finished project - https://github.com/alphagov/content-publisher/pull/977/commits/b20de581340b8b5d27c82e964775416c8295c87c.

And thought that `FileAttachment` could be renamed to just `Attachment` as there is multiple examples of having `File` as the prefix made names longwinded and confusing, for example `file_attachment_file_revisions` isn't very nice to read, where as `attachment_file_revisions` is reasonable.

Another naming suggestion is to rename `FileRevision` to `BlobRevision`, this would counter the example above with `file_attachment_blob_revisions` which reads much better.

app/views/attachments/index.html.erb
kevindew: Thanks sorted

spec/features/editing_attachments/upload_attachment_spec.rb
kevindew: Updated

app/services/versioning/revision_updater/file_attachment.rb
kevindew: Sorted

app/services/preview_service.rb
kevindew: Yeah I don't think that's needed. We do the same thing with images so I'd want to change them both at the same time (if at all).

app/controllers/attachments_controller.rb
kevindew: This one is a bit tied up in "attachments" being the public naming of the concern, but FileAttachment being the only type of attachment and a private concern.

Since we link to insert attachment I believe the URL should be `/attachments` and it feels more natural for the controller to follow that name.

kevindew:
> We've noticed that the Govspeak preview work has been started and not completed, and wondered if it it should be a separate pr with the finished project - [b20de58](https://github.com/alphagov/content-publisher/commit/b20de581340b8b5d27c82e964775416c8295c87c).

Makes sense, I've pulled that out

> And thought that `FileAttachment` could be renamed to just `Attachment` as there is multiple examples of having `File` as the prefix made names longwinded and confusing, for example `file_attachment_file_revisions` isn't very nice to read, where as `attachment_file_revisions` is reasonable.
> 
> Another naming suggestion is to rename `FileRevision` to `BlobRevision`, this would counter the example above with `file_attachment_blob_revisions` which reads much better.

As we discussed offline blob revision suggestion sounds good (and going into next card), attachment as a name felt too likely to cause confusion if we have a UI concern of attachments containing file attachments and HTML attachments.

app/controllers/attachments_controller.rb
benthorner: I think we were coming from the idea that we're using the concept 'file attachment' to make room for 'html attachment' when it comes along. Maybe we're planning to use this controller for HTML attachments as well, but if not, would it be reasonable to change the models to follow the public naming instead?

app/services/preview_service.rb
kevindew: I've removed them

app/models/file_attachment/revision.rb
```diff
@@ -16,7 +16,9 @@ class FileAttachment::Revision < ApplicationRecord
 
   delegate :title, to: :metadata_revision
   delegate :filename,
-           :file_asset,
+           :asset,
+           :assets,
+           :asset_url,
```
benthorner: Will `asset_url` work here?

spec/models/file_attachment/file_revision_spec.rb
```diff
@@ -2,23 +2,39 @@
 
 RSpec.describe FileAttachment::FileRevision do
   describe "#ensure_assets" do
-    it "doesn't change an asset that already exists" do
+    it "doesn't change the assets when they already exist" do
       file_revision = build(:file_attachment_file_revision)
-      file_asset = file_revision.file_asset
+      assets = file_revision.assets
```
benthorner: Just a minor thing, but as this is an unevaluated collection, I think this just checks `true == true`?

app/models/file_attachment/revision.rb
kevindew: Thanks - removed.

spec/models/file_attachment/file_revision_spec.rb
kevindew: I was a little bit confused by this comment, it did lead me to fix an issue and hopefully it's the same one you're referencing as they sound rather similar.

By comparing an instance of CollectionProxy in the expectation we weren't actually testing if anything changed, because even if the contents of the CollectionProxy change they'll still be the same object. 

I've resolved this by converting to an array.

benthorner: Sorry, I put it on the wrong line, which wasn't helpful :-|.

PR #984, opened 2019-04-24
Bump govuk_test to 0.4.2

benthorner:
This updates to the latest version of govuk_test in order to take
advantage of the new env var to disable the now-deprecated
chromedriver-helper.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -42,7 +42,7 @@ group :development, :test do
   gem "factory_bot_rails", "~> 5"
   gem "govuk-lint", "~> 3"
   gem "govuk_schemas", "~> 3.2"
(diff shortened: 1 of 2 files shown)


Gemfile
```diff
@@ -42,7 +42,7 @@ group :development, :test do
   gem "factory_bot_rails", "~> 5"
   gem "govuk-lint", "~> 3"
   gem "govuk_schemas", "~> 3.2"
-  gem "govuk_test", "~> 0.3"
```
kevindew: Any reason why we're doing a more specific pinning?

benthorner: What more specific pinning? ;-)

PR #985, opened 2019-04-24
Rename FileRevision to BlobRevision

benthorner:
https://trello.com/c/pd6fOgwf/800-follow-up-steps-on-minimal-workflow-for-inline-attachments

Previously we used FileRevision to encapsulate the versioning of the
bytes ('blob') associated with an image or a file attachment. This
renames FileRevision to BlobRevision to reduce the reuse of the word
'file' throughout the codebase, as the repetition was confusing.

This also updates the debug view for an edition so that it's
automatically available to devs and shows any file attachments.


docs/adr/0009-modelling-history.md
```diff
@@ -1,6 +1,6 @@
 # 9. Modelling history
 
-Date: 2018-12-17  
+Date: 2018-12-17
```
kevindew: Btw you've lost the new line here

docs/adr/0009-modelling-history.md
```diff
@@ -1,6 +1,6 @@
 # 9. Modelling history
```
kevindew: I don't think we should be changing the ADR for this, it represents a document at a point in time rather than a living document.

docs/adr/0009/revision-diagram.plantuml
```diff
@@ -20,7 +20,7 @@ class MetadataRevision {
 }
 
 class ImageRevision {
-  image_file_revision
+  image_blob_revision
```
kevindew: Note you'd also have to regenerate the png

db/migrate/20190424145213_rename_file_revision_to_blob_revision.rb
```diff
@@ -0,0 +1,12 @@
+class RenameFileRevisionToBlobRevision < ActiveRecord::Migration[5.2]
```
kevindew: I'd suggest that we do this migration in a separate PR and do the first deploy of the code with it using the old table names but new model names. 

Then we can do another one that flips things over and can break and would be a much smaller amount of changes.

db/schema.rb
```diff
@@ -90,24 +90,24 @@
...
 
-  create_table "file_attachment_file_revisions", force: :cascade do |t|
+  create_table "file_attachment_blob_revisions", force: :cascade do |t|
     t.bigint "blob_id", null: false
     t.bigint "created_by_id"
     t.string "filename", null: false
     t.datetime "created_at", null: false
     t.bigint "size", null: false
-    t.index ["blob_id"], name: "index_file_attachment_file_revisions_on_blob_id"
+    t.index ["blob_id"], name: "index_file_attachment_blob_revisions_on_blob_id"
```
kevindew: it doesn't look like there are migrations that rename this, and any other, indexes?

db/seeds.rb
```diff
@@ -4,6 +4,6 @@
 
 gds_organisation_content_id = "af07d5a5-df63-4ddc-9383-6a666845ebe9"
 
-user.update!(permissions: [User::PRE_RELEASE_FEATURES_PERMISSION],
+user.update!(permissions: [User::PRE_RELEASE_FEATURES_PERMISSION, User::DEBUG_PERMISSION],
```
kevindew: I'm a bit concerned about the test user having debug permission by default. Could we do that if you're in a dev environment?

Could this be a separate commit too?

app/controllers/debug_controller.rb
```diff
@@ -48,17 +48,30 @@ def revision_hash(revision)
...
+  end
+
+  def file_attachment_revision_hash(file_attachment_revision)
+    return nil unless file_attachment_revision
+
+    common_except = %i[id created_at created_by_id]
+    blob_revision = file_attachment_revision.blob_revision.as_json(except: common_except)
+    metadata_revision = file_attachment_revision.metadata_revision.as_json(except: common_except)
+
+    blob_revision.merge(metadata_revision)
```
kevindew: Thanks for getting file attachment stuff 👍 

I would suggest also a separate commit for this bit

docs/adr/0009-modelling-history.md
benthorner: I did wonder - undone.

db/seeds.rb
benthorner: Fair enough - done :-).

db/migrate/20190424145213_rename_file_revision_to_blob_revision.rb
benthorner: I had a quick think about possible approaches, but as it quickly became apparent the change is a breaking one, it seemed more valuable to keep everything in one PR to aid understanding. As changing table names should be fast, is there really much value in separating it?

db/schema.rb
benthorner: I doubted myself there for a moment, but thankfully this is done as part of renaming the column.

```
psql content_publisher_development -c '\di' | grep file_revision
```

db/migrate/20190424145213_rename_file_revision_to_blob_revision.rb
kevindew: Yup there is a breaking part of it whatever happens. What I'm encouraging is minimising the amount that is in the breaking piece of work.

Thus you'd have as much as possible in something that can be merged in as normal and avoids any special deployment considerations and is pretty safe to undo. Then you'd have a change that is db migration and model changes that needs to go from merge to deploy as quick as possible and is rather painful to undo.

db/schema.rb
kevindew: Oh TIL didn't realise the rename commands had extra side effects

db/migrate/20190424145213_rename_file_revision_to_blob_revision.rb
benthorner: How about now? (I didn't think this was doable for the column renames initially.)

PR #986, opened 2019-04-24
Force using system chromedriver everywhere

benthorner:
Hopefully we can remove this if https://github.com/alphagov/govuk_test/pull/19 goes down well.

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -2,6 +2,7 @@
 
 ENV["RAILS_ENV"] ||= "test"
 ENV["GOVUK_APP_DOMAIN"] = "test.gov.uk"
+ENV["GOVUK_TEST_USE_SYSTEM_CHROMEDRIVER"] = "true"
 
 require File.expand_path("../config/environment", __dir__)
 require "rspec/rails"


kevindew:
@benthorner could you explain why we'd want to force it?

Is there a discussion going on between the competing PRs of https://github.com/alphagov/govuk_test/pull/18 and https://github.com/alphagov/govuk_test/pull/19

By the number of reverts there seems to be a lot of problems with attempting to use system chromedriver and it doesn't look like any other apps use this env var. It's never been really explained what the benefit of using system chromedriver offers over using a provided one.

benthorner:
@kevindew sure - all of the problems we've had are actually due to the gem, rather than the system chromedriver. The gem theoretically allows us to lock to a version of chromedriver, but this has lead to two problems:

   - New releases of Chrome are installed on systems before the gem updates to cover the new version (my local build is failing because of this)
   - Google pulls all old releases of chromedriver, so version locks aren't actually feasible

It seems the main reason for using the gem was a 'handy' way to auto-install chromedriver, but it's actually on all our systems anyway. I can always set this locally, so perhaps this isn't necessary for everyone, although I expect your build will break soon too :-|.

kevindew:
@benthorner sure, do you know if you continue getting those problems if you use the webdrivers gem? I don't think that is tied to particular versions in the gem from what I can tell.

Since the idea to use the chromedriver path has caused problems each time, I'm a bit wary of it - and it's also another dependency to install and manage for a dev.

benthorner:
Closing in favour of further discussion on https://github.com/alphagov/govuk_test/pull/19.

PR #992, opened 2019-04-30
Show attachment snippet on separate page

benthorner:
https://trello.com/c/gf8wSm9P/825-build-page-for-inserting-inline-attachments-to-match-design

Previously we showed the markdown snippet for an attachment as part of
the list of attachments for an edition. This moves the snippet to a
separate page and makes use of the attachment_meta component to work
towards matching the design for both of these pages.

### Show page
Screen Shot 2019-04-30 at 09 27 09

### Index page
Screen Shot 2019-04-30 at 09 27 19

### Index page
Screen Shot 2019-04-30 at 09 28 27

### DIFF
--- a/app/models/file_attachment/revision.rb
+++ b/app/models/file_attachment/revision.rb
@@ -19,7 +19,8 @@ class FileAttachment::Revision < ApplicationRecord
            :asset,
            :assets,
            :ensure_assets,
-           :blob,
+           :content_type,
+           :byte_size,
            to: :blob_revision
 
   def readonly?
--- a/app/models/file_attachment/revision.rb
+++ b/app/models/file_attachment/revision.rb
@@ -19,6 +19,7 @@ class FileAttachment::Revision < ApplicationRecord
            :asset,
            :assets,
            :ensure_assets,
+           :blob,
            to: :blob_revision
 
   def readonly?
--- a/app/views/file_attachments/index.html.erb
+++ b/app/views/file_attachments/index.html.erb
@@ -39,7 +39,17 @@
       </h2>
 
       <% attachments.each do |attachment| %>
-        <%= render "file_attachment", attachment: attachment, document: @edition.document %>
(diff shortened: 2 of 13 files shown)


app/views/file_attachments/index.html.erb
```diff
@@ -39,7 +39,17 @@
       </h2>
 
       <% attachments.each do |attachment| %>
-        <%= render "file_attachment", attachment: attachment, document: @edition.document %>
+        <%= render "components/attachment_meta", {
+          title: attachment.title,
+          url: "#",
+          metadata_items: [
+            attachment.blob.content_type,
+            number_to_human_size(attachment.blob.byte_size),
```
kevindew: We should already have both of these without needing to go through blob.

We already have content_type as a method on a blob_revision which should be delegated to revision (as per [Image::Revision](https://github.com/alphagov/content-publisher/blob/master/app/models/image/revision.rb#L31) and should probably also delegate byte_size in the same way.

We have a size method on a FileAttachment::BlobRevision - which was used for this first. However I imagine Emma and I had overlooked that it was available on a blob so that should be removed if we're just going to use the blob one: https://github.com/alphagov/content-publisher/blob/03d45f97fc9b78e6b71fed1f7e0160c0300c914f/db/schema.rb#L109

benthorner: @kevindew ah, totally missed that. I've done as you suggested, and added a commit to remove the redundant field on the revision.

app/models/file_attachment/revision.rb
```diff
@@ -19,6 +19,7 @@ class FileAttachment::Revision < ApplicationRecord
            :asset,
            :assets,
            :ensure_assets,
+           :blob,
```
kevindew: does this need to be content_type and byte_size?

benthorner: Sorry - fixed.

PR #997, opened 2019-05-01
Add PDF page count

emmabeynon:
For https://trello.com/c/zMYmFTPx/801-show-metadata-for-attachments

This calculates the number of pages in a PDF so that it can eventually be displayed in a Govspeak preview, as part of information about PDFs that can be inserted in Content Publisher, and on the GOV.UK frontend.


Gemfile
```diff
@@ -19,6 +19,7 @@ gem "image_processing", "~> 1"
 gem "interactor", "~> 3"
 gem "kaminari", "~> 1"
 gem "notifications-ruby-client", "~> 3.1"
+gem "pdf-reader", "~> 2.2"
```
kevindew: It looks like it'd be more consistent with our other version pinning for this to be `~> 2`

app/models/file_attachment/revision.rb
```diff
@@ -21,6 +21,7 @@ class FileAttachment::Revision < ApplicationRecord
            :ensure_assets,
            :content_type,
            :byte_size,
+           :pdf_page_count,
```
kevindew: I'd suggest we don't prefix this with pdf, although this is the only format we support it on it could well be used for other formats if we could support it. I'm hoping this can help sidestep us away from having to handle file type metadata individually.

I'd also suggest we name this number_of_pages for Whitehall consistency. I suggest this because Whitehalls usage is already in govspeak and it'd be good to use same name in govuk_publishing_components.

app/services/file_attachment_upload_service.rb
```diff
@@ -40,4 +45,12 @@ def filename
...
     UniqueFilenameService.new(existing_filenames).call(file.original_filename)
   end
+
+  def page_count
+    return unless file.content_type == PDF_CONTENT_TYPE
+
+    PDF::Reader.new(file.tempfile).page_count
+  rescue PDF::Reader::MalformedPDFError, PDF::Reader::UnsupportedFeatureError, OpenSSL::Cipher::CipherError
+    nil
+  end
```
kevindew: This looks good - does it seem like something we could test with a fixture?

app/services/file_attachment_upload_service.rb
```diff
@@ -40,4 +45,12 @@ def filename
     existing_filenames = revision.file_attachment_revisions.map(&:filename)
     UniqueFilenameService.new(existing_filenames).call(file.original_filename)
   end
+
+  def page_count
+    return unless file.content_type == PDF_CONTENT_TYPE
```
kevindew: dare say at this point we might not need a constant since we're only using it here

app/models/file_attachment/revision.rb
emmabeynon: OK, I wasn't aware there would be other formats with multiple pages.  I figured prefixing it would make it clear what format it's for.  I'll change to `number_of_pages`.

app/services/file_attachment_upload_service.rb
emmabeynon: Sure, I figured we'd introduce the fixture when it's required for a test.  The File Attachment Upload Service doesn't have its own test file currently but I'd expect we'd introduce a test for page counts when we start displaying it in the app so the fixture can be added then.

app/services/file_attachment_upload_service.rb
emmabeynon: Might it make sense to make the constant an array to hold PDF and any future content types that might have pages?

app/models/file_attachment/revision.rb
kevindew: There's no plans for other formats to support it - and we may well never add another one. I meant more that there's no need to prefix it for pdf as if we were to ever support it for another format we could just re-use the field.

PR #1000, opened 2019-05-02
Render attachment link

emmabeynon:
For https://trello.com/c/zMYmFTPx/801-show-metadata-for-attachments

This renders the attachment link component for each uploaded attachment.

Screen Shot 2019-05-02 at 13 56 24


spec/features/editing_file_attachments/upload_file_attachment_spec.rb
```diff
@@ -30,15 +30,25 @@ def and_i_upload_a_file_attachment
...
   def then_i_can_see_the_attachment_markdown
-    expect(page).to have_content("[Attachment: attachment.csv]")
-    expect(page).to have_content("[AttachmentLink: attachment.csv]")
+    expect(page).to have_content("[Attachment: attachment.pdf]")
+  end
+
+  def and_i_can_see_a_preview_of_the_attachment_link
+    attachment = FileAttachment::Revision.last
+    file_size = number_to_human_size(attachment.byte_size)
+    number_of_pages = pluralize(attachment.number_of_pages, "page")
```
kevindew: I'd suggest we hard code these rather than having to use the helpers to look these up, as it has to dig a reasonable way into the code to pull out the values.

To make it clearer what they are they could put in the pdf file name. E.g. `10kb-2-page.pdf` ?

app/views/file_attachments/show.html.erb
```diff
@@ -12,11 +12,17 @@
...
+        attachment: {
+          title: @attachment.title,
+          url: "#",
+          content_type: @attachment.content_type,
+          file_size: @attachment.byte_size,
+          number_of_pages: @attachment.number_of_pages,
+        },
+        target: "_blank"
+      } %>
+    </div>
```
kevindew: Oh we should still have some the markdown provided for this - just noticed running locally.

spec/features/editing_file_attachments/upload_file_attachment_spec.rb
```diff
@@ -30,15 +30,25 @@ def and_i_upload_a_file_attachment
...
     click_on "Upload"
   end
 
   def then_i_can_see_the_attachment_markdown
-    expect(page).to have_content("[Attachment: attachment.csv]")
-    expect(page).to have_content("[AttachmentLink: attachment.csv]")
+    expect(page).to have_content("[Attachment: attachment.pdf]")
+  end
+
+  def and_i_can_see_a_preview_of_the_attachment_link
```
kevindew: Also noticed that this method isn't actually called.

app/views/file_attachments/show.html.erb
emmabeynon: I've added the attachment link markdown back in 👍

spec/features/editing_file_attachments/upload_file_attachment_spec.rb
emmabeynon: D'oh - added to the scenario block.

spec/features/editing_file_attachments/upload_file_attachment_spec.rb
emmabeynon: I've renamed the PDF and hardcoded the values 👍

emmabeynon:
Thanks @kevindew your comments should be addressed now.

PR #1001, opened 2019-05-02
Rename GTM vars to be consistent with Content Data

benthorner:
https://trello.com/c/NJajzaXL/925-tidy-up-our-gtm-config

https://github.com/alphagov/content-data-admin/blob/master/app/views/layouts/application.html.erb#L9

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -6,8 +6,8 @@
 
   <script>
     var dataLayer = [{
-      'dimension1': '<%= current_user.organisation_slug %>',
-      'cp_uid': '<%= current_user.uid %>'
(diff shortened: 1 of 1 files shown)


app/views/layouts/application.html.erb
```diff
@@ -6,8 +6,8 @@
 
   <script>
     var dataLayer = [{
-      'dimension1': '<%= current_user.organisation_slug %>',
-      'cp_uid': '<%= current_user.uid %>'
+      'user-organisation': '<%= current_user.organisation_slug %>',
+      'cp-uid': '<%= current_user.uid %>'
```
kevindew: Yay, much nicer. So kebab case all the way for dataLayer attributes?

benthorner: Yup :-).

PR #1002, opened 2019-05-02
Add validation requirements for attachments

1pretz1:
Currently file attachments can be uploaded without any checks on the file.The FileAttachmentUploadChecker has been created to check the validity of the file attempted to be uploaded. 

There are two input types, one being the file itself and the other the file's title. The file's title has two checks being its presence and has a max length of 255 chars. The file check itself validates its presence and whether its of a supported format and if it's a zip that it only contains files of supported formats.

Trello:
https://trello.com/c/EL9jPXIl/802-validation-requirements-for-uploading-file-attachments


benthorner:
Nice one @1pretz1 👍 . I forgot to suggest maybe editing the commit with the MIME type stuff to mention that we also disallow files with no extension where this is required to infer the type (e.g. text files), as only a few files seem to lack an extension in Asset Manager.

app/services/requirements/file_attachment_upload_checker.rb
```diff
@@ -28,6 +28,8 @@ class FileAttachmentUploadChecker
       "text/xml", # xml, xsd
     ].freeze
 
+    ALLOWED_ZIP = "application/zip" # zip
```
kevindew: I found this quite confusing so I had Ben explain to me about how that zip is a little special. 

I think there's probably a way to make this easier to understand.

I'd suggest we include this mime_type in the list of allowed ones and then just check that it's not an archive in the method.

app/services/requirements/file_attachment_upload_checker.rb
```diff
@@ -63,8 +70,23 @@ def issues
...
+        zf.each do |f|
+          type = Marcel::MimeType.for(f.get_input_stream, name: f.name)
+          return true if ALLOWED_FORMATS.exclude? type
+        end
+      end
+      false
+    end
+
+    def content_type
+      Marcel::MimeType.for(file, name: file.original_filename)
```
kevindew: I'd memo this method since it's quite expensive to magically determine the mime_type

app/services/requirements/file_attachment_upload_checker.rb
```diff
@@ -63,8 +70,23 @@ def issues
...
+    def invalid_zip?
+      return if content_type != ALLOWED_ZIP
+
+      Zip::File.open(file) do |zf|
+        zf.each do |f|
+          type = Marcel::MimeType.for(f.get_input_stream, name: f.name)
+          return true if ALLOWED_FORMATS.exclude? type
+        end
+      end
+      false
```
kevindew: Should we have any exception handling in this method if it fails to open the zip? does Whitehall have exception handling for that? 

I figure it's possibly to have an invalid or password protected zip that can't be opened.

app/services/requirements/file_attachment_upload_checker.rb
```diff
@@ -63,8 +70,23 @@ def issues
...
-      type = Marcel::MimeType.for(file, name: file.original_filename)
-      ALLOWED_FORMATS.exclude? type
+      (ALLOWED_FORMATS + [ALLOWED_ZIP]).exclude? content_type
+    end
+
+    def invalid_zip?
+      return if content_type != ALLOWED_ZIP
+
+      Zip::File.open(file) do |zf|
+        zf.each do |f|
```
kevindew: Is zf an enumerable? If so I'd suggest using a method like all? or any? rather than an each with a return inside it (my instinct when seeing that is thinking you meant a next)

app/services/requirements/file_attachment_upload_checker.rb
```diff
@@ -63,8 +70,23 @@ def issues
...
+          type = Marcel::MimeType.for(f.get_input_stream, name: f.name)
+          return true if ALLOWED_FORMATS.exclude? type
```
kevindew: With regards to the zip suggestion I'd check it's an allowed format and not a zip with a comment to say we don't allow nested archives.

app/services/requirements/file_attachment_upload_checker.rb
```diff
@@ -29,7 +48,23 @@ def issues
...
+        issues << Issue.new(:file_attachment_upload, :unsupported_type)
+      end
+
       CheckerIssues.new(issues)
     end
+
+  private
+
+    def unsupported_type?
+      type = Marcel::MimeType.for(file, name: file.original_filename)
```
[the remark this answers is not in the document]
1pretz1: Okay sure, there may still be a case where there isn't a content_type, an extension and marcel isn't able to determine the type but this scenario seems highly unlikely!

app/services/requirements/file_attachment_upload_checker.rb
```diff
@@ -29,7 +48,23 @@ def issues
...
+      ALLOWED_FORMATS.exclude? type
```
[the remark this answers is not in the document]
1pretz1: Yep

config/locales/en/requirements.yml
```diff
@@ -67,6 +67,8 @@ en:
     file_attachment_upload:
       no_file:
         form_message: Select a file to upload
+      unsupported_type:
+        form_message: Select a supported file type
```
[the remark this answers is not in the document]
1pretz1: Yep this is in need of a content review! 

As to your second question, as we discussed offline I've just had a look in the whitehall UI and tried to upload a) an invalid zip file and b) an invalid file, both of these attempts were met with the error:
`To save the file attachment please fix the following issues: File can't be blank`

So this workflow seems pretty poor in whitehall currently as it doesn't indicate the right sort of error!

app/services/requirements/file_attachment_upload_checker.rb
1pretz1: Yep good shout, didn't think to use the methods you suggested! :)

app/services/requirements/file_attachment_upload_checker.rb
1pretz1: 👍

app/services/requirements/file_attachment_upload_checker.rb
1pretz1: Sure, good spot!

app/services/requirements/file_attachment_upload_checker.rb
1pretz1: From what I can see whitehall itself doesn't catch any exceptions regarding the opening of a zip. I created a new test to see if the method would bomb out with a password protected zip, but it worked correctly, as the zip can be opened to see the metadata of the files contained without actually having to open each individual (protected) file.

The other checks whitehall does is for UTF8 encoding of the filename and the contents. That the files inside the zip are one of the allowed formats as well as validating it as a ArcGIS zip but we have decided not to include this sort of check as there are very few instances of this sort of file.

1pretz1:
@kevindew ready for review no.2! :)

config/locales/en/requirements.yml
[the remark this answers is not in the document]
kevindew: woa - that's a pretty low benchmark to beat at least.

1pretz1:
@kevindew I've made a few small changes since you last reviewed the pr, one being making sure the `file_attachment_upload_service` creates a blob with the same content-type as whats been checked in the checker.

Two, I've used `file.path` instead of just `file` for the arg when opening the zip as this previously was causing errors when running the app locally (however not in the test suite as the file upload object was different between environments).

PR #1004, opened 2019-05-02
Remove foreman gem

kevindew:
The author of this gem suggests not installing this in your application
gemfile (it's a source of long running arguments in the PRs) however we
had to install it for this app to work on GOV.UK dev VMs. That problem
is now fixed so this gem can be removed.

Instead of installing the gem as part of the bundle this is installed at
the point of running ./startup.sh with the --conservative flag which
means it won't be installed if it already exists.


kevindew:
Thanks @emmabeynon would you mind looking at https://github.com/alphagov/govuk-puppet/pull/9081 too? once that is merged I can merge this.

PR #1007, opened 2019-05-07
Render attachment

kevindew:
Trello: https://trello.com/c/zMYmFTPx/801-show-metadata-for-attachments

This is a draft as it waits for https://github.com/alphagov/govuk_publishing_components/pull/842 to be released.

This provides the rendering for attachments within the application and for govspeak.

A tricky part is that we have the same hash of attachment data that is used to render a component and is sent to govspeak. I tried to find a single place that could store that serialising logic. After initially going with a method on the model I tried out a Helper after a chat with Ben. I'm not quite sure where'd be best, there are disadvantages to all options.

See commits for more details

Screen Shot 2019-05-08 at 16 51 15

### DIFF
--- a/spec/features/editing_file_attachments/upload_file_attachment_spec.rb
+++ b/spec/features/editing_file_attachments/upload_file_attachment_spec.rb
@@ -7,7 +7,7 @@ RSpec.feature "Upload file attachment" do
     and_i_go_to_insert_an_attachment
     and_i_upload_a_file_attachment
     then_i_can_see_the_attachment_markdown
-    and_i_can_see_a_preview_of_the_attachment_link
+    and_i_can_see_previews_of_the_attachment
     and_the_attachment_has_been_uploaded_successfully
   end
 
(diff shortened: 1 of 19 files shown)


spec/features/editing_file_attachments/upload_file_attachment_spec.rb
```diff
@@ -32,21 +32,29 @@ def and_i_upload_a_file_attachment
     @publishing_api_request = stub_publishing_api_put_content(@edition.content_id, {})
 
     @attachment_filename = "13kb-1-page-attachment.pdf"
```
emmabeynon: I noticed we have an instance variable here but don't use it outside of this method.  I think we could either use it in lines 43 and 44, or make it a local variable.

spec/services/govspeak_document/payload_options_spec.rb
```diff
@@ -14,12 +14,42 @@
...
+          id: "filename.png",
+          alt_text: "Some alt text",
+          caption: "An optional caption",
+          credit: "An optional credit",
+          url: a_string_matching(%r{/media/.*/filename.png}),
+        ),
+      )
+    end
+
+    it "returns the remote asset options for a file attachment" do
```
benthorner: Just a minor thing that the it text reads differently to the one above - not really fussed which one we keep :-).

kevindew:
Thanks @benthorner coincidently I spotted the same just a few mins ago on the thumbnail work so I'm just working on addressing it in the component.

1pretz1:
Looks good from me too, only a small thing. I noticed the attachments on the index page are grouped quite close together meaning it was easy to miss clicking on the title and actually delete the attachment above (as there is no space between hovering over one link to the other): 
Screen Shot 2019-05-09 at 13 55 20

Changing the margin-bottom spacing from 1 to 2 here https://github.com/alphagov/content-publisher/blob/render-attachment/app/assets/stylesheets/components/_attachment-meta.scss#L13 created a gap between the links and generally looks nicer in my opinion.

spec/features/editing_file_attachments/upload_file_attachment_spec.rb
kevindew: Good point, I've used that variable on lines 43 and 44 now.

spec/services/govspeak_document/payload_options_spec.rb
kevindew: I've made the assertions consistent now, thanks

kevindew:
@benthorner I'd agree that the meta name in component is confusing. I'm not quite sure what the right name for them is though, they act as items in a list mostly. I'd like to look at resolving that outside this PR though since that's a non blocking aspect of this.

@1pretz1 you're right that it doesn't look great as it is, I don't think that margin suggestion is the right one though, I think we'd want this to be consistent with the image one where there are separating lines between items:

Screen Shot 2019-05-10 at 09 46 51

I've just had a word with @alex-ju and he can have a look at making these more consistent and the naming.

PR #1010, opened 2019-05-07
Prevent double clicking for create/destroy actions

alex-ju:
This prevents create/destroy buttons from performing an action twice which may result in a system error (double creations) or erroring not being able to remove a non-existent item.

[Trello card](https://trello.com/c/JLkfkNxQ)

### DIFF
--- a/app/views/documents/index.html.erb
+++ b/app/views/documents/index.html.erb
@@ -1,7 +1,12 @@
 <% content_for :title, t("documents.index.title") %>
 
 <% content_for :title_side, render("govuk_publishing_components/components/button", {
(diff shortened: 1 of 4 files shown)


benthorner:
Thanks @alex-ju :-). I had a look at the [original commit](https://github.com/alphagov/govuk-frontend/commit/884e7dd73ad943d71fb701da15626e6a7ad0b933) that put this in. Although they seemed cautious about making it the default in govuk-frontend, could we do this in gov_pub_cub? Maybe it makes the scope of this too big, but it would help with future maintenance...

alex-ju:
Closed in favour of https://github.com/alphagov/govuk_publishing_components/pull/849

PR #1012, opened 2019-05-08
Download attachments

benthorner:
https://trello.com/c/xpS0zlU4/803-download-an-attachment-file-through-content-publisher

This is the initial attempt at doing 'intelligent preview' for an asset, so that we redirect to Asset Manager only when we know a file is available, and potentially try to upload it if it's not in Asset Manager at all e.g. if it was uploaded as part of the edit form for a document. Further work:

   - Use the auth bypass token in the asset URL
   - Use the new preview path elsewhere in the app
   - Consider if we want special behaviour for infected files

I've done a bit of refactoring on the PreviewService, which had similar code for managing assets, to keep the tests manageable. The commit explains a bit more about the pros/cons of doing that - it seems to be a compromise between duplication and stubbing our own code.

### DIFF
--- a/app/interactors/file_attachments/preview_interactor.rb
+++ b/app/interactors/file_attachments/preview_interactor.rb
@@ -1,0 +1,50 @@
+# frozen_string_literal: true
+
+class FileAttachments::PreviewInteractor
+  include Interactor
+
+  delegate :params,
+           :edition,
+           :attachment_revision,
+           :asset,
+           to: :context
+
+  def call
+    Edition.transaction do
+      find_and_lock_edition
+      find_attachment
+      check_uploaded
+      check_available
+    end
+  end
+
+private
+
+  def find_and_lock_edition
+    context.edition = Edition.lock.find_current(document: params[:document])
+  end
+
+  def find_attachment
+    context.attachment_revision = edition.file_attachment_revisions
+      .find_by!(file_attachment_id: params[:file_attachment_id])
+  end
+
+  def check_uploaded
+    attachment_revision.ensure_assets
+    context.asset = attachment_revision.asset("file")
+
+    if asset.absent?
+      PreviewAssetService.new(edition).upload_asset(asset)
+      context.fail!(can_preview: false)
+    end
+  rescue GdsApi::BaseError
+    context.fail!(api_error: true)
+  end
+
+  def check_available
+    service = PreviewAssetService.new(edition)
+    context.can_preview = service.can_preview_asset?(asset)
+  rescue GdsApi::BaseError
+    context.fail!(api_error: true)
+  end
+end
--- a/app/services/preview_asset_service.rb
+++ b/app/services/preview_asset_service.rb
@@ -1,0 +1,27 @@
+# frozen_string_literal: true
+
+class PreviewAssetService
+  attr_reader :edition
+
+  def initialize(edition)
+    @edition = edition
+  end
+
+  def upload_asset(asset)
+    return unless asset.absent?
+
+    auth_bypass_id = EditionUrl.new(edition).auth_bypass_id
+    file_url = AssetManagerService.new.upload(asset, auth_bypass_id)
+    asset.update!(file_url: file_url, state: :draft)
+  rescue GdsApi::BaseError => e
+    GovukError.notify(e)
+    raise
+  end
+
+  def can_preview_asset?(asset)
+    AssetManagerService.new.get(asset)["state"] == "uploaded"
+  rescue GdsApi::BaseError => e
+    GovukError.notify(e)
+    raise
+  end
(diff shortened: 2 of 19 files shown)


app/interactors/file_attachments/preview_interactor.rb
```diff
@@ -0,0 +1,48 @@
...
+class FileAttachments::PreviewInteractor
+  include Interactor
+
+  delegate :params,
+           :edition,
+           :attachment_revision,
+           :asset,
+           to: :context
+
+  def call
```
kevindew: Do you mean for this to be in a transaction?

app/interactors/file_attachments/preview_interactor.rb
```diff
@@ -0,0 +1,48 @@
...
+    if asset.absent?
+      PreviewAssetService.new(edition).upload_asset(asset)
+      context.fail!(pending: true)
+    end
+  rescue GdsApi::BaseError
+    context.fail!(api_error: true)
+  end
+
+  def check_available
+    context.asset = attachment_revision.asset("file")
```
kevindew: Isn't this context already set by here?

app/services/preview_asset_service.rb
```diff
@@ -0,0 +1,39 @@
...
+
+    auth_bypass_id = EditionUrl.new(edition).auth_bypass_id
+    file_url = AssetManagerService.new.upload(asset, auth_bypass_id)
+    asset.update!(file_url: file_url, state: :draft)
+  rescue GdsApi::BaseError => e
+    GovukError.notify(e)
+    raise
+  end
+
+  def pending?(asset)
```
kevindew: minor consistency issue that the method above this one uses asset in it's name whereas this one has it implied by argument.

app/interactors/file_attachments/preview_interactor.rb
benthorner: At first I thought it wasn't necessary, but the call to 'ensure_assets' could create Asset records, which the PreviewAssetService would then save, so I think we do need it.

app/interactors/file_attachments/preview_interactor.rb
benthorner: Oops - thanks for spotting!

app/services/preview_asset_service.rb
benthorner: Good point - I've renamed this one to 'can_preview_asset?' and inverted the code where it was used; pending was a bit vague to be useful at this point, I think. Doing this also made me spot I'd missed the unit tests for this method :-).

kevindew: Heh I was going to suggest changing pending but I noticed it was used liberally in the trello card and that the Asset Manager alternative was "accessible" which was of course rather ambiguous.

app/interactors/file_attachments/preview_interactor.rb
```diff
@@ -0,0 +1,48 @@
...
+  def find_attachment
+    context.attachment_revision = edition.file_attachment_revisions
+      .find_by!(file_attachment_id: params[:file_attachment_id])
+  end
+
+  def check_uploaded
+    attachment_revision.ensure_assets
+    context.asset = attachment_revision.asset("file")
+
+    if asset.absent?
```
1pretz1: As we just discussed in person, is it better to upload the attachment here which would probably mean the user wouldn't be able to access the attachment immediately meaning we would create a page to handle to the pending status?

Or to upload the attachment to asset manager when the file is originally uploaded which would mitigate the preview issue mentioned above. A possible downside I can see with this approach is if an asset was uploaded and deleted before the draft was saved then multiple requests would be sent to asset manager (however this scenario seems unlikely).

Sorry for repeating what we discussed but I thought i should put it in writing here too. :)

benthorner: @1pretz1 as discussed in person: this is something we could provide in feature if we see people are getting the 'pending' page a lot, but we would still have check_uploaded here as a decoupled fallback :-). We should do this again ;-).

app/interactors/file_attachments/preview_interactor.rb
kevindew: You're right it can change state so should be within a transaction.

benthorner:
@kevindew I've rebased this against the now-merged rendering PR and added a new commit to use the new in-app preview path. Slight downside is we now have to pass document everywhere, but the only alternative seems to be `attachment_revision.revisions.first.document` :-|.

kevindew:
@benthorner that seems reasonable, I can't imagine we'd likely be in a a situation where we aren't already in a document context before looking up the attachment. Even if we did have a nice association based method it'd likely generate an extra query.

PR #1015, opened 2019-05-08
Content updates

alex-ju:
Content updates based on last round of user research.

[Trello card](https://trello.com/c/DaLtJhyA)

### DIFF
--- a/config/locales/en/images/crop.yml
+++ b/config/locales/en/images/crop.yml
@@ -2,5 +2,5 @@ en:
   images:
     crop:
       title: "Select the part of the image to use"
-      title_modal: "Crop image"
+      title_modal: "Select the part of the image to use"
       title_exact_dimensions: "This image cannot be cropped"
--- a/config/locales/en/images/crop.yml
+++ b/config/locales/en/images/crop.yml
@@ -2,5 +2,4 @@ en:
   images:
     crop:
       title: "Select the part of the image to use"
-      title_modal: "Select the part of the image to use"
       title_exact_dimensions: "This image cannot be cropped"
--- a/config/locales/en/images/crop.yml
+++ b/config/locales/en/images/crop.yml
@@ -3,3 +3,4 @@ en:
     crop:
       title: "Crop image for ‘%{title}’"
       title_modal: "Crop image"
+      title_exact_dimensions: "Image does not need to be cropped"
--- a/config/locales/en/images/crop.yml
+++ b/config/locales/en/images/crop.yml
@@ -1,6 +1,6 @@
 en:
   images:
     crop:
-      title: "Crop image for ‘%{title}’"
+      title: "Select the part of the image to use"
       title_modal: "Crop image"
-      title_exact_dimensions: "Image does not need to be cropped"
+      title_exact_dimensions: "This image cannot be cropped"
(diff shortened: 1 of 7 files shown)


config/locales/en/images/crop.yml
```diff
@@ -3,3 +3,4 @@ en:
     crop:
       title: "Crop image for ‘%{title}’"
```
benthorner: @benhazell do we need all 3 of these titles?

config/locales/en/images/edit.yml
```diff
@@ -12,5 +12,8 @@ en:
         requirements: You need to
       guidance:
         alt_text_govspeak: A simple and specific description of what the image shows. This helps screen-readers and search engines.
-        caption_govspeak: This text appears on the page under the image.
-        credit_govspeak: You must have the rights to use this image. Images should be credited to their source. For example, Getty Images. Open Government Licence (OGL) images do not need to be credited.
+        caption_govspeak: Optional. A description of what the image shows. Caption appears next to the image.
+        credit_govspeak: |
+          Images licensed from a 3rd party must be credited. Government images do not need a credit.
+
+          [Full image credit guidance](https://www.gov.uk/guidance/content-design/image-copyright-standards-for-gov-uk)
```
kevindew: We should open this with `target="_blank"` as otherwise clicking it in through the modal view will leave the edit.

config/locales/en/images/crop.yml
```diff
@@ -3,3 +3,4 @@ en:
...
       title_modal: "Crop image"
+      title_exact_dimensions: "Image does not need to be cropped"
```
kevindew: This one is a little jarring for when you click edit image and then are presented with "Image does not need to be cropped" as it feels like the title and what you requested don't match.

image-crop

config/locales/en/images/edit.yml
benhazell: I agree. But I don't know how to do that with markdown. Can you show me?

config/locales/en/images/crop.yml
benhazell: I didn't realise the crop screen in the lead image flow had a different title.
I think the lead image crop and the insert image crop should be the same. Then we only need 2 titles - one for cropping the image, and one for if the image is the exact size and can't be cropped.

config/locales/en/images/edit.yml
alex-ju: `[Full image credit guidance](https://www.gov.uk/guidance/content-design/image-copyright-standards-for-gov-uk){:target="_blank"}`

alex-ju: ☝️ the moment when markdown loses its scope

config/locales/en/images/crop.yml
alex-ju: @benthorner updated to get rid of the redundant modal title

config/locales/en/images/crop.yml
benhazell: I think this is acceptable for a quick improvement. 
If the user wants to edit the image they need reminding that they can't crop. They might be trying to adjust the crop, but in this situation they need to be told they can't.

The alternative would be to have separate flow for exact size images to skip this screen when edited, but the user wouldn't know why.
Or to go back to having separate links for edit crop and image title.

The prior content, Crop Image, was worse as it told users to do something that they couldn't do.

PR #1018, opened 2019-05-09
Add coverage checker for I18n locales

benthorner:
https://trello.com/c/s122Fg91/959-we-dont-monitor-unused-and-untested-i18n-translations

This adds a GOVUKI18nCov class* that generates a coverage report of the
I18n keys used after a script (e.g. rspec) has finished running. I've
removed a couple of unused locales that the report has identified as
being unused - the remainder are used but not covered by the tests.

I'm not really sure where this code should live, but it does seem to be 
beneficial to monitor coverage of our locales :-).

*there are other tools out there, such as i18n-tasks, but I couldn't find one 
that produces a good quality results for our use case - we have a lot of 
dynamic uses of I18n in our code, as well as simple uses in the views.

### DIFF
--- a/spec/support/i18n_cov.rb
+++ b/spec/support/i18n_cov.rb
@@ -1,0 +1,80 @@
+# frozen_string_literal: true
+
+require "singleton"
+
+class I18nCov
+  REPORT_PATH = Rails.root.join("coverage", "i18n")
+  IN_APP_LOCALE_FILES = Dir.glob(Rails.root.join("config", "locales", "**", "*.yml"))
+
+  include Singleton
+
+  attr_reader :locales, :used_keys
+
+  module I18nOverrides
+    def lookup(locale, key, scope = [], options = {})
+      I18nCov.log(key, scope)
+      super
+    end
+  end
+
+  class << self
+    delegate :start, :report, :log, to: :instance
+  end
+
+  def start(locales = nil)
+    @used_keys = []
+    @locales = locales || in_app_locales
+    I18n::Backend::Simple.include I18nOverrides
+  end
+
+  def log(key, scope)
+    key = (Array(scope || []) + [key]).compact.join(".")
+    used_keys << key unless used_keys.include?(key)
+  end
+
+  def report
+    return if used_keys.blank?
+
+    all = all_locale_keys
+    scoped_used = all.select { |as| used_keys.any? { |us| as.start_with?(us) } }
+
+    write_report(all - scoped_used)
+    print_message(all, scoped_used)
+  end
+
+private
+
+  def in_app_locales
+    IN_APP_LOCALE_FILES.map(&YAML.method(:load_file)).reduce({}, :deep_merge)
+  end
+
+  def all_locale_keys
+    locales.flat_map { |locale, ts| key_chains_for(locale, ts, is_root: true) }
+      .map { |key_chain| key_chain.join(".") }.uniq
+  end
+
+  def key_chains_for(key, value, is_root: false)
+    return [[key]] unless value.is_a? Hash
+
+    value.flat_map { |sub_key, sub_value| key_chains_for(sub_key, sub_value) }
+      .map { |sub_key| (is_root ? [] : [key]) + sub_key }
+  end
+
+  def write_report(unused)
+    FileUtils.mkdir_p(File.dirname(REPORT_PATH))
+
+    File.open(REPORT_PATH, "w") do |f|
+      f.puts("Keys not covered (potentially unused)\n\n")
+      unused.each { |key| f.puts key }
+    end
+  end
+
+  def print_message(all, used)
+    percent = ((used.count.to_f / all.count) * 100).round(2)
+
+    puts "Coverage report generated for I18n to #{Dir.pwd}/coverage/i18n. " +
+      "#{used.count} / #{all.count} keys (#{percent}%) covered."
+  end
+end
+
+END { I18nCov.report }
--- a/config/locales/en/documents/history.yml
+++ b/config/locales/en/documents/history.yml
@@ -16,7 +16,6 @@ en:
         image_updated: "Updated image"
         image_deleted: "Deleted image"
         new_edition: "New edition"
-        create_preview: "Create preview"
         withdrawn: "Withdrawn"
         withdrawn_updated: "Updated withdrawal explanation"
         removed: "Removed"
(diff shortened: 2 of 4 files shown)


spec/support/govuk_i18n_cov.rb
```diff
@@ -0,0 +1,81 @@
+# frozen_string_literal: true
+
+class GOVUKI18nCov
```
kevindew: It'd be more conventional for this to be capitalised as Govuk. Though I wonder if it needs a Govuk prefix as it's not specific to us.

spec/support/govuk_i18n_cov.rb
```diff
@@ -0,0 +1,81 @@
...
+    end
+  end
+
+  class << self
+    def start
+      return if @started
+
+      @started = true
+      @used_keys = []
+      ::I18n::Backend::Simple.send :include, I18n::Registry
```
kevindew: Looks like include is a public method: https://github.com/ruby-i18n/i18n/blob/0c5dab494d9b043e00662d8e789229c33045c024/lib/i18n/backend/simple.rb#L24

spec/support/govuk_i18n_cov.rb
```diff
@@ -0,0 +1,81 @@
...
+  module I18n
+    module Registry
+      def lookup(locale, key, scope = [], options = {})
+        GOVUKI18nCov.log(locale, key, scope)
+        super
+      end
+    end
+  end
+
+  class << self
```
kevindew: Note, operating on the eigenclass was discouraged in the GOV.UK style guide before it was replaced. Hence it's seen very little in GOV.UK code. I don't mind it myself, though I wonder if using a Singleton pattern might make it a bit easier to follow as a class.

spec/support/govuk_i18n_cov.rb
```diff
@@ -0,0 +1,81 @@
...
+      if unused.any?
+        write_report(unused)
+        print_message(all, unused)
+      end
+    end
+
+  private
+
+    def get_all_i18n_keys
+      ymls = Dir.glob("config/locales/**/*.yml")
```
kevindew: Should it use Rails to access these? We configure this here: https://github.com/alphagov/content-publisher/blob/57e6867e9033cef6c2c24f88e1ded8d7569de61e/config/application.rb#L32

Though after looking this up I see that it will include all the gems too so this might be better.

spec/support/govuk_i18n_cov.rb
```diff
@@ -0,0 +1,81 @@
...
+    end
+
+    def normalise_keys(keys)
+      keys.uniq.map(&:to_s).map { |s| s.sub(/^\./, "") }
+    end
+
+    def write_report(unused)
+      FileUtils.mkdir_p("coverage")
+
+      File.open("coverage/i18n", "w") do |f|
```
kevindew: I have a recollection that you said this file was in the same format as simplecov, hopefully I'm not making that up. I notice this is a list but simplecov drops out a HTML file.

spec/support/govuk_i18n_cov.rb
```diff
@@ -0,0 +1,81 @@
+# frozen_string_literal: true
+
+class GOVUKI18nCov
+  module I18n
+    module Registry
```
kevindew: Any reason this should be two modules deep?

spec/support/govuk_i18n_cov.rb
```diff
@@ -0,0 +1,81 @@
...
+
+    def key_chains_for(key, value)
+      return [[key]] unless value.is_a? Hash
+
+      value.flat_map { |sub_key, sub_value| key_chains_for(sub_key, sub_value) }
+        .map { |sub_key| [key] + sub_key }
+    end
+
+    def normalise_keys(keys)
+      keys.uniq.map(&:to_s).map { |s| s.sub(/^\./, "") }
```
kevindew: Just out of curiosity, what happens that would mean a key needs normalising?

spec/support/govuk_i18n_cov.rb
benthorner: Fair enough - I was being cautious :-).

spec/support/govuk_i18n_cov.rb
benthorner: Nope - you find the flaws in my pasted code ;-). I've revised it now, which I think has made it a bit clearer too!

spec/support/govuk_i18n_cov.rb
benthorner: Fair enough - SimpleCov seems to be doing this, but I like it better after moving to the Singleton pattern.

spec/support/govuk_i18n_cov.rb
benthorner: You got it in one! I've hopefully made this a bit clearer, plus making it somewhat configurable to support testing.

spec/support/govuk_i18n_cov.rb
benthorner: It's possible to use I18n in a number of ways, which include using scoped and symbolic arguments, as well as specifying the optional root '.' element. That said, it looks like that's not a problem for our in-app keys, so I've removed it for now :-).

spec/support/i18n_cov.rb
```diff
@@ -0,0 +1,80 @@
...
+
+  def key_chains_for(key, value)
+    return [[key]] unless value.is_a? Hash
+
+    value.flat_map { |sub_key, sub_value| key_chains_for(sub_key, sub_value) }
+      .map { |sub_key| [key] + sub_key }
+  end
+
+  def write_report(unused)
+    FileUtils.mkdir_p("coverage")
```
kevindew: Did you mean for this to be relative to Rails.root too?

spec/support/i18n_cov.rb
```diff
@@ -0,0 +1,80 @@
...
+private
+
+  def get_in_app_locales
+    ymls = Dir.glob(Rails.root.join("config", "locales", "**", "*.yml"))
+    ymls.map { |yml| YAML.load_file(yml) }.reduce({}, :deep_merge)
+  end
+
+  def get_all_i18n_keys
+    locales.flat_map { |locale, ts| key_chains_for(locale, ts) }
+      .map { |key_chain| key_chain.drop(1).join(".") }
```
kevindew: This might need some documentation - refactor, comment, or commit message - to explain the drop of the locale part

benthorner:
@kevindew thanks for taking another look. I've done a bit more refactoring: added some constants with Rails.join, renamed some of the methods, made the scoping of which used keys we care about more prominent, and made the drop(1) a concern of the key_chains_for method, which hopefully makes it clearer, given the updated commit message :-).

benthorner:
@kevindew ah, I hadn't thought about 'default' like that - it was meant to refer to it's usage in generating the default locales they aren't specified in '.start'. I've changed it to be 'in_app', which reflects the method it's used in :-).

PR #1020, opened 2019-05-10
Attachment preview auth bypass

benthorner:
https://trello.com/c/xpS0zlU4/803-download-an-attachment-file-through-content-publisher

This refactors the EditionUrl service in order to expose the secret token it uses and generate a similar kind of 'secret preview URL' for file attachments.


app/helpers/file_attachment_helper.rb
```diff
@@ -13,6 +13,12 @@ def file_attachment_payload_attributes(attachment_revision)
     )
   end
 
+  def file_attachment_secret_preview_url(attachment_revision, document)
```
kevindew: Might not need to flag this as secret?

app/services/edition_url.rb
```diff
@@ -1,29 +1,12 @@
 # frozen_string_literal: true
 
-class EditionUrl
+class PreviewSecretService
```
kevindew: I assume you mean to delete this file?

app/services/preview_secret_service.rb
```diff
@@ -0,0 +1,33 @@
+# frozen_string_literal: true
+
+# For more info, see https://docs.publishing.service.gov.uk/manual/content-preview.html#authentication
+class PreviewSecretService
```
kevindew: I feel sad suggesting this, as it's rather fun that we have a class called secret service, but since this doesn't create something that is actually secret it might be clearer to call it PreviewAuthBypassService?

app/helpers/edition_url_helper.rb
```diff
@@ -0,0 +1,23 @@
+# frozen_string_literal: true
+
+module EditionUrlHelper
+  def edition_public_url(edition)
+    return unless edition.base_path
+
+    Plek.new.website_root + edition.base_path
+  end
+
+  def edition_secret_preview_url(edition)
```
kevindew: Since we don't have a non secret form of this I'd suggest the secret part is redundant. It might not be necessary to have the private method here too since it's a one liner.

app/views/documents/show/_actions.html.erb
```diff
@@ -86,7 +86,7 @@
       <hr class="govuk-section-break govuk-section-break--m govuk-section-break--visible" />
       <ul class="govuk-list govuk-!-margin-0">
         <li>
-          <%= link_to "View on GOV.UK", EditionUrl.new(@edition.document.live_edition).public_url, class: "govuk-link govuk-link--no-visited-state" %>
+          <%= link_to "View on GOV.UK", edition_public_url(@edition.document.live_edition), class: "govuk-link govuk-link--no-visited-state" %>
```
kevindew: This looks nicer 👍

app/services/edition_url.rb
benthorner: Oops - thanks for spotting...deleted :-).

app/services/preview_secret_service.rb
benthorner: I'm a little confused about this, so would be good to clarify here. 

My impression of the draft stack is that it is protected by some kind of auth (secret), whether that's via Signon, or via auth bypass tokens. This makes sense especially for access limiting, where you could reasonably call the content/attachment 'secret' until it's published; in other cases, perhaps it's not 'secret', but it is 'confidential' among publishers. 

I can see the logic with 'AuthBypassService', but perhaps a downside with this name is that it suggests some external operation is being performed, but in reality it's more of a generator. Although I'm not too fussed about the name, is 'secret' good enough?

kevindew: Probably worth having a quick chat about this, will try grab you shortly.

benthorner: OK. I've gone with renaming to 'AuthBypass' - I couldn't decide between them, so hopefully this makes more sense :-).

spec/helpers/edition_url_helper_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+    it "returns nil without a base_path" do
+      edition = build(:edition, base_path: nil)
+      url = edition_public_url(edition)
+      expect(url).to be_nil
+    end
+  end
+
+  describe "#preview_url" do
+    let(:preview_secret_service) do
+      instance_double("PreviewSecretService", preview_token: "secret")
```
kevindew: Looks like this needs to be updated

benthorner: Ah, thanks for spotting - I've added an extra commit to help prevent this happening again. Could you take a look?

PR #1022, opened 2019-05-10
Edit File Attachment title

emmabeynon:
For https://trello.com/c/WvCPILCl/817-provide-means-to-edit-replace-an-attachment

This enables users to edit the title of an existing file attachment.  Functionality to replace an attachment file will be implemented in a future PR.

A couple of things that would be good to get some thoughts on:
- In https://github.com/alphagov/content-publisher/commit/76e297028db5c942d4e4df8d8ab1bfa57a9e136b we've made a direct copy of the Image Revision Updater.  So far no amendments have been required to support File Attachments, so should we think about renaming it to something more generic?  Asset Revision Updater?
- The File Attachment Upload Checker wasn't quite right for editing a title only, as it requires a file to be passed to it and will log an issue if it's not present.  I've extracted the title checks to its own File Attachment Title Checker.  This will also help when it comes to future work to allow users to replace the attachment file.  I've updated the File Attachment Create Interactor to call both checkers.  Thoughts whether this is the right approach appreciated.

Screen Shot 2019-05-10 at 18 02 50
Screen Shot 2019-05-10 at 18 03 17
Screen Shot 2019-05-10 at 18 03 40
Screen Shot 2019-05-10 at 18 03 57


app/controllers/file_attachments_controller.rb
```diff
@@ -58,6 +58,13 @@ def edit
   end
 
   def update
-    redirect_to file_attachments_path
+    result = FileAttachments::UpdateInteractor.call(params: params,
+                                                   user: current_user)
+    edition, attachment_revision = result.to_h
+                                         .values_at(:edition,
+                                                    :file_attachment_revision)
```
1pretz1: Indentation here seems slightly odd to me, how about this? 
```ruby
result.to_h.values_at(:edition,
                      :file_attachment_revision)
```

emmabeynon: I was trying to avoid going too far over the max character count for the line.  That implementation goes too far 😞

app/interactors/file_attachments/update_interactor.rb
```diff
@@ -0,0 +1,51 @@
...
+private
+
+  def find_and_lock_edition
+    context.edition = Edition.lock.find_current(document: params[:document])
+  end
+
+  def find_and_update_file_attachment
+    current_attachment_revision = edition.file_attachment_revisions
+                                         .find_by!(file_attachment_id: params[:file_attachment_id])
+    attachment_params = params.require(:file_attachment).permit(:title, :file)
```
1pretz1: Would we only need to permit the `title` at this point as we are not concerned about updating the `file` in this pr?

emmabeynon: Yes, that's one I've missed out removing - good catch.

spec/features/editing_file_attachments/edit_file_attachment_title_spec.rb
```diff
@@ -30,4 +34,19 @@ def and_i_click_on_edit_file
...
+
+  def when_i_enter_a_new_attachment_title_and_click_save
+    @put_content_request = stub_publishing_api_put_content(@edition.content_id, {})
+
+    fill_in "file_attachment[title]", with: "Another title"
+    click_on "Save"
+  end
+
+  def then_i_see_the_new_title_on_the_attachment_index_page
+    expect(page).to have_link("Another title", href: preview_file_attachment_path(@edition.document, @attachment_revision.file_attachment))
```
1pretz1: This line is quite long, this makes it a bit shorter. :) 
```ruby
expect(page).to have_link("Another title", href: preview_file_attachment_path(@edition.document,
                                                                              @attachment_revision.file_attachment))
```

app/services/versioning/file_attachment_revision_updater.rb
```diff
@@ -0,0 +1,38 @@
+# frozen_string_literal: true
+
+module Versioning
+  class FileAttachmentRevisionUpdater < BaseUpdater
```
kevindew: Is the code for this class the same as ImageRevisionUpdater? if so it might be worth having a brain storm with @benthorner on approaches/naming we could use so code and tests aren't duplicated.

app/interactors/file_attachments/create_interactor.rb
```diff
@@ -24,7 +24,10 @@ def call
 private
 
   def check_for_issues
-    issues = Requirements::FileAttachmentUploadChecker.new(params[:file], params[:title]).issues
+    issues = Requirements::FileAttachmentUploadChecker.new(params[:file]).issues
+    title_issues = Requirements::FileAttachmentTitleChecker.new(params[:title]).issues
```
kevindew: I wonder if we can have this be just one checker and that checker calls the title one?

Thinking a little bit a head I think we basically have two almost identical checkers for file and title but one requires file to exist other doesn't. Could be that we're able to have two simple classes FileAttachmentCreateChecker and FileAttachmentEditChecker and they both call out to file and title checker classes

app/services/versioning/file_attachment_revision_updater.rb
emmabeynon: Yep, I've made a suggestion for this in the PR description.

app/interactors/file_attachments/create_interactor.rb
emmabeynon: I like @benthorner's suggestion to create 'upload_issues' and 'update_issues' methods.  It will probably mean the class name should be more generic - FileAttachmentChecker I guess?

kevindew: Sounds good 👍

spec/features/editing_file_attachments/edit_file_attachment_title_spec.rb
emmabeynon: That path was so long...I've assigned it to a variable :)

app/interactors/file_attachments/create_interactor.rb
emmabeynon: This is now done - let me know if it's ok.

app/controllers/file_attachments_controller.rb
benthorner: If it helps, we do [break with this convention sometimes](https://github.com/alphagov/content-publisher/blob/master/app/controllers/images_controller.rb#L55) when it's not helpful :-).

app/interactors/file_attachments/create_interactor.rb
benthorner: That looks good - only thing that might be a bit confusing is 'update_issues' sounds like we're going to update them 😆. We use the 'pre_' prefix in other places, which might help with the ambiguity here i.e. 'pre_update_issues' and 'pre_create_issues'. Sorry, should have thought of this earlier :-|.

app/controllers/file_attachments_controller.rb
1pretz1: +1 to Ben's suggestion.

app/interactors/file_attachments/update_interactor.rb
1pretz1: cool

spec/features/editing_file_attachments/edit_file_attachment_title_spec.rb
1pretz1: aha cool! :)

app/services/versioning/file_attachment_revision_updater.rb
emmabeynon: @benthorner and I have had a chat about this.  We felt that renaming the original updater might be confusing and reduce clarity over what type of revision it's for.  We also thought that two similar classes is probably fine for now; if another model comes along that has metadata and a blob and would need a similar class for itself then we could look at extracting out some kind of base metadata and blob revision updater thing and have classes for each of the three models that inherit from the base class.  That's probably overkill right now, but if that need becomes apparent in the future then we can deal with it then.  I'll add this reasoning to the commit message for posterity.

app/controllers/file_attachments_controller.rb
emmabeynon: Updated as per Ben's suggestion 👍

app/interactors/file_attachments/create_interactor.rb
emmabeynon: Fair point - now updated :)

PR #1024, opened 2019-05-13
Align image and file upload

1pretz1:
Fix visual styles on attachments. This includes:
- Update image upload page to have the guidance as hint
- Update attachment form markup
- Fix hr styling an fix `</p>` when showing attachments
- Fix style attachment meta for group listing

Trello:
https://trello.com/c/zMYmFTPx

### DIFF
--- a/app/assets/stylesheets/components/_attachment-meta.scss
+++ b/app/assets/stylesheets/components/_attachment-meta.scss
@@ -1,3 +1,15 @@
+.app-c-attachment-meta {
+  @include govuk-font(19);
+
+  margin-bottom: govuk-spacing(6);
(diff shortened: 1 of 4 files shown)


kevindew:
Code looks good - any chance of screenshot to show how it is changed?

1pretz1:
@kevindew The only visual difference would be for the group spacing when displaying a list of file attachments and the font is also standardised so the links are all the same size.

## Before

Screen Shot 2019-05-14 at 11 36 34

## After

Screen Shot 2019-05-14 at 11 07 22

Other things that have changed are some code clean up and making the forms more accessible.

PR #1025, opened 2019-05-13
Add missing rubyzip gem in production

benthorner:
https://trello.com/c/EL9jPXIl/802-validation-requirements-for-uploading-file-attachments

Previously we were unknowingly relying on this depedency to be provided
by a gem that was only installed in the dev/test environments. This adds
the gem explicitly to ensure it's available in the prod environment.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -23,6 +23,7 @@ gem "pdf-reader", "~> 2"
 gem "pg", "~> 1"
 gem "plek", "~> 2"
 gem "rinku", "~> 2"
+gem "rubyzip", "~> 1"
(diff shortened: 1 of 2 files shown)


Gemfile
```diff
@@ -23,6 +23,7 @@ gem "pdf-reader", "~> 2"
 gem "pg", "~> 1"
 gem "plek", "~> 2"
 gem "rinku", "~> 2"
+gem "rubyzip", "~> 1.2"
```
kevindew: Any reason why we care about the .2?

benthorner: What .2? ;-)

PR #1040, opened 2019-05-16
Don't timeout for modal forms that upload files

kevindew:
Trello: https://trello.com/c/gPfQ2lBT/819-interface-with-a-documents-attachments-via-a-modal

I had a quick go at seeing if there was a simple way I could stop the risk of timeouts when uploading files since we can't reasonably guess whether the time spent is waiting on the client side or server side.

This was the least invasive/complex way I could think of. May be too simple, comments/suggestions welcome.

### DIFF
--- a/app/assets/javascripts/modal/modal-fetch.js
+++ b/app/assets/javascripts/modal/modal-fetch.js
@@ -5,7 +5,7 @@ window.ModalFetch.getLink = function (item) {
   var headers = { 'Content-Publisher-Rendering-Context': 'modal' }
   var options = { credentials: 'include', signal: controller.signal, headers: headers }
   var href = item.dataset.modalActionUrl || item.href
-  setTimeout(function () { controller.abort() }, 5000)
+  setTimeout(function () { controller.abort() }, 15000)
 
   return window.fetch(href, options)
     .then(function (response) {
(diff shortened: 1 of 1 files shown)


1pretz1:
@benthorner I've removed the comment replacing it with a function and increased the timeout to 15 seconds matching the nginx timeout limit as per your recommendation. :)

app/assets/javascripts/modal/modal-fetch.js
```diff
@@ -23,7 +23,10 @@ window.ModalFetch.getLink = function (item) {
 
 window.ModalFetch.postForm = function (form) {
   var controller = new window.AbortController()
-  setTimeout(function () { controller.abort() }, 10000)
+
+  if (window.ModalFetch._notFileUpload(form)) {
```
benthorner: This is looking better, but I think it would be more idiomatic to use '!' in the if statement, rather than 'not' in the method name. We could say 'isFileUpload' instead.

benthorner: We might also want to consider using the same timeout for 'getLink' :-).

1pretz1: Sure, I originally wrote it with the '!' statement but thought it may not be that obvious when it's prefixed onto the long method call.

1pretz1: And sure i'll increase the timeout for getLink too!

1pretz1:
@benthorner Great, I have all the powers now... 🔌

PR #1047, opened 2019-05-17
Fix image spec not passing on CI

benthorner:
This adds a selector check to make Capybara wait before trying to edit
    metadata for an inline image, where the page may not have finished
    loading asynchronously.


app/services/contacts_service.rb
```diff
@@ -2,6 +2,7 @@
 
 class ContactsService
   CACHE_OPTIONS = { expires_in: 15.minutes, race_condition_ttl: 30.seconds }.freeze
+
```
1pretz1: Did you mean to add a new line here?

benthorner: It's vital to the PR ;-). No, it was just to get a PR up to verify the issue. It's gone now, so I'll merge...

PR #1059, opened 2019-05-22
Insert file attachments via modal

emmabeynon:
For https://trello.com/c/gPfQ2lBT/819-interface-with-a-documents-attachments-via-a-modal

This enables users with Javascript enabled to auto-insert file attachment markdown via the modal.
I've tested inserting the markdown before text, after text and within text for both attachment types. All renders correctly bar the attachment link when inserted before text, however this is being dealt with here https://github.com/alphagov/govspeak/pull/154

Screen Shot 2019-05-22 at 14 46 51

### DIFF
--- a/app/assets/javascripts/modal/modal-editor.js
+++ b/app/assets/javascripts/modal/modal-editor.js
@@ -1,0 +1,12 @@
+function ModalEditor ($module) {
+  this.$module = $module
+  this.$editor = this.$module.closest('[data-module="markdown-editor"]')
+}
+
+ModalEditor.prototype.insertBlock = function (data) {
+  this.$editor.selectionReplace(data, { surroundWithNewLines: true })
+}
+
+ModalEditor.prototype.insertInline = function (data) {
+  this.$editor.selectionReplace(data)
+}
--- a/app/assets/javascripts/modal/modal-workflow.js
+++ b/app/assets/javascripts/modal/modal-workflow.js
@@ -11,9 +11,9 @@ ModalWorkflow.prototype.initComponents = function () {
 
 ModalWorkflow.prototype.overrideActions = function (actions) {
   var formItems = this.$modal.querySelectorAll('form[data-modal-action]')
-  var linkItems = this.$modal.querySelectorAll('a[data-modal-action]')
+  var clickItems = document.querySelectorAll('a[data-modal-action], button[data-modal-action]')
 
-  linkItems.forEach(function (item) {
+  clickItems.forEach(function (item) {
     item.addEventListener('click', function (event) {
       event.preventDefault()
       actions(item)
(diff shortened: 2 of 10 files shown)


app/assets/javascripts/modal/modal-workflow.js
```diff
@@ -11,9 +11,10 @@ ModalWorkflow.prototype.initComponents = function () {
 
 ModalWorkflow.prototype.overrideActions = function (actions) {
   var formItems = this.$modal.querySelectorAll('form[data-modal-action]')
-  var linkItems = this.$modal.querySelectorAll('a[data-modal-action]')
+  var linkItems = Array.prototype.slice.call(this.$modal.querySelectorAll('a[data-modal-action]'))
```
benthorner: I think we can do both at the same time, with `this.$modal.querySelectorAll('a[data-modal...], button[data-modal...]')`.

https://stackoverflow.com/questions/37552933/concat-two-nodelists

app/assets/javascripts/modules/inline-attachment-modal.js
```diff
@@ -36,6 +36,16 @@ InlineAttachmentModal.prototype.renderSuccess = function (result) {
   this.workflow.initComponents()
 }
 
+InlineAttachmentModal.prototype.insertAttachmentBlockSnippet = function (item) {
```
benthorner: It looks like inserting a snippet has become pretty standard among the modal modules, so I wonder if it's worth refactoring this into 'insertInline' and 'insertBlock' functions in a new 'ModalSnippet' utility - it doesn't really fit into 'ModalWorkflow', as we could have modals that don't insert snippets.

What do you think?

app/assets/javascripts/modal/modal-editor.js
```diff
@@ -0,0 +1,15 @@
+function ModalEditor ($module) {
+  this.$module = $module
+  if (this.$module == null) {
+    return
+  }
+  this.$editor = this.$module.closest('[data-module="markdown-editor"]')
```
benthorner: I like that we only have to do this once now :-).

app/assets/javascripts/modal/modal-editor.js
```diff
@@ -0,0 +1,15 @@
+function ModalEditor ($module) {
+  this.$module = $module
+  if (this.$module == null) {
```
benthorner: We could avoid the utility having to know to do this check by moving where we construct it in the consuming modules. I took this approach in [my refactoring PR](https://github.com/alphagov/content-publisher/pull/1061/files#diff-68f3cc805b20202812678437322d8725R11).

emmabeynon: Good idea - I've got rid of these and now initialize ModalEditor in the init function for each modal JS file.

1pretz1:
All looks good to me! My only comment would be that if we have a no js spec for inserting an inline attachment then we should probably have one for images too but thats not concerning this pr! :) ⭐️

benthorner:
@1pretz1 well spotted with the missing spec. We planned to add it, but forgot - @emmabeynon, do you think you could do a quick PR for this?

PR #1061, opened 2019-05-22
Refactor rendering/actions into modal workflow

benthorner:
https://trello.com/c/NwWjLtlX/724-user-can-embed-attachment-links-inline-attachment

Previously we duplicated some of the render and action code in each
modal module. It now appears three times, which is good evidence it
needs DRY-ing up. This refactors the render and action code into the
existing ModalWorkflow utility to avoid duplication.

Originally it was unclear how portable the code we wrote for inline
images would be, so leaving it in each modal module allowed it to be
easily customised in the future, which a refactoring like this could
make harder. But this flexibility has not been useful, so there's no
strong argument for continuing to duplicate the code between modules.

### DIFF
--- a/app/assets/javascripts/modal/modal-workflow.js
+++ b/app/assets/javascripts/modal/modal-workflow.js
@@ -1,7 +1,11 @@
 /* global $ */
 
-function ModalWorkflow ($modal) {
+function ModalWorkflow ($modal, actionCallback) {
   this.$modal = $modal
+  this.actionCallback = actionCallback
+
+  this.$multiSectionViewer = this.$modal
+    .querySelector('[data-module="multi-section-viewer"]')
 }
 
 ModalWorkflow.prototype.initComponents = function () {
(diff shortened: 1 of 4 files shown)


1pretz1:
Nice work refactoring! 

All looks good to me, my only comment would be concerning the methods both called `performAction` in ModalWorkflow and its respective modal class (e.g InlineAttachmentModal). This is a little confusing to me when reading through, maybe in the ModalWorkflow it could be renamed to `displayModal` or something that better reflects the behaviour of the method? :)

benthorner:
@1pretz1 that was a concern I had when I made the changes, since as you say, we now have two different kinds of method with the same name. However, the ModalWorkflow one doesn't really display the modal; it just contains some common code.

I was thinking about keeping the code duplicated, but then each module has to have a MultiSectionViewer again. How about 'actionCallback' for each of the modal modules? (This matches the parameter name in ModalWorkflow.)

1pretz1:
@benthorner Yep your suggestion sounds good to me and will make the code easier to understand! :)

benthorner:
@1pretz1 updated - how's it looking now?

1pretz1:
@benthorner LGTM! ⭐️

PR #1076, opened 2019-05-29
added attachments and file type guidance

benthorner:
https://trello.com/c/TI9uPeMu/855-design-and-copy-review-of-attachments-flow

This iterates the user journey when uploading attachments that are not in the list of supported types. We now link through to updated guidance that explains the supported types in detail, so it's clear which types are supported and which are preferred.


app/views/publisher_information/how_to_use_publisher.govspeak.md
```diff
@@ -45,6 +45,26 @@ Read full [guidance on markdown](https://www.gov.uk/guidance/how-to-publish-on-g
...
 
+##Attachments
+
+Attachments are files the user can download. You can upload file attachments and add link to them from the body text. The link can be either a simple text link or a larger icon.
+
+Use the Insert menu to upload files and add markdown which creates a link to the file. 
+
+All file attachments [must be accessible](https://www.gov.uk/guidance/how-to-publish-on-gov-uk/accessible-pdfs), or there must be an accessible version available.
+
+All file attachments must be published in an [open standards file format](https://www.gov.uk/guidance/content-design/planning-content#open-formats). These are file types which do not require specific software.
```
benthorner: @benhazell should we say 'proprietary' or 'paid-for' instead of 'specific'? Open formats still require 'specific' software to open them, it's just you don't have to pay for it if you don't want to.

benhazell: Good point. I'm asking the new open formats rep to review. 
I didn't use 'proprietary' as not plain english, and I'm not keen on 'paid-for' because lots of people get MS products free or included and don't always see it as something paid for.

benthorner:
@kevindew good point - I've changed back to href :-).

PR #1078, opened 2019-05-29
Make publishing requirement prompt target accurate

Tetrino:
Adjusted links found in the warning message box of a draft document such that they bring the user to the required location (e.g. the content edit field of the document) for further editing.

Changed the ID for the govuk content field to refer to revision_field.id instead of just field.id so we avoid conflicts in the use of anchors. Updated related tests that were affected by this.

Trello Card: https://trello.com/c/Ml6qt5on/845-make-publishing-requirement-prompt-target-accurate

### DIFF
--- a/app/views/documents/edit.html.erb
+++ b/app/views/documents/edit.html.erb
@@ -16,7 +16,7 @@
     gtm: "edit-document-submit",
   }) do |f| %>
   <div class="govuk-grid-row">
-    <div class="govuk-grid-column-two-thirds">
+    <div class="govuk-grid-column-two-thirds" id="title">
       <%= render "govuk_publishing_components/components/textarea", {
         label: {
           text: t("documents.edit.form_labels.title"),
@@ -65,7 +65,7 @@
   </div>
 
   <div class="govuk-grid-row">
-    <div class="govuk-grid-column-two-thirds">
+    <div class="govuk-grid-column-two-thirds" id="summary">
       <%= render "govuk_publishing_components/components/textarea", {
         label: {
           text: t("documents.edit.form_labels.summary"),
--- a/app/views/documents/edit/_change_notes.html.erb
+++ b/app/views/documents/edit/_change_notes.html.erb
@@ -12,7 +12,7 @@
     } %>
 <% end %>
 <div class="govuk-grid-row">
-  <div class="govuk-grid-column-two-thirds">
+  <div class="govuk-grid-column-two-thirds" id="change-note">
     <% legend_text = tag.p(
       tag.strong(t("documents.edit.update_type.title")),
       class: "govuk-!-margin-bottom-3")
--- a/app/views/documents/fields/_govspeak_input.html.erb
+++ b/app/views/documents/fields/_govspeak_input.html.erb
@@ -1,5 +1,5 @@
 <div class="govuk-grid-row">
-  <div class="govuk-grid-column-two-thirds">
+  <div class="govuk-grid-column-two-thirds" id="<%= field.id %>">
     <%= render "components/markdown_editor", {
       label: {
         text: field.label,
@@ -11,7 +11,7 @@
           gtm: "#{field.id}-input",
           "gtm-copy-paste-tracking": true,
         },
-        id: field.id,
+        id: "revision_#{field.id}",
         name: "revision[contents][#{field.id}]",
         value: revision.contents[field.id],
         rows: 20,
--- a/app/views/documents/show/_requirements.html.erb
+++ b/app/views/documents/show/_requirements.html.erb
@@ -1,13 +1,13 @@
 <% issue_params = {
   style: "summary",
   link_options: {
-    title: { href: "#content" },
-    summary: { href: "#content" },
-    body: { href: "#content" },
-    change_note: { href: "#content" },
+    title: { href: edit_document_path(@edition.document, anchor: "title") },
+    summary: { href: edit_document_path(@edition.document, anchor: "summary") },
+    body: { href: edit_document_path(@edition.document, anchor: "body") },
+    change_note: { href: edit_document_path(@edition.document, anchor: "change-note") },
     alt_text: { href: "#lead-image" },
     caption: { href: "#lead-image" },
-    topics: { href: "#topics" }
+    topics: { href: update_topics_path(@edition.document) },
   }
 } %>
 
--- a/spec/features/editing_content/insert_inline_file_attachment_spec.rb
+++ b/spec/features/editing_content/insert_inline_file_attachment_spec.rb
@@ -57,13 +57,13 @@ RSpec.feature "Insert inline file attachment", js: true do
     expect(page).to_not have_selector(".gem-c-modal-dialogue")
     snippet = I18n.t("file_attachments.show.attachment_markdown",
                      filename: @file_attachment_revision.filename)
-    expect(find("#body").value).to match snippet
+    expect(find("#revision_body").value).to match snippet
   end
 
   def then_i_see_the_attachment_link_snippet_is_inserted
     expect(page).to_not have_selector(".gem-c-modal-dialogue")
     snippet = I18n.t("file_attachments.show.attachment_link_markdown",
                      filename: @file_attachment_revision.filename)
-    expect(find("#body").value).to match snippet
+    expect(find("#revision_body").value).to match snippet
   end
 end
--- a/spec/features/editing_content/insert_inline_image_spec.rb
+++ b/spec/features/editing_content/insert_inline_image_spec.rb
@@ -37,6 +37,6 @@ RSpec.feature "Insert inline image", js: true do
 
   def then_i_see_the_snippet_is_inserted
     snippet = I18n.t("images.index.meta.inline_code.value", filename: @image_revision.filename)
-    expect(find("#body").value).to match snippet
+    expect(find("#revision_body").value).to match snippet
   end
 end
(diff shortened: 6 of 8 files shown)


app/views/documents/fields/_govspeak_input.html.erb
```diff
@@ -1,5 +1,5 @@
 <div class="govuk-grid-row">
-  <div class="govuk-grid-column-two-thirds">
+  <div class="govuk-grid-column-two-thirds" id="body">
```
kevindew: It looks like this conflicts with the id used for the label, see the test fails here: https://ci.integration.publishing.service.gov.uk/job/content-publisher/job/JS-prompt-targets/2/console

It's worth checking whether other ids you have created conflict with existing ones.

app/views/documents/show/_requirements.html.erb
```diff
@@ -1,13 +1,15 @@
...
   style: "summary",
-  link_options: {
-    title: { href: "#content" },
-    summary: { href: "#content" },
-    body: { href: "#content" },
-    change_note: { href: "#content" },
-    alt_text: { href: "#lead-image" },
-    caption: { href: "#lead-image" },
-    topics: { href: "#topics" }
+  hrefs: {
```
kevindew: It looks like this have overwritten the recent change to switch to link options. See https://github.com/alphagov/content-publisher/pull/1076/commits/32027cfde1e495f9fee337b1a720ea618c8963ab

Aside from the overwrite they look good though 👍

app/views/documents/show/_requirements.html.erb
```diff
@@ -1,13 +1,15 @@
...
+    title: edit_document_path(anchor: "title"),
+    summary: edit_document_path(anchor: "summary"),
+    body: edit_document_path(anchor: "body"),
+    change_note: edit_document_path(anchor: "changenote"),
+    #Come back to this at some point so we could link directly to the image?
+    alt_text: images_path(@edition.document, anchor:"imagelist"),
+    caption: images_path(@edition.document, anchor:"imagelist"),
+    #--
+    topics: update_topics_path,
```
kevindew: I notice that this one and the `edit_documents_path` don't receive document as an argument whereas `images_path` does. I assume the ones with a document are working based on some form of magic context in Rails.

I think it'd be better if these consistently took the document argument as that seems to be the most common precedent in the app.

app/views/documents/show/_requirements.html.erb
```diff
@@ -1,13 +1,15 @@
...
-    topics: { href: "#topics" }
+  hrefs: {
+    title: edit_document_path(anchor: "title"),
+    summary: edit_document_path(anchor: "summary"),
+    body: edit_document_path(anchor: "body"),
+    change_note: edit_document_path(anchor: "changenote"),
+    #Come back to this at some point so we could link directly to the image?
+    alt_text: images_path(@edition.document, anchor:"imagelist"),
+    caption: images_path(@edition.document, anchor:"imagelist"),
+    #--
```
kevindew: I'd have this just remain linking to lead image until we have a strategy for these, rather needing to leave a todo comment. 

We do have precedent for leaving TODO and FIXME comments, so if you do need to do them it should follow the existing format: https://github.com/alphagov/content-publisher/blob/04e2640b0092f930e058c1502a1a83a900b436af/app/controllers/editions_controller.rb#L7-L8 but in general we discourage them.

Also minor but you missed space characters after the colon of the hash in both of those.

kevindew:
I opened https://github.com/alphagov/content-publisher/pull/1086 for the image requirements

app/views/documents/edit/_change_notes.html.erb
```diff
@@ -12,7 +12,7 @@
     } %>
 <% end %>
 <div class="govuk-grid-row">
-  <div class="govuk-grid-column-two-thirds">
+  <div class="govuk-grid-column-two-thirds" id="changenote">
```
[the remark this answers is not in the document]
Tetrino: Shall update.

app/views/documents/edit.html.erb
```diff
@@ -16,7 +16,7 @@
     gtm: "edit-document-submit",
   }) do |f| %>
   <div class="govuk-grid-row">
-    <div class="govuk-grid-column-two-thirds">
+    <div class="govuk-grid-column-two-thirds" id="edit-document-title">
```
kevindew: Do these need to be as long as they are? It looks like we're now set up for pretty widespread collision prevention but I think there was only a couple of collisions.

kevindew: Doing some quick searching it looks like only the "body" id is used in places.

```
➜  content-publisher git:(attachment-upload) ✗ ag '#body'
spec/features/editing_images/upload_image_spec.rb
88:    expect(find("#body").value).to match snippet

spec/features/editing_content/insert_inline_image_spec.rb
40:    expect(find("#body").value).to match snippet

spec/features/editing_content/insert_video_embed_spec.rb
37:    expect(find("#body").value).to match snippet

spec/features/editing_content/insert_inline_file_attachment_spec.rb
60:    expect(find("#body").value).to match snippet
67:    expect(find("#body").value).to match snippet
```

We could just change the textarea id to be `#body-field` to make that id free for the anchor.

app/views/documents/fields/_govspeak_input.html.erb
```diff
@@ -1,5 +1,5 @@
 <div class="govuk-grid-row">
-  <div class="govuk-grid-column-two-thirds">
+  <div class="govuk-grid-column-two-thirds" id="edit-document-body">
```
kevindew: Oops something I forgot last review. This field is dynamic, it gets it's name from `field.id`

app/views/documents/show/_requirements.html.erb
```diff
@@ -1,13 +1,13 @@
...
-    change_note: { href: "#content" },
-    alt_text: { href: "#lead-image" },
-    caption: { href: "#lead-image" },
-    topics: { href: "#topics" }
+    title: { href: edit_document_path(@edition.document, anchor: "edit-document-title") },
+    summary: { href: edit_document_path(@edition.document, anchor: "edit-document-summary") },
+    body: { href: edit_document_path(@edition.document, anchor: "edit-document-body") },
+    change_note: { href: edit_document_path(@edition.document, anchor: "edit-document-change-note") },
+    alt_text: @edition.lead_image_revision == nil ? nil : { href: update_image_path(@edition.document, @edition.lead_image_revision.image_id, wizard: params[:wizard], anchor: "edit-image-alt-text") },
+    caption: @edition.lead_image_revision == nil ? nil : { href: update_image_path(@edition.document, @edition.lead_image_revision.image_id, wizard: params[:wizard], anchor: "edit-image-caption") },
```
kevindew: Ah nicely done!

You can also do `@edition.lead_image_revision.nil?` rather than needing an equality check. To be honest though, since https://github.com/alphagov/content-publisher/pull/1086 will resolve these issues I'd cut them out and have this PR not handle any images.

app/views/documents/edit.html.erb
Tetrino: Is there a specific issue with long ids? Or this just a standards thing?

app/views/documents/fields/_govspeak_input.html.erb
Tetrino: @kevindew The field does, but I'm ID'ing to the containing frame rather than the field itself, to avoid these issues. Shouldn't need changing here, just shortening.

app/views/documents/edit.html.erb
kevindew: Just a question of why make them long if we don't need to - particularly as these will form part of a URL whereas most ids a user never sees so can be long and ugly without a user ever seeing.

app/views/documents/fields/_govspeak_input.html.erb
kevindew: I mean that this partial is intended to be used by fields other than "body"

kevindew:
So from our quick chat just now things left to do for this would be:

- [ ] Fix commit message (currently has a trello card title)
- [ ] Trello card just in Pull Request description
- [ ] title one not id of title-field instead title and same for others
- [ ] change https://github.com/alphagov/content-publisher/blob/1d011d0aa4313e8defb0118d1b6b204dbc123a1a/app/views/documents/fields/_govspeak_input.html.erb#L2 to reference field.id
- [ ] change https://github.com/alphagov/content-publisher/blob/1d011d0aa4313e8defb0118d1b6b204dbc123a1a/app/views/documents/fields/_govspeak_input.html.erb#L14 to be "revision_#{field.id}”
- [ ] Fix tests that break because they rely on `#body` and instead will need `#revision_body`
- [ ] remove ids from: https://github.com/alphagov/content-publisher/pull/1078/files#diff-d29dcf175afecc3295b6b48c8e2779ae these are covered in https://github.com/alphagov/content-publisher/pull/1086

app/views/documents/fields/_govspeak_input.html.erb
```diff
@@ -1,5 +1,5 @@
 <div class="govuk-grid-row">
-  <div class="govuk-grid-column-two-thirds">
+  <div class="govuk-grid-column-two-thirds" id="<%= "#{field.id}" %>">
```
kevindew: Oh shoot I told you this totally wrong earlier didn't I? Anyway very minor but this doesn't need interpolation It can be: `<%= field.id %>`

Tetrino: Done

PR #1079, opened 2019-05-29
Refactor actions into helper methods

benthorner:
https://trello.com/c/B9rVHJTh/876-audit-and-fix-gtm-issues

This was a pain point when I attempted to add some data-gtm attrs...

Previously we had a long and complex 'if' statement to determine which
actions a user can perform on a document. Much of the length of the
statement was due to (duplicated) rendering of elements, passing various
options. This refactors all the rendering into a new 'ActionsHelper' and
an 'external_links' partial, leaving just a complex 'if' statement ;-).

### DIFF
--- a/app/helpers/actions_helper.rb
+++ b/app/helpers/actions_helper.rb
@@ -1,0 +1,76 @@
+# frozen_string_literal: true
+
+module ActionsHelper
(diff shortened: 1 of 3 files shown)


benthorner:
@kevindew oh yeah - thanks for spotting. I've cleaned it up now to pass edition 👍

PR #1080, opened 2019-05-29
Standardise copy/paste GTM attributes

benthorner:
https://trello.com/c/B9rVHJTh/876-audit-and-fix-gtm-issues

Previously we enabled tracking for copy/paste events using a particular
'data-gtm-copy-paste-tracking' attribute. In some cases, we also set the
'data-gtm' attribute, which we conventionally use the identify the
element. This ensures 'data-gtm' is populated in all places where we
track copy/paste events, and replaces the (now) redundant value of the
'data-gtm-copy-paste-tracking' attribute with 'true' so it still works.


benthorner:
True - I was thinking we could just carry on sending it, although I don't think it was ever used, either, so would be good to get rid of. Updated 👍

PR #1086, opened 2019-05-29
Links to image requirement issues

kevindew:
Trello: https://trello.com/c/Ml6qt5on/845-make-publishing-requirement-prompt-target-accurate

As a companion to https://github.com/alphagov/content-publisher/pull/1078 this adds in the links for image requirement issues. These were somewhat tricky as the URLs require the id of the image which isn't readily available. To resolve this I have made use of the context hash on an Issue and allowed passing in a lambda to generate the link.

May-29-2019 19-14-36

This also fixes a couple of issues I spotted along the way, such as missing credit link and errors. 

For further details see commit messages.

### DIFF
--- a/app/views/images/edit.html.erb
+++ b/app/views/images/edit.html.erb
@@ -88,6 +88,7 @@ content_for :title, t(title_key, title: @edition.title_or_fallback)
         },
         name: "image_revision[credit]",
         value: @image_revision.credit,
+        error_items: @issues&.items_for(:credit),
         data: {
           "contextual-guidance": "credit-guidance"
         }
--- a/app/views/images/edit.html.erb
+++ b/app/views/images/edit.html.erb
@@ -28,7 +28,7 @@ content_for :title, t(title_key, title: @edition.title_or_fallback)
   }
 ) do %>
   <div class="govuk-grid-row">
-    <div class="govuk-grid-column-two-thirds">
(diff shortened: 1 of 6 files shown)


app/views/images/edit.html.erb
```diff
@@ -28,7 +28,7 @@ content_for :title, t(title_key, title: @edition.title_or_fallback)
   }
 ) do %>
   <div class="govuk-grid-row">
-    <div class="govuk-grid-column-two-thirds">
+    <div class="govuk-grid-column-two-thirds" id="alt-text">
```
alex-ju: We should assign these IDs to the input component, not to presentation containers. By doing so we also get an automated focus on the specified field and, in our case, the associated guidance appearing on the right hand side.

kevindew: We didn’t do that as you can’t see the label, I hadn’t realised about the focus though. 

This suggestion has consequences for https://github.com/alphagov/content-publisher/pull/1078 too, so I’ll catch up with you about it later

alex-ju: The component is build so that the label is displayed when the ID of the associated input is focus for this exact scenario.

kevindew: As discussed IRL - looks like there is a bug that stops the scroll behaviour that you're hoping for, so we concluded this is the best user experience for now.

Alex may work on resolving this later.

PR #1088, opened 2019-05-30
Update autocomplete to allow narrow width needed for time select

alex-ju:
Update autocomplete to support a narrow version witch matches the date input component width so it can be used for time select.
Use tag helpers to generate markup for the autocomplete component.

[Trello card](https://trello.com/c/rGjDviSL)

### DIFF
--- a/app/views/components/_autocomplete.html.erb
+++ b/app/views/components/_autocomplete.html.erb
@@ -6,21 +6,20 @@
   type ||= nil
   size ||= nil
   hint ||= nil
+  width ||= nil
+  css_classes = %w(app-c-autocomplete govuk-form-group)
+  css_classes << "app-c-autocomplete--#{width}" if width
 %>
 
-<div class="app-c-autocomplete govuk-form-group" data-module="autocomplete" data-autocomplete-type="<%= type %>">
+<%= tag.div class: css_classes, data: { module: "autocomplete", "autocomplete-type": type } do %>
   <%= render "govuk_publishing_components/components/label", {
     html_for: id
   }.merge(label.symbolize_keys) %>
 
   <% if options.any? %>
-    <% if multiple %>
-      <span class="govuk-hint app-c-autocomplete__multiselect-instructions">To select multiple items in a list, hold down Ctrl (PC) or Command (Mac) key.</span>
-    <% end %>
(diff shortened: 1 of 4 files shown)


app/views/components/_autocomplete.html.erb
```diff
@@ -6,21 +6,20 @@
   type ||= nil
   size ||= nil
   hint ||= nil
+  width ||= nil
+  css_classes = %w(app-c-autocomplete govuk-form-group)
+  css_classes << "app-c-autocomplete--#{width}" if width
```
kevindew: could this just be called classes? I feel like the css prefix is redundant

alex-ju: It's modifier class, which means that it should be marked as such using the `--narrow` suffix according to the adopted BEM convention. We could explicitly have `app-c-autocomplete--narrow` instead of a variable passed within.

kevindew:
Oh heroku is broken due to: 

```
2019-05-30T14:41:28.554342+00:00 heroku[router]: at=info code=H82 desc="Free app running time quota exhausted" method=GET path="/" host=content-publisher-revi-pr-1090.herokuapp.com request_id=0785553d-4a46-42e0-b4a7-a4051afa97b8 fwd="81.152.8.99" dyno= connect= service= status=503 bytes= protocol=https
2019-05-30T14:41:28.925826+00:00 heroku[router]: at=info code=H82 desc="Free app running time quota exhausted" method=GET path="/favicon.ico" host=content-publisher-revi-pr-1090.herokuapp.com request_id=ad9c6b40-bc3b-4c29-8706-3436311c956a fwd="81.152.8.99" dyno= connect= service= status=503 bytes= protocol=https
```

I guess that means it'll be back tomorrow.

app/views/components/_autocomplete.html.erb
kevindew: I'm not sure we're talking about the same thing. 

I was implying that `css_classes` as a variable name might want to change as they're just classes for the HTML element and we don't need to differentiate them from other classes.

alex-ju: Ah, I got that wrong. We currently use a mix of `component_classes` and `css_classes` within our applications and a mix of `css_classes` and `[component_name]_classes` in govuk_publishing_components. Not very opinionated to be honest, but if we decide on one form I would update (at least the application level) for consistency.

kevindew: Sure, well if css_classes is already consistent then that's cool. I'd be voting we don't use that terminology though.

PR #1089, opened 2019-05-30
Retire FileAttachmentUploadService and refactor FileAttachmentBlobService

kevindew:
Trello: https://trello.com/c/BQ5bBkMw/881-wrap-up-technical-loose-ends-on-attachments

This removes the FileAttachmentUploadService and instead handles the
concerns of that in a create_initial method on FileAttachment::Revision.
This is to reflect that FileAttachmentUploadService no longer is
responsible for uploading and it's name was becoming redundant.

This also renames FileAttachmentBlobService to
FileAttachmentBlobAttributesService and changes it's interface from
being something that acts like a model into a service that returns a
hash of attributes - hence the name change.


app/interactors/file_attachments/update_interactor.rb
```diff
@@ -76,14 +76,9 @@ def attachment_params
   def blob_attributes(file_attachment_revision)
```
benthorner: I think this param is provided by the context.

app/interactors/file_attachments/create_interactor.rb
```diff
@@ -34,11 +34,14 @@ def find_and_lock_edition
   end
 
   def upload_attachment
-    context.attachment_revision = FileAttachmentUploadService.new(
-      params[:file],
-      edition.revision,
+    blob_attributes = FileAttachmentBlobAttributesService.new(file: params[:file],
+                                                              revision: edition.revision)
+                                                         .call
+    context.attachment_revision = FileAttachment::Revision.create_initial(
```
benthorner: It might be a little clearer to have named params here, since the first two are a bit of a mix.

app/interactors/file_attachments/update_interactor.rb
```diff
@@ -76,14 +76,9 @@ def attachment_params
...
     return {} unless attachment_params[:file]
```
benthorner: Do you think this should be a concern of the service?

app/interactors/file_attachments/update_interactor.rb
kevindew: It's not set at that point, which avoids setting it twice (once as the current revision, second as the updated one) - but I think we can avoid the need for this method call to hopefully make this moot.

app/interactors/file_attachments/create_interactor.rb
kevindew: Sounds good - it's inconsistent with the other create_initials but then the interfaces aren't the same so I think that's fine.

app/interactors/file_attachments/update_interactor.rb
kevindew: Yeah I think we can move that there.

kevindew:
> Thanks for pursuing this :-). I think the create_initial approach works well, since it re-uses an existing concept and removes an ambiguous one. The blob service is still surprising me each time I read it, which I'll try and articulate:

Thanks for the speedy review, you raise some interesting points. 

> * Some of the parameters are unexpected for the generic constructor/call combo.

It makes me sad that we have two interface points in these cases where we have a single purpose class. I don't ever see there being much consistency in our logic of what goes on `initialize` or `call`, though if we were going to spend energy on it I'd much rather we worked towards just a single static method to cut out the two places.

> * The name of the service still doesn't match my understanding of what it achieves (upload).

It's a funny thing really the service. It isn't uploading really, it's the conversion of an uploaded file into objects. So I guess the closest we've come to this having a clear role is when it was almost a BlobRevision factory. 

As that was a bit grim for widening scope then this creates the attributes needed to update a BlobRevision. 

> 
> More helpfully, I wonder if actually 'BlobService' was OK as a name, and we should actually be distinguishing between the two different (but strongly related) uses of the service: upload and replace. The first thing that put me onto this was noticing that the 'replacement' constructor param changes the behaviour of the service; if we were to explicitly tell it to do '#replace', then that would also be an opportunity to separate required vs. optional args, which might help with the above point.

Hmm I don't think upload and replace are the two distinct things as both of those involve uploading. It's more create and replace.

I'm not too keen on having a replace method though as the service itself isn't replacing anything, I'm hoping that intention can be reflected better with a parameter.

I've come up with this naming option here: `FileAttachmentBlobService.new(revision).attributes_for(file:, replacing: nil)` 

Which I think reflects the dealings with a blob, that you get attributes back and ability to specify it's a replacement syntactically.

Let me know what you think?

Some options I rejected are:

```
FileAttachmentBlobService.create(file: file, for: revision)
FileAttachmentBlobService.update(file: file, for: revision, replacing: file_attachment_revision) 
```

rejected as not actually updating anything and both seem to create

```
FileAttachmentBlobService.upload(file: file, for: revision, replacing: file_attachment_revision) 
```

rejected as we're not actually doing an upload.

Also, damn you for making me think so hard about this 😢

benthorner:
Fair enough - I'm a little confused about why 'create_after_upload!' doesn't upload anything, but as you suggest, there are other suitable verbs. The more I think about this in conjunction with the 'create_initial' methods, the more I prefer the factory (or suitable verb) idea. 

I reckon it's hard to think of an intuitive interface because the concept is unclear - it's not even a 'Params' processor, like we've thought about previously, since it does more than that. The word 'attributes' is also tricky, because we're passing in attributes, but also getting out attributes.

Would it be easier to think about this if we admit a new concept as one of the solutions? We're kind of doing that already with this service, as there's no precedent for it, so we may as well consider all options and the factory one seems like the one with the most evidence to support it existing.

Sorry about the pain :-|.

kevindew:
> Fair enough - I'm a little confused about why 'create_after_upload!' doesn't upload anything, but as you suggest, there are other suitable verbs. The more I think about this in conjunction with the 'create_initial' methods, the more I prefer the factory (or suitable verb) idea.

that method may upload to s3 or write to the local file system - environment dependent. I think rather than upload it's better to think of it as saving a file. As we're both persisting the file to s3 and writing a record in the database.

> I reckon it's hard to think of an intuitive interface because the concept is unclear - it's not even a 'Params' processor, like we've thought about previously, since it does more than that. The word 'attributes' is also tricky, because we're passing in attributes, but also getting out attributes.

Yup I'd agree, the whole thing makes sense as a utility function but doesn't feel like elegant object orientation.

> Would it be easier to think about this if we admit a new concept as one of the solutions? We're kind of doing that already with this service, as there's no precedent for it, so we may as well consider all options and the factory one seems like the one with the most evidence to support it existing.

We're at risk of blowing the timebox by broadening the scope. I think I'm pretty close to an approach that offers a net improvement and I think we should look to get that merged with only small tweaks - then address further if there's pain introduced or a complementary pattern develops.

kevindew:
Closing in favour of https://github.com/alphagov/content-publisher/pull/1092

PR #1090, opened 2019-05-30
Stop scheduled publishing

emmabeynon:
For https://trello.com/c/FndkZ8gG/843-allow-users-to-stop-a-scheduled-publishing

This extracts the existing unschedule controller action into an interactor and creates a dedicated Unschedule Service to take care of making the appropriate status changes to a scheduled edition and deleting the publish intent that had been created when the edition was scheduled.

### DIFF
--- a/app/controllers/unschedule_controller.rb
+++ b/app/controllers/unschedule_controller.rb
@@ -4,6 +4,11 @@ class UnscheduleController < ApplicationController
   def unschedule
     result = Unschedule::UnscheduleInteractor.call(params: params, user: current_user)
 
-    redirect_to document_path(params[:document])
+    if result.api_error
+      redirect_to document_path(params[:document]),
+                  alert_with_description: t("documents.show.flashes.unschedule_error")
+    else
+      redirect_to document_path(params[:document])
+    end
   end
 end
--- a/app/controllers/unschedule_controller.rb
+++ b/app/controllers/unschedule_controller.rb
@@ -2,18 +2,8 @@
 
 class UnscheduleController < ApplicationController
   def unschedule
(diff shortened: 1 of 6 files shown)


spec/services/unschedule_service_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+  let(:edition) { create(:edition, :scheduled, scheduling: scheduling) }
+  let(:user) { build(:user) }
+
+  describe "#unschedule" do
+    it "sets an edition's scheduled publishing datetime to nil" do
+      stub_publishing_api_destroy_intent(edition.base_path)
+      UnscheduleService.new(edition).unschedule(user: user)
+
+      expect(edition.scheduled_publishing_datetime).to be nil
+    end
```
kevindew: I assume this test is no longer relevant?

app/services/unschedule_service.rb
```diff
@@ -0,0 +1,29 @@
...
+  def unschedule(user:)
+    set_edition_status(user)
+    destroy_publishing_intent
+  end
+
+private
+
+  def set_edition_status(user)
+    updater = Versioning::RevisionUpdater.new(edition.revision, user)
+    edition.assign_revision(updater.next_revision, user).save!
```
kevindew: I think this code isn't needed anymore, I assume it's from when we were setting the date to nil?

emmabeynon: Yes, good catch thank you.

emmabeynon:
Thanks @kevindew.  Makes sense to keep everything in the Unschedule Interactor, especially now that I've taken another couple of lines of code out of it.  I've also added a feature test for when the Publishing API is down.

PR #1092, opened 2019-05-31
Services to create BlobRevision objects

kevindew:
Trello: https://trello.com/c/BQ5bBkMw/881-wrap-up-technical-loose-ends-on-attachments

This is a follow on from the draft PR of https://github.com/alphagov/content-publisher/pull/1089

The key changes this makes is:

- It provides a public method for Versioning::BaseUpdater to specify a new revision change - which can bypass the process of setting attributes, this is to allow updating a SubRevision directly.
- It updates the FileAttachmentRevisionUpdater to allow setting a SubRevision as an attribute in assign
- It removes the FileAttachmentUploadService in favour of using create_initial on FileAttachment::Revision
- It changes the FileAttachmentBlobService to create a blob revision
- It then makes the Image equivalent code for dealing with an uploaded file consistent with the FileAttachment code


spec/services/file_attachment_blob_service_spec.rb
```diff
@@ -2,50 +2,45 @@
...
   let(:file) { fixture_file_upload("files/text-file.txt") }
-  let(:revision) do
-    build(:revision, file_attachment_revisions: [file_attachment_revision])
-  end
-  let(:file_attachment_revision) do
-    build(:file_attachment_revision, filename: "text-file.txt")
-  end
+  let(:revision) { build(:revision) }
+  let(:user) { build(:user) }
+  let(:instance) { FileAttachmentBlobService.new(revision, user) }
```
benthorner: This reminds me of `subject` ;-).

app/services/versioning/file_attachment_revision_updater.rb
```diff
@@ -16,6 +16,10 @@ def assign(fields)
 
       sub_columns = sub_updaters.values.flat_map(&:column_names)
       dup_revision.assign_attributes(fields.except(*sub_columns))
+
+      fields.slice(*sub_updaters.keys).each do |key, value|
```
benthorner: I'm hoping there's a less intrusive way of achieving this. But to recap our discussion, I see the need for the changes reported by the sub-revision updater and the revision updater to be consistent. We should be able to remove this constraint, though, because it's not necessary to have a sub-revision updater for the blob revision, since it can only ever be replaced (as opposed to updated).

To make that a bit clearer, I'm suggesting we:

   - Remove https://github.com/alphagov/content-publisher/blob/0a99bef926d2ed1d9f3a367e9a397fa636ff39d3/app/services/versioning/file_attachment_revision_updater.rb#L38
   - Add `%[blob_revision]` to https://github.com/alphagov/content-publisher/blob/0a99bef926d2ed1d9f3a367e9a397fa636ff39d3/app/services/versioning/file_attachment_revision_updater.rb#L6

kevindew: Sorted.

spec/services/file_attachment_blob_service_spec.rb
kevindew: haha - but the subject is the method. I know, I know - let's not go there again.

spec/services/versioning/file_attachment_revision_updater_spec.rb
```diff
@@ -80,5 +64,15 @@
       expect(next_revision.blob_revision.blob_id).to eq(old_fields[:blob_id])
       expect(next_revision.number_of_pages).to eq(old_fields[:number_of_pages])
     end
+
+    it "can accept a blob_revision as an attribute" do
```
benthorner: I think we usually cover this sort of thing with the 'fields' in the other specs, but I can also see why we might want to call it out specifically...

PR #1094, opened 2019-06-03
ADR for interactors

kevindew:
This documents the decision made recently to move business logic outside
of controllers and to use the Interactor gem as a pattern to implement
our business logic concerns.

This builds on the discussion in https://github.com/alphagov/content-publisher/issues/740

### DIFF
--- a/docs/adr/00xx-moving-business-logic-out-of-controllers.md
+++ b/docs/adr/00xx-moving-business-logic-out-of-controllers.md
@@ -1,0 +1,124 @@
+# XX. Moving business logic out of controllers
+
+Date: 2019-06-03
+
+## Context
+
+Over time the controller classes for Content Publisher have become
+[increasingly complex][complex-controllers] as the application grew in size
+and functionality. This has led to action methods that are long, difficult to
+understand and expensive to test. Looking towards the future it can be
+anticipated that there will be more functionality in controllers, such as
+access limiting, concurrent editing protection and invalid state handling.
+Therefore there is an expectation that controllers would become an increasing
+pain point for the application due to their complexity.
+
+This is a common scenario for a growing Rails application. A customary
+approach to address it is to distinguish between business and rendering
+logic in an action - where business logic is the action the user requested,
+such as changing application state, and rendering logic is the process of
+building the response to be returned to the user - and move the business logic
+to another class.
+
+Within the Content Publisher team we have been evaluating
+a number of approaches to perform our business logic operations, such
+as [service objects][],
+[the dry-transaction gem][dry-transaction],
+[the interactor gem][interactor-gem],
+[trailblazer operations][trailblazer-operation] and
+the [Publishing API command pattern][pub-api-command]. We came to the
+following conclusions:
+
+- Service objects: This is a pattern already used in Content Publisher,
+  however it is used to perform distinct supporting tasks rather than being
+  coupled to particular controller actions. We felt the services directory had
+  become a mess in Content Publisher, with little consistency between service
+  objects.
+- Dry-transaction gem: This provides a nice interface for managing response
+  through [ruby blocks][calling-a-dry-transaction]. However dry-transaction
+  requires a strict adherence to a provided DSL which can make a transaction
+  class very different from a plain old Ruby object.
+- Interactor gem: This is a simple, flexible pattern which relies on
+  building up an enhanced open struct based response. It had disadvantages
+  that it wasn't particularly clear what input and output was expected.
+- Trailblazer operation: This gem provides a DSL for creating classes to
+  perform a business operation. It works best within a Rails application
+  that has embraced the collection of Trailblazer utilities.
+- Publishing API command pattern: This is an effective pattern at moving logic
+  out of controllers, it however has produced usage problems where the base
+  class is coupled to particular domain logic.
+
+## Decision
+
+Controller actions can use a dedicated class to perform the business logic of
+a request. Logic in an action is to be focused on producing the appropriate
+HTTP response. To meet this requirement it is expected that classes shall
+exist for actions that mutate application state (typically POST/PUT/DELETE
+requests) and may exist for actions which don’t mutate state (typically GET)
+depending on the request complexity.
+
+Content Publisher makes use of the [Interactor gem][interactor-gem] as the
+means to implement these classes. It was chosen over other community
+approaches due to it’s simplicity and flexibility. It was decided to use
+a community provided framework rather that writing our own, this is
+because:
+
+- a bespoke problem is not being solved, so it shouldn't require a bespoke
+  solution;
+- authoring a new pattern is something that requires negotiation,
+  documentation, maintenance and iteration;
+- It’s harder to [bikeshed][] on decisions made externally.
+
+The Interactor gem is relativity simple in comparison to similar gems. This
+offers the advantage that, were we to discover problems with the gem or a
+change in our needs, it would be realistic for us to adapt this to a self
+built or different community approach.
+
+Interactor classes are created with a coupling to a particular controller
+action. Interactors are stored in an app/interactors directory. Within here
+there are directories created for each controller to store the interactors for
+each action, in a pattern similar to views. Each interactor class name is
+suffixed with "Interactor", to make their purpose clear. As an example,
+given a create action on a DocumentsController there is a corresponding
+interactor to create a document, Document::CreateInteractor, which is stored
+in app/interactors/document.
+
+## Status
+
+Accepted
+
+## Consequences
+
+Many of the controller actions in Content Publisher have been refactored to
+make use of the interactor pattern. This has resulted in an increase in the
+number of classes of the application while the size and complexity of
+controller classes have decreased.
+
+Conventions have been established on a consistent approach to using the
+Interactor gem. These have focused on:
+
+- [initialising controllers with request parameters and a user][initialize-example];
+- [authoring an interactor's `call` method as a procedure of descriptive
+  methods][call-method-example];
+- [using delegation to reduce repetitive references to the `context`
+  variable][delegation-example];
+- [setting variables on context to be used to determine HTTP
+  response][response-example],
+  rather than using context.success / context.failure methods, this allows the
+  code to be more explicit about which success/failure scenario.
+
+Service classes that have a singular purpose could be refactored to use the
+interactor pattern. Before this can be done, we will need to decide on a
+directory convention for where they should be stored.
+
+[complex-controllers]: https://github.com/alphagov/content-publisher/blob/1eb067d35d557982f05601cde33c93f9ebea5694/app/controllers/images_controller.rb#L55-L101
+[service objects]: https://medium.com/@scottdomes/service-objects-in-rails-75ca74214b77
+[dry-transaction]: https://dry-rb.org/gems/dry-transaction/
+[calling-a-dry-transaction]: https://dry-rb.org/gems/dry-transaction/basic-usage/#calling-a-transaction
+[trailblazer-operation]: http://trailblazer.to/gems/operation/2.0/
+[bikeshed]: https://en.wiktionary.org/wiki/bikeshedding
+[interactor-gem]: https://github.com/collectiveidea/interactor
+[initialize-example]: https://github.com/alphagov/content-publisher/blob/e24ae71cb3a27889c1e09b0ec6135dfd20ffb7a1/app/controllers/documents_controller.rb#L31
+[call-method-example]: https://github.com/alphagov/content-publisher/blob/e24ae71cb3a27889c1e09b0ec6135dfd20ffb7a1/app/interactors/documents/destroy_interactor.rb#L12-L16
+[delegation-example]: https://github.com/alphagov/content-publisher/blob/e24ae71cb3a27889c1e09b0ec6135dfd20ffb7a1/app/interactors/documents/destroy_interactor.rb#L5-L9
+[response-example]: https://github.com/alphagov/content-publisher/blob/e24ae71cb3a27889c1e09b0ec6135dfd20ffb7a1/app/controllers/documents_controller.rb#L32-L37
--- a/docs/adr/00xx-moving-business-logic-out-of-controllers.md
+++ b/docs/adr/00xx-moving-business-logic-out-of-controllers.md
@@ -10,78 +10,82 @@ and functionality. This has led to action methods that are long, difficult to
 understand and expensive to test. Looking towards the future it can be
 anticipated that there will be more functionality in controllers, such as
 access limiting, concurrent editing protection and invalid state handling.
-Therefore there is an expectation that controllers would become an increasing
+Therefore, there is an expectation that controllers would become an increasing
 pain point for the application due to their complexity.
 
 This is a common scenario for a growing Rails application. A customary
-approach to address it is to distinguish between business and rendering
-logic in an action - where business logic is the action the user requested,
+approach to address this is to distinguish between business and rendering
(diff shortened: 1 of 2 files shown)


docs/adr/00xx-moving-business-logic-out-of-controllers.md
```diff
@@ -0,0 +1,124 @@
...
+building the response to be returned to the user - and move the business logic
+to another class.
+
+Within the Content Publisher team we have been evaluating
+a number of approaches to perform our business logic operations, such
+as [service objects][],
+[the dry-transaction gem][dry-transaction],
+[the interactor gem][interactor-gem],
+[trailblazer operations][trailblazer-operation] and
+the [Publishing API command pattern][pub-api-command]. We came to the
```
benthorner: Looks like this is an invalid reference :-|.

docs/adr/00xx-moving-business-logic-out-of-controllers.md
```diff
@@ -0,0 +1,124 @@
...
+  through [ruby blocks][calling-a-dry-transaction]. However dry-transaction
+  requires a strict adherence to a provided DSL which can make a transaction
+  class very different from a plain old Ruby object.
+- Interactor gem: This is a simple, flexible pattern which relies on
+  building up an enhanced open struct based response. It had disadvantages
+  that it wasn't particularly clear what input and output was expected.
+- Trailblazer operation: This gem provides a DSL for creating classes to
+  perform a business operation. It works best within a Rails application
+  that has embraced the collection of Trailblazer utilities.
+- Publishing API command pattern: This is an effective pattern at moving logic
```
benthorner: Minor grammar thing: can we change ', it however' to '. However, it'?

docs/adr/00xx-moving-business-logic-out-of-controllers.md
```diff
@@ -0,0 +1,124 @@
...
+  out of controllers, it however has produced usage problems where the base
```
benthorner: It might be helpful to elaborate why the coupling is bad, since we've referred to coupling in a neutral/positive way above ^^^.

docs/adr/00xx-moving-business-logic-out-of-controllers.md
```diff
@@ -0,0 +1,124 @@
...
+  class is coupled to particular domain logic.
+
+## Decision
+
+Controller actions can use a dedicated class to perform the business logic of
```
benthorner: By 'can', do we mean 'should'? I struggled slightly to understand which of these first two sentences states the requirement - I think it's the second one, but I was looking for a 'must' or a 'should' to mark it. Could we rephrase this to make the primary decision more prominent?

docs/adr/00xx-moving-business-logic-out-of-controllers.md
```diff
@@ -0,0 +1,124 @@
...
+a request. Logic in an action is to be focused on producing the appropriate
+HTTP response. To meet this requirement it is expected that classes shall
+exist for actions that mutate application state (typically POST/PUT/DELETE
+requests) and may exist for actions which don’t mutate state (typically GET)
+depending on the request complexity.
+
+Content Publisher makes use of the [Interactor gem][interactor-gem] as the
+means to implement these classes. It was chosen over other community
+approaches due to it’s simplicity and flexibility. It was decided to use
+a community provided framework rather that writing our own, this is
```
benthorner: Minor, but can we remove 'this is' (unless it's a new sentence)?

docs/adr/00xx-moving-business-logic-out-of-controllers.md
```diff
@@ -0,0 +1,124 @@
...
+because:
+
+- a bespoke problem is not being solved, so it shouldn't require a bespoke
+  solution;
+- authoring a new pattern is something that requires negotiation,
+  documentation, maintenance and iteration;
+- It’s harder to [bikeshed][] on decisions made externally.
```
benthorner: Minor, and somewhat ironic for this point: the capitalisation is inconsistent in these points. I'm wondering if this is really a detail of the first point. In fact, it seems like all three points are saying similar things in different ways - maybe that's the intention here :-).

docs/adr/00xx-moving-business-logic-out-of-controllers.md
```diff
@@ -0,0 +1,124 @@
...
+
+The Interactor gem is relativity simple in comparison to similar gems. This
```
benthorner: By 'similar gems', are we referring to the ones we mentioned above? Might help to use 'gems like xyz' to help clarify.

docs/adr/00xx-moving-business-logic-out-of-controllers.md
```diff
@@ -0,0 +1,124 @@
...
+offers the advantage that, were we to discover problems with the gem or a
+change in our needs, it would be realistic for us to adapt this to a self
+built or different community approach.
+
+Interactor classes are created with a coupling to a particular controller
+action. Interactors are stored in an app/interactors directory. Within here
+there are directories created for each controller to store the interactors for
+each action, in a pattern similar to views. Each interactor class name is
+suffixed with "Interactor", to make their purpose clear. As an example,
+given a create action on a DocumentsController there is a corresponding
```
benthorner: Minor, but maybe use code syntax for 'create' to help parsing the sentence.

docs/adr/00xx-moving-business-logic-out-of-controllers.md
```diff
@@ -0,0 +1,124 @@
...
+interactor to create a document, Document::CreateInteractor, which is stored
```
benthorner: I think it's actually Document**s** ;-).

docs/adr/00xx-moving-business-logic-out-of-controllers.md
```diff
@@ -0,0 +1,124 @@
...
+Many of the controller actions in Content Publisher have been refactored to
+make use of the interactor pattern. This has resulted in an increase in the
+number of classes of the application while the size and complexity of
+controller classes have decreased.
+
+Conventions have been established on a consistent approach to using the
+Interactor gem. These have focused on:
+
+- [initialising controllers with request parameters and a user][initialize-example];
+- [authoring an interactor's `call` method as a procedure of descriptive
```
benthorner: This one took a few readings :-). Could we say 'writing interactor `call` methods as a sequence of (no-argument) method calls`?

docs/adr/00xx-moving-business-logic-out-of-controllers.md
```diff
@@ -0,0 +1,124 @@
...
+  methods][call-method-example];
+- [using delegation to reduce repetitive references to the `context`
+  variable][delegation-example];
+- [setting variables on context to be used to determine HTTP
+  response][response-example],
+  rather than using context.success / context.failure methods, this allows the
```
benthorner: Minor, but code syntax might help a bit here. Also, I think it's conventional to write `Context#failure`, since `Context.failure` might be confused as referring to a class method (under this same convention). Pretty sure I read that in a book somewhere!

docs/adr/00xx-moving-business-logic-out-of-controllers.md
```diff
@@ -0,0 +1,124 @@
...
+  code to be more explicit about which success/failure scenario.
+
+Service classes that have a singular purpose could be refactored to use the
```
benthorner: Could we clarify this a bit? I'm thinking of the recent UnscheduleService example, where we just inlined the code into the interactor. Are we talking about bigger services that still need their own class?

docs/adr/00xx-moving-business-logic-out-of-controllers.md
benthorner: Also, since we're listing these below, should we just provide the links on each bullet point?

docs/adr/00xx-moving-business-logic-out-of-controllers.md
```diff
@@ -0,0 +1,124 @@
...
+- Service objects: This is a pattern already used in Content Publisher,
+  however it is used to perform distinct supporting tasks rather than being
+  coupled to particular controller actions. We felt the services directory had
+  become a mess in Content Publisher, with little consistency between service
+  objects.
+- Dry-transaction gem: This provides a nice interface for managing response
+  through [ruby blocks][calling-a-dry-transaction]. However dry-transaction
+  requires a strict adherence to a provided DSL which can make a transaction
+  class very different from a plain old Ruby object.
+- Interactor gem: This is a simple, flexible pattern which relies on
```
benthorner: This is a nice summary, but it seems a little low-level in comparison to the others. It might be more helpful to contrast it to the other DSL approaches (i.e. focus on the 'flexible' bit) e.g. 'This is a simple, relatively freeform pattern (no DSL) that uses a `context` object for basic control flow'.

docs/adr/00xx-moving-business-logic-out-of-controllers.md
```diff
@@ -0,0 +1,124 @@
...
+- Publishing API command pattern: This is an effective pattern at moving logic
+  out of controllers, it however has produced usage problems where the base
+  class is coupled to particular domain logic.
+
+## Decision
+
+Controller actions can use a dedicated class to perform the business logic of
+a request. Logic in an action is to be focused on producing the appropriate
+HTTP response. To meet this requirement it is expected that classes shall
+exist for actions that mutate application state (typically POST/PUT/DELETE
```
benthorner: This seems more like a consequence or observation, rather than part of the decision. If we happened to have a tiny POST action, would we use an interactor? I thought of this because I was wondering how we could phrase this to get to the interactor bit of the decision earlier, since it helps clarify what the 'classes' are - the suspense was palpable ;-).

docs/adr/00xx-moving-business-logic-out-of-controllers.md
```diff
@@ -0,0 +1,124 @@
...
+requests) and may exist for actions which don’t mutate state (typically GET)
+depending on the request complexity.
+
+Content Publisher makes use of the [Interactor gem][interactor-gem] as the
+means to implement these classes. It was chosen over other community
+approaches due to it’s simplicity and flexibility. It was decided to use
```
benthorner: Although this was part of the decision, does this fit better in the previous section, where we were comparing different approaches? It seems like another approach i.e. 'roll our own'.

docs/adr/00xx-moving-business-logic-out-of-controllers.md
```diff
@@ -0,0 +1,124 @@
...
+number of classes of the application while the size and complexity of
+controller classes have decreased.
+
+Conventions have been established on a consistent approach to using the
+Interactor gem. These have focused on:
+
+- [initialising controllers with request parameters and a user][initialize-example];
+- [authoring an interactor's `call` method as a procedure of descriptive
+  methods][call-method-example];
+- [using delegation to reduce repetitive references to the `context`
```
benthorner: This is a nice way to summarise the conventions :-).

docs/adr/00xx-moving-business-logic-out-of-controllers.md
1pretz1: Small thing but referring to the `PUT` part of `POST/PUT/DELETE`, shouldn't this actually be `PATCH` as the verbs used in our routes are get, post, patch and delete.

docs/adr/00xx-moving-business-logic-out-of-controllers.md
kevindew: Thanks, updated.

docs/adr/00xx-moving-business-logic-out-of-controllers.md
kevindew: Done, with a bit of rewording.

docs/adr/00xx-moving-business-logic-out-of-controllers.md
kevindew: Sure updated.

docs/adr/00xx-moving-business-logic-out-of-controllers.md
kevindew: Done

docs/adr/00xx-moving-business-logic-out-of-controllers.md
kevindew: I've rephrased this bit as part of a re-ordering.

docs/adr/00xx-moving-business-logic-out-of-controllers.md
kevindew: I've updated the second one. I think of the three things as distinct points.

docs/adr/00xx-moving-business-logic-out-of-controllers.md
kevindew: Done

docs/adr/00xx-moving-business-logic-out-of-controllers.md
kevindew: Done

docs/adr/00xx-moving-business-logic-out-of-controllers.md
kevindew: Thanks

docs/adr/00xx-moving-business-logic-out-of-controllers.md
kevindew: Done

docs/adr/00xx-moving-business-logic-out-of-controllers.md
kevindew: Thanks - yeah that's a common convention. I was referencing the instance with the lower case context but your suggestion looks an improvement.

docs/adr/00xx-moving-business-logic-out-of-controllers.md
kevindew: I removed this section since there aren't decisions made here.

I'm referring to how our pure-function-as-an-object esque services could easily be built as an interactor but we don't really know how that'd fit in directory wise.

docs/adr/00xx-moving-business-logic-out-of-controllers.md
kevindew: Done

docs/adr/00xx-moving-business-logic-out-of-controllers.md
kevindew: I've adjusted the section to address this.

docs/adr/00xx-moving-business-logic-out-of-controllers.md
kevindew: I've moved it up to the decision for better context relevance.

docs/adr/00xx-moving-business-logic-out-of-controllers.md
kevindew: Done.

docs/adr/00xx-moving-business-logic-out-of-controllers.md
```diff
@@ -0,0 +1,126 @@
...
+interactors. These have focused on:
+
+- [initialising controllers with request parameters and a user][initialize-example];
+- [writing interactor `call` methods as a sequence of (no-argument)
+  method calls][call-method-example];
+- [using delegation to reduce repetitive references to the `context`
+  variable][delegation-example];
+- [setting variables on context to be used to determine HTTP
+  response][response-example],
+  rather than using `Context#success` / `Context#failure` methods, this allows
```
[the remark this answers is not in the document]
benthorner: Also can we change: this -> which

kevindew:
@benthorner thanks all those amends made

PR #1095, opened 2019-06-03
Remove superfluous delegations in interactors

1pretz1:
There is a number of superfluous delegations in our interactors which
all concern our fail state variables. I'm not sure if this was a
conscious decision to delegate everything we set in the context but it
seems unnecessary when some of them are not used.

### DIFF
--- a/app/interactors/documents/destroy_interactor.rb
+++ b/app/interactors/documents/destroy_interactor.rb
@@ -5,7 +5,6 @@ class Documents::DestroyInteractor
   delegate :params,
            :user,
            :edition,
-           :api_error,
            to: :context
 
   def call
--- a/app/interactors/editions/create_interactor.rb
+++ b/app/interactors/editions/create_interactor.rb
@@ -6,7 +6,6 @@ class Editions::CreateInteractor
            :user,
            :live_edition,
            :next_edition,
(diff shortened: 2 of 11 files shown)


kevindew:
So the purpose of these was to also provide the API to the context object that is returned, otherwise you have to dig into code to see what possible options there are.

I'm in a bad place to judge myself as to whether these add value or not as I'm now quite familiar with the pattern. Do you think these add no value to you and are better removed?

1pretz1:
Now you explain their purpose I can see a value in delegating these. For me personally they just caused confusion as to why we were delegating something which didn't need to be delegated, however if you can see more value in having them I'm not hell bent on deleting them, I just though i'd try and tidy some code up. :)

kevindew:
> Now you explain their purpose I can see a value in delegating these. For me personally they just caused confusion as to why we were delegating something which didn't need to be delegated, however if you can see more value in having them I'm not hell bent on deleting them, I just though i'd try and tidy some code up. :)

I don't mind too much either way - I can see pros and cons - but would suggest if we're not too fussed we close it until we're clearer on an opinion.

1pretz1:
Sure! 🔥

PR #1096, opened 2019-06-03
Adapt scheduling flow

emmabeynon:
For https://trello.com/b/mn2MqH3M/publishing-workflow-doing-q4-2019-to
Collab with @1pretz1 ⭐️ 

This adapts the scheduling workflow to match the new design as per the [prototype](https://content-publisher-prototype.cloudapps.digital/content-summary-draft).

Once https://github.com/alphagov/content-publisher/pull/1088 is merged we can use this to resize the time selector and match the prototype.


app/controllers/schedule_controller.rb
```diff
@@ -2,33 +2,38 @@
...
-      end
-
-      set_scheduled_publishing_datetime(edition, checker.parsed_datetime)
-
+    result = Schedule::SaveDatetimeInteractor.call(params: params, user: current_user)
+    edition, issues = result.to_h.values_at(:edition, :issues)
+
+    if issues
+      flash.now["alert_with_items"] = {
+        "title" => I18n.t!("requirements.scheduled_datetime.title"),
```
kevindew: I think this is something we might want to change since it seems tailored to a particular error.

app/interactors/schedule/save_datetime_interactor.rb
```diff
@@ -0,0 +1,67 @@
...
+  end
+
+  def set_scheduled_publishing_datetime
+    updater = Versioning::RevisionUpdater.new(edition.revision, user)
+    updater.assign(scheduled_publishing_datetime: context.datetime)
+
+    if updater.changed?
+      edition.assign_revision(updater.next_revision, user).save!
+      create_timeline_entry
+    end
```
kevindew: I'd suggest we use fail to abort if there aren't any changes, as per: https://github.com/alphagov/content-publisher/blob/93d067aea5b3f1b11c854b692efc38753a59dc8b/app/interactors/images/update_interactor.rb#L57

Then we can do timeline entry as a separate step

app/interactors/schedule/save_datetime_interactor.rb
```diff
@@ -0,0 +1,67 @@
...
+  def find_and_lock_edition
+    context.edition = Edition.lock.find_current(document: params[:document])
+  end
+
+  def check_for_issues
+    unless edition.draft? || edition.submitted_for_review?
+      raise "Can't save a scheduling date unless edition is a draft or has been submitted for 2i"
+    end
+
+    checker = Requirements::ScheduledDatetimeChecker.new(params[:schedule])
```
kevindew: To save this failing on a potential nil, and other request nasties, I think this could have a method of schedule_params with a `params.require(:schedule).permit()` with the available fields.

app/interactors/schedule/save_datetime_interactor.rb
```diff
@@ -0,0 +1,67 @@
...
+      edition.assign_revision(updater.next_revision, user).save!
+      create_timeline_entry
+    end
+  end
+
+  def create_timeline_entry
+    TimelineEntry.create_for_revision(
+      entry_type: :scheduled_publishing_datetime_set,
+      edition: edition,
+    )
```
kevindew: I don't think this is needed as per Ben Hazell's comment: https://trello.com/c/5vJhi2a8/691-update-schedule-workflow-based-on-user-research#comment-5cbf2e90ae40db85431fe988 

apologies if that isn't very clear on the trello card I notice the link is only in the description but not the AC

app/interactors/schedule/save_datetime_interactor.rb
```diff
@@ -0,0 +1,67 @@
...
+  end
+
+  def action_issues
+    action = params[:schedule][:action] || params[:schedule][:update]
```
kevindew: I wonder if there is a cleaner way to check whether it is an update or not than having a parameter passed from the form?

Could we be checking on whether a date is saved or not?

app/controllers/schedule_controller.rb
```diff
@@ -2,33 +2,38 @@
...
   def clear_scheduled_publishing_datetime
     Edition.find_and_lock_current(document: params[:document]) do |edition|
-      set_scheduled_publishing_datetime(edition)
+      updater = Versioning::RevisionUpdater.new(edition.revision, current_user)
+      updater.assign(scheduled_publishing_datetime: nil)
+
+      if updater.changed?
+        edition.assign_revision(updater.next_revision, current_user).save!
+        create_datetime_cleared_timeline_entry(edition)
+      end
```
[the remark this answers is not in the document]
emmabeynon: Sure, I thought about doing this in this PR but we've got a card to review clearing so I think it can be done as part of that.  Will add as an A/C to that.

app/controllers/schedule_controller.rb
```diff
@@ -2,33 +2,38 @@
...
-
+    result = Schedule::SaveDatetimeInteractor.call(params: params, user: current_user)
+    edition, issues = result.to_h.values_at(:edition, :issues)
+
+    if issues
+      flash.now["alert_with_items"] = {
+        "title" => I18n.t!("requirements.scheduled_datetime.title"),
+        "items" => issues.items(
+          link_options: {
+            scheduled_datetime: { href: "#scheduled_publishing_datetime" },
```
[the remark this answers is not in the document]
emmabeynon: I think this is because we check the date and time combination rather than the separate date and time inputs, so we don't have checks that produce issues for the date and time separately.  This will probably make it difficult to highlight the fields that need correcting.  Would be good to have a chat about how to tackle this.

app/controllers/schedule_controller.rb
emmabeynon: This is now using the generic "You must" title and the individual error messages have been adapted to work with it.

app/controllers/schedule_controller.rb
```diff
@@ -2,33 +2,38 @@
 
 class ScheduleController < ApplicationController
   def save_scheduled_publishing_datetime
```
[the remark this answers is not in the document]
emmabeynon: Done 👍

app/helpers/actions_helper.rb
```diff
@@ -38,10 +38,24 @@ def remove_link(edition)
...
+           data_attributes: { gtm: "schedule" },
+           href: scheduling_confirmation_path(edition.document)
+  end
+
+  def create_schedule_date_link(edition, extra_classes = [])
+    link_to "Schedule",
+            scheduled_publishing_datetime_path(edition.document),
+            class: %w(govuk-link govuk-link--no-visited-state) + Array(extra_classes),
             data: { gtm: "schedule" }
   end
```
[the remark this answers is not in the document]
emmabeynon: Done 👍

app/controllers/schedule_controller.rb
benthorner: I thought we were going with "Thou shalt" ;-).

app/interactors/schedule/save_datetime_interactor.rb
emmabeynon: Done 👍

app/interactors/schedule/save_datetime_interactor.rb
emmabeynon: Done 👍

app/interactors/schedule/save_datetime_interactor.rb
emmabeynon: Now removed

app/interactors/schedule/save_datetime_interactor.rb
emmabeynon: I've removed the hidden tag and we now just check if the edition already has scheduled_publishing_datetime set.

app/controllers/schedule_controller.rb
emmabeynon: Well you know, that would be my preference 🤷‍♀

app/controllers/schedule_controller.rb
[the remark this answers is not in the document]
kevindew: Ok cool 👍

app/controllers/schedule_controller.rb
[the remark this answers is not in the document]
kevindew: ok cool - just ping me when you get a chance. I'll need some more context.

kevindew:
@emmabeynon is this ready for another review or shall I wait for a shout?

app/interactors/schedule/schedule_interactor.rb
```diff
@@ -0,0 +1,51 @@
...
+  def check_for_issues
+    unless edition.schedulable?
+      # FIXME: this shouldn't be an exception but we've not worked out the
+      # right response - maybe bad request or a redirect with flash?
+      raise "Can't schedule an edition which isn't schedulable"
+    end
+
+    if params[:review_status].blank?
+      issues = Requirements::CheckerIssues.new([
+        Requirements::Issue.new(:scheduling, :not_selected),
```
kevindew: I wonder if this should have a more descriptive naming like "schedule_review" as scheduling feels quite broad

app/views/documents/show/_actions.html.erb
```diff
@@ -12,32 +12,36 @@
...
-        <%= publish_link(@edition) %>
+      <% if @edition.schedulable? %>
+        <%= delete_draft_link(@edition) %>
+        <%= schedule_confirmation_link(@edition, "app-link--right") %>
+      <% elsif @edition.editable? && current_user.has_permission?(User::PRE_RELEASE_FEATURES_PERMISSION) %>
+        <p class="govuk-body govuk-!-margin-bottom-3">
+          <%= publish_link(@edition) %>
+          <%= create_schedule_date_link(@edition, "app-link--right") %>
+        </p>
+        <%= delete_draft_link(@edition) %>
```
kevindew: I feel like we should have a schedule link when revision hasn't synced too - since that can be used to set a date even if we can't write to Publishing API

app/views/documents/show/_scheduling_notice.html.erb
```diff
@@ -0,0 +1,49 @@
+<% datetime = @edition.scheduled_publishing_datetime %>
+<% time = datetime&.strftime("%l:%M%P")&.strip %>
```
kevindew: Do you know what the strip is for here?

app/views/schedule/confirmation.html.erb
```diff
@@ -1,30 +1,30 @@
 <% content_for :back_link, render_back_link(href: document_path(@edition.document)) %>
 <% content_for :title, t("schedule.confirmation.title") %>
-
-<% scheduled_time = @edition.scheduled_publishing_datetime.strftime("%l:%M%P").strip %>
-<% scheduled_date = @edition.scheduled_publishing_datetime.strftime("%-d %B %Y") %>
+<% datetime = @edition.scheduled_publishing_datetime %>
+<% time = datetime&.strftime("%l:%M%P")&.strip %>
+<% date = datetime&.strftime("%-d %B %Y") %>
```
kevindew: as we have these a few times it could make sense to have scheduled_date and scheduled_time helper methods

app/views/documents/show/_actions.html.erb
kevindew: Oh and should remember to hide that from users without permission

app/controllers/schedule_controller.rb
```diff
@@ -40,36 +45,24 @@ def confirmation
...
+    edition, issues = result.to_h.values_at(:edition, :issues)
+
+    if issues
+      flash.now["alert_with_items"] = {
+        "title" => I18n.t!("schedule.scheduled_publishing_datetime.flashes.requirements"),
+        "items" => issues.items,
+      }
+
+      render :confirmation,
+             assigns: { review_status_issues: issues.items.first[:text], edition: edition },
```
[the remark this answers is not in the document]
emmabeynon: There's only one possible issue that the ScheduleInteractor could generate, so we're just passing that one to as an `error_message` argument to the radio component.  We could switch that to `error_items` and pass the array if that's cleaner?

app/views/documents/show/_scheduling_notice.html.erb
emmabeynon: I think that's a hangover from another view where we'd sometimes get additional spacing.  That doesn't seem to be a problem with this view so I'll get rid of it.

app/interactors/schedule/schedule_interactor.rb
emmabeynon: Done

app/views/documents/show/_actions.html.erb
emmabeynon: Done

app/views/schedule/confirmation.html.erb
emmabeynon: Now in the Scheduled Datetime Helper.

app/controllers/schedule_controller.rb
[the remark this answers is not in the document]
kevindew: Yup I think that'd more consistent with other places. For example: https://github.com/alphagov/content-publisher/blob/master/app/controllers/publish_controller.rb#L33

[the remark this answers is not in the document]
kevindew: Aside from this I think this is now good to merge

app/views/schedule/scheduled_publishing_datetime.html.erb
```diff
@@ -0,0 +1,89 @@
+<% content_for :back_link, render_back_link(href: document_path(@edition.document)) %>
+<% content_for :title, t("schedule.scheduled_publishing_datetime.title") %>
+
+<% current_scheduled_publish_datetime = @edition.revision.scheduled_publishing_datetime %>
+<% default_date = Time.zone.now.tomorrow %>
+<% publish_date = current_scheduled_publish_datetime || default_date %>
+<% schedule_param = params["schedule"] ? params["schedule"] : {} %>
+<% selected_time = schedule_param["time"] || scheduled_time(current_scheduled_publish_datetime) %>
+<% action_not_selected = @issues&.select { |i| i.issue_key == :action_not_selected }&.first&.to_item %>
```
benthorner: Could we use `@issues&.items_for` here, to avoid the manual select?

spec/features/workflow/change_scheduled_publishing_datetime_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+  def when_i_visit_the_summary_page
+    visit document_path(@edition.document)
+  end
+
+  def and_i_click_on_change_date
+    click_on "Change date"
+  end
+
+  def then_i_see_the_current_date_and_time_selected
+    expect(page).to have_selector(
```
benthorner: Minor, but I think we can use `find_field(...).value` here e.g. as in the [image specs](https://github.com/alphagov/content-publisher/blob/master/spec/features/editing_images/choose_lead_image_spec.rb#L39).

spec/features/workflow/clear_scheduled_publishing_datetime_spec.rb
```diff
@@ -4,57 +4,47 @@
...
 
   def given_there_is_an_edition_with_a_set_scheduled_publishing_datetime
     @datetime = Time.zone.now.advance(days: 2).midday
+    @scheduled_date = @datetime.strftime("%-d %B %Y")
     @edition = create(:edition, scheduled_publishing_datetime: @datetime)
   end
 
   def when_i_visit_the_summary_page
     visit document_path(@edition.document)
+    expect(page).to have_content("Proposed to publish at 12:00pm on #{@scheduled_date}")
```
benthorner: Could you help me understand why this line is here? I think it would be confusing for the tests to fail   on this line, as it's a `when` method.

app/controllers/schedule_controller.rb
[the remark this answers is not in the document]
emmabeynon: Should be sorted now 👍

spec/features/workflow/schedule_set_datetime_passes_spec.rb
```diff
@@ -5,6 +5,7 @@
     given_there_is_an_edition_with_past_scheduled_publishing_datetime
```
benthorner: I'm wondering if this spec is covered by the ones for changing the date?

app/views/schedule/scheduled_publishing_datetime.html.erb
kevindew: That's something that is resolved in my PR

app/interactors/schedule/save_datetime_interactor.rb
```diff
@@ -0,0 +1,67 @@
...
+    updater = Versioning::RevisionUpdater.new(edition.revision, user)
+    updater.assign(scheduled_publishing_datetime: context.datetime)
+
+    context.fail! unless updater.changed?
+
+    edition.assign_revision(updater.next_revision, user).save!
+  end
+
+  def action_issues
+    return [] if scheduled_datetime_already_exists?
```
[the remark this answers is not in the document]
emmabeynon: So `already_scheduled?` wouldn't be quite right because the user doesn't need to have scheduled the document for publishing in order to change the scheduled_publishing_datetime.  They could have just set it already and then gone in and changed it again and not scheduled it.  According to the prototype if a scheduled_publishing_datetime has been set and then they go and change it again there is no choice of action to take; they just save the changed scheduled_publishing_datetime.

We'll be covering changing the scheduled_publishing_datetime after a document has been scheduled in this card https://trello.com/c/5n7oZvXZ/851-change-a-scheduled-publishing-date

app/interactors/schedule/schedule_interactor.rb
```diff
@@ -0,0 +1,51 @@
...
+           :edition,
+           :issues,
+           to: :context
+
+  def call
+    Edition.transaction do
+      find_and_lock_edition
+      check_for_issues
+      schedule_publishing
+      create_scheduling
```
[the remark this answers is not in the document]
emmabeynon: I'd rather not combine them because they are doing different things, however I agree the naming is not clear.  Perhaps we could have `set_scheduled_publishing_job` instead of `schedule_publishing`?

kevindew:
Thanks @benthorner my thoughts on the things to fix:

> * I think we should see if we can reduce the feature tests for scheduling. I've left comments here and there, but I think it's worth spending some time going through them after we merge this

Yup got this flagged to look at on epic I believe

> * I found a bug where text is accepted in the time field, which is weird as there's a test explicitly about that (screenshot below)

I believe that's fixed my PR

> * I found a potential bug where the 'Back' button on the confirmation page goes back to the summary, which was unexpected but might be intentional

Lets make a note to fix as part of story but not this PR

> * Small style thing: 'Delete draft' is all lonesome on the right when the document is submitted for 2i (screenshot below)

Lets make a note to fix as part of story but not this PR

> * Another small thing: the page to set the date/time doesn't remember the radio selection if there are other requirements issues

Thanks - I'll fix that in my PR

spec/features/workflow/clear_scheduled_publishing_datetime_spec.rb
emmabeynon: It's to make it clear that it's present at that point, and we are then expecting it not to be present in line 47.

spec/features/workflow/change_scheduled_publishing_datetime_spec.rb
emmabeynon: We use `have_selector` elsewhere in the specs so I'm going to leave it as is.

spec/features/workflow/schedule_set_datetime_passes_spec.rb
emmabeynon: I don't think so because this spec is testing that the user is unable to click on the schedule button if the scheduled_publishing_datetime is in the past.

app/interactors/schedule/save_datetime_interactor.rb
```diff
@@ -0,0 +1,67 @@
...
+    context.datetime = checker.parsed_datetime
+  end
+
+  def schedule_params
+    params.require(:schedule).permit(:day, :month, :year, :time)
+  end
+
+  def set_scheduled_publishing_datetime
+    updater = Versioning::RevisionUpdater.new(edition.revision, user)
+    updater.assign(scheduled_publishing_datetime: context.datetime)
```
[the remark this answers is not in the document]
emmabeynon: Done 👍

app/interactors/schedule/schedule_interactor.rb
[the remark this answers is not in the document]
benthorner: I think it's actually the `create_scheduling` one that was puzzling me, as 'scheduling' isn't a noun, I'm not sure how we could create one. Looking at the methods, I'm now wondering what they do that's different - they both look like they're doing scheduling, but in different ways (one is a job, one is a service; they both say schedule). I reckon this might take a bit of time to come to a conclusion, so I'll note it down on the card to think about later.

app/interactors/schedule/save_datetime_interactor.rb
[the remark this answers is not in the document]
benthorner: Ah, that's a shame - I was hopeful it would be simple. I think this will come up again if we try to address the 'Back' button bug (the assumption will break under that condition), so I'm happy to leave it - I'm afraid it's still confusing for me to read, both here and in the view, so it'd be great to have a deeper think about what the concept behind this check is.

kevindew:
Thanks @benthorner 👍

PR #1104, opened 2019-06-05
Show scheduling errors with their respective fields

kevindew:
Trello: https://trello.com/c/rGjDviSL/848-adapt-scheduling-implementation-for-iterated-workflow

This allows each of the requirements issues to link directly to a field. It does this by breaking up the datetime errors into ones that are date or time. To complete this the autocomplete component needed updating to accept errors.

There are few things that aren't quite ideal in this which I'll explain:

- The fields for scheduling are most prefixed with `schedule_` this is done to disambiguate them amongst the many fields in the requirements yaml file, it would have been nicer to have date, time and action as names but they looked rather generic in amongst that file;
- The time parsing is quite strict to limit checking that a time input is actually a time, the alternative is pretty wild west parsing wise without further complication - unless anyone has better ideas?
- the date_input doesn't accept error_items as input, hopefully we can fix that soon
- the date_input doesn't have a label inside it so that's not caught in an error state

Screenshot:

Screen Shot 2019-06-05 at 09 50 04

### DIFF
--- a/app/services/requirements/scheduled_datetime_checker.rb
+++ b/app/services/requirements/scheduled_datetime_checker.rb
@@ -44,13 +44,13 @@ module Requirements
         day, month, year = params[:date]&.values_at(:day, :month, :year)
         Date.strptime("#{day}-#{month}-#{year}", DATE_FORMAT)
       rescue ArgumentError
-        issues << Issue.new(:scheduled_date, :invalid)
+        issues << Issue.new(:schedule_date, :invalid)
       end
 
       begin
         Time.strptime(params[:time].to_s, TIME_FORMAT)
       rescue ArgumentError
-        issues << Issue.new(:scheduled_time, :invalid)
+        issues << Issue.new(:schedule_time, :invalid)
       end
 
       issues
@@ -59,7 +59,7 @@ module Requirements
     def scheduled_in_the_past_issues
       return [] unless parsed_datetime < Time.current
 
-      field = parsed_datetime.today? ? :scheduled_time : :scheduled_date
+      field = parsed_datetime.today? ? :schedule_time : :schedule_date
       [Issue.new(field, :in_the_past)]
     end
 
@@ -68,7 +68,7 @@ module Requirements
       return [] unless parsed_datetime.between?(Time.current, minimum_time)
 
       [
-        Issue.new(:scheduled_time,
+        Issue.new(:schedule_time,
                   :too_close_to_now,
                   time_period: time_period_for_issue(MINIMUM_FUTURE_TIME_PERIOD)),
       ]
@@ -79,7 +79,7 @@ module Requirements
       return [] unless parsed_datetime > maximum_future_time
 
       [
-        Issue.new(:scheduled_date,
+        Issue.new(:schedule_date,
                   :too_far_in_future,
                   time_period: time_period_for_issue(MAXIMUM_FUTURE_TIME_PERIOD)),
       ]
--- a/app/services/requirements/scheduled_datetime_checker.rb
+++ b/app/services/requirements/scheduled_datetime_checker.rb
@@ -4,8 +4,6 @@ module Requirements
   class ScheduledDatetimeChecker
     attr_reader :params
 
-    DATE_FORMAT = "%d-%m-%Y"
-    TIME_FORMAT = "%l:%M%P"
     MAXIMUM_FUTURE_TIME_PERIOD = { months: 14 }.freeze
     MINIMUM_FUTURE_TIME_PERIOD = { minutes: 15 }.freeze
 
@@ -27,10 +25,13 @@ module Requirements
 
     def parsed_datetime
       @parsed_datetime ||= begin
-                             day, month, year = params[:date]&.values_at(:day, :month, :year)
-                             Time.zone.strptime(
-                               "#{day}-#{month}-#{year} #{params[:time]}",
-                               "#{DATE_FORMAT} #{TIME_FORMAT}",
+                             time = time_grouping(params[:time])
+                             Time.current.change(
+                               day: params.dig(:date, :day).to_i,
+                               month: params.dig(:date, :month).to_i,
+                               year: params.dig(:date, :year).to_i,
+                               hour: time[:hour].to_i + (time[:period]&.downcase == "pm" ? 12 : 0),
+                               min: time[:minute].to_i,
                              )
                            end
     end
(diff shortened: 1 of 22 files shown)


app/services/requirements/scheduled_datetime_checker.rb
```diff
@@ -2,76 +2,82 @@
...
+          time_period: time_period_for_issue(MINIMUM_FUTURE_TIME_PERIOD),
+        )
+      end
     end
 
-    def too_far_in_the_future?
-      now = Time.zone.now
-      parsed_datetime > now.advance(MAXIMUM_FUTURE_TIME_PERIOD).end_of_day
-    end
+    def check_datetime_is_foreseeable
```
emmabeynon: I don't think foreseeable is the right word here; it's more about predictability and we've got a concrete future time period that we don't need to predict.

app/views/schedule/scheduled_publishing_datetime.html.erb
```diff
@@ -1,49 +1,53 @@
...
         hint: t("schedule.scheduled_publishing_datetime.date_hint_text"),
-        id: date_input_id,
+        id: "scheduled-date",
+        error_message: error_message.presence,
         items: [
           {
             name: "day",
             width: 2,
-            value: schedule_param["day"] || publish_date&.day
+            value: params.dig(:schedule, :scheduled_date, :day) || publish_time&.day
```
emmabeynon: I wonder if `publish_time` should be renamed `publish_datetime`?  When I first saw the variable I assumed it was the time only, and got confused when I saw it appear as a date value.

app/services/requirements/scheduled_datetime_checker.rb
kevindew: Sure, any suggestions on the wording? I was hoping I could get something with an is rather than isn't and found isnt_too_far_in_the_future a bit wordy.

somethings considered and rejected: is_realistic is_soonish is_sane

app/views/schedule/scheduled_publishing_datetime.html.erb
kevindew: Sure, I'll have a look at that - it's a bit of a pain as a Time object in Ruby always represents a date so that kind of makes sense to name it a time. But I know we also deal with time as a string.

benthorner:
Was just trying to have a look at this - does it need rebasing?

kevindew:
> Was just trying to have a look at this - does it need rebasing?

Just done so

app/views/schedule/scheduled_publishing_datetime.html.erb
```diff
@@ -1,49 +1,53 @@
...
-<% publish_date = current_scheduled_publish_datetime || default_date %>
-<% schedule_param = params["schedule"] ? params["schedule"] : {} %>
-<% selected_time = schedule_param["time"] ||  current_scheduled_publish_datetime&.strftime("%l:%M%P")&.strip %>
-<% action_not_selected = @issues&.select { |i| i.issue_key == :action_not_selected }&.first&.to_item %>
-
+<%
+  current_scheduled_publish_datetime = @edition.scheduled_publishing_datetime
+  default_time = Time.current.tomorrow
+  publish_time = current_scheduled_publish_datetime || default_time
+  action_not_selected = @issues&.select { |i| i.issue_key == :action_not_selected }&.first&.to_item
```
benthorner: Looks like it's not part of this PR, but the requirements system has a `items_for` method especially for this :-).

app/views/schedule/scheduled_publishing_datetime.html.erb
```diff
@@ -1,49 +1,53 @@
...
-<% default_date = Time.zone.now.tomorrow %>
-<% publish_date = current_scheduled_publish_datetime || default_date %>
-<% schedule_param = params["schedule"] ? params["schedule"] : {} %>
-<% selected_time = schedule_param["time"] ||  current_scheduled_publish_datetime&.strftime("%l:%M%P")&.strip %>
-<% action_not_selected = @issues&.select { |i| i.issue_key == :action_not_selected }&.first&.to_item %>
-
+<%
+  current_scheduled_publish_datetime = @edition.scheduled_publishing_datetime
+  default_time = Time.current.tomorrow
+  publish_time = current_scheduled_publish_datetime || default_time
```
benthorner: If we had two variables here - `publish_date` and `publish_time` - could we avoid all the `||`ing we do later on? This might be a follow-up refactor, though.

app/services/requirements/scheduled_datetime_checker.rb
emmabeynon: `within_time_limit`? I appreciate it still has multiple words but I don't think there is a single word to describe this.

app/services/requirements/scheduled_datetime_checker.rb
```diff
@@ -2,76 +2,82 @@
...
 
     def pre_submit_issues
-      issues = []
+      validity_issues = check_datetime_valid
+      return CheckerIssues.new(validity_issues) if validity_issues.any?
 
-      if day.blank? || month.blank? || year.blank?
-        issues << Issue.new(:scheduled_datetime, :invalid, field: "date")
-      end
+      issues = [check_datetime_is_future,
```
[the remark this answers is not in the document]
kevindew: Would you mind grabbing me a good example?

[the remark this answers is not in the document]
benthorner: I think [the attachment checker](https://github.com/alphagov/content-publisher/blob/master/app/services/requirements/file_attachment_checker.rb#L93) is pretty good, as it too has some quite complex checks to deal with. Hope that helps.

app/views/schedule/scheduled_publishing_datetime.html.erb
kevindew: Yup I think I squash this later in this PR

app/views/schedule/scheduled_publishing_datetime.html.erb
kevindew: I think that's mostly unavoidable here as we have both user input and a time object. Would prefer looking at outside of PR.

spec/services/requirements/scheduled_datetime_checker_spec.rb
```diff
@@ -1,101 +1,118 @@
...
-      datetime_params[:time] = "def"
+    it "returns an issue if the datetime is too far into the future" do
+      time_period = { days: 1, months: 14 }
+      future_date = Time.current.advance(time_period)
 
-      issue = I18n.t!("requirements.scheduled_datetime.invalid_input.form_message")
-      issues = Requirements::ScheduledDatetimeChecker.new(datetime_params).pre_submit_issues
+      issues = Requirements::ScheduledDatetimeChecker.new(
+        date: { day: future_date.day,
+                          month: future_date.month,
```
emmabeynon: Same indentation issue here and line 107.

spec/services/requirements/scheduled_datetime_checker_spec.rb
```diff
@@ -1,101 +1,118 @@
...
 
-    it "returns an issue if the date or time fields are blank" do
-      datetime_params[:day] = ""
-      datetime_params[:time] = ""
+    it "returns a date issue if the date is in the past" do
+      two_days_ago = Time.current - 2.days
+
+      issues = Requirements::ScheduledDatetimeChecker.new(
+        date: { day: two_days_ago.day,
+                          month: two_days_ago.month,
```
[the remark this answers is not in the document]
kevindew: Oh good catch!

kevindew:
Just fixed those @emmabeynon

PR #1105, opened 2019-06-05
Search taxonomy

alex-ju:
- Configure a new autocomplete type which uses `millerColumns.taxonomy` as source of data,  highlights matching substrings from the input query and triggers a topic click in the miller columns onConfirm
- Update edit topics page to add a autocomplete with `type: "search-topics"`
- Hide 'No options available' for `search-topics` type as the source of it is coming from JavaScript
- Update SCSS to add styles for highlight

#### Preview
Screen Shot 2019-06-05 at 13 28 22

#### Note 
e5bcbf1 will be replaced with a commit which consumes miller-columns-element via npm after https://github.com/alphagov/miller-columns-element/pull/10 and https://github.com/alphagov/miller-columns-element/pull/11 are merged.

[Trello card](https://trello.com/c/qECJohOD)


app/assets/javascripts/components/autocomplete.js
```diff
@@ -133,5 +135,60 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
...
+      topicSuggestions.push({
+        index: index,
+        topic: breadcrumbs[breadcrumbs.length - 1].trim(),
+        breadcrumbs: breadcrumbs
+      })
+    })
+
+    new window.accessibleAutocomplete({ // eslint-disable-line no-new, new-cap
+      id: 'topics-autocomplete',
+      name: 'topics-autocomplete',
```
kevindew: Can avoid these being hardcoded at all? or take them from an element initialising this?

app/assets/javascripts/components/autocomplete.js
```diff
@@ -133,5 +135,60 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
...
+        },
+        suggestion: function (result) {
+          return result && result.breadcrumbs.join(' › ')
+        }
+      },
+      onConfirm: function (result) {
+        if (result && result.index && !topics[result.index].selected) {
+          millerColumns.taxonomy.topicClicked(topics[result.index])
+        }
+      }
```
kevindew: Ahh ok I see how the allTopicNames fits into this. Can I make a counter suggestion, which fits into how I was pondering this.

syncResults -> runs filter on allTopics and checks topic name, returns an array of topics
template methods -> convert topics into appropriate string
onConfirm -> result is a topic so no need to use index

Does that sound workable? I could be missing a key part where this crashes the autocomplete of course

alex-ju: I'll give it a try

app/assets/javascripts/components/autocomplete.js
alex-ju: The thing with the autocomplete component (the erb bit in our app) is that we're not generating a select element unless we pass options to the component. We're in this case where we're not enhancing on top of a set of suggestions, that's why I hardcoded them (as they're required by the autocomplete).

kevindew: If we're only putting it there because it needs it maybe go for something random?

app/assets/javascripts/components/autocomplete.js
alex-ju: I tried the suggestions. It remove the duplicated source and simplifies the code overall, but requires a bit of logic for the suggestions, as we don't want to edit the topics themselves, so we work with a copy of `topicNames` to do the changes.

app/assets/javascripts/components/autocomplete.js
```diff
@@ -133,5 +135,55 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
...
+              var queryRegex = new RegExp('(' + query + ')', 'ig')
+              result.highlightedTopicName = result.topicName.replace(queryRegex, '<mark>$1</mark>')
+            }
+            return resultContainsQuery
+          }) : []
+        )
+      },
+      templates: {
+        inputValue: function (result) {
+          return result && result.topicName && ''
```
kevindew: What does this do? I'd have thought it always returned '' ?

app/assets/javascripts/components/autocomplete.js
```diff
@@ -133,5 +135,55 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
...
+        },
+        suggestion: function (result) {
+          var suggestions = result.topicNames
+          suggestions[suggestions.length - 1] = result.highlightedTopicName
+          return result && suggestions.join(' › ')
```
kevindew: Also a bit confused by the result && here

app/assets/javascripts/components/autocomplete.js
kevindew: Ah great! I'd suggest rather than modifying the object you return an object like 

`{ topic: result, highlightedTopicName: result.topicName.replace(queryRegex, '<mark>$1</mark>' }`

(or could even extend that to create the template string in advance since it's basically doing that under a different key)

then you could avoid having to remove stuff but would have to do result.topic in other places.

app/assets/javascripts/components/autocomplete.js
alex-ju: Default to an empty string as an input value (if `result.topicName` doesn't return anything), which enables the autocomplete to be cleared after selection.

kevindew: Not sure if I'm missing something but this will return an empty string even if topicName does return something won't it.

e.g. 

```
"this" && "that" && ""
> ""
```

```
"this" && "that" && ""
> ""

```
"this" && null && ""
> null
```

File: app/assets/javascripts/components/autocomplete.js (hunk 4 above)
Comment (alex-ju) [reply app/assets/javascripts/components/autocomplete.js]: I'm using conditional statement in boolean algebra consistent with other custom autocompletes and examples. We can expand it as a conditional expression, but I'll have to update the rest of custom inits too.

Essentially if we have a result we return the suggestions as a string of breadcrumbs items.

Comment (kevindew) [reply app/assets/javascripts/components/autocomplete.js]: but won't that already error as you've tried to call `.topicNames` on result earlier?

File: app/assets/javascripts/components/autocomplete.js (hunk 1 above)
Comment (alex-ju) [reply app/assets/javascripts/components/autocomplete.js]: We could add [a helper function to deal with this](https://github.com/alphagov/govuk-frontend/blob/ad596de756ef51619c6e5bc8f22f3b93479773dd/src/common.js#L18), I didn't thought we're expecting to have more than one search topics per page.

File: app/assets/javascripts/components/autocomplete.js (hunk 4 above)
Comment (alex-ju) [reply app/assets/javascripts/components/autocomplete.js]: Yep, that's true, I forgot we expanded the logic within this function.

File: app/assets/javascripts/components/autocomplete.js (hunk 3 above)
Comment (alex-ju) [reply app/assets/javascripts/components/autocomplete.js]: Right, we don't need to check for `result` in our case, only to return the empty string.

File: app/assets/javascripts/components/autocomplete.js (hunk 2 above)
Comment (alex-ju) [reply app/assets/javascripts/components/autocomplete.js]: I would avoid creating a new array of objects for every typed letter, especially since there will be a lot of repetition in terms of matched results.

Comment (alex-ju) [reply app/assets/javascripts/components/autocomplete.js]: Updated as discussed, to re-create the `topicSuggestions` source and operate changes over it.

File: app/assets/javascripts/components/autocomplete.js [hunk 5]
```diff
@@ -133,5 +135,66 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
...
+      name: 'topics-autocomplete',
+      element: this.$module,
+      minLength: 3,
+      autoselect: false,
+      source: function (query, syncResults) {
+        var results = topicSuggestions
+
+        syncResults(query
+          ? results.filter(function (result) {
+            var indexOf = result.topic.topicName.toLowerCase().indexOf(query.toLowerCase())
```
kevindew: minor, but could be nice to set a variable of topicName so we can reduce this method chain?

By the way, this is all looking good now 👍

app/assets/javascripts/components/autocomplete.js
alex-ju: Updated to grab the id and name from as it's now enhancing an input element

alex-ju:
@kevindew swapped vendor files with code from npm and added a few SCSS adjustments needed for the font baseline on label and arrow.

kevindew:
Just given this a play, looks really neat 👍 Nice that the search bar isn't present without JS.

I noticed a bug when selecting an item where a child of it is already selected. If you have say "Regional and local government > Local government > Data collection and reporting" and then search and select "Regional and local government > Local government" you'll lose the "Local government selection"

Also it'd be nice if it could be part of a feature test if that's not super hard.

alex-ju:
@kevindew rebased again, feature test added, feature flag added.

app/views/components/_autocomplete.html.erb
```diff
@@ -44,7 +44,7 @@
         multiple: multiple
       %>
     <% else %>
-      <%= tag.p "No options available", class: "govuk-hint" %>
+      <%= tag.p "No options available", class: "govuk-hint" unless type=='search-topics' %>
```
kevindew: Minor, `type == 'search-topics' `

alex-ju: Ah, that bit of code is outdated. We're only showing this message for type `multiple`, which is more adequate.

PR #1110, opened 2019-06-06
Notify users about scheduled publishing

1pretz1:
We need to send users who have edited an edition a notification that their scheduled content has been published.

Trello:
https://trello.com/c/utiuH5Q9/736-notify-users-that-scheduled-content-has-been-published

### DIFF
--- a/app/jobs/scheduled_publishing_job.rb
+++ b/app/jobs/scheduled_publishing_job.rb
@@ -9,15 +9,16 @@ class ScheduledPublishingJob < ApplicationJob
   discard_and_log(ActiveRecord::RecordNotFound)
 
   def perform(id)
-    Edition.find_and_lock_current(id: id) do |edition|
-      next if no_longer_schedulable?(edition)
+    published_edition = Edition.find_and_lock_current(id: id) do |edition|
+      return if no_longer_schedulable?(edition) # rubocop:disable Lint/NonLocalExitFromIterator
 
       user = edition.status.created_by
       reviewed = edition.status.details.reviewed
-
       PublishService.new(edition)
                     .publish(user: user, with_review: reviewed)
     end
+
+    ScheduledPublishMailer.success_email(published_edition, user).deliver_later
   end
 
 private
--- a/app/jobs/scheduled_publishing_job.rb
+++ b/app/jobs/scheduled_publishing_job.rb
@@ -18,7 +18,7 @@ class ScheduledPublishingJob < ApplicationJob
                     .publish(user: user, with_review: reviewed)
     end
 
-    ScheduledPublishMailer.success_email(published_edition, user).deliver_later
+    send_notifications(published_edition)
   end
 
 private
@@ -34,6 +34,12 @@ private
     if schedule > Time.zone.now
       Rails.logger.warn("Cannot publish an edition (\##{edition.id}) scheduled in the future")
       return true
+    end
+  end
+
+  def send_notifications(edition)
+    edition.editors.each do |editor|
+      ScheduledPublishMailer.success_email(edition, editor).deliver_later
     end
   end
 end
--- a/app/mailers/scheduled_publish_mailer.rb
+++ b/app/mailers/scheduled_publish_mailer.rb
@@ -1,0 +1,31 @@
+# frozen_string_literal: true
+
+class ScheduledPublishMailer < ApplicationMailer
+  self.delivery_job = EmailDeliveryJob
+
+  add_template_helper(EditionUrlHelper)
+
+  def success_email(edition, user)
+    @user = user
+    @edition = edition
+    @status = edition.status
+
+    unless @status.published? || @status.published_but_needs_2i?
+      raise "Cannot send publish email with a non-published state"
+    end
+
+    mail(to: user.email, subject: subject)
+  end
+
+private
+
+  def subject
+    if @edition.number > 1
+      I18n.t("scheduled_publish_mailer.success_email.subject.update",
+             title: @edition.title)
+    else
+      I18n.t("scheduled_publish_mailer.success_email.subject.#{@status.state}",
+             title: @edition.title)
+    end
+  end
+end
--- a/spec/jobs/scheduled_publishing_job_spec.rb
+++ b/spec/jobs/scheduled_publishing_job_spec.rb
@@ -10,12 +10,16 @@ RSpec.describe ScheduledPublishingJob, inline: true do
       publish_service = double(:publish_service, publish: nil)
 
       expect(PublishService).to receive(:new).with(edition) { publish_service }
+      message_delivery = double(ActionMailer::MessageDelivery, deliver_later: nil)
 
       expect(publish_service)
         .to receive(:publish)
         .with(user: user, with_review: edition.status.details.reviewed)
 
-      ScheduledPublishingJob.new.perform(edition.id)
+      expect(ScheduledPublishMailer).to receive(:success_email).with(edition, user)
+                                                               .and_return(message_delivery)
+
+      ScheduledPublishingJob.perform_now(edition.id)
     end
 
     it "aborts the job if the edition does not exist (e.g. env sync)" do
--- a/spec/jobs/scheduled_publishing_job_spec.rb
+++ b/spec/jobs/scheduled_publishing_job_spec.rb
@@ -2,24 +2,45 @@
 
 RSpec.describe ScheduledPublishingJob, inline: true do
   describe "#perform" do
-    it "calls PublishService if edition can be published" do
-      edition = create(:edition,
-                       :scheduled,
-                       scheduled_publishing_datetime: Time.current)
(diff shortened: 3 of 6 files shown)


app/views/scheduled_publish_mailer/publish_email.text.erb
```diff
@@ -0,0 +1,35 @@
+<%= @edition.document_type.label %> ‘<%= @edition.title %>’
+
+<% if @edition.number > 1 -%>
+<%= t("scheduled_publish_mailer.publish_email.details.update",
+  time: @status.created_at.strftime("%l:%M%P").strip,
```
kevindew: If we use a formatting of `"%-l:%M%P"` we don't need to do the strip

app/jobs/scheduled_publishing_job.rb
```diff
@@ -9,15 +9,19 @@ class ScheduledPublishingJob < ApplicationJob
...
   def perform(id)
-    Edition.find_and_lock_current(id: id) do |edition|
-      next if no_longer_schedulable?(edition)
+    edition = Edition.find_current(id: id)
+    user = edition.status.created_by
 
-      user = edition.status.created_by
-      reviewed = edition.status.details.reviewed
+    Edition.transaction do
+      edition.lock!
```
kevindew: I'd suggest we set edition here with `edition = edition.find_and_lock_current(id: id)` this'll still be in scope after the transaction block.

app/mailers/scheduled_publish_mailer.rb
```diff
@@ -0,0 +1,31 @@
+# frozen_string_literal: true
+
+class ScheduledPublishMailer < ApplicationMailer
+  self.delivery_job = EmailDeliveryJob
+
+  add_template_helper(EditionUrlHelper)
+
+  def publish_email(edition, user)
```
kevindew: Might want to call this `success_email` since we're likely to have a `failure_email` in this class?

spec/jobs/scheduled_publishing_job_spec.rb
```diff
@@ -10,12 +10,41 @@
...
+
+      publish_service = double(:publish_service, publish: nil)
+      message_delivery = double(ActionMailer::MessageDelivery, deliver_later: nil)
+
+      expect(PublishService).to receive(:new).with(edition) { publish_service }
+
+      edition.editors.each do |editor|
+        expect(ScheduledPublishMailer).to receive(:publish_email).with(edition, editor)
+                                                                 .and_return(message_delivery)
+      end
```
kevindew: Rather than using the edition.editors I think it'd be better to check each editor - as we're not actually testing here that each editor gets it. If editors returned an empty array we'd never run these expectations.

I'd set an`allow(ScheduledPublishMailer).to receive(:publish_email).and_return(message_delivery)` first

and then just have an expectation for each editor.

spec/jobs/scheduled_publishing_job_spec.rb
```diff
@@ -10,12 +10,41 @@
...
+      ScheduledPublishingJob.perform_now(edition.id)
+    end
+
+    it "calls ScheduledPublishMailer for each editor of the edition" do
+      editor_one = create(:user, email: "someone@example.com")
+      metadata_revision = create(:metadata_revision, scheduled_publishing_datetime: Time.current)
+      editor_two = create(:user, email: "someone-else@example.com")
+      revision = create(:revision,
+                        created_by: editor_two,
+                        metadata_revision: metadata_revision)
```
kevindew: I think you can set `scheduled_publishing_datetime` here rather than needing to create a metadata_revision.

benthorner:
I'm wondering how the ScheduleMailer is different from the PublishMailer. The ScheduleMailer talks a lot about publishing, and not so much about scheduling; would it be possible to use the PublishMailer?

app/views/scheduled_publish_mailer/publish_email.text.erb
1pretz1: Sure, good spot!

app/jobs/scheduled_publishing_job.rb
1pretz1: Done, sort of! I've gone with the normal approach `Edition.find_and_lock_current(id: ..) { ... }` and just assigned the return to a variable, we can't actually call `find_and_lock_current` without passing in a block.

app/mailers/scheduled_publish_mailer.rb
1pretz1: Yep, good shout!

spec/jobs/scheduled_publishing_job_spec.rb
1pretz1: Yep I agree, done!

spec/jobs/scheduled_publishing_job_spec.rb
1pretz1: You're correct!

1pretz1:
@kevindew 
Do we just have the one rogue line about "scheduled_by" that differentiates this from publish email?

Nope, there's another subtle difference being whether the document has been first published or updated. The difference is the presence of the user in the sentence, publish email includes who it was published / updated by, whereas the scheduled publish email doesn't as this is taken care of by the sentence underneath. If that was the case I would've probably combined the templates and just added a conditional. See below.

### Scheduled publish email
Screen Shot 2019-06-06 at 15 30 54

### Publish email
Screen Shot 2019-06-06 at 15 31 45

@benthorner 
The ScheduledPublishMailer and PublishMailer are very similar, I first went down the road of combining  the two but as the need for two distinct templates / locales emerged I decided to create a separate mailer for scheduled publishing. I could potentially create another action in the PublishMailer which dealt with scheduled publishing and maybe combine the locales but it starts to feel a little messy?

benthorner:
@1pretz1 thanks for explaining - that makes sense. The only thing I'm not sure we need is the spec for the mailer, since the PublishMailer seems to get by without one, and this doesn't affect our coverage. Is there a particular edge case it's covering?

1pretz1:
@benthorner The actual email produced by PublishMailer is inadvertently tested by the `publish_spec.rb` feature in the workflow directory, the conditional logic in here isn't tested however. 

Seen as though we can't do the same sort of testing for a scheduled publishing, the only reasonable place to test the output of the email was to create a spec for the mailer. I also added a test for some of the logic because why not! :)

benthorner:
@kevindew oh yes, I see what you mean there as well. This does look like the best place to test this, but I'm not sure we should write tests just because we write code - look what happened to Whitehall!

Sorry for the somewhat iterative review here - I should probably just talk to you about approach-level stuff before writing comments. Now that's all clear, I've written some comments that might actually be useful ;-).

spec/jobs/scheduled_publishing_job_spec.rb
```diff
@@ -10,12 +10,38 @@
       publish_service = double(:publish_service, publish: nil)
 
       expect(PublishService).to receive(:new).with(edition) { publish_service }
+      message_delivery = double(ActionMailer::MessageDelivery, deliver_later: nil)
```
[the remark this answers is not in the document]
benthorner: Ooo...maybe we should add this (perhaps as separate PR, though!)...

https://github.com/varvet/rule_io/blob/master/.rubocop.yml#L682

spec/jobs/scheduled_publishing_job_spec.rb
```diff
@@ -10,12 +10,38 @@
...
+      allow(ScheduledPublishMailer).to receive(:success_email).and_return(message_delivery)
 
+      expect(ScheduledPublishMailer).to receive(:success_email).with(edition, user)
```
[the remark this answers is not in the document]
1pretz1: Yep good point!

spec/jobs/scheduled_publishing_job_spec.rb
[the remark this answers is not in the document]
1pretz1: Instance doubles are now implemented! 🎉And yes I think the check would be a good thing to have.

spec/jobs/scheduled_publishing_job_spec.rb
```diff
@@ -10,12 +10,38 @@
...
+      revision = create(:revision,
+                        created_by: editor_two,
+                        scheduled_publishing_datetime: Time.current)
+      edition = create(:edition,
+                       :scheduled,
+                       revision: revision,
+                       created_by: editor_one)
+
+      publish_service = double(:publish_service, publish: nil)
+      message_delivery = double(ActionMailer::MessageDelivery, deliver_later: nil)
```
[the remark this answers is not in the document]
1pretz1: Okay sure! I've separated the spec into two `contexts`, one for when the jobs succeeds and the other for when it fails. I've put the common code for the doubles into a before block which has made the spec more dry. :)

1pretz1:
@benthorner I forgot to notify you that i've made the changes you've requested... here it is!

PR #1113, opened 2019-06-06
Create Clear Datetime interactor

emmabeynon:
For https://trello.com/c/bFymOdvj/850-allow-updating-and-clearing-of-a-proposed-scheduling-date

This extracts the logic for clearing a scheduled publishing datetime into an
interactor.  We have made a content decision that we no longer need to create a
timeline entry when a scheduled publishing datetime is clearer as this is an
edit. This is consistent with not creating a timeline entry when a scheduled
publishing datetime is updated.  We have also decided not to check if a
scheduled publishing datetime is present before clearing it, as if it's not
present the updater won't assign a new revision as it won't have changed.

### DIFF
--- a/app/controllers/schedule_controller.rb
+++ b/app/controllers/schedule_controller.rb
@@ -28,17 +28,10 @@ class ScheduleController < ApplicationController
   end
 
   def clear_scheduled_publishing_datetime
-    Edition.find_and_lock_current(document: params[:document]) do |edition|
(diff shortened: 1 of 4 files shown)


emmabeynon:
@kevindew now done :-)

benthorner:
Should we consider removing [this](https://github.com/alphagov/content-publisher/blob/master/config/locales/en/documents/history.yml#L28) as well? I guess it might cause a translation issue on integration if someone has done a 'clear' today, but the data sync should get rid of that.

PR #1119, opened 2019-06-10
Update autocomplete component to allow empty input

alex-ju:
This PR updates the vendored autocomplete code to the latest version which contains an update to allow empty inputs (or cleared inputs). For this to be working fully it needs that the `<select>` element that is enhanced to contain an option with an empty value and any label `<option value="">No option selected</option>`.

[Review in the component guide](https://content-publisher-revi-pr-1119.herokuapp.com/component-guide/autocomplete)

[Trello card](https://trello.com/c/4Uc4JL4T)


kevindew:
@alex-ju Do you know where the unminified code is for this? I did some investigating of this myself so would be good to compare notes.

alex-ju:
@kevindew I'm having trouble pushing to the accessible-autocomplete repo atm, will try and fix that and update the PR with a diff on the source code.

kevindew:
> @kevindew I'm having trouble pushing to the accessible-autocomplete repo atm, will try and fix that and update the PR with a diff on the source code.

Oh I wonder if that is related to the govuk-github team deletion from last week? I can't push my branch to this either. Probably worth emailing senior tech to sort out. I also don't seem to have miller-columns-element access anymore but still seem to have paste-html-to-govspeak access.

alex-ju:
@kevindew. Yep, I think that's the cause. There are a few missing repos from GOV.UK groups.

alex-ju:
This is already captured by https://github.com/alphagov/accessible-autocomplete/commit/b3d0e9b2009fed92c6bc491010f0cf753bb6a2ac

PR #1120, opened 2019-06-10
Reschedule

benthorner:
https://trello.com/c/5n7oZvXZ/851-change-a-scheduled-publishing-date

This enables changing the scheduled publishing date/time for a scheduled edition. Some of the tests have been refactored to better distinguish between the new feature and changing a proposed schedule, as well as make both sets consistent. Please see the commits for more info...


spec/features/scheduling/change_proposed_schedule_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+    when_i_visit_the_summary_page
+    and_i_click_on_change_date
+    then_i_see_the_current_date_and_time_selected
+
+    when_i_set_a_new_proposed_schedule_datetime
+    then_i_see_the_new_proposed_schedule_datetime
+  end
+
+  def given_there_is_an_edition_with_a_proposed_schedule
+    @current_datetime = Time.current.tomorrow.change(hour: 10)
```
kevindew: minor, but how would you feel about avoiding `current` here and instead something like `proposed_time`. Just that I read current as now.

kevindew: Also apologies, I've noticed that you didn't introduce this so don't worry if it's a pain.

spec/features/scheduling/schedule_spec.rb
```diff
@@ -43,7 +43,7 @@ def then_i_see_the_edition_has_been_scheduled
     expect(job[:args].first).to eq @edition.id
     expect(job[:at].to_i).to eq @datetime.to_i
 
-    assert_requested @request
+    expect(@request).to have_been_requested
```
kevindew: 👍

app/services/schedule_service.rb
```diff
@@ -8,11 +8,27 @@ def initialize(edition)
...
+  end
+
+  def reschedule(user: nil)
+    unless edition.scheduled?
+      raise "Edition must be scheduled in order to reschedule"
+    end
+
+    scheduling = edition.status.details
+    set_edition_status(scheduling, user)
+    create_timeline_entry(scheduling)
```
kevindew: I'd have thought we'd have a different timeline entry for updating it.

app/services/schedule_service.rb
```diff
@@ -8,11 +8,27 @@ def initialize(edition)
...
+
+    create_publish_intent
```
kevindew: Are we using create to replace since the eventual request is a put? If so I wonder if we'd want to rename the method.

app/services/schedule_service.rb
benthorner: Sure - it makes sense to me both ways, but the flexibility is good.

app/services/schedule_service.rb
benthorner: `create_or_update...` looked weird to me, so I've gone with an alias. Would you prefer just a method name change?

spec/features/scheduling/change_proposed_schedule_spec.rb
benthorner: It didn't bother me particularly, but I've gone with just plain 'datetime', which hopefully stops people making any assumptions about it ;-).

benthorner:
@kevindew I think that's all your comments addressed. The one about the timeline entry made me spot that we're not testing these in the feature tests, so I added it for a couple.

PR #1121, opened 2019-06-10
Standardise tracking for MD toolbar buttons

benthorner:

    Previously we set the same data-gtm attribute for all MD toolbar
    buttons, which was used to scope the associated GTM trigger. This
    modifies the data-gtm attributes to consistently reflect the action
    performed, making events for these elements equivalent to other
    data-gtm actions, avoiding premature grouping. Using the common
    'markdown-toolbar' prefix to scope the GTM trigger should suffice.

### DIFF
--- a/app/views/components/_markdown_editor.html.erb
+++ b/app/views/components/_markdown_editor.html.erb
@@ -82,22 +82,22 @@
       </div>
       <div class="app-c-markdown-editor__toolbar">
         <markdown-toolbar class="app-c-markdown-editor__toolbar-group" for="<%= textarea[:id] %>">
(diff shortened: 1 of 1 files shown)


benthorner:
@kevindew I didn't encounter that issue when I tried it. Presumably the extensions work the same as a normal button with it's content 🤷‍♂.

kevindew:
> @kevindew I didn't encounter that issue when I tried it. Presumably the extensions work the same as a normal button with it's content 🤷‍♂.

Great thanks for checking it 👍It's possible I'm muddling memories.

PR #1123, opened 2019-06-10
Replace Edition#schedulable? with inline checks

benthorner:

Previously we had a schedulable? method on each edition that was used
for sanity checking in a couple of controller actions, as well as to
modify the actions shown to the user. The latter behaviour conflicts
with the new design for schedule requirements, as the 'Schedule' action
would disappear if the proposed time drifted into the past.

This replaces the schedulable? check for user actions by simply checking
the scheduled_publishing_datetime, noting the edition is editable in
both scenarios. Since it is valid to enqueue a job in the past (?!),
this also inlines and simplifies the schedulable? check when scheduling
a document, and removes it when harmlessly rendering the form.

### DIFF
--- a/app/controllers/schedule_controller.rb
+++ b/app/controllers/schedule_controller.rb
@@ -3,12 +3,6 @@
 class ScheduleController < ApplicationController
   def new
     @edition = Edition.find_current(document: params[:document])
-
-    unless @edition.schedulable?
-      # FIXME: this shouldn't be an exception but we've not worked out the
-      # right response - maybe bad request or a redirect with flash?
-      raise "Document is not in a schedulable state."
-    end
   end
 
   def edit
--- a/app/interactors/schedule/create_interactor.rb
+++ b/app/interactors/schedule/create_interactor.rb
@@ -24,7 +24,7 @@ private
   end
 
   def check_for_issues
-    unless edition.schedulable?
+    unless edition.editable? && edition.scheduled_publishing_datetime.present?
       # FIXME: this shouldn't be an exception but we've not worked out the
       # right response - maybe bad request or a redirect with flash?
       raise "Can't schedule an edition which isn't schedulable"
--- a/app/models/edition.rb
+++ b/app/models/edition.rb
@@ -130,13 +130,6 @@ class Edition < ApplicationRecord
     !live? && !scheduled?
   end
 
-  def schedulable?
-    return false unless editable?
-    return false if scheduled_publishing_datetime.nil?
-
-    scheduled_publishing_datetime > Time.zone.now.advance(MINIMUM_SCHEDULING_TIME)
-  end
-
   def resume_discarded(live_edition, user)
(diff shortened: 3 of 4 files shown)


app/interactors/schedule/create_interactor.rb
```diff
@@ -24,7 +24,7 @@ def find_and_lock_edition
   end
 
   def check_for_issues
-    unless edition.schedulable?
+    unless edition.editable? && edition.scheduled_publishing_datetime.present?
```
kevindew: I think we'd need something here that checks the date is correct - though could be considered part of https://trello.com/c/a4sDMtqD/844-ensure-publishers-meet-requirements-before-they-schedule

app/controllers/schedule_controller.rb
```diff
@@ -3,12 +3,6 @@
 class ScheduleController < ApplicationController
   def new
     @edition = Edition.find_current(document: params[:document])
-
-    unless @edition.schedulable?
-      # FIXME: this shouldn't be an exception but we've not worked out the
-      # right response - maybe bad request or a redirect with flash?
-      raise "Document is not in a schedulable state."
-    end
```
kevindew: I'd have thought this still needs something so users can't just access this via URL.

app/interactors/schedule/create_interactor.rb
benthorner: Agreed it's a bit unusual - perhaps a full-blown requirements checker for this particular action would be the way to go here? I guess we want to force users down the path to success, even though submitting with a date in the past doesn't actually cause an error?!

app/controllers/schedule_controller.rb
benthorner: Ah, my hope here was that we'd agree this wasn't particularly important, since it's just a GET action. If it helps, I don't think we have this kind of check on any other actions - just for modifier ones. Is there something we're particularly concerned about here?

app/interactors/schedule/create_interactor.rb
kevindew: Let's cover this when we chat about error stuff tomorrow. We can start by looking at whether we now know everything we need for the requirements card.

app/controllers/schedule_controller.rb
kevindew: I'd imagine if you were to go to this route without a scheduled_publishing_datetime value you'd get an exception. Whereas I think if you went back to this form after scheduling with the back button you'd end up seeing the form again even if you're scheduled.

Most consistent approach we have is raising an exception when users access pages outside of the state they're supposed to be in. I think removing this will make https://trello.com/c/8PsWxNzY/651-audit-how-we-cope-with-duplicate-user-actions-eg-in-two-tabs harder

benthorner:
Discussed with @kevindew and we've decided to merge this and address how we do checks in future work.

PR #1124, opened 2019-06-10
Revisit analytics for radio buttons

benthorner:
https://trello.com/c/fjVo0aK9/877-analytics-for-scheduling

Previously we had a custom event listener for submitting a form with
radio buttons. While this gave us accurate submission data, it seems
redundant to have a bespoke data capture method for certain forms when
this and other data are available in the database for the app.

This replaces the previous listener with a data-gtm approach on the
individual radio buttons, the aim being to capture an event for every
radio selection made. One use of this data could be to monitor the
effectiveness of the options for e.g. creating a new document.


benthorner:
@kevindew no worries - we don't seem to be using this data anywhere. I can't find any reference to specific document types on our DataStudio dashboards.

I was a bit worried about losing this, since the new version isn't comparable with the old version (e.g. if a user switches selections), but  on the whole, I don't think we should be scraping our own form data when we have a database of it, so pushing on...

PR #1126, opened 2019-06-11
Notify users of failed scheduled publishing

1pretz1:
We need to send users who have edited an edition a notification that
their scheduled content has failed publish. 

The retry limit for the publishing of a scheduled edition is 5 times before a failed schedule
publishing email will be sent to all the editors of that edition.

I have a question around one of the A/C on the trello card  - 'Notify Sentry when the email fails to send'. Looking at the [ADR](https://github.com/alphagov/content-publisher/blob/master/docs/adr/0010-sending-email-notifications.md) for sending email notifications it states, 

`We already use Raven to report errors to Sentry, which automatically integrates with Sidekiq. Note that Raven also integrates with Active Job, but this is disabled when Sidekiq is present. Active Job catches exceptions as part of its retry behaviour, so an error will only get reported to Sentry when the exception is not handled by Active Job, or the retries for an exception we do handle are exhausted`.

In this PR ActiveJob handles the exception thrown if a publishing fails and after all the retries have been exhausted executes some logic to send failure emails, which I assume from reading the above a sentry notification would not be sent but don't really know how to test this? 

Trello:
https://trello.com/c/Bdccoiqd/735-notify-users-that-scheduled-content-has-failed-to-publish

Content for email:
https://docs.google.com/document/d/1K5Ny-wSdpUpiK79ei1NfQrE2DF63EiS6Zi4_ptg-390/edit#

### DIFF
--- a/spec/jobs/scheduled_publishing_job_spec.rb
+++ b/spec/jobs/scheduled_publishing_job_spec.rb
@@ -23,7 +23,7 @@ RSpec.describe ScheduledPublishingJob, inline: true do
         ScheduledPublishingJob.perform_now(edition.id)
       end
 
-      it "calls ScheduledPublishMailer for each editor of the edition" do
+      it "calls ScheduledPublishMailer's success_email method for each editor of the edition" do
         editor_one = create(:user, email: "someone@example.com")
         editor_two = create(:user, email: "someone-else@example.com")
         revision = create(:revision,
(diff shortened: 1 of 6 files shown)


spec/jobs/scheduled_publishing_job_spec.rb
```diff
@@ -23,7 +23,7 @@
         ScheduledPublishingJob.perform_now(edition.id)
       end
 
-      it "calls ScheduledPublishMailer for each editor of the edition" do
+      it "calls ScheduledPublishMailer's success_email method for each editor of the edition" do
```
benthorner: Minor, but I think we can afford to be more concise here. How about 'delegates to the ScheduledPublishMailer for each editor'

spec/mailers/scheduled_publish_mailer_spec.rb
```diff
@@ -72,4 +72,24 @@
       )
     end
   end
+
+  describe "sending an email when scheduled edition has failed to published" do
```
benthorner: Minor, but I think it's more conventional to describe method names for unit tests like these e.g. `describe "#failure_email"`.

kevindew:
This is superseded by https://github.com/alphagov/content-publisher/pull/1140 which incorporates this.

PR #1132, opened 2019-06-13
Fix parsing of time

kevindew:
This was having problems with times around the hour of 12 in both am and
pm. This was due to the previous code (which I wrote 😭) being a bit too
simplistic with time handling and handling every hour but 12. There are
now extra tests to catch this.

I've also added some extra tests to make clear some of the time
scenarios that should be rejected, these are mostly types of parsing
that are more complex than this time should accept but `Time.parse`
would.

@benthorner I imagine there is a crossover with this and what you've been working on so I guess we may need to co-ordinate somewhat.

### DIFF
--- a/app/services/requirements/schedule_datetime_checker.rb
+++ b/app/services/requirements/schedule_datetime_checker.rb
@@ -25,13 +25,13 @@ module Requirements
 
     def parsed_datetime
       @parsed_datetime ||= begin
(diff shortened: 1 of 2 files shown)


benthorner:
@kevindew the main change looks simple enough, so it shouldn't be much work to incorporate it :-).

kevindew:
@benthorner great - I'll merge away

PR #1035, opened 2019-05-15
Enable email links in markdown toolbar

alex-ju:
This PR:
- updates markdown-toolbar-element to [0.2.0](https://github.com/alphagov/markdown-toolbar-element/pull/9) which updates the link button to render email markdown when the selected text matches a basic email pattern (name@domain.tld).
- removes email guidance from sidebar

Demo: https://content-publisher-revi-pr-1035.herokuapp.com/component-guide/markdown_editor/default/preview

[Trello card](https://trello.com/c/W1EYHDBn)

PR #1133, opened 2019-06-13
Add back the guidance for email markdown

alex-ju:
We recently [removed the email markdown link](https://github.com/alphagov/content-publisher/pull/1035/commits/44efa4fcbea2ac105e077683bb04963e7f73c16b) from the  sidebar and [updated the link button so it can be used for email](https://github.com/alphagov/markdown-toolbar-element/pull/8). During research, most users did not use the link button and tried to find the markdown instructions.

In this PR we want re-add this instruction while we explore other ways of helping users add email correctly.

[Trello card](https://trello.com/c/srCexKrZ)

### DIFF
--- a/app/views/components/_markdown_guidance.html.erb
+++ b/app/views/components/_markdown_guidance.html.erb
@@ -50,11 +50,13 @@
         data_attributes: { gtm: "markdown-guidance-details" },
         margin_bottom: 0
   } do %>
-    <p>Select an email address and use the Link button to make a Mailto link.</p>
-    <p>Or you can add < and > before and after an email address.</p>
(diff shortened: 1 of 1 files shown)


alex-ju:
> @alex-ju could you update this based on the latest GTM change...
> 
> https://github.com/alphagov/content-publisher/pull/1135/files#diff-0b8a2be908ee86edd0b24da1ca39dcedR21

Yep, I wasn't sure who was going to merge first.

alex-ju:
> Looks good - wonder why we don't have these in a translation file

I wondered the same. Happy to do a stretch on this to update all.

PR #1134, opened 2019-06-13
More schedule requirements

benthorner:

This ensures the user can see publishing issues (e.g. no summary) and scheduling issues (e.g. publish time has drifted into the past) on the document summary, and that they meet both before they attempt to schedule a document. Part of this involved a refactoring of the existing ScheduleDatetimeChecker - please see the commits for more details.


app/services/requirements/schedule_checker.rb
```diff
@@ -0,0 +1,45 @@
...
+      @revision = revision
+    end
+
+    def pre_schedule_issues
+      issues = []
+
+      if publish_at > MAX_PUBLISH_DELAY.from_now
+        issues << Issue.new(:schedule_date,
+                            :too_far_in_future,
+                            time_period: distance_of_time_in_words(MAX_PUBLISH_DELAY))
```
kevindew: Is this where we get "about a year" ?

I don't really think we should use the `distance_of_time_in_words` helper as any changes to the constants may make it's outputs rather confusing which then means we may as well have hardcoded the value as that'd at least be clearer that something had to change with the number.

spec/services/datetime_parser_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+    end
+
+    context "when the time is invalid" do
+      it "throws an exception with the issue" do
+        params = { date: { day: "10", month: "1", year: "2019" }, time: "13421" }
+
+        expect { DatetimeParser.new(params).parse }.to raise_error do |error|
+          form_message = error.issues.items_for(:schedule_time).first[:text]
+          expect(form_message).to eq(I18n.t!("requirements.schedule_time.invalid.form_message"))
+        end
```
kevindew: It'd be good to have an example here that reflects a value that would be bad if Time.parse was used. 

It's easy to think "why didn't they just use `Time.parse`" without consideration of just how broad a range of values could be accepted and cause odd outcomes.

app/controllers/schedule_controller.rb
```diff
@@ -2,14 +2,13 @@
 
 class ScheduleController < ApplicationController
   def new
-    @edition = Edition.find_current(document: params[:document])
-
-    issues = Requirements::EditionChecker.new(@edition)
-                                         .pre_publish_issues(rescue_api_errors: false)
+    result = Schedule::NewInteractor.call(params: params)
```
kevindew: woa, do we have our first GET interactor?

app/views/documents/show/_actions.html.erb
```diff
@@ -1,17 +1,17 @@
...
-    <% elsif @edition.status.published? %>
+    <% elsif @edition.published? %>
       <%= create_edition_button(@edition) %>
       <%= withdraw_link(@edition) %>
       <%= remove_link(@edition) %>
-    <% elsif @edition.status.removed? %>
+    <% elsif @edition.removed? %>
       <%= create_edition_button(@edition) %>
-    <% elsif @edition.status.submitted_for_review? %>
+    <% elsif @edition.submitted_for_review? %>
```
kevindew: ⭐️ for this and the ones above

spec/services/datetime_parser_spec.rb
benthorner: Oof #notmyspec ;-). But yes, this did cross my mind when I first read the implementation - I assumed the '1234' example did that already. I've gone with a different one.

app/interactors/schedule_proposal/update_interactor.rb
```diff
@@ -23,39 +26,43 @@ def find_and_lock_edition
     context.edition = Edition.lock.find_current(document: params[:document])
   end
 
-  def check_for_issues
+  def parse_publish_time
+    context.publish_time = DatetimeParser.new(schedule_params).parse
+  rescue DatetimeParser::ParseError => e
+    context.fail!(issues: e.issues)
```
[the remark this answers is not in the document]
benthorner: So I did try this kind of value-driven approach, but my initial attempt seemed harder to work with. I've tried again and it seems better now - what do you think?

app/services/datetime_parser.rb
```diff
@@ -0,0 +1,88 @@
+# frozen_string_literal: true
+
+class DatetimeParser
```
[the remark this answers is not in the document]
benthorner: I'm afraid I don't agree here. Just to make sure we both have the same knowledge: we do have [places where we use issues](https://github.com/alphagov/content-publisher/blob/master/app/interactors/publish/publish_interactor.rb#L35) outside of the requirements namespace.

For me, the most important thing is that this is not principally a checker - that's what I found unintuitive about the original implementation. I wouldn't expect to find anything but a checker in requirements; or to put it the other way round, this seems like it's providing a generic service to the app and not a requirements-specific one.

I guess I could also say 'this feels like it should like inside the models directory since it's rather coupled to...' ;-).

app/services/datetime_parser.rb
```diff
@@ -0,0 +1,88 @@
...
+  class ParseError < StandardError
+    attr_reader :issues
+
+    def initialize(issues)
+      @issues = issues
+    end
+  end
+
+  def initialize(params)
+    @params = params
```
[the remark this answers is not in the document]
benthorner: Fair enough - I'm curious to see our first dual time form. Updated.

app/services/requirements/schedule_checker.rb
benthorner: Agreed, that did annoy me in the tests, although I would prefer not to go back to the original either, as it was somewhat confusing to me. Fortunately, I think I found an alternative - see what you think :-).

benthorner:
@kevindew I think I've addressed all your comments now, except for the one where where I haven't 😜 .

app/services/requirements/schedule_checker.rb
kevindew: That seems good 👍 are they covered by a test? 

My only concern would be that we're relying on the ActiveSupport::Duration#inspect method in a way that it's not really intended for, but as long as we have a test that blows up in the very unlikely even that it changes I'm happy. It's a shame the `to_s` method isn't set up to do this as that feels far more natural than it's string integer return 🤔

app/services/datetime_parser.rb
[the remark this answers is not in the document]
kevindew: Thanks! Maybe on speeches I'll insist we have the delivery date and time be named `delivery_date` and `delivery_time` just to feel a vindication 😁

app/interactors/schedule_proposal/update_interactor.rb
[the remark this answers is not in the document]
kevindew: Like it 👍

app/services/requirements/schedule_checker.rb
benthorner: OK. I've specified the periods explicitly in the test - I forgot to change one of them anyway. That OK?

app/services/datetime_parser.rb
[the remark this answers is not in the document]
kevindew: > I'm afraid I don't agree here. Just to make sure we both have the same knowledge: we do have [places where we use issues](https://github.com/alphagov/content-publisher/blob/master/app/interactors/publish/publish_interactor.rb#L35) outside of the requirements namespace.

We do those to avoid writing classes, if those things were moved out of an interactor they would be in requirements

> For me, the most important thing is that this is not principally a checker - that's what I found unintuitive about the original implementation. I wouldn't expect to find anything but a checker in requirements; or to put it the other way round, this seems like it's providing a generic service to the app and not a requirements-specific one.

Sure, I don't think of requirements as a place strictly for checkers as we do already use it for Issue and CheckerIssues classes. I'd agree it's not principally a checker, though I wouldn't go as far as it being a general service since we don't have a use case outside of user input.

app/services/requirements/schedule_checker.rb
kevindew: Sounds good

app/services/datetime_parser.rb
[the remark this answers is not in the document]
benthorner: Discussed and we're going to defer on this (https://trello.com/c/f1QaCH7x/231-our-services-directory-is-a-mess).

PR #1135, opened 2019-06-13
Gtm mega bundle

benthorner:

This makes a few iterations to further simplify our analytics:

   - A more generic way to track unusual clicks (e.g. MD toolbar)
   - A more generic / less verbose way to specify event actions
   - A more generic way to track based on visibility (for guidance and filters)

Please see the commits for more details...

### DIFF
--- a/app/views/documents/index/_filters.html.erb
+++ b/app/views/documents/index/_filters.html.erb
@@ -6,7 +6,12 @@
     },
     name: "title_or_url",
     value: params[:title_or_url],
-    type: "search"
+    type: "search",
+    data: {
+      gtm: "input-title-or-url",
+      "gtm-visibility-tracking": true,
+      "gtm-action": params[:title_or_url],
+    }
   } %>
 
   <div class="govuk-form-group">
@@ -27,7 +32,12 @@
       options_for_select(organisation_select, [params[:organisation]]),
       include_blank: true,
       id: "document-organisation-filter",
-      class: "govuk-select"
+      class: "govuk-select",
+      data: {
+        gtm: "select-organisation",
+        "gtm-visibility-tracking": true,
+        "gtm-action": params[:organisation]
+      }
     %>
   </div>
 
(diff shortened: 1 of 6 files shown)


app/views/new_document/choose_document_type.html.erb
```diff
@@ -12,7 +12,10 @@
             value: document_type.id,
             text: document_type.label,
             hint_text: document_type.description,
-            data_attributes: { gtm: "choose-document-type" },
+            data_attributes: {
+              gtm: "choose-document-type",
+              "gtm-action": document_type.label,
```
kevindew: by the way you can do these as `gtm_action` and they'll be converted to `data-gtm-action` attributes

app/views/documents/index/_filters.html.erb
```diff
@@ -6,7 +6,12 @@
     },
     name: "title_or_url",
     value: params[:title_or_url],
-    type: "search"
+    type: "search",
+    data: {
+      gtm: "input-title-or-url",
+      "gtm-visibility-tracking": true,
+      "gtm-action": params[:title_or_url],
```
kevindew: Will this and the others fire for blank attributes? if so do we want that?

app/views/new_document/choose_document_type.html.erb
benthorner: I wondered about that, but I recall we do it this way elsewhere. I think I prefer this way, as it's clearer what the output will be to me.

app/views/documents/index/_filters.html.erb
benthorner: Indeed it will. We do this at the moment and I'm relying on that to make this generic. I think it's OK to keep firing as it helps us understand, in a relative way how, often each of the filters is actually used; if we didn't want to see blank events, we can just filter them out later on. If we want to avoid blank events, we could do that with a special tag, but perhaps that can be a future iteration?

PR #1137, opened 2019-06-14
Replace scheduled_publishing_datetime field

kevindew:

This replaces the `scheduled_publishing_datetime` field with two fields: 

- `proposed_publish_time` on a MetadataRevision
- `publish_time` on a Scheduling

This is done so that once an edition is scheduled it no longer holds the proposed time as part of the revision. This then allows us to handle a failed publishing scenario where we need to know the information on the failed publishing while allowing means to re-schedule this document. 

This also feels like a slightly more appropriate separation between when the date is content and when it is part of a status.

The `publish_time` and `proposed_publish_time` naming began as `publish_at` and `proposed_publish_at` to embrace the `_at` convention for timestamp naming in Rails. This however felt a bit unnatural in terminology for feature tests. As there is already precedence for the phrase [`publish_time` in scheduling](https://github.com/alphagov/publishing-api/blob/master/doc/api.md#put-publish-intentbase_path) this seemed a more natural fit.

This also does some work on refactoring the scheduling tests to embrace the change in terminology and reduce repetition.


spec/features/scheduling/schedule_unpublishable_spec.rb
```diff
@@ -0,0 +1,29 @@
+# frozen_string_literal: true
```
benthorner: I think this one should still be `schedule_requirements_spec`. For me, 'unpublishable' is pretty ambiguous and my first thought was that we would never be able to publish it for some reason :-|.

spec/features/scheduling/schedule_unpublishable_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+
+RSpec.feature "Schedule an unpublishable edition" do
+  scenario do
+    given_there_is_an_edition_with_issues
+    when_i_go_to_schedule_the_edition
```
benthorner: Minor - I think it would help clarify this spec to separate visiting the summary page, or rename the subsequent step since we haven't actually left the page from the user's point of view ;-).

spec/features/scheduling/schedule_spec.rb
```diff
@@ -1,82 +1,68 @@
 # frozen_string_literal: true
 
-RSpec.feature "Schedule an edition" do
+RSpec.feature "Schedule a publishing" do
```
benthorner: Minor, but neither of these make much sense to me 😆 . How about 'Schedule to publish'?

spec/features/scheduling/update_proposed_publish_time_spec.rb
```diff
@@ -0,0 +1,53 @@
+# frozen_string_literal: true
+
+RSpec.feature "Change proposed publish time" do
```
benthorner: Minor, but can we keep the feature title consistent with the file name?

spec/features/scheduling/update_publish_time_spec.rb
```diff
@@ -32,29 +36,36 @@ def and_i_click_on_change_date
...
-                                         date: @new_datetime.strftime("%-d %B %Y")))
+                                         date: "15 August 2019"))
   end
 
-  def and_the_edition_is_scheduled_to_publish
+  def and_the_publish_intent_has_been_updated
+    expect(@request).to have_been_requested
+  end
+
+  def and_a_job_is_queued
```
benthorner: Minor - could we say `and_a_new_job_is_queued` to emphasise the change?

spec/features/scheduling/update_publish_time_spec.rb
```diff
@@ -32,29 +36,36 @@ def and_i_click_on_change_date
...
+    fill_in "schedule[date][month]", with: "8"
+    fill_in "schedule[date][year]", with: "2019"
     select "11:00pm", from: "schedule[time]"
+
     click_on "Save date"
   end
 
-  def then_i_see_the_edition_is_rescheduled
-    visit document_path(@edition.document)
+  def then_i_see_the_edition_is_scheduled
```
benthorner: Sadness ;-). I did quite like 'rescheduled', since it was already scheduled before.

app/jobs/scheduled_publishing_job.rb
```diff
@@ -9,32 +9,37 @@ class ScheduledPublishingJob < ApplicationJob
...
     end
 
-    send_notifications(published_edition)
+    send_notifications(edition)
   end
 
 private
 
-  def no_longer_schedulable?(edition)
+  def publishable?(edition)
```
benthorner: Minor, but the new name is a little bit misleading, since the edition is 'publishable'. I was wondering whether it would be helpful to rename this to something like 'cancelled?', since we rely on it for that?

app/views/documents/show/_scheduling_notice.html.erb
```diff
@@ -1,52 +1,51 @@
...
+                schedule_proposal_path(@edition.document),
+                class: "govuk-link govuk-link--no-visited-state" %>
+  <% end %>
+
   <%= render "govuk_publishing_components/components/notice", {
     description: proposed_scheduling_description,
     margin_bottom: 4,
   } %>
 <% end %>
+
```
benthorner: Newline ;-).

spec/features/scheduling/schedule_spec.rb
kevindew: I don't really know how this makes much difference but I've gone with your suggestion

spec/features/scheduling/update_proposed_publish_time_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+
+RSpec.feature "Update proposed publish time" do
+  include ActiveSupport::Testing::TimeHelpers
+
+  around do |example|
+    travel_to(Time.zone.parse("2019-06-13")) { example.run }
+  end
+
+  scenario do
+    given_there_is_an_edition_with_a_proposed_publish_time
```
benthorner: Minor, but looks like we've now got a concise 'schedulable' term we can use here :-).

spec/jobs/scheduled_publishing_job_spec.rb
```diff
@@ -12,7 +12,7 @@
       it "calls PublishService if edition can be published" do
         edition = create(:edition,
                          :scheduled,
-                         scheduled_publishing_datetime: Time.current)
+                         scheduling: build(:scheduling, publish_time: Time.current))
```
benthorner: Just double-checking we still want to manually build a scheduling here?

kevindew: Oops missed.

PR #1140, opened 2019-06-17
Handle publishing failing

kevindew:

This builds off of #1138

I'm marking this as a draft as I'm certainly going to have to rebase this one.

I'll add some more details soon.


app/services/scheduled_publishing_failed_service.rb
```diff
@@ -0,0 +1,31 @@
...
+      update_status(edition)
+    end
+
+    notify_editors(edition)
+  end
+
+private
+
+  def update_status(edition)
+    raise "Expected edition to be scheduled" unless edition.scheduled?
```
benthorner: Could you help me understand which edge case are we trying to protect against here? I'm still worried about this kind of statement proliferating into every class, without any definite value.

app/mailers/publish_mailer.rb
```diff
@@ -1,20 +1,19 @@
...
-  def publish_email(edition, user)
-    @user = user
+  def publish_email(recipient, edition, status)
     @edition = edition
-    @status = edition.status
+    @status = status
 
-    unless @status.published? || @status.published_but_needs_2i?
-      raise "Cannot send publish email with a non-published state"
+    if !status.published? && !status.published_but_needs_2i?
```
benthorner: Minor, but I'm surprised the linter accepts this, since the previous version was more readable to me. How come it changed?

spec/integration/publishing_a_scheduled_edition_spec.rb
```diff
@@ -0,0 +1,85 @@
...
+                      scheduling: scheduling,
+                      created_by: create(:user, email: "user@example.com"),
+                      base_path: "/news/breaking-story")
+  end
+
+  def when_the_scheduled_publishing_job_runs_successfully
+    @request = stub_publishing_api_publish(@edition.content_id,
+                                           update_type: nil,
+                                           locale: @edition.locale)
+    perform_enqueued_jobs { ScheduledPublishingJob.perform_later(@edition.id) }
```
benthorner: Should we consider using this instead of Sidekiq::Testing in the feature specs?

spec/integration/publishing_a_scheduled_edition_spec.rb
```diff
@@ -0,0 +1,85 @@
...
+
+    expect(message.body).to include("https://www.test.gov.uk/news/breaking-story")
+    expect(message.body)
+      .to include(I18n.t("scheduled_publish_mailer.success_email.details.publish",
+                         time: "10:00am",
+                         date: "17 June 2019"))
+  end
+
+  def when_the_scheduled_publishing_job_fails
+    allow(PublishService).to receive(:new).and_raise(RuntimeError)
```
benthorner: It would be great if we could do this in a more realistic way by stubbing an API call error. Would that work here?

spec/services/scheduled_publishing_failed_service_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+        .from("scheduled")
+        .to("failed_to_publish")
+    end
+
+    it "maintains the scheduling details from the previous status" do
+      expect { ScheduledPublishingFailedService.new.call(edition.id) }
+        .not_to(change { edition.reload.status.details })
+    end
+
+    context "when an edition is not scheduled" do
```
benthorner: As we discussed previously, I don't think we have a precedent for these kinds of tests, since they are basically just assertions.

spec/integration/publishing_a_scheduled_edition_spec.rb
```diff
@@ -0,0 +1,85 @@
+# frozen_string_literal: true
+
+RSpec.describe "Publishing a scheduled edition" do
```
benthorner: Yay - does this mean we can lose https://github.com/alphagov/content-publisher/pull/1140/commits/be778ef0181346c907d9a31fddd719dc390f38c7#diff-a7f748b67cad60e7c2600e6be3d0896b, since it should be covered by this test?

benthorner: I'm also wondering if we can lose the mailer spec, or maybe we should create one for the PublishMailer and simplify the associated feature specs.  Not part of this PR really, but it would be helpful if we used the same approach for each :-).

app/services/scheduled_publishing_failed_service.rb
kevindew: Understand your concern. I think this one is valid as there is nothing to ensure the state of this edition.

Consider:

- Final scheduling fails and we leave this transaction: https://github.com/alphagov/content-publisher/blob/handle-publishing-failing/app/jobs/scheduled_publishing_job.rb#L22
- Some period of time occurs between this and https://github.com/alphagov/content-publisher/blob/handle-publishing-failing/app/jobs/scheduled_publishing_job.rb#L6
- In that time any changes to the state of edition would make this failed to publish make the data model inconsistent

It is super unlikely as the period of time is super short, but all it takes is a request to occur during the attempt of scheduling and that request would have the next lock on the edition.

app/mailers/publish_mailer.rb
kevindew: Heh, didn't realise this was a controversial one. With [some](http://www.railstips.org/blog/archives/2008/12/01/unless-the-abused-ruby-conditional/) [googling](https://github.com/rubocop-hq/rubocop/issues/5388) I realise it's a point of controversy in the ruby community.

But yeah I thought the linter might have ruled that the previous one was bad.

To me the `unless` reads wrong once you have more than one condition as it reads as "unless published or published but needs 2i" which makes the second clause ambiguous. Whereas "if not published and not published but needs 2i" you lack that ambiguity so the code is more readable.

I'll change it back though since it's not unanimous.

spec/integration/publishing_a_scheduled_edition_spec.rb
kevindew: We could do, it'd be a shift of our model though rather than a drop in replacement. We currently use `Sidekiq::Testing.inline` so the jobs run automatically whereas if we did this we'd always have to call them.

It might be better to have them always called though as it helps breaks the assumptive behaviour that async jobs can be treated as synchronous calls.

spec/integration/publishing_a_scheduled_edition_spec.rb
kevindew: Yeah, that's a better idea than what I did.

spec/services/scheduled_publishing_failed_service_spec.rb
kevindew: Will nuke

spec/integration/publishing_a_scheduled_edition_spec.rb
kevindew: I'll see if that test can be killed. 

I'd suggest the approach of having a PublishMailer test and loosen feature specs as there is quite a lot of conditionals and translation keys involved in the mails.

app/services/scheduled_publishing_failed_service.rb
benthorner: OK. It's good to see a potential scenario, and it helps me to think about this as a state machine transition - in that sense, we should be validating the origin (scheduled?) in order to transition to the destination (failed_to_publish).

spec/mailers/scheduled_publish_mailer_spec.rb
```diff
@@ -1,75 +1,133 @@
...
+      mail = ScheduledPublishMailer.success_email(recipient, edition, edition.status)
+
+      expect(mail.to).to eq([recipient.email])
+
+      mail_subject = I18n.t!("scheduled_publish_mailer.success_email.subject.published",
+                             title: edition.title)
+      expect(mail.subject).to eq(mail_subject)
+    end
+
+    it "errors for an unexpected status" do
```
benthorner: Oops - didn't spot this one earlier. Again, there doesn't seem to be a precedent for this kind of assertion test.

spec/mailers/scheduled_publish_mailer_spec.rb
```diff
@@ -1,75 +1,133 @@
...
+      mail = ScheduledPublishMailer.failure_email(recipient, edition, edition.status)
+
+      expect(mail.to).to eq([recipient.email])
+
+      mail_subject = I18n.t!("scheduled_publish_mailer.failure_email.subject",
+                             title: edition.title)
+      expect(mail.subject).to eq(mail_subject)
+    end
+
+    it "errors for an unexpected status" do
```
benthorner: Same here :-|.

spec/integration/publishing_a_scheduled_edition_spec.rb
benthorner: 👍 - should we have a follow-up task to make the approaches consistent?

spec/integration/publishing_a_scheduled_edition_spec.rb
benthorner: So I had a little play with this, and I think we can get to only using one mechanism to run jobs inline. Turns out we only rely on `inline!` in the publish specs, which we could replace with `around { perform_enqueued_jobs { |example| example.run } }`. Perhaps after this PR we could do a quick PR to use this common approach in all our specs?

spec/integration/publishing_a_scheduled_edition_spec.rb
kevindew: Yup, I've got that on my list that I need to move to trello.

spec/integration/publishing_a_scheduled_edition_spec.rb
kevindew: Sure, that's an interesting idea, does feel a bit unconventional.

I had a good ponder yesterday if there were means to run these scheduling integration tests as part of feature tests and was a bit dismayed there wasn't a perform_enqueued_jobs that takes no argument and just runs what is backlogged. The best approach I got for that was to use `Sidekiq::Worker.drain_all` (although I'm not sure if that is compatible with activejob. Then a test could be like:

```
# methods to set up scheduling
travel_to(Time.current.advance(days: 2)
Sidekiq::Worker.drain_all
# methods to check scheduling workd
```

I'm mostly referencing this thought process to indicate concerns that `perform_enqueud_jobs` could be an all purpose fit.

kevindew:
Thanks for catching those 2 @benthorner. Are you happy if I merge this now?

benthorner:
Sure - go ahead :-).

spec/integration/publishing_a_scheduled_edition_spec.rb
benthorner: You'll have to help me with your thoughts on it being unconventional. It actually seems better to me: we're using a generic job framework, so it makes sense that our tests use the method of that framework (and not the backing job system), like we do when checking what's on the queue. I guess it would be clearer to perform the jobs at the point of creation, as you do in your example, but it should still be pretty visible if we use the 'around' approach in the small number of tests where this is a concern.

We can discuss a bit tomorrow if you like :-).

kevindew: Sounds good - In the meantime I'll have a go at my scheduling feature test too so we can consider that.

PR #1141, opened 2019-06-17
Remove gtm-doc-status and gtm-doc-type attributes

benthorner:

Previously we had two custom GTM attributes to communicate the visible
document status and type, although the latter was never actually used.
Both can now be inferred from other events, but we should ideally use
the app DB as the source of this kind of non-eventful information.

This removes both attributes and the associated field of the metadata
component, as it is now redundant.

### DIFF
--- a/app/views/components/_metadata.html.erb
+++ b/app/views/components/_metadata.html.erb
@@ -3,12 +3,11 @@
   inline ||= false
   list_classes = %w(app-c-metadata__list)
   list_classes << 'app-c-metadata__list--inline' if inline
-  data_attributes ||= nil
   margin_bottom ||= 0
   component_classes = %w(app-c-metadata)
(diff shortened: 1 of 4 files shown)


app/views/components/_metadata.html.erb
```diff
@@ -3,12 +3,11 @@
   inline ||= false
   list_classes = %w(app-c-metadata__list)
   list_classes << 'app-c-metadata__list--inline' if inline
-  data_attributes ||= nil
```
kevindew: This also needs to be removed from the example: https://github.com/alphagov/content-publisher/blob/dbe3451b3c6ce3979b20e2a0014b7ab97cb32bc5/app/views/components/docs/metadata.yml#L8-L10

benthorner: Good spot - I've removed that now.

PR #1146, opened 2019-06-19
Backdate content

emmabeynon:
For https://trello.com/c/x6ImOUSP/918-create-a-form-so-user-can-save-the-backdate-for-their-content

This creates a form for users to backdate content and, if the date is valid, sends the date as `first_published_at` to the Publishing API.  If the date is not valid errors are displayed on the form.

**No backdate set**
> Screen Shot 2019-06-18 at 17 13 51

**Backdating form**
> Screen Shot 2019-06-18 at 17 14 42

**Backdating form with errors**
> Screen Shot 2019-06-18 at 17 14 24

**Backdated**
> Screen Shot 2019-06-19 at 14 26 07

### DIFF
--- a/app/interactors/backdate/update_interactor.rb
+++ b/app/interactors/backdate/update_interactor.rb
@@ -10,6 +10,7 @@ class Backdate::UpdateInteractor
   def call
     Edition.transaction do
       find_and_lock_edition
+      check_for_issues
       update_edition
     end
   end
@@ -34,5 +35,13 @@ private
     Time.zone.local(backdate_params[:year].to_i,
                     backdate_params[:month].to_i,
                     backdate_params[:day].to_i)
+  end
+
+  def check_for_issues
+    unless edition.number == 1
+      # FIXME: this shouldn't be an exception but we've not worked out the
+      # right response - maybe bad request or a redirect with flash?
+      raise "Only first editions can be backdated."
+    end
   end
 end
--- a/app/interactors/backdate/update_interactor.rb
+++ b/app/interactors/backdate/update_interactor.rb
@@ -1,0 +1,38 @@
+# frozen_string_literal: true
+
+class Backdate::UpdateInteractor
+  include Interactor
+  delegate :params,
+           :user,
+           :edition,
+           to: :context
+
+  def call
+    Edition.transaction do
+      find_and_lock_edition
+      update_edition
+    end
+  end
+
+private
+
+  def find_and_lock_edition
+    context.edition = Edition.lock.find_current(document: params[:document])
+  end
+
+  def update_edition
+    updater = Versioning::RevisionUpdater.new(edition.revision, user)
+    updater.assign(backdated_to: submitted_date)
+    edition.assign_revision(updater.next_revision, user).save!
+  end
+
+  def backdate_params
+    params.require(:backdate).permit(:day, :month, :year)
+  end
+
+  def submitted_date
+    Time.zone.local(backdate_params[:year].to_i,
+                    backdate_params[:month].to_i,
+                    backdate_params[:day].to_i)
+  end
+end
--- a/app/interactors/backdate/update_interactor.rb
+++ b/app/interactors/backdate/update_interactor.rb
@@ -5,6 +5,7 @@ class Backdate::UpdateInteractor
   delegate :params,
            :user,
            :edition,
+           :date,
            to: :context
 
   def call
@@ -23,18 +24,12 @@ private
 
   def update_edition
     updater = Versioning::RevisionUpdater.new(edition.revision, user)
-    updater.assign(backdated_to: submitted_date)
+    updater.assign(backdated_to: date)
     edition.assign_revision(updater.next_revision, user).save!
   end
 
   def backdate_params
-    params.require(:backdate).permit(:day, :month, :year)
-  end
-
-  def submitted_date
-    Time.zone.local(backdate_params[:year].to_i,
-                    backdate_params[:month].to_i,
-                    backdate_params[:day].to_i)
+    params.require(:backdate).permit(date: %i[day month year])
   end
 
   def check_for_issues
@@ -43,5 +38,12 @@ private
       # right response - maybe bad request or a redirect with flash?
       raise "Only first editions can be backdated."
     end
+
+    checker = Requirements::BackdateChecker.new(backdate_params[:date])
+    issues = checker.pre_submit_issues
+
+    context.fail!(issues: issues) if issues.any?
+
+    context.date = checker.parsed_date
   end
 end
--- a/app/interactors/backdate/update_interactor.rb
+++ b/app/interactors/backdate/update_interactor.rb
@@ -11,7 +11,10 @@ class Backdate::UpdateInteractor
   def call
     Edition.transaction do
       find_and_lock_edition
+
+      parse_date
       check_for_issues
+
       update_edition
       create_timeline_entry
     end
@@ -21,6 +24,12 @@ private
 
   def find_and_lock_edition
     context.edition = Edition.lock.find_current(document: params[:document])
+
+    unless edition.first? && edition.editable?
+      # FIXME: this shouldn't be an exception but we've not worked out the
+      # right response - maybe bad request or a redirect with flash?
+      raise "Only editable first editions can be backdated."
+    end
   end
 
   def update_edition
@@ -33,19 +42,18 @@ private
     params.require(:backdate).permit(date: %i[day month year])
   end
 
-  def check_for_issues
-    unless edition.number == 1
-      # FIXME: this shouldn't be an exception but we've not worked out the
-      # right response - maybe bad request or a redirect with flash?
-      raise "Only first editions can be backdated."
-    end
+  def parse_date
+    parser = DateParser.new(date: backdate_params[:date], issue_prefix: :backdate)
+    context.date = parser.parse
 
-    checker = Requirements::BackdateChecker.new(backdate_params[:date])
+    context.fail!(issues: parser.issues) if parser.issues.any?
+  end
+
+  def check_for_issues
+    checker = Requirements::BackdateChecker.new(date)
     issues = checker.pre_submit_issues
 
     context.fail!(issues: issues) if issues.any?
-
-    context.date = checker.parsed_date
   end
 
   def create_timeline_entry
--- a/app/interactors/backdate/update_interactor.rb
+++ b/app/interactors/backdate/update_interactor.rb
@@ -13,6 +13,7 @@ class Backdate::UpdateInteractor
       find_and_lock_edition
       check_for_issues
       update_edition
+      create_timeline_entry
     end
   end
 
@@ -45,5 +46,14 @@ private
     context.fail!(issues: issues) if issues.any?
 
     context.date = checker.parsed_date
+  end
+
+  def create_timeline_entry
+    TimelineEntry.create_for_revision(
+      entry_type: :backdated,
+      revision: edition.revision,
+      edition: edition,
+      created_by: user,
+    )
   end
 end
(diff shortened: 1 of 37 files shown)


app/interactors/backdate/update_interactor.rb
```diff
@@ -0,0 +1,70 @@
...
+    end
+  end
+
+private
+
+  def find_and_lock_edition
+    context.edition = Edition.lock.find_current(document: params[:document])
+  end
+
+  def check_if_first_edition
```
benthorner: Minor - I think we now tend to do this in the 'find' method, as it's really just an assertion about the edition that we've found e.g. https://github.com/alphagov/content-publisher/blob/master/app/interactors/schedule/create_interactor.rb#L25

app/models/edition.rb
```diff
@@ -130,6 +131,10 @@ def editable?
     !live? && !scheduled?
   end
 
+  def first?
```
benthorner: This looks useful, and I can see you've replaced it in a few places already. Do you think we could have a subsequent PR to make use of it throughout the rest app? Possible use case: https://github.com/alphagov/content-publisher/blob/master/app/views/documents/show/_contents.html.erb#L10

app/services/requirements/backdate_checker.rb
```diff
@@ -0,0 +1,27 @@
...
+  class BackdateChecker
+    def initialize(backdate)
+      @backdate = backdate
+    end
+
+    def pre_submit_issues
+      issues = []
+
+      if in_the_future?
+        issues << Issue.new(:backdate, :in_the_future)
```
benthorner: Minor, but I wonder if we should also prevent publishers from backdating to 'today'? Probably doesn't matter very much, but it might be worth checking with @benhazell to see if it's sensible.

app/interactors/backdate/update_interactor.rb
benthorner: I think @kevindew is also keen on an 'editable?' check, although we're not really consistent on this.

https://trello.com/c/eAGtt4ry/322-we-dont-enforce-our-view-restrictions-in-our-controllers

app/views/backdate/edit.html.erb
```diff
@@ -0,0 +1,50 @@
+<% content_for :back_link, render_back_link(href: document_path(@edition.document)) %>
+<% content_for :title, t("backdate.edit.title") %>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <%= render_govspeak(t("backdate.edit.intro_paragraph")) %>
```
benthorner: We conventionally suffix things like this with '_govspeak' to make it clear to people like @benhazell how the content will be interpreted.

(https://github.com/alphagov/content-publisher/blob/master/docs/adr/0004-editing-microcopy.md#decision)

app/views/backdate/edit.html.erb
```diff
@@ -0,0 +1,50 @@
...
+  </div>
+</div>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <% legend = capture do %>
+      <span class="govuk-heading-s govuk-!-margin-bottom-0">
+        <%= t("backdate.edit.date") %>
```
benthorner: Minor, but could we pass this into the translation instead of rendering it separately? I almost didn't spot it on the end ;-).

app/views/documents/show/_content_settings.html.erb
```diff
@@ -0,0 +1,24 @@
+<% backdating_status = capture do %>
+  <% if @edition.backdated_to.present? %>
+    <% @edition.backdated_to.strftime("%-d %B %Y") %>
```
benthorner: Another minor thing - do you think we could re-use the (suitably renamed) SchedulingHelper here? Possibly something for a follow-up PR, though!

app/views/documents/show/_history_tab.html.erb
```diff
@@ -28,6 +28,11 @@
           <%= entry.details.public_explanation %>
         <% end %>
 
+        <% if entry.backdated? %>
+          <% date = entry.edition.backdated_to.strftime("%d %B %Y") %>
```
benthorner: Same here for the helper.

app/views/documents/show/_summary_tab.html.erb
```diff
@@ -20,6 +20,14 @@
     <% end %>
 
     <%= render "documents/show/tags" %>
+
+    <%
+      # TODO: Once Access Limiting has been implemented, this logic should only
```
benthorner: I don't think we need this here, since we will certainly change this when the feature is built, and the specs protect us from breaking it for backdating.

app/views/backdate/edit.html.erb
```diff
@@ -0,0 +1,50 @@
...
+            name: "year",
+            width: 4,
+            value: params.dig(:backdate, :year) || date&.year,
+          }
+        ]
+      } %>
+      <%= render "govuk_publishing_components/components/button", {
+        text: "Save",
+        margin_bottom: true
+      } %>,
```
benthorner: Stray comma here!

Screen Shot 2019-06-19 at 14 37 33

emmabeynon: I had spotted this before and meant to fix it, slipped my mind!

app/services/requirements/backdate_checker.rb
emmabeynon: I had pondered this, and wondered if possibly they would want to backdate it to today and then actually publish it at a future date?  Not sure if it's a thing.

app/models/edition.rb
emmabeynon: Oh sure, I'd looked for instances of `edition.number == 1` but hadn't considered where the `<` and `>` operators may be used.

app/interactors/backdate/update_interactor.rb
emmabeynon: Makes sense, we do check if the edition is editable before rendering the context settings section so we should make that consistent here.

app/services/backdate_parser.rb
```diff
@@ -0,0 +1,36 @@
...
+    check_date_is_valid
+    return if issues.any?
+
+    Time.current.change(
+      day: raw_date[:day].to_i,
+      month: raw_date[:month].to_i,
+      year: raw_date[:year].to_i,
+      hour: 0,
+      minute: 0,
+    )
```
kevindew: It might be cleaner to have this as a Date object at this point in the code rather than a Time object. 

This might also make this class simpler as we already create a Date object in line 32.

It would mean having to then convert the date to a time in the interactor, but I think that would actually be better as that feels a more appropriate place to set a magic time on it (I believe the requested magic time is 1 minute past midnight)

app/views/backdate/edit.html.erb
```diff
@@ -0,0 +1,50 @@
...
+        <%= t("backdate.edit.date") %>
+      </span>
+    <% end %>
+    <% date = @edition.backdated_to %>
+
+    <%= form_tag backdate_path(@edition.document) do %>
+      <%= render "govuk_publishing_components/components/date_input", {
+        legend_text: legend,
+        hint: t("backdate.edit.date_hint_text"),
+        name: "backdate",
```
kevindew: I think this would be more consistent with other forms if it was backdate[date], as these day, month and year are just properties of this date field.

app/services/publish_time_parser.rb
```diff
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-class DatetimeParser
+class PublishTimeParser
```
kevindew: Is this renamed due to coupling to schedule with the issues?

I wonder if we could make the issue prefix a property of initialize (and default to schedule since this is the only one) so that we can have DateTimeParser and DateParser as generic classes.

app/interactors/backdate/update_interactor.rb
kevindew: Yup 👍 looking this up in find the most consistent place we have.

app/views/backdate/edit.html.erb
emmabeynon: As discussed offline, moved the legend variable up to the top.

app/views/backdate/edit.html.erb
emmabeynon: Sure, I figured since we were only dealing with the date on this page it didn't need additional labelling.

config/locales/en/backdate/edit.yml
```diff
@@ -0,0 +1,7 @@
+en:
+  backdate:
+    edit:
+      title: Backdate content
+      intro_paragraph: If this content has previously been published on another webpage you must provide the date for when it was first published online. This prevents old information being presented as new.
```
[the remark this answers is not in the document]
emmabeynon: We do have `description_govspeak` elsewhere so I can change that.

app/views/backdate/edit.html.erb
kevindew: Yup, you're right it doesn't actually need it, but does slip into inconsistency without it.

db/schema.rb
```diff
@@ -197,6 +197,7 @@
     t.datetime "created_at"
     t.bigint "created_by_id"
     t.datetime "proposed_publish_time"
+    t.datetime "backdated_to"
```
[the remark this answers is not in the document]
kevindew: With regards users, they may be only entering a date but we are storing a time in the database so the field does reference a time.

If I remember right the Rails conventions are `_at` for a time and `_on` for a date, but they only really make sense for past tense.

I do agree it reads well linguistically though 👍

app/services/backdate_parser.rb
emmabeynon: Do we have any rationale behind setting the time to 00:01?  I think it would be helpful to include this in the commit message to explain it.

kevindew: Not a great bit of rationale, could be worth asking Tobi for more info on that tomorrow - I think it's based on an expectation that users get confused by 00:00 (but that never confuses me so I don't relate) - there's this comment about it: https://trello.com/c/aBHXR82t/932-user-can-backdate-or-set-document-as-new#comment-5cf8dbb949a3fa4dfbe0291d

app/services/publish_time_parser.rb
emmabeynon: @benthorner and I had a fairly lengthy discussion about this earlier and tried exploring a few different options.  Might be worth us having a chat face-to-face tomorrow.

spec/features/workflow/backdate_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+RSpec.feature "Backdate the first edition of a document" do
+  scenario do
+    given_there_is_a_document_with_a_first_edition
+    when_i_visit_the_summary_page
+    and_i_click_to_backdate_the_content
+    and_i_enter_a_date_to_backdate_the_content_to
+    and_i_click_save
+  end
+
+  def given_there_is_a_document_with_a_first_edition
```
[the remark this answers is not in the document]
emmabeynon: This is all modified in a later commit to specify various editions so will leave for now.

app/models/edition.rb
emmabeynon: I've searched for the other instances and changed them - hopefully I found them all 🤞

app/views/documents/show/_content_settings.html.erb
emmabeynon: I've added this is to 'Things we need to do' on the card.

app/services/backdate_parser.rb
emmabeynon: Discussed with Tobi and since the time isn't going to be shown in Content Publisher or on the frontend, we're going to stick with midnight.

benthorner: I think we should use 10:10 ;-).

https://petapixel.com/2013/06/27/why-photographs-of-watches-and-clocks-show-the-time-1010/

PR #1151, opened 2019-06-19
Gtm for errors

benthorner:

This introduces separate tracking for requirements issues vs. the flashes they're displayed in.

### DIFF
--- a/app/controllers/backdate_controller.rb
+++ b/app/controllers/backdate_controller.rb
@@ -10,8 +10,7 @@ class BackdateController < ApplicationController
     edition, issues = result.to_h.values_at(:edition, :issues)
 
     if issues
-      flash.now["alert_with_items"] = {
-        "title" => I18n.t!("backdate.edit.flashes.requirements"),
+      flash.now["requirements"] = {
         "items" => issues.items(
           link_options: { backdate_date: { href: "#backdate-date" } },
         ),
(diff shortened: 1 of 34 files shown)


benthorner:
Reopened with an additional commit.

benthorner:
@kevindew the same thought did go through my head. I think what we might be grappling with here is that the Issue class is both a data store for issues, and a presenter for them as objects to various components. 

One potential downside of putting the data attributes in this class, which I should have commented about, is that if any of the input components are iterated to support data attributes on individual error items, then we will get duplicate events for each requirements issue.

We could perhaps have a helper for when issues are being fed into the inset/error_summary components, which extends them with these extra attributes, something like `track_requirements(issues)`. How does that sound?

kevindew:
A helper sounds good :+1:

benthorner:
@kevindew I've had a go at this, but unfortunately it didn't turn out very well - including the helper and using it in most controllers felt pretty onerous and prone to being missed in future. An alternative might be to use a specific flash for requirements and inject the attributes there; a potential advantage here is that we lose the duplicate "You need to" translations. On the other hand, the only concern I really have here is around event uniqueness, but we can mostly fix that at the GA end. Any preferences?

kevindew:
> @kevindew I've had a go at this, but unfortunately it didn't turn out very well - including the helper and using it in most controllers felt pretty onerous and prone to being missed in future. An alternative might be to use a specific flash for requirements and inject the attributes there; a potential advantage here is that we lose the duplicate "You need to" translations. On the other hand, the only concern I really have here is around event uniqueness, but we can mostly fix that at the GA end. Any preferences?

Alternative flash for requirements sounds good.

benthorner:
@kevindew this should now be updated as discussed. How does it look now?

PR #1156, opened 2019-06-19
Missing scheduling timeline entries

kevindew:

These had got missed last time around

### DIFF
--- a/config/locales/en/documents/history.yml
+++ b/config/locales/en/documents/history.yml
@@ -26,6 +26,7 @@ en:
         scheduled: "Scheduled to publish"
         scheduled_publishing_succeeded: "Published on schedule"
         scheduled_publishing_without_review_succeeded: "Published on schedule but needs 2i review"
+        scheduled_publishing_failed: "Scheduled publishing failed"
         schedule_updated: "Rescheduled to publish"
         unscheduled: "Scheduled publishing stopped"
         file_attachment_deleted: "Attachment deleted"
--- a/config/locales/en/documents/history.yml
+++ b/config/locales/en/documents/history.yml
@@ -24,6 +24,8 @@ en:
         draft_reset: "Draft reset to last published edition"
         unwithdrawn: "Unwithdrawn"
         scheduled: "Scheduled to publish"
+        scheduled_publishing_succeeded: "Published on schedule"
+        scheduled_publishing_without_review_succeeded: "Published on schedule but needs 2i review"
         schedule_updated: "Rescheduled to publish"
         unscheduled: "Scheduled publishing stopped"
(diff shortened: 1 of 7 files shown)


config/locales/en/documents/history.yml
```diff
@@ -24,8 +25,11 @@ en:
         draft_reset: "Draft reset to last published edition"
         unwithdrawn: "Unwithdrawn"
         scheduled: "Scheduled to publish"
+        scheduled_publishing_succeeded: "Published on schedule"
+        scheduled_publishing_without_review_succeeded: "Published on schedule but needs 2i review"
```
benthorner: Minor, but I notice the 'needs 2i review' is inconsistent compared to 'without review' above. Is that intentional?

spec/integration/failed_scheduled_publishing_spec.rb
```diff
@@ -36,6 +37,11 @@ def then_the_edition_is_set_to_failed_to_publish
     expect(@edition.reload).to be_failed_to_publish
   end
 
+  def and_a_timeline_entry_is_created
+    timeline_entry = @edition.timeline_entries.last
+    expect(timeline_entry.entry_type).to eq("scheduled_publishing_failed")
```
benthorner: Would it be possible to do this through the document summary page? I think we conventionally try to avoid expectations on the database in feature tests, since it isn't something users see. And it would help to keep our I18n coverage up ;-). I suppose it's a bit unclear whether an integration test should really be visiting a view - does that sound reasonable to you?

config/locales/en/documents/history.yml
kevindew: I felt that it was more important to have the key consistent with `publishing_without_review` than it was to be consistent with the copy of the text. I want to check in with @benthorner about the text copy as a step of completing this feature.

config/locales/en/documents/history.yml
```diff
@@ -24,8 +25,11 @@ en:
         draft_reset: "Draft reset to last published edition"
         unwithdrawn: "Unwithdrawn"
         scheduled: "Scheduled to publish"
+        scheduled_publishing_succeeded: "Published on schedule"
```
emmabeynon: Looks like we've got a duplicate here; the same translation is on line 11 already.

spec/integration/failed_scheduled_publishing_spec.rb
kevindew: That's something that is done through https://github.com/alphagov/content-publisher/pull/1160

config/locales/en/documents/history.yml
kevindew: Oops - thanks for catching that!

PR #1157, opened 2019-06-20
Handle scheduled and revision not synced

kevindew:

Previously this edge case was falling into an expectation that the
document was a draft. This provides actions that are not applicable for
a scheduled document.


benthorner:
I'm a bit confused about this change: how can a scheduled edition not be synced?

kevindew:
@benthorner Sounds like we worked out this can happen in our chat today - still have a schedule link even if revision isn't synced and schedule point doesn't check whether revision is synced.

benthorner:
@kevindew thanks for explaining - I'm replacing this with https://github.com/alphagov/content-publisher/pull/1170.

PR #1158, opened 2019-06-20
Link and highlight issues for edit and tag forms

kevindew:
Trello: https://trello.com/c/4Uc4JL4T/870-add-a-publishing-requirement-to-select-a-primary-organisation

This corrects these two forms where we weren't linking to and highlighting issues.

## Edit form

Before:
Screen Shot 2019-06-20 at 09 47 07

After:
Screen Shot 2019-06-20 at 09 47 46

## Tags Form

Before:
Screen Shot 2019-06-20 at 09 51 52

After:
Screen Shot 2019-06-20 at 09 52 27

### DIFF
--- a/app/controllers/documents_controller.rb
+++ b/app/controllers/documents_controller.rb
@@ -44,11 +44,11 @@ class DocumentsController < ApplicationController
     if issues
       flash.now["alert_with_items"] = {
         "title" => I18n.t!("documents.edit.flashes.requirements"),
-        "items" => issues.items,
+        "items" => issues.items(link_options: issues_link_options(edition)),
       }
 
       render :edit,
-             assigns: { edition: edition, revision: revision },
+             assigns: { edition: edition, revision: revision, issues: issues },
              status: :unprocessable_entity
     elsif params[:submit] == "add_contact"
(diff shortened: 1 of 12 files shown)


benthorner:
@soniaturcotte just to double check, is this the approach we want to take for all forms in the app? One slight exception to this will be the file selector for images/attachments, which doesn't support showing errors at the moment, so I'm guessing we wouldn't want to link to it either.

soniaturcotte:
> @soniaturcotte just to double check, is this the approach we want to take for all forms in the app? One slight exception to this will be the file selector for images/attachments, which doesn't support showing errors at the moment, so I'm guessing we wouldn't want to link to it either.

Yes looks great, from the screenshots I'm assuming that this is the standard [Design System pattern](https://design-system.service.gov.uk/components/error-summary/). 

For the file upload, is there a reason we can't do this: https://design-system.service.gov.uk/components/file-upload/ ?

benthorner:
@soniaturcotte oops, that was incorrect of me. Embarrassingly, I actually worked on showing errors on the file upload component! Anyway, although it can support the pattern, we don't do it in the app for some reason. @kevindew, do you think we should have a tidy-up card for this stuff?

PR #1159, opened 2019-06-20
Remove gtm-topic-listener

benthorner:

Previously we had a 'submit' listener that scraped the selected topics
and sent each one to GTM as an individual event. This is not consistent
with our analytics strategy, since it's purpose is to (inaccurately) try
to replicate data that we store in the Publishing API database.

This removes the submit listener, but we will continue to use GTM to
track actual user actions that relate to an interaction with the page.


benthorner:
@kevindew thanks for spotting - I'll look at that.

PR #1163, opened 2019-06-20
Standardise copy/paste listener

benthorner:
https://trello.com/c/tAXP6pf1/948-analytics-for-topics

Previously we had a single listener to emit custom copy/paste events to
GTM. In order to accommodate more listeners to support analytics for
topics, this modifies the copy/paste listener to emit a new 'dataGtm'
variable directly on the dataLayer, with the same meaning as the current
'data-gtm' attribute present on many elements, but formed dynamically
out of the event being tracked by the listener, and a 'data-gtm-suffix'
attribute coming from the element itself; in contrast, using 'data-gtm'
statically on an element generally implies it only supports one event.

### DIFF
--- a/app/assets/javascripts/modules/gtm-copy-paste-listener.js
+++ b/app/assets/javascripts/modules/gtm-copy-paste-listener.js
@@ -1,28 +1,26 @@
 var GtmCopyPasteListener = {}
 
-GtmCopyPasteListener.handleCopy = function (event) {
-  var element = event.target
+GtmCopyPasteListener.handleCopyPaste = function (prefix) {
+  return function (event) {
+    var element = event.target
 
-  if (!element.hasAttribute('data-gtm-copy-paste-tracking')) {
-    return
+    if (!element) {
+      return
+    }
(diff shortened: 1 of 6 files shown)


benthorner:
@alex-ju thanks :-). Could you help me understand where you're seeing the major difference here. As far as I can see, we're still listening for and emitting events in the same way; it's just the names of the attributes that have changed.

With regard to the scoping in GTM, I did toy with pushing a boolean flag onto the dataLayer, which the trigger could look for; but this is the same as the presence of the 'dataGtm' variable being defined. Does that answer your question there?

alex-ju:
I wrote 'emitting' but I meant sending to GTM (or storing in the dataLayer then sending them to GTM, to be more precise). I'm referring to this line https://github.com/alphagov/content-publisher/pull/1163/files#diff-09bd3e260c59a4ec26e759ad508d42d5L6 which was looking for a specific data attribute to exist on the element that triggered the event, which now I see that we moved into a separate condition.

PR #1165, opened 2019-06-20
This adds a spec for the PublishMailer

1pretz1:
As the ScheduledPublishMailer needs a mailer spec, we thought it would
be a good idea to include a PublishMailer spec. This also means we can
make the feature for publishing simpler by removing alternative publishing
scenarios which actually only tested the different variations of the
email (now covered by the tests in the publish_mailer_spec).

Trello:
https://trello.com/c/uC28c63Y/949-tie-up-scheduling-loose-ends

### DIFF
--- a/spec/mailers/publish_mailer_spec.rb
+++ b/spec/mailers/publish_mailer_spec.rb
@@ -1,0 +1,91 @@
+# frozen_string_literal: true
+
+RSpec.describe PublishMailer do
+  let(:recipient) { build(:user) }
+
+  describe "#publish_email" do
+    it "creates an email for the user" do
+      edition = build(:edition, :published, base_path: "/news/some-great-news")
+
+      mail = PublishMailer.publish_email(recipient, edition, edition.status)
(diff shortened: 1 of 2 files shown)


spec/mailers/publish_mailer_spec.rb
```diff
@@ -0,0 +1,88 @@
...
+        edition = build(:edition, :published, state: :published_but_needs_2i)
+
+        mail = PublishMailer.publish_email(recipient, edition, edition.status)
+
+        review_notice = I18n.t!("publish_mailer.publish_email.2i_warning")
+        expect(mail.body.to_s).to include(review_notice)
+      end
+    end
+
+    context "when a subsequent edition is published" do
```
kevindew: I think we should also check the subject here so that i18n key is executed. Might be able to include as part of one of the existing tests with a little rewording.

1pretz1: Yep you're correct, I've added it to the test `"includes when the edition was updated"` as naming wise it makes sense for it to be there.

PR #1168, opened 2019-06-21
Enhance input element rather than select for time input

kevindew:
Trello: https://trello.com/c/uC28c63Y/949-tie-up-scheduling-loose-ends

This refactors the autocomplete component to allow initialising based off on an input or a select element. See commits for further details.

### DIFF
--- a/app/views/components/docs/autocomplete.yml
+++ b/app/views/components/docs/autocomplete.yml
@@ -1,7 +1,9 @@
 name: Autocomplete
 description: An autocomplete component, built to be accessible
 body: |
-  This component is build using [Accessible Autocomplete](https://github.com/alphagov/accessible-autocomplete)
+  This component is build using [Accessible Autocomplete](https://github.com/alphagov/accessible-autocomplete).
+
+  Typically it is used to enhance a select element with autocomplete options.
 part_of_admin_layout: true
 accessibility_criteria: |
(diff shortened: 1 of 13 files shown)


app/views/components/docs/autocomplete.yml
```diff
@@ -12,190 +14,176 @@ examples:
...
+            - de
 
-  multiselect_with_selected_values:
+  autocomplete_narrow_width:
     data:
-      id: autocomplete-multiselect-selected
-      name: autocomplete-multiselect-selected
+      id: autocomplete-narrow
+      name: autocomplete-narrow
+      type: without-narrowing-results
```
alex-ju: I don't think we're using this type anymore. We probably want to set this to
```
data_attributes: {
  "autocomplete-without-narrowing-results": true
},
```
especially since we don't have an example of how an enhanced select element works with this feature.

kevindew:
@alex-ju Thanks a lot, I've fixed both of those things you've addressed.

kevindew:
Thanks @alex-ju - that scenario didn't work previously as we remove the select element: https://github.com/alphagov/content-publisher/blob/master/app/assets/javascripts/components/autocomplete.js#L135

I can't really see why we'd need that though since it's basically just a normal select.

alex-ju:
@kevindew it worked because the input was retaining the name and value of the select, but indeed it was only a visual enhancement. We'll approach it properly when its time comes. Thanks for the work on this one!

PR #1171, opened 2019-06-21
More consistent app-side__action padding

kevindew:
Trello: https://trello.com/c/uC28c63Y/949-tie-up-scheduling-loose-ends

This removes a persistent margin-bottom from app-side__actions box,
which doesn't seem to have a particular use and makes the padding on
this box bottom heavy.

To consistently make the padding even we need to remove the margins from
buttons when they are the last item in the actions. This can be
complicated because the buttons could be embedded in forms, which makes
the usage of :last-child tricky.

There is also a margin removed on .govuk-link elements which isn't
needed since it only has an effect when this is floated.

@alex-ju this all got a bit hairier than I'd have liked. You'll have to let me know if
I've made any CSS sins.

## Before:
Screen Shot 2019-06-21 at 15 31 25
Screen Shot 2019-06-21 at 15 31 18
Screen Shot 2019-06-21 at 15 31 12
Screen Shot 2019-06-21 at 15 31 07
Screen Shot 2019-06-21 at 15 31 01

## After:

Screen Shot 2019-06-21 at 16 21 49
Screen Shot 2019-06-21 at 16 21 44
Screen Shot 2019-06-21 at 16 21 36
Screen Shot 2019-06-21 at 16 21 31
Screen Shot 2019-06-21 at 16 21 25

### DIFF
--- a/app/assets/stylesheets/objects/_side.scss
+++ b/app/assets/stylesheets/objects/_side.scss
@@ -7,15 +7,17 @@
 }
 
 .app-side__actions {
-  @include govuk-responsive-margin(3, "bottom");
-
   .govuk-button {
     width: 100%;
     margin-bottom: govuk-spacing(3);
   }
 
-  .govuk-link {
-    margin-bottom: govuk-spacing(3);
+  > .govuk-button:last-child {
+    margin-bottom: 0;
+  }
+
+  > .app-side__form:last-child .govuk-button {
+    margin-bottom: 0;
   }
 
   .govuk-section-break {
--- a/app/helpers/actions_helper.rb
+++ b/app/helpers/actions_helper.rb
@@ -2,7 +2,9 @@
 
 module ActionsHelper
   def create_edition_button(edition, secondary: false)
-    form_tag create_edition_path(edition.document), data: { gtm: "create-new-edition" } do
+    form_tag create_edition_path(edition.document),
+             class: "app-side__form",
+             data: { gtm: "create-new-edition" } do
       render "govuk_publishing_components/components/button",
              text: "Create new edition",
              data_attributes: { gtm: "create-edition" },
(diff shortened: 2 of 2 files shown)


app/assets/stylesheets/objects/_side.scss
```diff
@@ -7,15 +7,17 @@
...
-  @include govuk-responsive-margin(3, "bottom");
-
   .govuk-button {
     width: 100%;
     margin-bottom: govuk-spacing(3);
   }
 
-  .govuk-link {
-    margin-bottom: govuk-spacing(3);
+  > .govuk-button:last-child {
```
alex-ju: We generally try to avoid using child combinators as they heavily rely on markup structure, but given the fact that we're unable to be in control on how the markup looks for an action (it could be inside our outside a form) I don't see a better option here..

app/assets/stylesheets/objects/_side.scss
```diff
@@ -7,15 +7,17 @@
...
 .app-side__actions {
-  @include govuk-responsive-margin(3, "bottom");
-
   .govuk-button {
     width: 100%;
     margin-bottom: govuk-spacing(3);
   }
 
-  .govuk-link {
-    margin-bottom: govuk-spacing(3);
```
alex-ju: I get that this is a leftover from when we introduced the paragraph with `govuk-body govuk-!-margin-bottom-3` classes (with which I'm not very happy with) to wrap these links.

app/assets/stylesheets/objects/_side.scss
kevindew: Yeah I felt it wasn't ideal but only adding very specific classes seemed to be an alternative.

app/assets/stylesheets/objects/_side.scss
kevindew: Sure, I think the problem we have on the links is the margin only has an effect on the floated ones.

PR #1173, opened 2019-06-24
ADR for rendering file attachments

kevindew:
Trello: https://trello.com/c/R1Erlbpa/878-write-an-adr-about-how-we-render-attachments-with-govspeak

[View Rendered 👀](https://github.com/alphagov/content-publisher/blob/file-attachment-rendering-adr/docs/adr/00xx-file-attachment-rendering.md)

### DIFF
--- a/docs/adr/00xx-file-attachment-rendering.md
+++ b/docs/adr/00xx-file-attachment-rendering.md
@@ -66,11 +66,12 @@ Pros:
 
 - single place for file attachment presentation logic;
 - HTML of file attachment can be customised for different scenarios;
-- allows wider use of the component beyond Content Publisher.
+- allows wider use of the components beyond Content Publisher.
 
 Cons:
 
-- most complex approach;
+- most complex approach for implementation, as GOV.UK Publishing Component
+  releases and integration need to be managed;
 - additional dependency for Govspeak.
 
 ## Decision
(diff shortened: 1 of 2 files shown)


docs/adr/00xx-file-attachment-rendering.md
```diff
@@ -0,0 +1,121 @@
...
+
+Govspeak now has a dependency on GOV.UK Publishing Components. In doing so
+this has introduced a [new approach to rendering a Govspeak markdown
+extension][govspeak-approach], which may cause confusion as there were
+already multiple. On the positive side, this will allow other components to be
+rendered through Govspeak which may allow future consolidation of presentation
+logic.
+
+GOV.UK Publishing Components no longer has a dependency on Govspeak. This
+needed to be removed to avoid a cyclic dependency. GOV.UK Publishing Components
```
emmabeynon: Minor rewording suggestion:

GOV.UK Publishing Components had been using Govspeak to render markdown.  However we needed to remove Govspeak as a dependency in order to avoid a cyclic dependency.  Since GOV.UK Publishing Components wasn't using any specific Govspeak extensions we decided to replace Govspeak with [Kramdown](https://kramdown.gettalong.org/), which incidentally is the library that Govspeak uses.

docs/adr/00xx-file-attachment-rendering.md
```diff
@@ -0,0 +1,121 @@
...
+To allow rendering markdown the [Kramdown](https://kramdown.gettalong.org/)
+library is used (which is the same library Govspeak uses).
+
+The rendering of attachments for [publication][publication-schema] documents
+could now be done on the frontend by using the file attachment components.
+This was previously impossible and the rendering of attachments has been
+achieved by Whitehall providing the Publishing API with [an array of HTML
+strings][attachment-html] to represent the attachments. This HTML based approach
+has made it very difficult for teams to extract attachment data from the
+[GOV.UK Content API][govuk-content-api].
```
emmabeynon: Do we expect that the new file attachment components could make it easier for teams to extract attachment data from Content API?

docs/adr/00xx-file-attachment-rendering.md
```diff
@@ -0,0 +1,121 @@
...
+
+Pros:
+
+- single place for file attachment presentation logic;
+- HTML of file attachment can be customised for different scenarios;
+- allows wider use of the component beyond Content Publisher.
+
+Cons:
+
+- most complex approach;
```
benthorner: It might be good to clarify why this is more complex. Presumably we mean for Govspeak specifically?

docs/adr/00xx-file-attachment-rendering.md
kevindew: It's somewhat convoluted. With the components there is now the logic to render the components on the frontend so we could stop sending blocks of HTML and replace that with data. When the Content API then had data it'd be easier to extract the information.

docs/adr/00xx-file-attachment-rendering.md
kevindew: I've updated the text to clarify.

kevindew:
@emmabeynon @benthorner comments addressed if you'd like to take another peak.

PR #1175, opened 2019-06-24
Rename scheduling helper

emmabeynon:
For https://trello.com/c/x6ImOUSP/918-create-a-form-so-user-can-save-the-backdate-for-their-content

This renames the Scheduling Helper to Date Time Helper, removes padding for dates and times and introduced a couple of new helpers to support date and time combinations.


emmabeynon:
@kevindew No tests are breaking when I remove the safe navigation operator so I will remove it 👍

PR #1183, opened 2019-06-24
Fix tracking attribute for index filters

benthorner:

This ensures we really don't set the 'data-gtm-visibility-tracking'
attribute when the value of the index filter is not present.


benthorner:
@kevindew 'fraid so :-(. I thought about detecting this in GTM, but then there are quite a few 'false' value we'd end up looking for.

PR #1189, opened 2019-06-25
Clear the set backdate if content has not been published

Tetrino:
Trello Card: https://trello.com/c/TBs5E8vv/919-clear-the-set-backdate-if-content-has-not-been-published

To allow the user to remove a backdate that has been applied to some content,
this commit does the following:

- Creates a Destroy Interactor class and associated route for the Backdate, tied into
the controller, to allow clean removal of the backdate value from the Edition.
The backdate is now set to Nil when prompted.

- Adds a user action on the Edit Backdate page to allow the backdate to be removed
if there is one set. This action is not present if no date has been set.

- Adds an entry in the history tab for when a backdate is removed. Wording to be
confirmed. Along these lines, fixed a bug wherein we were referring to the backdate
from the Edition directly rather than an appropriate Revision. Fix will make more
calls to the database but prevents a Nil error when we nullify the backdate value.

- Adds an error for if the user hits an issue when editing the backdate, similar to
scheduling.

### DIFF
--- a/app/controllers/backdate_controller.rb
+++ b/app/controllers/backdate_controller.rb
@@ -23,4 +23,15 @@ class BackdateController < ApplicationController
       redirect_to document_path(edition.document)
     end
   end
+
+  def destroy
+    result = Backdate::DestroyInteractor.call(params: params, user: current_user)
+
+    if result.api_error
+      redirect_to document_path(params[:document]),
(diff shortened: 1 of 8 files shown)


Tetrino:
Closed it for now, I've found an issue with how things are rendered.

Tetrino:
Fixed, reopening.

PR #1190, opened 2019-06-25
Clear the set backdate if content has not been published

Tetrino:
For https://trello.com/c/TBs5E8vv/919-clear-the-set-backdate-if-content-has-not-been-published

This allows users to clear a set backdate.  We also fix a bug where we used the edition's `backdate_to`
value instead of the `backdated_to` value associated with a revision, which caused nil errors and incorrect dates to appear in timeline entries.

### DIFF
--- a/app/controllers/backdate_controller.rb
+++ b/app/controllers/backdate_controller.rb
@@ -23,4 +23,11 @@ class BackdateController < ApplicationController
       redirect_to document_path(edition.document)
     end
   end
+
+  def destroy
+    result = Backdate::DestroyInteractor.call(params: params, user: current_user)
+    edition = result.edition
+
+    redirect_to document_path(edition.document)
+  end
 end
--- a/app/interactors/backdate/destroy_interactor.rb
+++ b/app/interactors/backdate/destroy_interactor.rb
@@ -9,6 +9,7 @@ class Backdate::DestroyInteractor
     Edition.transaction do
       find_and_lock_edition
       update_edition
+      create_timeline_entry
       update_preview
     end
   end
@@ -31,6 +32,15 @@ private
     context.fail! unless updater.changed?
 
     edition.assign_revision(updater.next_revision, user).save!
+  end
+
+  def create_timeline_entry
+    TimelineEntry.create_for_revision(
+      entry_type: :backdate_cleared,
+      revision: edition.revision,
+      edition: edition,
+      created_by: user,
+    )
   end
 
   def update_preview
--- a/app/interactors/backdate/destroy_interactor.rb
+++ b/app/interactors/backdate/destroy_interactor.rb
@@ -1,0 +1,39 @@
+# frozen_string_literal: true
+
+class Backdate::DestroyInteractor
+  include Interactor
+
+  delegate :params, :edition, :user, to: :context
+
+  def call
+    Edition.transaction do
+      find_and_lock_edition
+      update_edition
+      update_preview
+    end
+  end
+
+private
+
+  def find_and_lock_edition
+    context.edition = Edition.lock.find_current(document: params[:document])
+
+    unless edition.editable? && edition.first?
+      # FIXME: this shouldn't be an exception but we've not worked out the
+      # right response - maybe bad request or a redirect with flash?
+      raise "Only editable backdated first editions can have their backdated date cleared."
+    end
+  end
+
+  def update_edition
+    updater = Versioning::RevisionUpdater.new(edition.revision, user)
+    updater.assign(backdated_to: nil)
+    context.fail! unless updater.changed?
+
+    edition.assign_revision(updater.next_revision, user).save!
+  end
+
+  def update_preview
+    PreviewService.new(edition).try_create_preview
+  end
+end
--- a/app/views/documents/show/_history_tab.html.erb
+++ b/app/views/documents/show/_history_tab.html.erb
@@ -29,7 +29,7 @@
         <% end %>
 
         <% if entry.backdated? %>
-          <% date = format_date(entry.edition.backdated_to) %>
+          <% date = format_date(entry.revision.backdated_to) %>
           <%= t "documents.history.entry_content.backdated", date: date %>
         <% end %>
 
--- a/config/locales/en/documents/history.yml
+++ b/config/locales/en/documents/history.yml
@@ -33,5 +33,6 @@ en:
         file_attachment_uploaded: "Attachment uploaded"
         file_attachment_updated: "Attachment updated"
         backdated: "Backdated document"
+        backdate_cleared: "Backdate cleared"
       entry_content:
         backdated: "First published date backdated to %{date}"
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -74,6 +74,7 @@ Rails.application.routes.draw do
 
     get "/backdate" => "backdate#edit", as: :backdate
     post "/backdate" => "backdate#update"
+    delete "/backdate" => "backdate#destroy"
 
     post "/editions" => "editions#create", as: :create_edition
 
--- a/spec/features/workflow/clear_backdate_spec.rb
+++ b/spec/features/workflow/clear_backdate_spec.rb
@@ -1,0 +1,41 @@
+# frozen_string_literal: true
+
+RSpec.feature "Clear backdate" do
+  scenario do
+    given_there_is_an_editable_backdated_edition
+    when_i_visit_the_summary_page
+    i_can_see_the_date_the_content_has_been_backdated_to
+    when_i_click_to_edit_the_backdate
+    and_i_click_clear_backdate
+    then_i_see_the_content_is_no_longer_backdated
+  end
+
+  def given_there_is_an_editable_backdated_edition
+    @backdated_to = Time.current.yesterday
+    @edition = create(:edition, backdated_to: @backdated_to)
+  end
+
+  def when_i_visit_the_summary_page
+    visit document_path(@edition.document)
+  end
+
+  def i_can_see_the_date_the_content_has_been_backdated_to
+    backdate_title = I18n.t("documents.show.content_settings.backdate.title")
+    date = @backdated_to.strftime("%-d %B %Y")
+    expect(page).to have_content("#{backdate_title} #{date}")
+  end
+
+  def when_i_click_to_edit_the_backdate
+    click_on "Edit Backdate"
+  end
+
+  def and_i_click_clear_backdate
+    @request = stub_publishing_api_put_content(@edition.content_id, {})
+    click_on "Clear backdate"
+  end
+
+  def then_i_see_the_content_is_no_longer_backdated
+    expect(@request).to have_been_requested
+    expect(page).not_to have_content(@backdated_to.strftime("%-d %B %Y"))
+  end
+end
--- a/spec/features/workflow/clear_backdate_spec.rb
+++ b/spec/features/workflow/clear_backdate_spec.rb
@@ -36,6 +36,10 @@ RSpec.feature "Clear backdate" do
 
   def then_i_see_the_content_is_no_longer_backdated
     expect(@request).to have_been_requested
+
     expect(page).not_to have_content(@backdated_to.strftime("%-d %B %Y"))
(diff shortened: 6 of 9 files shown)


kevindew:
Just a heads up to anyone else on the team that I did a quick face-to-face review with @Tetrino on this.

app/controllers/backdate_controller.rb
```diff
@@ -23,4 +23,15 @@ def update
       redirect_to document_path(edition.document)
     end
   end
+
+  def destroy
+    result = Backdate::DestroyInteractor.call(params: params, user: current_user)
+
+    if result.api_error
```
kevindew: As this interactor can't return an api_error we don't need this code.

app/interactors/backdate/destroy_interactor.rb
```diff
@@ -0,0 +1,36 @@
...
+      find_and_lock_edition
+      update_edition
+      create_timeline_entry
+    end
+  end
+
+private
+
+  def find_and_lock_edition
+    context.edition = Edition.lock.find_current(document: params[:document])
```
kevindew: I think we should have a check for whether the document is editable here as per: https://github.com/alphagov/content-publisher/blob/9eb34fc4080532947c0d3c3ae6b9cae246cefba1/app/interactors/backdate/update_interactor.rb#L28-L32

app/interactors/backdate/destroy_interactor.rb
```diff
@@ -0,0 +1,36 @@
...
+  end
+
+  def update_edition
+    updater = Versioning::RevisionUpdater.new(edition.revision, user)
+    updater.assign(backdated_to: nil)
+
+    edition.assign_revision(updater.next_revision, user)
+           .save!
```
kevindew: super minor, but there's likely room for this method on the previous line.

app/interactors/backdate/destroy_interactor.rb
```diff
@@ -0,0 +1,36 @@
...
+class Backdate::DestroyInteractor
+  include Interactor
+
+  delegate :params, :edition, :user, to: :context
+
+  def call
+    Edition.transaction do
+      find_and_lock_edition
+      update_edition
+      create_timeline_entry
```
kevindew: We should update the preview here. It looks like that is a step that should be done in https://github.com/alphagov/content-publisher/blob/9eb34fc4080532947c0d3c3ae6b9cae246cefba1/app/interactors/backdate/update_interactor.rb also

app/interactors/backdate/destroy_interactor.rb
```diff
@@ -0,0 +1,36 @@
...
+    updater = Versioning::RevisionUpdater.new(edition.revision, user)
+    updater.assign(backdated_to: nil)
+
+    edition.assign_revision(updater.next_revision, user)
+           .save!
+  end
+
+  def create_timeline_entry
+    TimelineEntry.create_for_status_change(
+      entry_type: :backdate_deleted,
```
kevindew: minor but I think backdate_cleared would be more consistent with the terminology we present a user.

app/interactors/backdate/destroy_interactor.rb
```diff
@@ -0,0 +1,36 @@
...
+  def update_edition
+    updater = Versioning::RevisionUpdater.new(edition.revision, user)
+    updater.assign(backdated_to: nil)
+
+    edition.assign_revision(updater.next_revision, user)
+           .save!
+  end
+
+  def create_timeline_entry
+    TimelineEntry.create_for_status_change(
```
kevindew: Also this shouldn't be a `status_change` but a `create_for_revision`

app/views/documents/show/_history_tab.html.erb
```diff
@@ -29,10 +29,14 @@
         <% end %>
 
         <% if entry.backdated? %>
-          <% date = format_date(entry.edition.backdated_to) %>
+          <% date = format_date(entry.revision.backdated_to) %>
```
kevindew: Good catch! It'd be good if this bug fix was in a separate commit.

app/views/documents/show/_history_tab.html.erb
```diff
@@ -29,10 +29,14 @@
...
           <%= t "documents.history.entry_content.backdated", date: date %>
         <% end %>
 
+        <% if entry.backdate_deleted? %>
+          <%= t "documents.history.entry_content.backdate_deleted", date: date %>
```
kevindew: this date variable doesn't exist / isn't used, also I don't think we need an entry content for this action.

config/locales/en/documents/history.yml
```diff
@@ -33,5 +33,7 @@ en:
         file_attachment_uploaded: "Attachment uploaded"
         file_attachment_updated: "Attachment updated"
         backdated: "Backdated document"
+        backdate_deleted: "Backdate deleted"
```
kevindew: Again comment around cleared

config/locales/en/documents/history.yml
```diff
@@ -33,5 +33,7 @@ en:
...
       entry_content:
         backdated: "First published date backdated to %{date}"
+        backdate_deleted: "Document backdate removed"
```
kevindew: I don't think we need this.

config/locales/en/documents/show.yml
```diff
@@ -97,3 +97,6 @@ en:
         unschedule_error:
           title: Something has gone wrong
           description_govspeak: Something went wrong when stopping the scheduling of this document. Please try again or [raise a support request](https://support.publishing.service.gov.uk/technical_fault_report/new).
+        backdate_error:
+          title: Something has gone wrong
+          description_govspeak: Something went wrong when changing the backdate of this document. Please try again or [raise a support request](https://support.publishing.service.gov.uk/technical_fault_report/new).
```
kevindew: We shouldn't need this as this error can't occur

config/routes.rb
```diff
@@ -74,6 +74,7 @@
 
     get "/backdate" => "backdate#edit", as: :backdate
     post "/backdate" => "backdate#update"
+    delete "/backdate" => "backdate#destroy", as: :destroy_backdate
```
kevindew: I don't think we need to set a path name for this.

app/interactors/backdate/destroy_interactor.rb
emmabeynon: We're now raising an error if the edition is not editable and a first edition.

app/interactors/backdate/destroy_interactor.rb
emmabeynon: Now calling the Preview Service to update the Publishing API 👍

app/views/documents/show/_history_tab.html.erb
emmabeynon: This is now in a separate commit with a test file for editing a backdate which checks that there are two timeline entries with different dates.

app/controllers/backdate_controller.rb
emmabeynon: Removed

app/interactors/backdate/destroy_interactor.rb
emmabeynon: Changed to `backdated_cleared`.

app/interactors/backdate/destroy_interactor.rb
emmabeynon: Now using `create_for_revision`.

app/views/documents/show/_history_tab.html.erb
emmabeynon: Removed

config/locales/en/documents/history.yml
emmabeynon: As above

config/locales/en/documents/history.yml
emmabeynon: Removed

config/locales/en/documents/show.yml
emmabeynon: Removed

config/routes.rb
emmabeynon: Removed

emmabeynon:
This should be ready for re-review @kevindew

app/interactors/backdate/destroy_interactor.rb
```diff
@@ -0,0 +1,38 @@
...
+      # right response - maybe bad request or a redirect with flash?
+      raise "Only editable backdated first editions can have their backdated date cleared."
+    end
+  end
+
+  def update_edition
+    updater = Versioning::RevisionUpdater.new(edition.revision, user)
+    updater.assign(backdated_to: nil)
+
+    edition.assign_revision(updater.next_revision, user).save!
```
kevindew: We should check whether the updater has changed and abort if not - not sure if this is covered on the update interactor too

spec/features/workflow/clear_backdate_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+    @backdated_to = Time.current.yesterday
+    @edition = create(:edition, backdated_to: @backdated_to)
+  end
+
+  def when_i_visit_the_summary_page
+    visit document_path(@edition.document)
+  end
+
+  def i_can_see_the_date_the_content_has_been_backdated_to
+    expect(page).to have_content(@backdated_to.strftime("%-d %B %Y"))
```
kevindew: Is there any we can more specific for this and the other check?

I wonder if we can be checking for `expect(page).to have_content("#{backdate_title} #{backdate_time}")` as I think that would match?

app/interactors/backdate/destroy_interactor.rb
```diff
@@ -32,6 +33,15 @@ def update_edition
...
   end
 
+  def create_timeline_entry
+    TimelineEntry.create_for_revision(
+      entry_type: :backdate_cleared,
+      revision: edition.revision,
+      edition: edition,
+      created_by: user,
+    )
+  end
```
kevindew: 👍

spec/features/workflow/edit_backdate_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+    formatted_old_backdate = @old_backdate.strftime("%-d %B %Y")
+
+    expect(page).to have_content(
+      I18n.t!("documents.history.entry_types.backdated", date: formatted_old_backdate),
+    )
+    expect(page).to have_content(
+      I18n.t!("documents.history.entry_types.backdated", date: "1st January 2019"),
+    )
+  end
+end
```
kevindew: I'm a bit confused here as to how this test works. Since we created the backdated edition using FactoryBot and set to a backdate I don't know how generated the timeline entry that we confirm exists in line 38 :-/

Regardless of that I'm wondering whether we should have a test for this, mostly as it's whole extra feature test for a small tweak. I can definitely see why you put it together as there doesn't seem to be a clear way to test this exact scenario. I'd be satisfied though if you just didn't have a specific test for it (it's already covered in the general case in other tests so it has coverage)

app/interactors/backdate/destroy_interactor.rb
emmabeynon: It's not in the update interactor.  A quick look through the interactors shows that there are quite a few that don't abort if the updated has changed.  I will add it to this one, and raise a separate PR to add it consistently across the interactors.

kevindew: Thanks. Yeah we should try avoid creating timeline entries when they don't reflect any change to the underlying system state.

spec/features/workflow/clear_backdate_spec.rb
emmabeynon: Good idea.

spec/features/workflow/edit_backdate_spec.rb
emmabeynon: Fair enough, I thought that since it was a bug fix we should add a test to catch it, but agree it's not adding a tonne of value (and now you mention it, it is surprising that the timeline entry we check for in line 38 is there!).

emmabeynon:
@kevindew your comments should be addressed now 👍

PR #1194, opened 2019-06-26
The big one

benthorner:

The aim of this PR is to uplift the app so that it (mostly) fulfils ours tracking strategy, which is to "track everything" and still needs to be documented thoroughly. However, all the tracking that is added in this PR has a precedent elsewhere in the app. It's possible there are still some bits missing, but this should cover the majority of "user actions".

### DIFF
--- a/app/views/backdate/edit.html.erb
+++ b/app/views/backdate/edit.html.erb
@@ -13,9 +13,10 @@
     <%= render_govspeak(t("backdate.edit.description_govspeak")) %>
   </div>
 </div>
+
 <div class="govuk-grid-row">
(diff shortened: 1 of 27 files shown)


benthorner:
@kevindew thanks for reviewing :-). I've just pushed a small addition to track the sort header on the index page. That OK with you?

kevindew:
> @kevindew thanks for reviewing :-). I've just pushed a small addition to track the sort header on the index page. That OK with you?

Yup looks good

benthorner:
Thanks @alex-ju :-).

PR #1196, opened 2019-06-26
Fix 2i link tracking to use standard attributes

benthorner:

Previously we just used the 'utm_content' attribute, which isn't
sufficient to support tracking in GA. This swaps out the old attribute
for a new set that's consistent with the way we do this for email links.

### DIFF
--- a/app/views/documents/show/_submitted_for_review.html.erb
+++ b/app/views/documents/show/_submitted_for_review.html.erb
@@ -1,7 +1,10 @@
 <% copy_to_clipboard = render(
   "govuk_publishing_components/components/copy_to_clipboard",
     label: t("documents.show.flashes.submitted_for_review.label"),
-    copyable_content: document_url(@edition.document, utm_content: "2i-link"),
+    copyable_content: document_url(@edition.document,
+                                   utm_source: "2i-link",
+                                   utm_medium: "anonymous",
+                                   utm_campaign: "govuk-publishing"),
     button_text: "Copy link",
     button_data_attributes: {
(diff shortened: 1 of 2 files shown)


benthorner:
@emmabeynon sure - it's pretty unintuitive. Google Analytics has a bunch of fields to track the origin of inbound links to a site, which is useful to us as we want to track the usage of the links we give out. Unfortunately, this mechanism was only designed with paid advertising as the use case, so the fields don't make much sense for us (Wikipedia exemplifies this). My take on it:

   - **source** is the place they found the link; usually this is on another site like 'Google', but in our case it makes more sense to refer to the feature they were using, such as '2i-link' or 'publish-email'
   - **medium** is the context in which they use the link; usually this is something like 'email', or 'cpc' for paid adverts; perhaps 'copy-paste' would be more intuitive than 'anonymous', which is more related to the behaviour of the feature; splitting it into 'before-publish' and 'after-publish' could also work
   - **campaign** is just a way of grouping the other attributes, in case other people pollute our analytics with their own source/medium for some reason

I went with 'anonymous', as that does somewhat express the context in which they use the link. How about the alternatives I suggested above? This is yet another thing that really ought to go into that long-awaited documentation for how we do all this stuff :-).

kevindew:
Sorry I approved without noticing the conversation (and a rather bláse approach to the shrug to the "anonymous") - don't let my approval stop the conversation.

benthorner:
@kevindew and I merged, forgetting about the conversation. @emmabeynon happy to continue talking about this - it certainly made me think.

PR #1201, opened 2019-06-26
Fix withdraw guidance

benthorner:
This fixes a few issues with the withdraw feature:

   - Requirements issues should not be persistent flashes
   - The guidance should be confined to the central page
   - The README should correctly say how to enable the feature

### DIFF
--- a/app/views/withdraw/new.html.erb
+++ b/app/views/withdraw/new.html.erb
@@ -7,9 +7,11 @@
   </div>
 </div>
 
-<div class="govuk-grid-row">
-  <div class="govuk-grid-column-two-thirds">
-    <%= form_tag withdraw_path(@edition.document), data: { gtm: "confirm-withdraw" } do %>
+<%= form_tag withdraw_path(@edition.document),
+             class: "app-c-contextual-guidance__form",
+             data: { gtm: "confirm-withdraw" } do %>
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-two-thirds">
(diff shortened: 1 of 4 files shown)


benthorner:
@alex-ju I did a quick fix for the font size of the code example in the guidance, to stop it spilling outside the newly constrained guidance width. Is there a better way to approach this?

app/views/withdraw/new.html.erb
```diff
@@ -24,33 +26,34 @@
...
-        <%= render "govuk_publishing_components/components/button", {
-          text: "Withdraw document", margin_bottom: true
-        } %>
+    <div class="govuk-grid-column-one-third">
+      <%= render "components/contextual_guidance", {
+        id: "withdrawal-public-explanation-guidance",
+        title: t("withdraw.new.public_explanation.guidance_title"),
+      } do %>
+        <p>Explain to users why the page has been withdrawn. You can link to alternative content with markdown.</p>
+        <pre class="app-c-markdown-guidance__code-snippet" style="font-size: 12px">
```
alex-ju: Thanks for flagging this, I missed it when reviewing.
I would recommend adding `word-break: break-word;` to the `.app-c-markdown-guidance__code-snippet` class to deal with text overflowing issues and preserve the font-size as is.

benthorner:
@alex-ju thanks :-). I've pushed a fix - OK to merge?

alex-ju:
@benthorner. Yep, looks good now.

PR #1203, opened 2019-06-27
Initial model for Access Limit

kevindew:
Trello: https://trello.com/c/gdQrj56v/956-design-the-modelling-of-access-limiting

This creates a model for AccessLimit, sets up a method for assigning an access limit and removes an existing one at the point of publishing.

One thing I wrestled with a bit was the syntax for methods to add and remove access limits to an edition. I ended up going with a single method that can be used to add and remove:

```
edition.assign_access_limit(limit_type: :all_organisations, user: user) # to add
edition.assign_access_limit(limit_type: nil, user: user) # to remove
```

Another option could be `edition.remove_access_limit(user)` or `edition.assign_no_access_limit(user)`, I wasn't so keen on dropping the assign convention and felt `assign_no_access_limit` was a bit weird.

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,10 +10,18 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2019_06_17_142340) do
+ActiveRecord::Schema.define(version: 2019_06_26_155821) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
+
+  create_table "access_limits", force: :cascade do |t|
+    t.datetime "created_at"
+    t.bigint "created_by_id"
+    t.bigint "edition_id", null: false
+    t.bigint "revision_at_creation_id", null: false
+    t.string "limit_type", null: false
+  end
 
   create_table "active_storage_attachments", force: :cascade do |t|
     t.string "name", null: false
@@ -70,6 +78,8 @@ ActiveRecord::Schema.define(version: 2019_06_17_142340) do
     t.bigint "status_id", null: false
     t.bigint "revision_id", null: false
     t.boolean "revision_synced", default: false, null: false
+    t.bigint "access_limit_id"
+    t.index ["access_limit_id"], name: "index_editions_on_access_limit_id"
     t.index ["created_by_id"], name: "index_editions_on_created_by_id"
     t.index ["document_id", "current"], name: "index_editions_on_document_id_and_current", unique: true, where: "(current = true)"
     t.index ["document_id", "live"], name: "index_editions_on_document_id_and_live", unique: true, where: "(live = true)"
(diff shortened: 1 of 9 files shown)


spec/models/edition_spec.rb
```diff
@@ -207,4 +207,40 @@
       end
     end
   end
+
+  describe "#assign_access_limit" do
+    it "can set an access limit" do
+      edition = build(:edition)
+      user = build(:user)
+
+      expect { edition.assign_access_limit(limit_type: :all_organisations, user: user) }
```
benthorner: Minor, but would it be OK to lose this particular expectation? It looks like it's better covered by the subsequent ones, and would make the setup/run/test phases clearer.

spec/models/edition_spec.rb
```diff
@@ -207,4 +207,40 @@
...
+
+      expect(edition.access_limit).to be_all_organisations
+      expect(edition.access_limit.created_by).to eq(user)
+      expect(edition.access_limit.revision_at_creation).to eq(edition.revision)
+    end
+
+    it "can remove an access limit" do
+      edition = build(:edition, :access_limited)
+
+      expect { edition.assign_access_limit(limit_type: nil, user: build(:user)) }
```
benthorner: This seems a bit counter-intuitive, depending on how you interpret 'nil'. Perhaps a dedicated `remove_access_limit` method would be clearer, or an explicit 'everyone' option?

spec/models/edition_spec.rb
```diff
@@ -207,4 +207,40 @@
...
+        .from(an_instance_of(AccessLimit))
+        .to(nil)
+    end
+
+    it "updates the edition last edited information" do
+      edition = build(:edition)
+      user = build(:user)
+
+      travel_to(Time.current) do
+        expect { edition.assign_access_limit(limit_type: :all_organisations, user: user) }
```
benthorner: Similarly here, could we separate the setup/run/test phases in here, so it's a bit clearer?

db/schema.rb
```diff
@@ -10,11 +10,19 @@
...
-ActiveRecord::Schema.define(version: 2019_06_17_142340) do
+ActiveRecord::Schema.define(version: 2019_06_26_155821) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
 
+  create_table "access_limits", force: :cascade do |t|
+    t.datetime "created_at"
+    t.bigint "created_by_id"
+    t.bigint "edition_id", null: false
```
benthorner: I'm somewhat surprised this works as well, noticing you mentioned the expected race condition in your commit. However, should it actually b consistent with statuses, which don't have a not null requirement?

spec/models/edition_spec.rb
kevindew: Did you read PR description?

benthorner: Indeed I did - and the commit message that invites some comments about it ;-). But it's quite a minor thing, really, so I'm not fussed.

spec/models/edition_spec.rb
kevindew: I've removed this as I agree about the subsequent assertions.

spec/models/edition_spec.rb
kevindew: After much pondering I've gone with remove_access_limit and just dealt with it not specifying assign. It had the advantage that I could remove the keyword arguments which seemed inconsistent with most similar methods.

spec/models/edition_spec.rb
kevindew: I think this an appropriate way to test whether a method changes the state of an object. It has advantages that it checks something actually changes and the syntax is more declarative of what we're testing.

db/schema.rb
kevindew: This works because Edition has an optional association with an Access Limit. Which allows us to have a required one on the AccessLimit. 

This isn't the case with Status as an Edition requires it.

PR #1204, opened 2019-06-27
BH backdate release

benhazell:
For https://trello.com/c/MciuKE6C/970-release-backdating

updates to 
What's New
What Content Publisher Can and Cannot do
How to use Content Publisher

Date may need updating.

### DIFF
--- a/app/views/publisher_information/publisher_updates.govspeak.md
+++ b/app/views/publisher_information/publisher_updates.govspeak.md
@@ -1,3 +1,8 @@
+### Backdating <span class="govuk-caption-m">27 June 2019</span>
+You can now backdate on content if it has previously been available on another web page.
+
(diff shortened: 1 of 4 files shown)


app/views/publisher_information/publisher_updates.govspeak.md
```diff
@@ -1,3 +1,6 @@
+### Backdating <span class="govuk-caption-m">28 June 2019</span>
+You can now backdate the First Published date on content if it has previously been available on another website. It is important that old information is not presented a new.
```
emmabeynon: "is not presented a" = > "is not presented as"

emmabeynon: Should we mention it's only available on first editions?

benhazell: fixed typo, updated How to page to explain first editions.

PR #1206, opened 2019-06-27
Fail interactors when an updater hasn't changed

emmabeynon:
This will ensure additional actions, such as creating a timeline entry, don't
happen if an edition hasn't actually changed.


emmabeynon:
Closed in favour of https://github.com/alphagov/content-publisher/pull/1227

PR #1211, opened 2019-06-28
Trigger search topic event on autocomplete for search topics

alex-ju:
Trigger an event when searching for topics using the autocomplete component, so it can be tracked with analytics. It uses a `triggerEvent` function which acts as a polyfill for browsers that don't support CustomEvents..

It can be used/tested as shown in the example below:

```js
var autocomplete = document.querySelector('[data-autocomplete-type="topics"]')

autocomplete.addEventListener('search-topic', function(e){
  console.log(e.detail.topicNames.join(' > '))
})
```

[Trello card](https://trello.com/c/tAXP6pf1)

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -29,6 +29,7 @@
 //= require modal/modal-workflow.js
 
 //= require modules/gtm-copy-paste-listener.js
+//= require modules/gtm-topics-listener.js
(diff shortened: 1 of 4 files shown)


alex-ju:
@benthorner added the event listener to this PR. let me know if we need to update it in any way to make it consistent with how we currently track things.

alex-ju:
@benthorner I think this is now good for a final review.

PR #1224, opened 2019-07-09
Fix video woes

kevindew:
A couple of issues were caught with video embeds in the scenario testing round yesterday:

- an invalid URI caused the video embed requirements checker to exception
- an error response wasn't being handled by the modals.

It does look like we have a testing black spot with regards to the modal flows.

### DIFF
--- a/app/assets/javascripts/modal/modal-workflow.js
+++ b/app/assets/javascripts/modal/modal-workflow.js
@@ -6,6 +6,9 @@ function ModalWorkflow ($modal, actionCallback) {
 
   this.$multiSectionViewer = this.$modal
     .querySelector('[data-module="multi-section-viewer"]')
+
+  this.renderSuccess = this.renderSuccess.bind(this)
+  this.renderError = this.renderError.bind(this)
 }
 
 ModalWorkflow.prototype.initComponents = function () {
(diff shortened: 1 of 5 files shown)


kevindew:
> Could you explain a little more about the 'bind' issue? I'm curious how the modal ever worked if this was incorrect 🤔 .

Sure, so where we render an error in a modal module we were calling: `this.workflow.renderError.bind(this)` which needed to be binded to `this.workflow` rather than the modal module. This was not a problem for render success as we called that with: https://github.com/alphagov/content-publisher/blob/c0f3227021cb9ebfe33d5fe2b9d5cf5a25831293/app/assets/javascripts/modules/inline-image-modal.js#L60

which would have `this.workflow` as the `this` context so did work. The other place where we rendered an error: https://github.com/alphagov/content-publisher/blob/c0f3227021cb9ebfe33d5fe2b9d5cf5a25831293/app/assets/javascripts/modal/modal-workflow.js#L41

was already in the context of the workflow object so was correctly binded.

benthorner:
Ah, I see. So it was a specific case, but exemplified a bad/redundant pattern elsewhere 👍 . Oops.

PR #1226, opened 2019-07-09
Show Content Data URL when document is live

emmabeynon:
For https://trello.com/c/UPrhfB0e/966-link-to-content-data-admin-for-published-documents-without-a-delay

Previously we were only able to show a link to a document's Content Data URL
under certain conditions, otherwise users would be directed to a 404 page. This
issue has since been resolved in the Content Data app so we remove the logic
for displaying the link as it's no longer necessary. Now, we show the link when
a document has a live edition.


1pretz1:
All looks good to me however I just have one question, seen as though we only use the helper method once does it need to be a in separate helper or could the code just reside in the view?

emmabeynon:
@1pretz1 I did start out constructing the URL in the view code but it seemed a bit messy, so I decided to extract it out to a helper to clean it up.  It is only used once, however we do have a precedent with that in the Time Options Helper (though I appreciate the code for that is a bit longer).

1pretz1:
Ah yes I see, just seems to me a bit overkill to create a helper with one simpl'ish method which is only used once but I'm not overly fussed so i'll leave it up to you. :) 

As for your comment around having to use `@edition.document.live_edition.base_path`, I agree that its a bit horrible but you could assign a variable like `live_edition = @edition.document.live_edition` and use it in both the url logic and the conditional which would be nicer.

PR #1230, opened 2019-07-10
Move controller raises to interactors

benthorner:

Previously we had 3 sanity checks ('raise' statements) in our
controllers, whereas most are found our interactors. This moves
them into their associated interactors to be more consistent.


benthorner:
Superseded by https://github.com/alphagov/content-publisher/pull/1241.

PR #1237, opened 2019-07-12
WIP

benthorner:


app/services/assertions.rb
```diff
@@ -0,0 +1,42 @@
+module Assertions
+  class Error < StandardError
+    def initialize(assertion)
+      super("Assertion failed: #{assertion}")
+    end
+  end
+
+  class ErrorWithEdition < Error
```
theseanything: This feels these errors should be more specific i.e. Edition::NotEditableError,  Edition::BackdateRemovalError etc... (probably can think of better names) 

This would provide better descriptions of what happened and allow for more specific control flow depending on the issue.

app/models/edition.rb
```diff
@@ -214,4 +214,37 @@ def editors
...
+
+  #def assert_state(method)
+    #return if public_send(method)
+    #raise StateAssertionError.new(self, method)
+  #end
+
+  #def assert_block(message:, &block)
+    #return if block.call
+    #raise BlockAssertionError.new(self, message)
+  #end
```
theseanything: Commented code

app/services/assertions.rb
```diff
@@ -0,0 +1,42 @@
...
+  end
+
+  def assert_access(edition, user)
+    # TODO
+  end
+
+  def assert_permission(user, permission)
+    return if user.has_permission?(permission)
+    raise PermissionError.new(permission)
+  end
```
theseanything: Whats the difference between access and permission?

app/controllers/editions_controller.rb
```diff
@@ -3,12 +3,6 @@
...
   def create
     result = Editions::CreateInteractor.call(params: params, user: current_user)
-    if result.draft_current_edition
-      # FIXME: this shouldn't be an exception but we've not worked out the
-      # right response - maybe bad request or a redirect with flash?
-      raise "Can't create a new edition when the current edition is a draft"
-    else
-      redirect_to edit_document_path(params[:document])
-    end
+    redirect_to edit_document_path(params[:document])
```
theseanything: I somewhat feel that this may obfuscates the underlying check that happens - i.e. it reads that the controller will always redirect to the "edit document" path and wouldn't be clear that users can get rerouted to just the "document" path

app/controllers/application_controller.rb
```diff
@@ -10,6 +11,26 @@ class ApplicationController < ActionController::Base
...
 
+  #rescue_from Edition::IntegrityAssertionError do |e|
+    #Rails.logger.error(e.message)
+    #redirect_to document_path(params[:document])
+  #end
+
+  rescue_from Assertions::ErrorWithEdition do |e|
+    Rails.logger.error(e.message)
+    redirect_to document_path(e.edition.document)
+  end
```
theseanything: I feel this rescue is too generalised - would there be edition errors where we didn't want to redirect to document_path?

PR #1240, opened 2019-07-12
Add page preview script to hide the cookie banner in iframes

alex-ju:
This PR adds a script to the page preview component which send a message to the child document of an iframe to hide the cookie banner.

The message is a stringified Object `{'hideCookieBanner': 'true'}` with the indent to use a similar approach for other changes needed in the child document (e.g. making the links open in a new window)

## Events chain
→ `PagePreview.start`
→ `CookieBanner.start`
→ `CookieBanner.listenForCrossOriginMessages`
→ `PagePreview` iframe loads
→ `PagePreview.sendMessage`
→ `CookieBanner.receiveMessage`

## Dependencies
This PR is marked as draft as it should be merged after govuk_publishing_components is released with [an updated version of `CookieBanner` which captures messages](https://github.com/alphagov/govuk_publishing_components/pull/988).

## Tests
This behaviour was tested locally using an updated version of the cookie banner component.
Cross browser testing: Chrome, Edge, Firefox, IE8-11, Safari

### Before
localhost_3221_component-guide_page_preview_default_preview
localhost_3221_component-guide_page_preview_default_preview (1)

### After
localhost_3221_component-guide_page_preview_default_preview (2)
localhost_3221_component-guide_page_preview_default_preview (3)

[Trello card](https://trello.com/c/QXbQKBG3)

### DIFF
--- a/app/views/components/docs/page_preview.yml
+++ b/app/views/components/docs/page_preview.yml
@@ -8,7 +8,7 @@ accessibility_criteria: |
 examples:
   default:
     data:
-      title: "A carrot cake has been baked"
-      base_path: "/news/cake-baked"
-      description: A carrot cake was baked for the publishing workflow team's fika, the government announced today.
-      url: "https://www.gov.uk/browse"
+      title: Foreign Secretary heads to Brussels to discuss Iran
+      base_path: "/government/news/foreign-secretary-heads-to-brussels-to-discuss-iran"
+      description: Ahead of the Foreign Affairs Council in Brussels, Jeremy Hunt said he would be building on the leadership shown by the E3 to maintain the nuclear deal.
+      url: "https://draft-origin.publishing.service.gov.uk/government/news/foreign-secretary-heads-to-brussels-to-discuss-iran"
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -16,11 +16,12 @@
 //= require whatwg-fetch/dist/fetch.umd.js
 
 //= require components/autocomplete.js
-//= require components/toolbar-dropdown.js
 //= require components/image-cropper.js
 //= require components/input-length-suggester.js
 //= require components/markdown-editor.js
 //= require components/multi-section-viewer.js
+//= require components/page-preview.js
+//= require components/toolbar-dropdown.js
 //= require components/url-preview.js
 //= require miller-columns-element/dist/index.umd.js
 
--- a/app/assets/javascripts/components/page-preview.js
+++ b/app/assets/javascripts/components/page-preview.js
@@ -1,0 +1,30 @@
+window.GOVUK = window.GOVUK || {}
+window.GOVUK.Modules = window.GOVUK.Modules || {};
+
+(function (Modules) {
+  function PagePreview () { }
+
+  PagePreview.prototype.start = function ($module) {
+    this.$module = $module[0]
+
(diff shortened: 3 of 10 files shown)


app/views/components/docs/page_preview.yml
```diff
@@ -8,7 +8,7 @@ accessibility_criteria: |
...
   default:
     data:
-      title: "A carrot cake has been baked"
-      base_path: "/news/cake-baked"
-      description: A carrot cake was baked for the publishing workflow team's fika, the government announced today.
-      url: "https://www.gov.uk/browse"
+      title: Foreign Secretary heads to Brussels to discuss Iran
+      base_path: "/government/news/foreign-secretary-heads-to-brussels-to-discuss-iran"
+      description: Ahead of the Foreign Affairs Council in Brussels, Jeremy Hunt said he would be building on the leadership shown by the E3 to maintain the nuclear deal.
+      url: "https://draft-origin.publishing.service.gov.uk/government/news/foreign-secretary-heads-to-brussels-to-discuss-iran"
```
kevindew: This seems a lot less fun 😢

app/helpers/document_url_helper.rb
```diff
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-module EditionUrlHelper
+module DocumentUrlHelper
```
kevindew: Was there a reason not to go with just UrlHelper? It wouldn't surprise me too much if that caused problems.

I'm not too sure `DocumentUrlHelper` really explains this well, maybe it's ExternalUrlHelper since everything here is to URL's outside the app.

@benthorner any thoughts/ideas/preferences?

benthorner: Hmm...it looks like this change was made so we can use the new method in a view. Since the method is only one line, can we just duplicate it into the view?

app/views/components/docs/page_preview.yml
alex-ju: I wanted to be able to do manual testing using an item from the draft stack. Happy to keep the old title and description though.

app/helpers/document_url_helper.rb
kevindew: Can do but then we need to remember an interdependency between the host name used in the edition_preview_url method and in the view since they need to point to the same host.

app/views/components/docs/page_preview.yml
kevindew: No no, it's better to have something real. We'll just refer to you as the fun police 😉

app/helpers/document_url_helper.rb
alex-ju: `UrlHelper` felt too generic. Not very opinionated about the naming as long as it makes sense to everyone.

@benthorner I think the idea was to avoid duplication, but if it doesn't fit within the helper I'm fine duplicating it into the view.

alex-ju:
@kevindew yep, tested in intergration. Compared to the local testing we seem to have a bit of flickering (more prominently on first load).

hide-cookie-banner-on-iframes

app/helpers/document_url_helper.rb
alex-ju: Updated and renamed the helper to `ExternalUrlHelper`

kevindew:
@alex-ju should we be closing this in favour of a different PR?

alex-ju:
Yep, waited for an approval on the new approach. Closing this in favour of https://github.com/alphagov/govuk_publishing_components/pull/995

PR #1241, opened 2019-07-12
Assertions

benthorner:

This replaces and supplements our previous ad-hoc 'raise' assertions with a new EditionAssertions module and a new redirect behaviour for when they happen. Please see the commits for more details.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -11,6 +11,18 @@ class ApplicationController < ActionController::Base
 
   add_flash_types :alert_with_description, :alert_with_items, :confirmation, :tried_to_publish, :tried_to_preview
 
+  rescue_from EditionAssertions::StateError do |e|
+    Rails.logger.warn(e.message)
+
+    if rendering_context == "modal"
+      render nothing: true, status: :bad_request
+    elsif e.edition.first? && e.edition.discarded?
+      redirect_to documents_path
+    else
+      redirect_to document_path(e.edition.document)
+    end
+  end
+
   def rendering_context
     request.headers["Content-Publisher-Rendering-Context"] || "application"
   end
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -2,6 +2,7 @@
 
 class ApplicationController < ActionController::Base
   include GDS::SSO::ControllerMethods
+  include EditionAssertions
 
   helper_method :rendering_context
 
--- a/app/controllers/backdate_controller.rb
+++ b/app/controllers/backdate_controller.rb
@@ -3,6 +3,8 @@
 class BackdateController < ApplicationController
   def edit
     @edition = Edition.find_current(document: params[:document])
+    assert_edition_state(@edition, &:editable?)
+    assert_edition_state(@edition, &:first?)
   end
 
   def update
@@ -27,7 +29,6 @@ class BackdateController < ApplicationController
   def destroy
     result = Backdate::DestroyInteractor.call(params: params, user: current_user)
     edition = result.edition
-
     redirect_to document_path(edition.document)
   end
 end
--- a/app/services/edition_assertions.rb
+++ b/app/services/edition_assertions.rb
@@ -1,0 +1,30 @@
+# frozen_string_literal: true
+
+module EditionAssertions
+  class Error < StandardError
+    attr_reader :edition
+
+    def initialize(edition, assertion)
+      @edition = edition
+      super("Assertion failed: #{assertion}")
+    end
+  end
+
+  class StateError < Error
+    def initialize(edition, assertion)
+      super(edition, assertion || "meets state requirements")
+    end
+  end
+
+  def assert_edition_state(edition, options = {}, &block)
+    return if block.call(edition)
+
+    assertion = options[:assertion]
+    assertion ||= block.to_s if block.to_s =~ /&:/ # &:editable?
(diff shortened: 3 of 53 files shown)


app/services/edition_assertions.rb
```diff
@@ -0,0 +1,25 @@
...
+      super("Assertion failed: #{assertion}")
+    end
+  end
+
+  class StateError < Error; end
+
+  def assert_edition_state(edition, options = {}, &block)
+    return if block.call(edition)
+
+    assertion = options[:assertion] || block.inspect
```
kevindew: I wonder if we actually have a need for custom messaging? I'd also be keen if didn't use `block.inspect` since it's either hard or impossible to distinguish.

I think we'd be ok with just a generic message of "Edition does not meet state requirements"

app/controllers/backdate_controller.rb
```diff
@@ -3,6 +3,8 @@
 class BackdateController < ApplicationController
   def edit
     @edition = Edition.find_current(document: params[:document])
+    assert_edition_state(@edition, &:editable?)
+    assert_edition_state(@edition, &:first?)
```
kevindew: Is this intentionally 2 separate calls? I remember we touched on this briefly in conversation before and didn't conclude.

app/controllers/application_controller.rb
```diff
@@ -10,6 +11,18 @@ class ApplicationController < ActionController::Base
 
   add_flash_types :alert_with_description, :alert_with_items, :confirmation, :tried_to_publish, :tried_to_preview
 
+  rescue_from EditionAssertions::StateError do |e|
+    Rails.logger.warn(e.message)
```
kevindew: One thing we should do as part of this is just double check that this makes it to kibana. I get a bit confused whether it's supposed to or not. If it doesn't we might want to add a splash of request information to explain where it came from.

app/controllers/application_controller.rb
```diff
@@ -10,6 +11,18 @@ class ApplicationController < ActionController::Base
...
+
+    if rendering_context == "modal"
+      render status: :bad_request
```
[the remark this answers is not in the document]
benthorner: Oh yes - good spot. The response code is sufficient, but it doesn't make sense to be rendering a view for no reason.

app/controllers/application_controller.rb
```diff
@@ -10,6 +11,18 @@ class ApplicationController < ActionController::Base
...
+    elsif e.edition.reload.first? && e.edition.discarded?
```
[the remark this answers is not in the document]
benthorner: Yup - I forgot the reason I put it in :-|.

app/services/edition_assertions.rb
benthorner: The current mechanism has a small benefit, but perhaps it's unclear:

```
assert_edition_state(edition, &:editable?)
=> "Assertion failed: Proc<dfdfdfsdf>(:editable?)"

assert_edition_state(edition) { |e| e.editable? }
=> "Assertion failed: Proc<dfsdfdf>"
```

So the `assertion:` option is to support messaging for more custom blocks. Especially where we have complex assertions, I think it's helpful to write what we're actually asserting.

In order to make the logs useful, I think we should have something non-static in them. The stacktrace seems overkill for a non-error. I guess this approach has the problem of it being unclear when `assertion:` should be used.

app/controllers/backdate_controller.rb
benthorner: Not really - I just found it more intuitive to write this way. It has the small benefit of automatically logging the check (see above comment).

app/services/edition_assertions.rb
benthorner: I've amended with something a little more deterministic - how does it look now?

kevindew: Yeah that's what I meant about hard to distinguish, and impossible would be if you included a normal block without an assertion message. 

Yeah - that's what I suggested about some request information potentially at the log point, that'd give you something more useful in the log.

benthorner:
@kevindew as discussed, we can see these logs are popping out in `app.err.log` without any surrounding context. We investigated this and found its due to the [somewhat dubious way we setup our Rails logger](https://github.com/alphagov/govuk_app_config/blob/master/lib/govuk_app_config/govuk_logging.rb#L23) accept that further work is needed to improve our app config to make these logs practical for debugging purposes.

kevindew:
> @kevindew as discussed, we can see these logs are popping out in `app.err.log` without any surrounding context. We investigated this and found its due to the [somewhat dubious way we setup our Rails logger](https://github.com/alphagov/govuk_app_config/blob/master/lib/govuk_app_config/govuk_logging.rb#L23) accept that further work is needed to improve our app config to make these logs practical for debugging purposes.

By the way, from everything I can tell so far about logstasher won't show request info in the warn method. This is what I've seen so far: `{"level":"warn","message":"a","source":"unknown","tags":["log"],"@timestamp":"2019-07-15T11:06:11.529Z","@version":"1"}` so we may need to sprinkle that in, but can be a later PR

app/controllers/backdate_controller.rb
benthorner: Discussed - we may prefer a one-liner depending on how much we repeat these checks.

app/controllers/application_controller.rb
benthorner: Checked (see main comments on PR).

PR #1244, opened 2019-07-12
Move bits of withdraw service into interactor

benthorner:
Previously the WithdrawService managed the TimelineEntry for a withdraw
action, as well as change detection when editing a withdrawal; both of
these are standard tasks for an interactor. This moves both of these
responsibilities out of the service and into the associated interactor.

### DIFF
--- a/app/interactors/withdraw/create_interactor.rb
+++ b/app/interactors/withdraw/create_interactor.rb
@@ -7,6 +7,7 @@ class Withdraw::CreateInteractor < ApplicationInteractor
            :no_permission,
            :issues,
            :api_error,
+           :already_withdrawn,
            to: :context
 
   def call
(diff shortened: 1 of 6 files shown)


app/interactors/withdraw/create_interactor.rb
```diff
@@ -39,10 +42,26 @@ def check_for_issues
     context.fail!(issues: issues) if issues.any?
   end
 
-  def withdraw
+  def check_previous_withdrawal
+    context.already_withdrawn = edition.withdrawn?
+    return unless already_withdrawn
+
+    previous_explanation = edition.status.details.public_explanation
+    context.fail!(unchanged: true) if previous_explanation == params[:public_explanation]
```
kevindew: I think this unchanged may be superfluous - unless I missed something.

benthorner: True - I had a vague intention of referring to it in the controller. Removed.

PR #1250, opened 2019-07-15
Move schedule timeline entries into interactors

benthorner:

This makes scheduling consistent with other actions, noting that the
creation of timeline entries is not inherently part of the scheduling of
an edition.

### DIFF
--- a/app/interactors/schedule/create_interactor.rb
+++ b/app/interactors/schedule/create_interactor.rb
@@ -12,6 +12,7 @@ class Schedule::CreateInteractor < ApplicationInteractor
       find_and_lock_edition
       check_for_issues
       schedule_to_publish
+      create_timeline_entry
     end
   end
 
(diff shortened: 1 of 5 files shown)


app/interactors/schedule/create_interactor.rb
```diff
@@ -45,4 +46,12 @@ def schedule_to_publish
     reviewed = params[:review_status] == "reviewed"
     ScheduleService.new(edition).schedule(user: user, reviewed: reviewed)
   end
+
+  def create_timeline_entry
+    TimelineEntry.create_for_status_change(
+      entry_type: :schedule_updated,
```
kevindew: I think this should be `:scheduled`

kevindew: Shame we're not catching that on a test - if I'm right.

benthorner: Oops - good spot! This is an unfortunate side-effect of the same copy being used in the scheduling notice. I've amended the test to catch it now.

PR #1251, opened 2019-07-16
Access limit form

benthorner:

This allows the user to set an access limit for an edition and for this to be superficially enforced. Please see the commits for more details. Further work:

   - Showing organisation details when editing access limits
   - Showing organisation details on the 403 page
   - Allowing the user to clear an access limit


app/views/access_limit/edit.html.erb
```diff
@@ -0,0 +1,41 @@
...
+          },
+          {
+            value: "all_organisations",
+            text: t("access_limit.edit.type.all_organisations"),
+            data_attributes: {
+              gtm: "choose-access-limit",
+              "gtm-value": t("access_limit.edit.type.all_organisations")
+            }
+          }
+        ]
```
kevindew: Is it intentional at this point that this doesn't show which option is selected?

spec/support/authentication_helper.rb
```diff
@@ -11,5 +12,6 @@ def current_user
 
   def reset_authentication
     GDS::SSO.test_user = nil
+    Capybara.reset_sessions!
```
kevindew: It's a bit of a pain we need to add these as it couples this to Capybara. I guess it's necessary since warden must be not getting called if there is a user.

app/services/edition_assertions.rb
```diff
@@ -16,6 +16,12 @@ def initialize(edition, assertion)
     end
   end
 
+  class AccessError < Error
+    def initialize(edition, limit_type)
+      super(edition, "user is in #{limit_type}")
```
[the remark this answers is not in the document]
benthorner: Nope - the superclass expects an 'assertion' message, so it should be the thing we were checking.

app/services/edition_assertions.rb
```diff
@@ -24,7 +30,18 @@ def assert_edition_state(edition, options = {}, &block)
...
+  def assert_edition_access(edition, user)
+    return unless edition.access_limit
+
+    org_id = user.organisation_content_id
+    limit_type = edition.access_limit.limit_type
+
+    return if limit_type == "primary_organisation" &&
+      org_id == edition.primary_publishing_organisation_id
+
+    return if limit_type == "all_organisations" &&
```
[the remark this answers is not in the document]
benthorner: As discussed, it seems like there's no obvious improvement to make here, so I didn't want to make it a point of contention for this PR. However, I do think we should consider 'edition_organisations'.

spec/support/authentication_helper.rb
benthorner: Unfortunately, yes - seems to only work sometimes, otherwise.

app/services/edition_assertions.rb
```diff
@@ -24,7 +30,18 @@ def assert_edition_state(edition, options = {}, &block)
...
 
-  def assert_edition_access(_edition, _user)
-    raise "not implemented"
+  def assert_edition_access(edition, user)
+    return unless edition.access_limit
+
+    org_id = user.organisation_content_id
+    limit_type = edition.access_limit.limit_type
+
+    return if limit_type == "primary_organisation" &&
```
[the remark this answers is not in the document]
benthorner: Oh yes, sure. I was under the impression we did special delegation for Status#state that wasn't available here.

app/interactors/access_limit/update_interactor.rb
```diff
@@ -0,0 +1,36 @@
...
+    edition.assign_access_limit(params[:limit_type], user).save!
+  end
+
+  def create_timeline_entry
+    TimelineEntry.create_for_edition(
+      entry_type: :access_limit_updated,
+      created_by: user,
+      edition: edition,
+      details: edition.access_limit,
+    )
```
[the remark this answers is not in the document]
benthorner: The card didn't say ;-). I think that'll help if we want to distinguish the timeline entries in future, so we now support 'create' and 'update'.

app/views/access_limit/edit.html.erb
benthorner: Ah, no - fixed with a special 'edit' spec to cover this and the different timeline entry.

app/interactors/access_limit/update_interactor.rb
```diff
@@ -0,0 +1,43 @@
...
+      find_and_lock_edition
+      update_edition
+      create_timeline_entry
+    end
+  end
+
+private
+
+  def find_and_lock_edition
+    context.edition = Edition.find_current(document: params[:document])
```
kevindew: we seem to be missing the lock on this

app/views/documents/show/_content_settings.html.erb
```diff
@@ -6,19 +6,40 @@
...
+  <% if @edition.access_limit.present? %>
+    <% t("documents.show.content_settings.access_limit.type.#{@edition.access_limit.limit_type}") %>
+  <% else %>
+    <% t("documents.show.content_settings.access_limit.no_access_limit") %>
+  <% end %>
+<% end %>
+
+<% items = [] %>
+
+<% if current_user.has_permission?(User::PRE_RELEASE_FEATURES_PERMISSION) %>
```
kevindew: Well remembered - totally slipped my mind 🤦‍♂

app/interactors/access_limit/update_interactor.rb
```diff
@@ -0,0 +1,43 @@
...
+    assert_edition_state(edition, &:editable?)
+  end
+
+  def update_edition
+    context.fail! if edition.access_limit&.limit_type == params[:limit_type]
+    edition.assign_access_limit(params[:limit_type], user).save!
+  end
+
+  def create_timeline_entry
+    entry_type = if edition.access_limit_id_before_last_save.nil?
```
kevindew: Ooh didn't know Rails created these methods

app/interactors/access_limit/update_interactor.rb
benthorner: Oops - this alerted me to an unused method: Edition#find_and_lock_current. I've added a commit to remove it.

kevindew: Cool - sounds good 👍

PR #1252, opened 2019-07-16
Use Rails for rendering errors

kevindew:
When going to add another error page (400) I decided it would be better
if Rails was used as the rendering mechanism for errors. This has the
advantages that:

- Our error pages look more like our app
- Virtually no code to maintain for error pages
- Resolves issues such as 404s and missing links in the error pages
- Error text can be managed with translations

And has the disadvantage that:

- If an error occurs when rendering a 500 page a very bare page is
returned [1]

As far as I'm aware we've never had an instance where we've thrown an
exception during an error and if we choose to want to fix this we can
always add a block to the ErrorsController of:

```
rescue_from StandardError do
  render file: "#{Rails.root}/public/500.html", status: :internal_server_error
end
```

However I don't really think that's necessary given the very low
likelihood a user will receive that and that we don't provide nice
experiences for the much more likely scenarios of 502 or 504 responses.

---

This adds in error pages for 400 and 422 responses that should have their
content reviewed.

A user is likely to see a 400 response due to Rails StrongParams
failing, which may happen in the following cases:

- Developer error, requiring a form input that is missing
- User submitting a form that is old (where the backend code has since
changed)
- User meddling with HTML
- An attack on a Content Publisher endpoint

A user is likely to see a 422 response due to a CSRF check failing, which may
happen in the following cases:

- User submitting a form that is old, where their session key has since
changed
- User deleting their cookies
- User meddling with HTML
- An attempt to exploit a user by submitting a form from another site

---

Example of error page changes:

400 before:
Screen Shot 2019-07-16 at 14 19 22

400 after:

Screen Shot 2019-07-23 at 12 10 52

404 before:

Screen Shot 2019-07-16 at 14 18 02

404 after:

Screen Shot 2019-07-16 at 14 24 18

422 before:
Screen Shot 2019-07-16 at 14 19 37

422 after:

Screen Shot 2019-07-23 at 12 10 59

500 before:

Screen Shot 2019-07-16 at 14 18 08

500 after:

Screen Shot 2019-07-23 at 12 10 54

Catastrophic 500:

Screen Shot 2019-07-16 at 14 30 28

More details in the commits

[1]: https://github.com/rails/rails/blob/98a57aa5f610bc66af31af409c72173cdeeb3c9e/actionpack/lib/action_dispatch/middleware/show_exceptions.rb#L20-L24


kevindew:
> This works for me - good to get rid of those errors for our error pages! The behaviour for other, unhandled error codes appears to be the same, too. I believe the scope for this is distinct from access limiting, since it's domain-specific; but then again, we do handle the exception in the app controller 🤔 .

Yup, only cross over with access limiting is that in theory it could re-use the partial, but it's so simple that's not a biggie.

kevindew:
Notes from Ben H:

Drop the request reference in 400 users might not understand.

Change 422 page to be be:

Sorry, there was a problem
Please try again or raise a support request

Remove "later" from 500 text.

kevindew:
Now updating with @benhazell's suggestions - screenshots also updated.

1pretz1:
All looks good to me @kevindew!

PR #1258, opened 2019-07-17
Allow users to remove a set access limit

benthorner:

### DIFF
--- a/app/interactors/access_limit/update_interactor.rb
+++ b/app/interactors/access_limit/update_interactor.rb
@@ -22,12 +22,21 @@ private
   end
 
   def update_edition
-    context.fail! if edition.access_limit&.limit_type == params[:limit_type]
-    edition.assign_access_limit(params[:limit_type], user).save!
+    limit_type = params.require(:limit_type)
+
+    if limit_type == "none"
+      context.fail! if edition.access_limit.nil?
+      edition.remove_access_limit(user).save!
+    else
+      context.fail! if edition.access_limit&.limit_type == params[:limit_type]
+      edition.assign_access_limit(limit_type, user).save!
+    end
   end
 
   def create_timeline_entry
-    entry_type = if edition.access_limit_id_before_last_save.nil?
+    entry_type = if edition.access_limit.nil?
+                   :access_limit_removed
+                 elsif edition.access_limit_id_before_last_save.nil?
                    :access_limit_created
                  else
                    :access_limit_updated
--- a/app/views/access_limit/edit.html.erb
+++ b/app/views/access_limit/edit.html.erb
@@ -17,6 +17,15 @@
         name: "limit_type",
         items: [
           {
+            value: "none",
+            text: t("access_limit.edit.no_access_limit"),
+            checked: @edition.access_limit.nil?,
+            data_attributes: {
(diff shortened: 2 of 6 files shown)


app/views/access_limit/edit.html.erb
```diff
@@ -16,6 +16,15 @@
       <%= render "govuk_publishing_components/components/radio", {
         name: "limit_type",
         items: [
+          {
+            value: "none",
+            text: t("access_limit.edit.no_access_limit"),
+            checked: @edition.access_limit.nil?,
```
kevindew: This won't show correct value if a user has to re-submit the form, I assume we don't have requirements yet so can't currently manifest.

app/interactors/access_limit/update_interactor.rb
```diff
@@ -22,12 +22,19 @@ def find_and_lock_edition
...
 
   def update_edition
-    context.fail! if edition.access_limit&.limit_type == params[:limit_type]
-    edition.assign_access_limit(params[:limit_type], user).save!
+    if params[:limit_type] == "none"
+      context.fail! if edition.access_limit.nil?
+      edition.remove_access_limit(user).save!
+    else
+      context.fail! if edition.access_limit&.limit_type == params[:limit_type]
+      edition.assign_access_limit(params[:limit_type], user).save!
```
kevindew: Only thing of note here, that I didn't notice last time around, is that an unexpected limit_type value would cause app to exception quite far down the stack.

benthorner: True. The message from ActiveRecord seems helpful enough, so I don't think we don't need to check it again.

app/views/access_limit/edit.html.erb
benthorner: This is familiar :-). I think it would be better to cover this in the requirements spec - that it replays the set value correctly - when it comes along. How does that sound?

kevindew: Yup, no problem.

app/interactors/access_limit/update_interactor.rb
kevindew: You know me, I don't like that an error with client input causes an internal server error rather than a client error.

benthorner: It's good to think about it. Looks like it's hard to do this with params, but I've gone with a compromise to give a bad request if the param isn't there.

kevindew: Sure, I was expecting this to be something thought about requirements point since we could check it's there and an expected value in one check.

benthorner: Not sure I follow there - what would we do if the check failed? It doesn't sound like a requirements issue to me.

kevindew: I was thinking of it in the same vein as: https://github.com/alphagov/content-publisher/blob/master/app/interactors/publish/publish_interactor.rb#L37-L39

benthorner: Still a bit confused :-). Let's talk in person. Anyway, it looks like this particular change is good to merge, so I'll get it in now.

PR #1259, opened 2019-07-17
Access limit everywhere

benthorner:

    This adds assertions for access limited editions in all places where
    (part of) the edition is exposed but the edition is not yet public. One
    exception to this is the 'debug' view, which has a developer-specific
    permission associated with it, which we will treat as an override.

### DIFF
--- a/app/controllers/access_limit_controller.rb
+++ b/app/controllers/access_limit_controller.rb
@@ -3,6 +3,7 @@
 class AccessLimitController < ApplicationController
   def edit
     @edition = Edition.find_current(document: params[:document])
+    assert_edition_access(@edition, current_user)
   end
 
   def update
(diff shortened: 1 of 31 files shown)


app/views/tags/edit.html.erb
```diff
@@ -18,7 +18,7 @@
       <div class="govuk-grid-column-two-thirds" id="<%= tag_field.id %>">
         <%= render "tags/tags/#{tag_field.type}_input",
           tag_field: tag_field,
-          tags: @revision.tags,
+          tags: @edition.tags,
```
kevindew: Was this meant to be in this commit? 

Anywho I believe it should still be `@revision` as that reflects the users input if validation fails whereas edition would always be the saved version.

benthorner: Ah true - I didn't consider the 'update' action.

benthorner: Reverted it back.

PR #1260, opened 2019-07-17
Revert "Remove redundant id attributes"

benthorner:
This reverts commit 82405a760da05d70efa318856c3deabb089d5c58.


benthorner:
Good spot - I've put in another commit to fix that. Doing this makes me wonder if it would be simpler to re-use the existing IDs on (most of) the input elements. Anyway, let's fix it for now...

PR #1261, opened 2019-07-17
Filter editions to respect access limiting

benthorner:
https://trello.com/c/OJwGOdvd/988-dont-show-access-limited-documents-in-index-page-when-user-is-not-part-of-a-relevant-organisation

This adds an additional filtering step to the EditionFilter that scopes
the query to only include access limited editions the user can access.

### DIFF
--- a/app/services/edition_filter.rb
+++ b/app/services/edition_filter.rb
@@ -9,20 +9,23 @@ class EditionFilter
   SORT_KEYS = %w[last_updated].freeze
   DEFAULT_SORT = "-last_updated"
 
-  attr_reader :filters, :sort, :page, :per_page
+  attr_reader :filters, :sort, :page, :per_page, :user
 
-  def initialize(params)
+  def initialize(user, **params)
     @filters = params[:filters].to_h.symbolize_keys
     @sort = allowed_sort?(params[:sort]) ? params[:sort] : DEFAULT_SORT
     @page = params.fetch(:page, 1).to_i
     @per_page = params[:per_page]
+    @user = user
   end
 
   def editions
     revision_joins = { revision: %i[content_revision tags_revision] }
     scope = Edition.where(current: true)
+                   .left_joins(:access_limit)
                    .joins(revision_joins, :status, :document)
                    .preload(revision_joins, :status, :document, :last_edited_by)
+    scope = access_limited_scope(scope)
     scope = filtered_scope(scope)
     scope = ordered_scope(scope)
     scope.page(page).per(per_page)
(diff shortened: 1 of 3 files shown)


app/services/edition_filter.rb
```diff
@@ -41,6 +44,22 @@ def allowed_sort?(sort)
...
+      "(
+         access_limits.limit_type = 'primary_organisation' AND
+         #{TAG_CONTAINS_QUERY % { tag: 'primary_publishing_organisation' }}
+       )" + " OR " +
+      "(
+         access_limits.limit_type = 'all_organisations' AND
+         (#{TAG_CONTAINS_QUERY % { tag: 'primary_publishing_organisation' }} OR
+          #{TAG_CONTAINS_QUERY % { tag: 'organisations' }})
+       )",
+      value: user.organisation_content_id,
```
kevindew: Do you know if this works ok if the value of organisation_content_id is nil?

app/services/edition_filter.rb
```diff
@@ -9,20 +9,23 @@ class EditionFilter
...
     @sort = allowed_sort?(params[:sort]) ? params[:sort] : DEFAULT_SORT
     @page = params.fetch(:page, 1).to_i
     @per_page = params[:per_page]
+    @user = user
   end
 
   def editions
     revision_joins = { revision: %i[content_revision tags_revision] }
     scope = Edition.where(current: true)
+                   .joins("LEFT OUTER JOIN access_limits ON access_limits.id = access_limit_id")
```
kevindew: I think we can use `.left_outer_joins(:access_limit)` or `.left_joins(:access_limit)` rather than writing the whole join.

app/services/edition_filter.rb
```diff
@@ -41,6 +44,22 @@ def allowed_sort?(sort)
     SORT_KEYS.flat_map { |item| [item, "-#{item}"] }.include?(sort)
   end
 
+  def access_limited_scope(scope)
+    scope.where(
+      "access_limit_id IS NULL" + " OR " +
+      "(
+         access_limits.limit_type = 'primary_organisation' AND
+         #{TAG_CONTAINS_QUERY % { tag: 'primary_publishing_organisation' }}
```
kevindew: This constant seems to have proved rather handy at concealing the rather grim SQL 👍

app/services/edition_filter.rb
benthorner: It does, but I've added a couple of extra tests to make sure.

app/services/edition_filter.rb
benthorner: Phew - I couldn't find this when I looked initially 🎆 .

benthorner:
@kevindew this should be ready for another review now :-).

PR #1262, opened 2019-07-17
Improve confirmation pages

alex-ju:
This PR updates the confirmation pages (for scheduled and published workflows) to render markup consistently and use inset text and links for URLs. It also breaks words in URLs for both links within inset-text components and for URLs in the url-preview component. This prevents long URLs from overflowing the containers (which is currently creating issues on mobile devices).

#### Confirmation for published
confirmation-published

#### Confirmation for scheduled
confirmation-scheduled

### Mobile (similar results for scheduled and published)
confirmation-published-mobile

[Trello card](https://trello.com/c/X9vsbFTC)

### DIFF
--- a/app/views/publish/published.html.erb
+++ b/app/views/publish/published.html.erb
@@ -6,6 +6,11 @@
 
 <% content_for :browser_title, t("publish.published.title") %>
 <% public_url = edition_public_url(@edition) %>
+<% published_edition_link = render("govuk_publishing_components/components/inset_text", {
+    text: link_to(strip_scheme_from_url(public_url), public_url,
+                  class: "govuk-link govuk-link--no-visited-state",
+                  data: { gtm: "view-published-edition" })
+}) %>
 
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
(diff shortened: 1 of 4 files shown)


app/views/publish/published.html.erb
```diff
@@ -44,13 +45,13 @@
...
 
-      <%= render "govuk_publishing_components/components/govspeak" do %>
-        <%= govspeak_to_html t("publish.published.reviewed.body_govspeak", title: @edition.title) %>
+      <%= render_govspeak(t("publish.published.reviewed.body_govspeak", title: @edition.title)) %>
 
-        <%= link_to(strip_scheme_from_url(public_url), public_url,
-                    class: "govuk-link govuk-link--no-visited-state",
-                    data: { gtm: "view-published-edition" }) %>
-      <% end %>
+      <%= render "govuk_publishing_components/components/inset_text", {
```
brucebolt: Have you considered moving this into a partial to DRY up the code, since the same code for the inset text is also used on L20-24?

alex-ju: Yes, I noticed the repetition. I guess storing it into a variable using a `capture do` block could be a decent option.

alex-ju: @brucebolt updated to use a variable for the published edition link

PR #1264, opened 2019-07-18
Access limit hints

benthorner:

This adds organisation 'hints' when the user is editing the access limit for an edition, and when they see a 'forbidden' page, so they can follow-up with the owner org.

### DIFF
--- a/app/views/access_limit/edit.html.erb
+++ b/app/views/access_limit/edit.html.erb
@@ -6,11 +6,12 @@
 
 <% if @edition.primary_publishing_organisation_id %>
   <% primary_org_id = @edition.primary_publishing_organisation_id %>
-  <% primary_org = service.by_content_id(primary_org_id)["internal_name"] %>
+  <% primary_org = service.by_content_id(primary_org_id)&.fetch("internal_name") %>
 <% end %>
 
 <% supporting_orgs = @edition.organisations
-  .map { |org_id| service.by_content_id(org_id)["internal_name"] } %>
+  .map { |org_id| service.by_content_id(org_id)&.fetch("internal_name") }
+  .compact %>
 
 <% tagged_orgs = ([primary_org] + supporting_orgs).compact %>
 
--- a/app/views/access_limit/edit.html.erb
+++ b/app/views/access_limit/edit.html.erb
@@ -2,6 +2,26 @@
 <% content_for :browser_title, t("access_limit.edit.browser_title") %>
 <% content_for :back_link, render_back_link(href: document_path(@edition.document)) %>
 
+<% service = LinkablesService.new("organisation") %>
+
+<% if @edition.primary_publishing_organisation_id %>
+  <% primary_org_id = @edition.primary_publishing_organisation_id %>
+  <% primary_org = service.by_content_id(primary_org_id)["internal_name"] %>
+<% end %>
+
+<% supporting_orgs = @edition.organisations
+  .map { |org_id| service.by_content_id(org_id)["internal_name"] } %>
+
+<% tagged_orgs = ([primary_org] + supporting_orgs).compact %>
+
+<% tagged_orgs_ul = capture do %>
+  <ul class="govuk-list">
+    <% tagged_orgs.each do |org| %>
+      <%= tag.li org %>
+    <% end %>
+  </ul>
+<% end %>
+
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
     <div class="govuk-body" %>
@@ -32,7 +52,8 @@
             data_attributes: {
               gtm: "choose-access-limit",
               "gtm-value": t("access_limit.edit.type.primary_organisation")
-            }
+            },
+            hint_text: primary_org
           },
           {
             value: "tagged_organisations",
@@ -41,7 +62,8 @@
             data_attributes: {
               gtm: "choose-access-limit",
               "gtm-value": t("access_limit.edit.type.tagged_organisations")
-            }
+            },
+            hint_text: tagged_orgs.any? && tagged_orgs_ul
           }
         ]
       } %>
--- a/app/views/documents/forbidden.html.erb
+++ b/app/views/documents/forbidden.html.erb
@@ -4,7 +4,11 @@
   <%= t("documents.forbidden.description") %>
 </div>
 
-<% primary_org = LinkablesService.new("organisation")
- .by_content_id(@edition.primary_publishing_organisation_id) %>
+<div class="govuk-body">
+  <% begin %>
+   <% primary_org = LinkablesService.new("organisation")
+     .by_content_id(@edition.primary_publishing_organisation_id) %>
 
-<%= t("documents.forbidden.owner", primary_org: primary_org["internal_name"]) %>
+   <%= t("documents.forbidden.owner", primary_org: primary_org["internal_name"]) %>
+  <% rescue GdsApi::BaseError; end %>
+</div>
--- a/app/views/documents/forbidden.html.erb
+++ b/app/views/documents/forbidden.html.erb
@@ -6,9 +6,12 @@
 
 <div class="govuk-body">
   <% begin %>
-   <% primary_org = LinkablesService.new("organisation")
-     .by_content_id(@edition.primary_publishing_organisation_id) %>
+    <% primary_org = LinkablesService.new("organisation")
+      .by_content_id(@edition.primary_publishing_organisation_id) %>
 
-   <%= t("documents.forbidden.owner", primary_org: primary_org["internal_name"]) %>
+    <% if primary_org %>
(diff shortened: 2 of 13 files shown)


app/views/access_limit/edit.html.erb
```diff
@@ -41,7 +52,8 @@
             data_attributes: {
               gtm: "choose-access-limit",
               "gtm-value": t("access_limit.edit.type.all_organisations")
-            }
+            },
+            hint_text: supporting_orgs.any? && "[#{supporting_orgs.join(",")}]"
```
kevindew: If I understand right, given a primary org of GDS and supporting orgs of HMRC and DWP this would show: [HMRC, DWP]. Whereas my understanding from the [design](https://content-publisher-prototype.cloudapps.digital/access-limit) is that this should show:

[GDS]
[HMRC]
[DWP]

app/views/documents/forbidden.html.erb
```diff
@@ -3,3 +3,8 @@
 <div class="govuk-body">
   <%= t("documents.forbidden.description") %>
 </div>
+
+<% primary_org = LinkablesService.new("organisation")
+ .by_content_id(@edition.primary_publishing_organisation_id) %>
+
+<%= t("documents.forbidden.owner", primary_org: primary_org["internal_name"]) %>
```
kevindew: This text is coming through as unstyled as it's not within a `.govuk-body`, also wondering what about nil case again.

app/views/access_limit/edit.html.erb
```diff
@@ -2,6 +2,16 @@
 <% content_for :browser_title, t("access_limit.edit.browser_title") %>
 <% content_for :back_link, render_back_link(href: document_path(@edition.document)) %>
 
+<% service = LinkablesService.new("organisation") %>
+
+<% if @edition.primary_publishing_organisation_id %>
+  <% primary_org_id = @edition.primary_publishing_organisation_id %>
+  <% primary_org = service.by_content_id(primary_org_id)["internal_name"] %>
```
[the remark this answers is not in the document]
benthorner: Ah, I forgot to talk with you about this earlier. Yes, it is a risk and we [have a precedent for not dealing with it](https://github.com/alphagov/content-publisher/blob/e576108992a71af9c9ff7fc065fbdebb6bceeb5b/app/views/documents/tags/_multi_tag.html.erb#L4). I was wondering if we should do something slightly better (e.g. `.fetch("internal_name", "something")`) to be more friendly to people like me, who don't have a data import.

app/views/access_limit/edit.html.erb
benthorner: You're correct. I've documented in the commit that we will support this in further work, as it depends on a [change to gov pub cub](https://github.com/alphagov/govuk_publishing_components/pull/996).

kevindew: Ah sure, might be worth including the primary one in this style too for the time being so that doesn't risk getting missed.

app/views/documents/forbidden.html.erb
benthorner: Well spotted! I forgot to check the updated page after I finished grappling with the linkables. Fixed, but same comment about the nil case as above.

app/views/access_limit/edit.html.erb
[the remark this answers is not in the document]
kevindew: Ah sure, yeah I see how that other example could crash too. 

I don't think fetch is the right approach as that isn't defined on nil. It's a shame that the safe navigation operator is a bit weird with a hash as could use `&.[]("internal_name")` but that's not the most intuitive syntax. Maybe using fetch without a fallback would be ok? and then mapped result could have compact executed on it while we wait for [`#filter_map`](https://blog.saeloun.com/2019/05/25/ruby-2-7-enumerable-filter-map.html) to arrive in 2.7.

app/views/access_limit/edit.html.erb
alex-ju: Hint component renders in a `<div>` element now and Content Publisher is using the latest version of govuk_publishing_components so this could be rebased and updated to display organisations in a list element `<ul class="govuk-list">`. We may need a colour overwrite though, let me know if that's the case.

benthorner: I've updated to use the new gov pub cub release, but you're right that we need an override for the colour.

Screen Shot 2019-07-19 at 13 43 23

alex-ju: Can you add the following attribute in `app/assets/stylesheets/utilities/_overwrites.scss`?

```
.govuk-hint {
  .govuk-list {
    color: $govuk-secondary-text-colour;
  }
}
```

I will raise it upstream, but to have it working now.

app/views/access_limit/edit.html.erb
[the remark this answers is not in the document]
benthorner: Sure - I've made that change where we're using it.

benthorner:
@kevindew I think this is ready for another review now :-)

app/views/access_limit/edit.html.erb
```diff
@@ -2,6 +2,25 @@
...
+<% service = LinkablesService.new("organisation") %>
+
+<% if @edition.primary_publishing_organisation_id %>
+  <% primary_org_id = @edition.primary_publishing_organisation_id %>
+  <% primary_org = service.by_content_id(primary_org_id).fetch("internal_name") %>
+<% end %>
+
+<% supporting_orgs = @edition.organisations
+  .map { |org_id| service.by_content_id(org_id).fetch("internal_name") }
+  .compact %>
```
kevindew: This list should start with the primary org

app/views/access_limit/edit.html.erb
```diff
@@ -2,6 +2,25 @@
 <% content_for :browser_title, t("access_limit.edit.browser_title") %>
 <% content_for :back_link, render_back_link(href: document_path(@edition.document)) %>
 
+<% service = LinkablesService.new("organisation") %>
+
+<% if @edition.primary_publishing_organisation_id %>
+  <% primary_org_id = @edition.primary_publishing_organisation_id %>
+  <% primary_org = service.by_content_id(primary_org_id).fetch("internal_name") %>
```
kevindew: This will still error on a nil, you'll need to use the safe navigation operator before the fetch

benthorner: Oops - going to quickly. Fixed now.

app/views/access_limit/edit.html.erb
benthorner: Of course it should :-). The component makes this a bit tricky to test for - see what you think.

benthorner:
@kevindew could you take another look?

spec/features/editing_content_settings/set_access_limit_spec.rb
```diff
@@ -58,11 +64,14 @@ def and_i_edit_the_access_limit
   end
 
   def and_i_limit_to_my_organisation
+    expect(page).to have_content("Primary org")
     choose I18n.t!("access_limit.edit.type.primary_organisation")
     click_on "Save"
   end
 
   def and_i_limit_to_tagged_organisations
+    expect(page).to have_content("Primary org", count: 2)
```
kevindew: Ah I see this is what you meant in your comment earlier - it does feel a bit of a loose test as any other places where we showed the organisation name would pass.

An alternative approach you could take is to check for "Primary org Supporting org" since that'll only match this grouping. 

Otherwise could use a within though that'd be a bit tricky to match the element, for example:

```
radio = page.find(:radio_button, I18n.t!("access_limit.edit.type.tagged_organisations"))
within radio.find(:xpath, "..") do 
 radio.choose
 expect(page).to have_content("Primary org")
 expect(page).to have_content("Supporting org")
end
```

benthorner: I agree. But I'd like to find another way:

   - The concatenated text approach did occur to me, but it just looks weird
   - The xpath approach also occurred, but it makes the tests more complex

I've pushed up a compromise that hopefully resolves the concern identified, although I can't say I like it much. This does seem like an integral part of the feature, so worth the testing effort 💪 .

app/views/documents/forbidden.html.erb
```diff
@@ -3,3 +3,13 @@
...
   <%= t("documents.forbidden.description") %>
 </div>
+
+<div class="govuk-body">
+  <% begin %>
+   <% primary_org = LinkablesService.new("organisation")
+     .by_content_id(@edition.primary_publishing_organisation_id) %>
+
+   <%= t("documents.forbidden.owner",
+         primary_org: primary_org&.fetch("internal_name")) %>
```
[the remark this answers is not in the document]
benthorner: 2ndline is lowering my standards - it wouldn't explode ;-). Fixed.

PR #1265, opened 2019-07-18
Default to all access if user subverts form

benthorner:

Previously we raised an ArgumentError if the user managed to submit the
access limit form without a selection, or with an incorrect limt_type
value. This iterates the approach to be more comparable with e.g. the
PublishInteractor, so that we remove any access limit by default.

### DIFF
--- a/app/interactors/access_limit/update_interactor.rb
+++ b/app/interactors/access_limit/update_interactor.rb
@@ -1,6 +1,8 @@
 # frozen_string_literal: true
 
 class AccessLimit::UpdateInteractor < ApplicationInteractor
+  LIMIT_TYPES = AccessLimit.limit_types.keys
+
   delegate :params,
            :user,
            :edition,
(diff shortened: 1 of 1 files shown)


app/interactors/access_limit/update_interactor.rb
```diff
@@ -24,12 +26,12 @@ def find_and_lock_edition
   def update_edition
     limit_type = params.require(:limit_type)
 
-    if limit_type == "none"
-      context.fail! if edition.access_limit.nil?
-      edition.remove_access_limit(user).save!
-    else
+    if LIMIT_TYPES.include?(limit_type)
       context.fail! if edition.access_limit&.limit_type == params[:limit_type]
```
kevindew: Shall we use the limit_type variable we assigned here too?

app/interactors/access_limit/update_interactor.rb
```diff
@@ -1,6 +1,8 @@
 # frozen_string_literal: true
 
 class AccessLimit::UpdateInteractor < ApplicationInteractor
+  LIMIT_TYPES = AccessLimit.limit_types.values
```
kevindew: Minor, this probably should refer to keys rather than values. Since the key refers to the enum text and the value is what the item is stored as in the DB.

Has no effect for us since we have the same value for key and value.

app/interactors/access_limit/update_interactor.rb
benthorner: Oh yes - thanks for spotting :-).

app/interactors/access_limit/update_interactor.rb
benthorner: Interesting. I went with 'values' for the same reason. I guess we have different interpretations!

benthorner:
@kevindew this should be ready for another review now :-).

PR #1266, opened 2019-07-18
Rename all_organisations -> tagged_organisations

benthorner:

Previously we used 'all_organisations' to refer to an access limited
edition where users in the primary and any supporting orgs had access,
which was a little confusing due to the use of 'all'. This renames it to
'tagged_organisations', based on the copy we show to the user.


benthorner:
It could, but I had the same thought that the problem will be limited to a couple of people only, so I don't think it's worth doing a data migration for this - probably should have mentioned that!

PR #1270, opened 2019-07-18
SQL script to remove access limited content

kevindew:
Trello: https://trello.com/c/X8hmEQKk/990-data-syncing-access-limited-content-to-integration

This introduces a SQL script, tests for that script and database migrations that allow for content to be removed as part of the [production to integration data sync](https://docs.publishing.service.gov.uk/manual/alerts/data-sync.html). The SQL file should be copied to https://github.com/alphagov/env-sync-and-backup/tree/master/sanitising-sql, it is included here so that it can be tested as part of any database schema changes.

The approach used here is very similar to [Email Alert API](https://github.com/alphagov/email-alert-api/blob/2d8c09117cd8bea1c696940283ccc64ec2be6e14/lib/data_hygiene/anonymise_email_addresses.sql) with the following differences:

- SQL file is put in `./db` for Content Publisher rather than `./lib/data_hygiene`, lib didn't seem an intuitive place for an SQL file and I'm not sure how this relates to data hygiene;
- SQL comments are used instead of Ruby ones, not really sure why that decision was made in Email Alert API.

Testing on my local machines I can see it successfully removes records:

Before

```
content_publisher_development# SELECT schemaname,relname,n_live_tup
  FROM pg_stat_user_tables
  ORDER BY n_live_tup DESC;
┌────────────┬─────────────────────────────────────┬────────────┐
│ schemaname │               relname               │ n_live_tup │
├────────────┼─────────────────────────────────────┼────────────┤
│ public     │ revisions_statuses                  │       3122 │
│ public     │ timeline_entries                    │       2796 │
│ public     │ revisions                           │       2400 │
│ public     │ editions_revisions                  │       2400 │
│ public     │ revisions_image_revisions           │       1836 │
│ public     │ content_revisions                   │       1428 │
│ public     │ statuses                            │       1166 │
```

After

```
content_publisher_development# SELECT schemaname,relname,n_live_tup
  FROM pg_stat_user_tables
  ORDER BY n_live_tup DESC;
┌────────────┬─────────────────────────────────────┬────────────┐
│ schemaname │               relname               │ n_live_tup │
├────────────┼─────────────────────────────────────┼────────────┤
│ public     │ revisions_statuses                  │       3117 │
│ public     │ timeline_entries                    │       2788 │
│ public     │ revisions                           │       2395 │
│ public     │ editions_revisions                  │       2395 │
│ public     │ revisions_image_revisions           │       1833 │
│ public     │ content_revisions                   │       1426 │
│ public     │ statuses                            │       1163 │
```

This SQL runs very fast as our database size is small, I'm not really sure how well the queries that check for orphans will scale though. 

I'll plan to follow this up with an ADR that updates our deletion strategy to reflect this use case.

### DIFF
--- a/db/scrub_access_limited.sql
+++ b/db/scrub_access_limited.sql
@@ -1,0 +1,24 @@
+-- This script is to be executed by https://github.com/alphagov/env-sync-and-backup
+-- (private repo) as part of the data sync from production to integration.
+-- It is used to remove any access limited content from Content Publisher at
+-- the point of copying to integration.
+-- It is kept in this repo to be tested with the application, whenever changes
+-- are made it should be re-copied to the env-sync-and-backup repository.
+
+-- Remove current status of any access limited editions
+UPDATE editions
+SET current = false
+WHERE access_limit_id IS NOT NULL;
+
+-- Set the live editions to current for any docs with an access limited draft
+UPDATE editions
+SET current = true
+WHERE live = true
+AND document_id IN (
+  SELECT document_id
+  FROM editions
+  WHERE access_limit_id IS NOT NULL
+);
+
+-- Delete any access limited editions
+DELETE FROM editions WHERE access_limit_id IS NOT NULL;
--- a/db/scrub_access_limited.sql
+++ b/db/scrub_access_limited.sql
@@ -22,3 +22,74 @@ AND document_id IN (
 
 -- Delete any access limited editions
 DELETE FROM editions WHERE access_limit_id IS NOT NULL;
+
+-- Delete orphaned revisions
+DELETE FROM revisions WHERE NOT EXISTS (
+  SELECT 1
+  FROM editions_revisions
+  WHERE editions_revisions.revision_id = revisions.id
+);
+
+DELETE FROM content_revisions WHERE NOT EXISTS (
+  SELECT 1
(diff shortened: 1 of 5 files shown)


db/scrub_access_limited.sql
```diff
@@ -0,0 +1,24 @@
...
+-- It is kept in this repo to be tested with the application, whenever changes
+-- are made it should be re-copied to the env-sync-and-backup repository.
+
+-- Remove current status of any access limited editions
+UPDATE editions
+SET current = false
+WHERE access_limit_id IS NOT NULL;
+
+-- Set the live editions to current for any docs with an access limited draft
+UPDATE editions
```
brucebolt: In the case that a document only has one edition, which is access limited, I assume we'll end up with a document that has no editions.

How does content publisher deal with this?  Will it cause exceptions to be raised in integration (and when the data is re-used locally) if someone attempts to view this document?

kevindew: That scenario is ok in Content Publisher. It's quite similar to when someone discards the first draft of a document and we end up with a document without a current or live edition. Our queries are based on an edition that joins on document rather than the other way round.

Thanks for reviewing @brucebolt 👍

PR #1274, opened 2019-07-19
Return suggestions when querying a mix of result and hint

alex-ju:
Update the `syncResults` function of the `AutoCompleteWithHintOnOptions` to return suggestions when a mix of text and hint terms is used in the query string.

[Trello card](https://trello.com/c/vPEvzSY9)

### DIFF
--- a/app/assets/javascripts/components/autocomplete.js
+++ b/app/assets/javascripts/components/autocomplete.js
@@ -52,9 +52,23 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
         })
         syncResults(query
           ? results.filter(function (result) {
-            var valueContains = result.text.toLowerCase().indexOf(query.toLowerCase()) !== -1
-            var hintContains = result.hint.toLowerCase().indexOf(query.toLowerCase()) !== -1
-            return valueContains || hintContains
+            var queryLowerCase = query.toLowerCase()
+            var resultTextLowerCase = result.text.toLowerCase()
+            var resultHintLowerCase = result.hint.toLowerCase()
+            var resultLowerCase = resultTextLowerCase + ' ' + resultHintLowerCase
+            var resultWords = resultLowerCase.split(' ')
+            var resultIncludesWord = false
+
+            var valueContains = resultTextLowerCase.indexOf(queryLowerCase) !== -1
+            var hintContains = resultHintLowerCase.indexOf(queryLowerCase) !== -1
+
+            resultWords.forEach(function (resultWord) {
+              if (resultWord && queryLowerCase.includes(resultWord)) {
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/components/autocomplete.js
```diff
@@ -52,9 +52,23 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
...
+            var valueContains = resultTextLowerCase.indexOf(queryLowerCase) !== -1
+            var hintContains = resultHintLowerCase.indexOf(queryLowerCase) !== -1
+
+            resultWords.forEach(function (resultWord) {
+              if (resultWord && queryLowerCase.includes(resultWord)) {
+                resultIncludesWord = true
+              }
+            })
+
+            return valueContains || hintContains || resultIncludesWord
```
kevindew: I wonder if it's doing something more complicated than it needs to.

If you match on resultIncludesWord doesn't that mean you'll definitely match on valueContains or hintContains?

This switch does seem to be getting rather close to having results appear in a somewhat random order because of lack of relevance - for example if one of your words was "of" you'd be matching tons of organisations and it'd be very hard to get something for more specific. 

I'd wonder if this would be better to be along the lines of `(resultTextLowerCase + ' ' + resultHintLowerCase).indexOf(queryLowerCase.trim()) || (resultHintLowerCase + ' ' + resultTextLowerCase`).indexOf(queryLowerCase.trim())`?

I feel like if we want to actually start replicating a search then we want to have means to order by relevancy and use a library that does things like stemming and stop words.

alex-ju: > I feel like if we want to actually start replicating a search then we want to have means to order by relevancy and use a library that does things like stemming and stop words.

I agree with this. The case this card wants to solve (as I understand it) is when having `John Doe, Department for Transport` and the user types `Doe Transport` we will return the result.

Happy to go with the suggested solution if I got the requirements wrong.

kevindew: Sure, I think you understood the requirements correctly but I don't think they're considering the slippery slope we'll be heading down. 

I don't think we can meet the requirements of the example you listed there without either a) using a relevancy based search or b) creating weird cases.

As I imagine the example you've listed there probably works great with code here whereas it would be rather surprising if you did "Doe Department for Transport"

The other aspect to consider as well if we're considering using a relevancy based search is what effect on UX it'll be for users who use the same widget in different scenarios and it works differently.

Maybe it's worth having a chat with Sonia about this from a UX angle?

alex-ju: Had a chat with designer and concluded to move forward with the suggested approach. This is now good for a second review.

PR #1275, opened 2019-07-19
Make our format definitions more flexible

benthorner:
https://trello.com/c/nGfCX2c6/996-identify-work-needed-for-multiple-formats

Things still to try:

   - Flexible requirements processing in different scenarios (pre-update, pre-preview, pre-publish)
   - Rendering multiple values for a field on the doc summary?


app/views/documents/show/_contents.html.erb
```diff
@@ -1,7 +1,7 @@
 <% contents = @edition.document_type.contents.map { |field|
   {
     field: field.label,
-    value: render("documents/fields/#{field.type}", field: field, edition: @edition)
+    value: render("documents/fields/#{field.id}", field: field, edition: @edition)
   }
 } %>
```
kevindew: Rather than using a loop here I wonder if we could have the document_type return the hash, this would then allow multiple fields for a single thing (such as title and base_path as we recently discussed)

benthorner:
Closing since we have a card to work through this now.

PR #1284, opened 2019-07-24
Show a warning on the tags page for access limited editions

kevindew:
Trello: https://trello.com/c/RUUozLeb/1006-warn-users-before-changing-organisations-on-an-access-limited-edition

This adds a warning for when an edition is access limited that users may experience undesirable consequences if they change an organisation.

It looks like this:

Screen Shot 2019-07-24 at 16 24 40

Few areas where I had to make some decisions:

- I assumed it was to be placed above the field rather than the intro text, this would lead well into having this warning be part of the view for primary_publishing_organisation field when we move forward with formats;
- the copy doesn't cater for the different types of access limit, but I don't think that's a big problem;
- the copy doc specified inset or warning component, I picked warning since we were doing a warning;
- feature testing this is a bit tricky as it's the side effect of a different feature, I considered adding this as steps onto the "Set access limit" feature but that's already rather long so thought it was better standalone.

### DIFF
--- a/app/views/tags/edit.html.erb
+++ b/app/views/tags/edit.html.erb
@@ -12,6 +12,12 @@
                  },
                  class: "app-c-contextual-guidance__form" do %>
       <% @edition.document_type.tags.each do |tag_field| %>
+        <% if @edition.access_limit.present? && tag_field.id == "primary_publishing_organisation" %>
+          <%= render "govuk_publishing_components/components/warning_text", {
+            text: t("tags.edit.organisation_warning"),
+          } %>
+        <% end %>
+
         <div id="<%= tag_field.id %>">
           <%= render "tags/tags/#{tag_field.type}_input",
             tag_field: tag_field,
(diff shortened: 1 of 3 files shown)


app/views/tags/edit.html.erb
```diff
@@ -12,6 +12,12 @@
                  },
                  class: "app-c-contextual-guidance__form" do %>
       <% @edition.document_type.tags.each do |tag_field| %>
+        <% if @edition.access_limit.present? && tag_field.id == "primary_publishing_organisation" %>
```
benthorner: It's a bit strange that this only activates for the primary org field, knowing that the supporting org field also affects this behaviour. On the other hand, I get that we don't necessarily want to show the same warning twice. This makes me wonder if we should actually have it as part of the intro text.

kevindew:
> This looks good to me. I'm just wondering if we've got the most intuitive approach for the placement/conditional. What would be the downside of it going below the intro text?

I'm assuming you mean code placement rather than visual placement.

I put it there since it was closer to primary_publishing_organisation field and therefore would lend itself better to being with that if that was to be it's own view.  It also seemed a bit weird to iterate through the `document_type.tags` to see if it contained a particular one only a line or two above another iteration through that collection. It's not something I'm particularly precious about though.

kevindew:
> I was thinking of a simple `keys.include?`. My concern, albeit vague, is that the organisation tag might become visually separated from the primary org one in future. It does seem unlikely, so I guess it's not blocking.

We don't actually have a keys I believe, I think we have to map the ids of each tag.

Yes - to be honest I'm hoping we come to a clearer resolution to this with the formatting work that can be applied more consistently for any future concerns.

PR #1286, opened 2019-07-24
Stop user locking themselves out

benthorner:

This adds a requirements checker to ensure a user cannot set an access
limit that would lock them out of an edition.

Screen Shot 2019-07-24 at 17 40 44

### DIFF
--- a/spec/services/requirements/access_limit_checker_spec.rb
+++ b/spec/services/requirements/access_limit_checker_spec.rb
@@ -1,0 +1,90 @@
+# frozen_string_literal: true
+
+RSpec.describe Requirements::AccessLimitChecker do
+  describe "#pre_update_issues" do
+    let(:user) { build :user, organisation_content_id: "my-org" }
+
+    it "returns no issues when there is no access limit" do
+      edition = build(:edition)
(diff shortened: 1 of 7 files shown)


spec/services/requirements/access_limit_checker_spec.rb
```diff
@@ -0,0 +1,90 @@
...
+                        tags: {
+                          primary_publishing_organisation: %w[another-org],
+                          organisations: %w[my-org],
+                        })
+
+        issues = Requirements::AccessLimitChecker.new(edition, user).pre_update_issues
+        expect(issues.items).to be_empty
+      end
+
+      it "returns no issues when the user is in the primary org" do
```
emmabeynon: Wondering if this test is necessary given that we're testing for the user being part of the primary organisation already on line 35?

benthorner: This is checking a different bit of functionality, as the context is different: we're using a limit_type of `tagged_organisations` instead of `primary_organisation`.

emmabeynon: Ah I missed that the type of access limit was checked before checking that the user wasn't part of any organisation.

PR #1289, opened 2019-07-25
Add array-like methods to CheckerIssues

benthorner:
This makes the CheckerIssues behave more like an array, so that we can
always work with this class instead of going backwards and forwards
between arrays in order to emulate '<<', '+=' and 'empty?'.


benthorner:
Thanks :-). Subclassing Array turns out to lead to surprises for anything that mutates the instances, like `#+`, because the inherited methods create a new Array and not a ContentIssues :-(.

PR #1291, opened 2019-07-25
ADR on permitted deletion

kevindew:
Trello: https://trello.com/c/X8hmEQKk/990-data-syncing-access-limited-content-to-integration

[Rendered version](https://github.com/alphagov/content-publisher/blob/deletion-adr/docs/adr/00xx-permitted-deletion.md)

Given that we now allow deletion it seemed prudent to provide a follow
up to the ADR on restricted deletion.

### DIFF
--- a/docs/adr/0013-permitted-deletion.md
+++ b/docs/adr/0013-permitted-deletion.md
@@ -1,4 +1,4 @@
-# XX. Permitted Deletion
+# 13. Permitted Deletion
 
 Date: 2019-07-24
 
--- a/docs/adr/00xx-permitted-deletion.md
+++ b/docs/adr/00xx-permitted-deletion.md
@@ -1,0 +1,55 @@
+# XX. Permitted Deletion
+
+Date: 2019-07-24
+
+## Context
+
+[ADR 8: Restricted deletion][ADR-8] explained the decision that
+[foreign key][foreign-keys] associations in the Content Publisher database
+will all use the `restrict` constraint. At that point in time it was
+identified that we do not have any use cases for deleting data from the Content
+Publisher database.
+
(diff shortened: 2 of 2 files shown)


kevindew:
Thanks for the review @benthorner - this is my second piece of writing in a row to get through without a spelling or grammar. Who knew that maybe there is some value in proof reading?

> This is great - I can't find any spelling or grammatical errors 😱. Just wondering if it's worth clarifying a couple of points:
> 
> * The discussion about editions vs. revision vs. revisions (vs. other stuff) is a bit tricky to understand without any contextual code. It might be helpful to point to the SQL PR that's associated with this decision, or to enumerate the different tables/models we're referring to explicitly.

That's a good idea. I found it difficult to decide what the level of assumed knowledge of a reader is, which essentially needs to be quite high to understand any of this. Linking to commits seems a reasonable way of providing further context without having to provide a lot of qualifying information.

> * There seems to be a grey area around Assets. It looks like we're not deleting these, perhaps because they might be shared with a non-access-limited edition. Would it be true to say that "we delete everything associated with an access limited edition that isn't shared with another edition"?

These are deleted automatically. When a blob revision deletes it cascades the removal of an asset.

You are correct in your statement there, though I feel like we've already captured that with: 
"The approach taken to achieve this is to delete all access limited edition records and then to delete any revisions that have no association with an edition."

benthorner:
@kevindew thanks :-). With the last suggestion, I know it would help me to have the 'more vague' summary of what we're deleting - it just makes the detailed version a bit easier to understand.

kevindew:
@benthorner sure, I've tweaked this to have your preferred phrasing in.

PR #1295, opened 2019-07-26
Insert contacts using a modal workflow

alex-ju:
This PR brings the existing contact page into a modal experience and preserves the old page as a fallback for when JavaScript is disabled (we will iterate this in a separate PR).

[Trello card](https://trello.com/c/8NN6jQiI)


spec/features/editing_content/insert_contact_no_js_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+
+RSpec.feature "Insert contact without Javascript" do
+  include AccessibleAutocompleteHelper
+
+  scenario do
+    given_there_is_an_edition
+    when_i_go_to_edit_the_edition
+    and_i_go_to_add_a_contact
+    when_i_select_a_contact
+    then_i_can_see_the_contact_in_the_body
```
benthorner: I think we need to change the behaviour here, as we'll be trying to insert the snippet into a potentially out-of-date version of the body field - we're no longer saving the page and redirecting.

alex-ju: Thank for the review Ben! I updated as discussed. Do let me know if something feels shady, especially on the Rails side

config/routes.rb
```diff
@@ -38,7 +38,7 @@
     get "/debug" => "debug#index", as: :debug_document
 
     get "/search-contacts" => "contacts#search", as: :search_contacts
-    post "/search-contacts" => "contacts#insert", as: :insert_contact
+    post "/search-contacts" => "contacts#search", as: :insert_contact
```
benthorner: Can we rename the 'as' part here? I think we can actually leave it out for both of them and use 'search_contacts_path'.

spec/features/editing_content/insert_contact_no_js_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+
+  scenario do
+    given_there_is_an_edition
+    when_i_go_to_edit_the_edition
+    and_i_go_to_add_a_contact
+    when_i_select_a_contact
+    then_i_see_the_contact_markdown_snippet
+  end
+
+  let(:organisation) { { "content_id" => SecureRandom.uuid, "internal_name" => "Organisation" } }
```
benthorner: Worth mentioning that we don't use lets or befores in our feature specs. Things like this usually get shoved in one of the early steps, before they are needed. If we need to refer to them later, we use instance '@' variables.

alex-ju: Hmm.. again these are the sort of things I carried this from our previous feature spec. Happy to update them, but I will need help.

config/routes.rb
alex-ju: Updated and removed both

spec/features/editing_content/insert_contact_no_js_spec.rb
alex-ju: Refactored as discussed. Do let me know if they need any updates.

app/assets/javascripts/components/autocomplete.js
```diff
@@ -85,6 +83,10 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
         if (options.length) {
           options[0].selected = true
         }
+
+        if ($insertButton && value) {
```
benthorner: This looks very specific to the behaviour for contacts - is there something we could do to call this out at a higher level? I'm also wondering if we should be using the translation file to render the snippet here? Sorry for not mentioning this the first time - I think I've started avoiding subconciously the autocomplete ;-).

config/routes.rb
```diff
@@ -37,8 +37,8 @@
 
     get "/debug" => "debug#index", as: :debug_document
 
-    get "/search-contacts" => "contacts#search", as: :search_contacts
-    post "/search-contacts" => "contacts#insert", as: :insert_contact
+    get "/search-contacts" => "contacts#search"
```
benthorner: Probably something for another PR - it would be good to rename this to just `contacts#index`, as we're not really searching contacts here. Happy to defer, as discussed.

app/controllers/contacts_controller.rb
```diff
@@ -6,14 +6,9 @@ def search
     assert_edition_state(@edition, &:editable?)
     assert_edition_access(@edition, current_user)
     @contacts_by_organisation = ContactsService.new.all_by_organisation
+    render layout: rendering_context
   rescue GdsApi::BaseError => e
     GovukError.notify(e)
     render "search_api_down", status: :service_unavailable
```
benthorner: Sorry for not spotting this the first time round: this doesn't actually get rendered in the modal as it's not a 2xx response code. Noting that the response code is somewhat inaccurate (our app server is available), I wonder if we should actually use the default 200 status and make it work with the modal - it's the only use of the modal which relies on an external API being up.

spec/features/editing_content/insert_contact_no_js_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+    @edition = create(:edition, document_type_id: document_type.id)
+  end
+
+  def when_i_go_to_edit_the_edition
+    visit document_path(@edition.document)
+    click_on "Edit Content"
+  end
+
+  def and_i_go_to_add_a_contact
+    @organisation = {
```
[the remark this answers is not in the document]
alex-ju: Yep, updated.

app/assets/javascripts/components/autocomplete.js
alex-ju: Renamed the init function to `initAutoCompleteContacts` to reflect its used case and using the translation file for the snippet template via a `data` attribute.

config/routes.rb
alex-ju: Agree, this doesn't reflect the action (which is meant to be insert contact). I'd rather do it as a separate PR if that's fine with you.

alex-ju: On a second thought it seems quite couples with these changes, so added two PRs to deal with the route rename and the locales.

app/controllers/contacts_controller.rb
alex-ju: Updated to use the API down message for both the modal and no-JS experience.

app/controllers/contacts_controller.rb
```diff
@@ -1,19 +1,14 @@
...
   rescue GdsApi::BaseError => e
     GovukError.notify(e)
-    render "search_api_down", status: :service_unavailable
-  end
-
-  def insert
-    Contacts::InsertInteractor.call(params: params, user: current_user)
-
-    redirect_to edit_document_path(params[:document], anchor: "body")
+    render "index_api_down", layout: rendering_context
```
benthorner: I think we just need to modify the existing API down feature to cover this, and we're there 🎆 .

alex-ju: Yup, I've split our `*_api_down` test into `*_spec` and `*_no_js_spec` testing both scenarios.

spec/features/editing_content/insert_contact_publishing_api_down_no_js_spec.rb
```diff
@@ -0,0 +1,35 @@
+# frozen_string_literal: true
+
+RSpec.feature "Insert contact when the Publishing API down" do
+  scenario do
+    given_there_is_an_edition
+    when_i_go_to_edit_the_edition
+    and_i_go_to_add_a_contact_with_the_publishing_api_down
```
benthorner: The API bit can just be an extra `and` step.

spec/features/editing_content/insert_contact_publishing_api_down_no_js_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+  end
+
+  def when_i_go_to_edit_the_edition
+    visit document_path(@edition.document)
+    click_on "Edit Content"
+  end
+
+  def and_i_go_to_add_a_contact_with_the_publishing_api_down
+    publishing_api_isnt_available
+    # Add contact will submit the form and update the content before redirecting
```
benthorner: I think we'll be OK without the comment ;-).

spec/features/editing_content/insert_contact_publishing_api_down_no_js_spec.rb
```diff
@@ -0,0 +1,35 @@
+# frozen_string_literal: true
+
+RSpec.feature "Insert contact when the Publishing API down" do
```
benthorner: Minor, but the "Publishing API *is* down" ;-).

spec/features/editing_content/insert_contact_spec.rb
```diff
@@ -36,21 +22,32 @@ def when_i_go_to_edit_the_edition
     click_on "Edit Content"
   end
 
-  def and_i_go_to_add_a_contact
-    @request = stub_publishing_api_put_content(@edition.content_id, {})
+  def and_i_click_to_insert_a_contact
+    @organisation = {
```
benthorner: Minor, but I think we could loose the '@' here.

alex-ju:
@benthorner updated addressing the comment, which are all valid. Let me know if the updates regarding `The API bit can just be an extra and step.` make sense to you.

PR #1297, opened 2019-07-26
Automatically enforce access limiting feature

benthorner:
https://trello.com/c/P3z1xFBW/347-investigate-methods-to-prevent-access-to-document-based-on-access-limit-and-state

    Previously we had to remember to add access limit assertions for each
    new user action where they apply - we've already missed a few. This
    replaces the individual assertions with a global before_action:
    
       - We use 'request.path_parameters' to work out if the URL is trying
    to access an edition under the 'documents/:document' scope
       - We replace the assertion/exception mechanism with a boolean check,
    so that all the access limit code is in one place
    
    While changing the mechanism may be pragmatic to overcome the issues
    with repetition and forgetfulness, we recognise that it also has a few
    downsides, which may cause pain in the future:
    
       - We now retrieve an edition twice in order to process a request
       - Our approach to enforcing access limits is now inconsistent with
    other kinds of assertion
       - We now implicitly rely on the 'document' scope in the routes
    
    This also moves the 'AccessLimit#organisation_ids' method onto the
    Edition class to better reflect the variety of data it uses, and to
    avoid doing even more extra queries when an edition is access limited -
    our current relationships prevent using inverse_of memoisation.


kevindew:
It'd be good to have a chat about this change. 

It does seem to present a u-turn from the strategy that was the outcome of https://trello.com/c/P3z1xFBW/347-investigate-methods-to-prevent-access-to-document-based-on-access-limit-and-state, which seems a bit disappointing given that was done as work to support this feature.

I can see how this approach is appealing, as the approach designed is repetitive which leaves us prone to forget. However that then leads me to question whether that is going to lead to similar problems for dealing with edition state and when we move towards checking document type features. So it might be an indication that we have designed too explicit an interface to create something that scales.

With this approach we introduce some undesirable elements:

- we do an extra 3 queries per request;
- we initialise a bunch of data, discard it and then load it up again in the action;
- we create a landmine in the code for any "document" param, it effectively becomes a reserved value for form/query params for any unrelated views without surprising consequence;
- as this occurs outside the lock there is the risk of race conditions where users can alter access limited data.

Finally, given this actual implementation, I would question the value in using the assertion system for access limiting since the access limit check is only done in a single place.

benthorner:
@kevindew I've made a few updates following our conversation earlier. I've changed the commits to document the pros and cons - what do you think?

app/controllers/application_controller.rb
```diff
@@ -23,12 +28,23 @@ class ApplicationController < ActionController::Base
...
-
   def rendering_context
     request.headers["Content-Publisher-Rendering-Context"] || "application"
   end
+
+  def check_user_access
+    document_id = request.path_parameters[:document]
+    return if document_id.blank?
+
+    access_limit = Edition.includes(:access_limit)
```
kevindew: It'd be good to have this join also include `revision: :tags_revision` to avoid the extra query for that.

app/controllers/application_controller.rb
```diff
@@ -23,12 +28,23 @@ class ApplicationController < ActionController::Base
...
+      .current(document: document_id)
```
kevindew: I'd just use `find_current` rather than this method as the only thing we won't load is the status and I'm not sure we need the extra scope methods.

app/controllers/application_controller.rb
```diff
@@ -23,12 +28,23 @@ class ApplicationController < ActionController::Base
...
-    Rails.logger.warn(e.message)
-    render "documents/forbidden", status: :forbidden, assigns: { edition: e.edition }
-  end
-
   def rendering_context
     request.headers["Content-Publisher-Rendering-Context"] || "application"
   end
+
+  def check_user_access
+    document_id = request.path_parameters[:document]
```
kevindew: nice find 👍 minor, but I'd probably call this document_param than document_id since that'd imply an integer whereas this is what you get when you run to_param on a document.

benthorner:
@kevindew I've updated as we discussed - should be ready for another look now.

app/controllers/application_controller.rb
```diff
@@ -23,12 +28,23 @@ class ApplicationController < ActionController::Base
...
+    return if document_id.blank?
+
+    edition = Edition
+      .includes(:access_limit, revision: [:tags_revision])
+      .find_current(document: document_id)
+
+    return unless edition
+      &.access_limit_organisation_ids
```
kevindew: minor - but edition won't be nil.

benthorner: 👍 - fixed.

app/controllers/application_controller.rb
benthorner: Also forgot to do this earlier - fixed.

PR #1299, opened 2019-07-29
Add check for user's organisation to set access limit

theseanything:
This adds a requirement that a user has an organisation associated with their account to set access limits. This is to prevent a user locking themselves out of content.

### DIFF
--- a/spec/services/requirements/access_limit_checker_spec.rb
+++ b/spec/services/requirements/access_limit_checker_spec.rb
@@ -34,5 +34,16 @@ RSpec.describe Requirements::AccessLimitChecker do
         expect(issues).to be_empty
       end
     end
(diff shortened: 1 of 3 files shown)


spec/services/requirements/access_limit_checker_spec.rb
```diff
@@ -34,5 +34,16 @@
         expect(issues).to be_empty
       end
     end
+
+    context "when user has no organisation associated with account" do
+      let(:edition) { build(:edition, :access_limited, created_by: user) }
```
benthorner: This surprised me a little, as the context and the let are somewhat unrelated. I think there's a way we could tie them together better, but it's just a minor thing.

PR #1302, opened 2019-07-30
Move edition methods into other classes

benthorner:

This moves a few of the methods in the Edition class to the (mostly) single points where they are used, making space to add another method in https://github.com/alphagov/content-publisher/pull/1297. Please see the commits for more details.

### DIFF
--- a/app/interactors/access_limit/update_interactor.rb
+++ b/app/interactors/access_limit/update_interactor.rb
@@ -28,14 +28,20 @@ private
   def update_access_limit
     limit_type = params.require(:limit_type)
 
-    if LIMIT_TYPES.include?(limit_type)
-      context.fail! if edition.access_limit&.limit_type == limit_type
-      edition.assign_access_limit(limit_type, user)
+    if LIMIT_TYPES.exclude?(limit_type)
+      context.fail! if edition.access_limit.nil?
+      edition.assign_as_edit(user, access_limit: nil)
       return
     end
 
-    context.fail! if edition.access_limit.nil?
-    edition.remove_access_limit(user)
+    context.fail! if edition.access_limit&.limit_type == limit_type
+
+    access_limit = AccessLimit.new(created_by: user,
+                                   edition: edition,
+                                   limit_type: limit_type,
+                                   revision_at_creation: edition.revision)
+
+    edition.assign_as_edit(user, access_limit: access_limit)
   end
 
   def check_for_issues
--- a/app/interactors/editions/create_interactor.rb
+++ b/app/interactors/editions/create_interactor.rb
@@ -5,16 +5,16 @@ class Editions::CreateInteractor < ApplicationInteractor
            :user,
            :live_edition,
            :next_edition,
-           :draft_current_edition,
            :discarded_edition,
+           :next_revision,
            to: :context
 
   def call
     Edition.transaction do
       find_and_lock_live_edition
+      create_next_revision
       create_next_edition
       create_timeline_entry
-      update_preview
     end
   end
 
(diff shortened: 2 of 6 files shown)


app/interactors/editions/create_interactor.rb
```diff
@@ -25,10 +25,23 @@ def find_and_lock_live_edition
     context.live_edition = edition
 
     assert_edition_state(edition, assertion: "can create new edition") do
-      edition.published? || edition.published_but_needs_2i? || edition.removed?
+      edition.published? ||
+        edition.published_but_needs_2i? ||
+        edition.removed? ||
+        edition.discarded?
```
kevindew: Not necessary for this PR but I wonder if we want to encapsulate this as a method on the edition since it's reached 4 states. Naming could be a pain though - best that comes immediately to mind is `supersedable?`

app/models/edition.rb
```diff
@@ -161,23 +129,25 @@ def assign_status(state,
...
-      attributes[:last_edited_by] = user
+      assign_as_edit(user, attributes)
+    else
+      assign_attributes(attributes)
+      self
     end
+  end
 
-    assign_attributes(attributes)
+  def assign_as_edit(user, attributes)
```
kevindew: I didn't find this the easiest method name to understand, but I don't have a better suggestion. Seems good to have it encapsulated in a method though 👍

spec/models/edition_spec.rb
```diff
@@ -53,59 +53,21 @@
...
-                       document: live_edition.document,
-                       change_note: "Changes",
-                       update_type: :minor)
-
-      edition.resume_discarded(live_edition, user)
-
-      expect(edition.change_note).to be_empty
-      expect(edition.update_type).to eq("major")
+    it "updates the edition metadata" do
+      freeze_time do
```
kevindew: TIL - this looks nicer than our travel.to(Time.current)

app/interactors/access_limit/update_interactor.rb
```diff
@@ -28,14 +28,18 @@ def find_and_lock_edition
...
+      context.fail! if edition.access_limit.nil?
+      edition.assign_as_edit(user, access_limit: nil)
       return
     end
 
-    context.fail! if edition.access_limit.nil?
-    edition.remove_access_limit(user)
+    access_limit = AccessLimit.build_initial(created_by: user,
+                                             edition: edition,
+                                             limit_type: limit_type)
```
kevindew: If this is the only place we're calling this method and we're only saving one keyword argument from creating the model directly we may as well just call new rather than build_initial - particularly as we don't have a precedent for a `build_initial` pattern

app/interactors/access_limit/update_interactor.rb
```diff
@@ -28,14 +28,18 @@ def find_and_lock_edition
...
+
+    context.fail! if edition.access_limit&.limit_type == limit_type
```
kevindew: Feels like this guard clause would be more intuitive before the object creation.

app/interactors/access_limit/update_interactor.rb
benthorner: Indeed we could. The problem I had was trying to make sure it's covered and wanting to avoid writing a spec for the entire interactor to do so. It's a tenuous one because the attribute is purely archival. Would a spec for the interactor seem surprising?

app/interactors/editions/create_interactor.rb
benthorner: As discussed, we decided we're happy for this to include withdrawn editions, so the check can be reduced to "live or discarded".

app/interactors/access_limit/update_interactor.rb
kevindew: I'd have thought a spec for the interactor would be fine - we already have a few and they seem a good practice. I'm also not massively concerned about this particular thing being covered since it's enforced at a db level so only thing we can really test is that the expected revision made it as an association.

benthorner: Oh yeah, it is enforced - I mixed that up with the 'created_by_id'. In that case, I believe we're reasonably covered and don't see the motivation for additional testing of the interactor. Is that something you feel is important here? Anyway, I've removed the `build_initial` method 👍 .

app/interactors/access_limit/update_interactor.rb
benthorner: Sure - moved it 👍 .

app/models/edition.rb
benthorner: Agreed - it's simply the first thing that came to mind.

benthorner:
@kevindew this should be ready for another look :-).

PR #1303, opened 2019-07-31
Re-work WhitehallNewsImporter for new Whitehall Export API

brucebolt:
A new document export API has been created in Whitehall Publisher (https://github.com/alphagov/whitehall/commit/e1ed1f4b474db34cbe4e56eadbfb44d4d2217633).  These changes update the existing WhitehallNewsImporter to work with this export, whilst giving the potential to import any document type.

Additionally stores the raw output from Whitehall into a database table, sets a status and logs any errors that occur during the import.

The task is invoked using `rake import:whitehall[document_id]` where `document_id` is Whitehall's internal document identifier.

It should be noted that this importer only beings in the skeleton of a document and it's editions.  Further follow-up work will be required to fully import the document.  This is discussed in https://docs.google.com/document/d/1lnCQZ_qbIzgwL94gL4hqIGyy8TS8suiFe3_5XoBaHdM/edit.

Trello card: https://trello.com/c/CnYgD1U3/1041-spike-importing-data-into-content-publisher


db/migrate/20190729145805_add_whitehall_import.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+class AddWhitehallImport < ActiveRecord::Migration[5.2]
+  def change
+    create_table :whitehall_imports do |t|
+      t.bigint :whitehall_document_id
+      t.json :payload
+      t.uuid :content_id
```
kevindew: I imagine all the above fields should be set to null: false

db/migrate/20190729145805_add_whitehall_import.rb
```diff
@@ -0,0 +1,14 @@
...
+      t.string :import_status
```
kevindew: I'd suggest naming this `state` to be consistent with existing fields we have such as: https://github.com/alphagov/content-publisher/blob/master/db/schema.rb#L273 and https://github.com/alphagov/content-publisher/blob/master/db/schema.rb#L105

It'd also be good for this to have enums defined for what the values could be in the model: https://github.com/alphagov/content-publisher/blob/master/app/models/file_attachment/asset.rb#L16-L19

lib/tasks/whitehall_importer.rb
```diff
@@ -0,0 +1,116 @@
...
+
+module Tasks
+  class WhitehallImporter
+    SUPPORTED_WHITEHALL_STATES = %w(draft published rejected submitted superseded).freeze
+    SUPPORTED_LOCALES = %w(en).freeze
+
+    def import(whitehall_document)
+      ActiveRecord::Base.transaction do
+        @most_recent_edition = most_recent_edition(whitehall_document)
+        @document = create_or_update_document(whitehall_document)
```
kevindew: It's rather unusual for an method to be setting instance variables that apply just for the duration of that methods running, I'd suggest this indicates that these are items that the class should be initialized with and then set to `attr_reader`

lib/tasks/import.rake
```diff
@@ -1,24 +1,22 @@
...
+    whitehall_export = JSON.parse(URI.parse("#{host}/government/admin/export/document/#{document_id}").open.read)
+    importer = Tasks::WhitehallImporter.new
+    whitehall_import = importer.store_json_blob(document_id, whitehall_export)
+    begin
+      importer.import(whitehall_export)
+    rescue StandardError => e
+      importer.log_error(whitehall_import.id, e.message)
+      importer.update_import_status(whitehall_import.id, "failed")
+    else
+      importer.update_import_status(whitehall_import.id, "completed")
```
kevindew: I didn't realise you could do an `else` on a begin / rescue - TIL ⚡️ 

Though I do wonder why we don't just have this statement after the `importer.import(whitehall_export)` line?

brucebolt: I've removed the `else`  and push the status update immediately after the import method executes

lib/tasks/whitehall_importer.rb
brucebolt: Refactored this class to move the instance variables into an initialize method

brucebolt:
> Is the outcome of this that we get a document from Whitehall that has a single revision for each edition?

That's the plan.  Since we don't have all the historic content edits stored in Whitehall, it seemed a logical way of tracking what we do have.

lib/tasks/import.rake
```diff
@@ -1,24 +1,20 @@
 # frozen_string_literal: true
 
-require "tasks/whitehall_news_importer"
+require "tasks/whitehall_importer"
 
 namespace :import do
   desc "Import news documents from JSON e.g. import:whitehall_news INPUT=export_file"
```
emmabeynon: Just a tiny one, I guess we'll need to remove the references to news from this description 😸

emmabeynon: Ignore - I see this is done in a later commit!

brucebolt: Good spot.  It's in the wrong commit.  I'll move that change.

lib/tasks/whitehall_importer.rb
```diff
@@ -0,0 +1,121 @@
...
+      @whitehall_document_id = whitehall_document_id
+      @whitehall_document = whitehall_document
+      @whitehall_import = store_json_blob
+    end
+
+    def import
+      ActiveRecord::Base.transaction do
+        document = create_or_update_document
+
+        @whitehall_document["editions"].each_with_index do |edition, edition_number|
```
kevindew: It'd be more idiomatic with GOV.UK ruby to set this (and the other instance variables) with an attr_reader and not use the `@` to access. For example: https://github.com/alphagov/content-publisher/blob/9e3c55538077b375341f053aa350ca7f2e4ea486/lib/govspeak_document.rb#L4-L9

app/models/whitehall_import.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+# Represents the raw import of a document from Whitehall Publisher and
+# the import status of the document into Content Publisher
+class WhitehallImport < ApplicationRecord
+  enum state: { importing: "importing",
+                completed: "completed",
+                failed: "failed" }
```
kevindew: 👍

lib/tasks/import.rake
```diff
@@ -1,24 +1,20 @@
 # frozen_string_literal: true
 
-require "tasks/whitehall_news_importer"
+require "tasks/whitehall_importer"
```
kevindew: Minor, but since https://github.com/alphagov/content-publisher/commit/d2d82b9ea4b3de2081a5c473005bbd6f867817b0 we can drop this require (same for https://github.com/alphagov/content-publisher/pull/1303/files#diff-88c72fb190dd2a4c9d8510058c97382bR3)

lib/tasks/whitehall_importer.rb
brucebolt: Thanks for the example.  I've updated this PR now.

PR #1305, opened 2019-07-31
Allow GDS users to override access limits

benthorner:
https://trello.com/c/nRJAyOex/983-access-limiting

This adds a new permission that overrides the normal 'foridden' and
filtering behaviours for access limited editions, so that users are
still able to provide support with requests associated with these
editions, which they would otherwise be unable to find/view/edit.

While some existing apps use more role-like permissions, such as 'GDS
Admin', the permission here is specific to its functionality, as well as
which users it should be applied to. This helps avoids confusion about
the behaviour and applicability of permissions that are actually roles.

Previously it was only possible to test this behaviour at the feature
level, which would be complex due to the three variables - permission,
limit type, organisations - involved. This extracts the conditional
into a new User#can_access? method in order to fully test the code.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -7,4 +7,16 @@ class User < ApplicationRecord
   PRE_RELEASE_FEATURES_PERMISSION = "pre_release_features"
   DEBUG_PERMISSION = "debug"
   MANAGING_EDITOR_PERMISSION = "managing_editor"
+  ACCESS_LIMIT_OVERRIDE_PERMISSION = "access_limit_override"
+
+  def can_access?(edition)
+    return true unless edition.access_limit
+
+    return true if has_permission?(
+      ACCESS_LIMIT_OVERRIDE_PERMISSION,
+    )
+
+    edition.access_limit_organisation_ids
+      .include?(organisation_content_id)
+  end
 end
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -40,11 +40,9 @@ class ApplicationController < ActionController::Base
       .includes(:access_limit, revision: [:tags_revision])
       .find_current(document: document_param)
 
-    return unless edition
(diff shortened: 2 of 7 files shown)


app/models/user.rb
```diff
@@ -7,4 +7,16 @@ class User < ApplicationRecord
   PRE_RELEASE_FEATURES_PERMISSION = "pre_release_features"
   DEBUG_PERMISSION = "debug"
   MANAGING_EDITOR_PERMISSION = "managing_editor"
+  GDS_ACCESS_LIMIT_OVERRIDE_PERMISSION = "gds_access_limit_override"
```
kevindew: I'd suggest dropping the GDS prefix of this, there's nothing that forces it to be GDS and we have other permissions (debug, pre release) that don't have a prefix.

app/models/user.rb
```diff
@@ -7,4 +7,16 @@ class User < ApplicationRecord
...
+
+  def can_access?(edition)
+    return true unless edition.access_limit
+
+    return true if has_permission?(
+      User::GDS_ACCESS_LIMIT_OVERRIDE_PERMISSION,
```
kevindew: minor, but you shouldn't need the `User::` qualifier here.

app/models/user.rb
benthorner: Fair enough. My hope was to make it clear to the person applying it that it's only supposed to be used by GDS people, to help avoid it being applied accidentally. But it also seems reasonable that the people applying the permissions (who are currently entirely within GDS) should be aware of and responsible for their effects.

app/models/user.rb
benthorner: 👍

app/models/user.rb
kevindew: Yup, I also think prefixing them with GDS slips it more towards role than effects. Thanks for changing.

PR #1306, opened 2019-07-31
Try to create preview when editing access limits

benthorner:
https://trello.com/c/jvJZCHPG/989-present-access-limited-documents-to-publishing-api-and-asset-manager

This extends the existing feature to do API calls with the necessary
access limit metadata. In order to better distinguish between the the
different tests for access limits (the form vs. the behaviour), this
also renames the previous 'set_access_limit_spec' to 'enforce_'.

### DIFF
--- a/spec/services/publishing_api_payload_spec.rb
+++ b/spec/services/publishing_api_payload_spec.rb
@@ -49,6 +49,20 @@ RSpec.describe PublishingApiPayload do
       expect(payload).not_to include("first_published_at")
     end
 
+    it "specifies an auth bypass ID for anonymous previews" do
+      edition = build(:edition)
+      allow_any_instance_of(PreviewAuthBypassService).to receive(:auth_bypass_id) { "id" }
+      payload = PublishingApiPayload.new(edition).payload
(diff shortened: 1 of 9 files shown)


spec/services/publishing_api_payload_spec.rb
```diff
@@ -49,6 +49,13 @@
       expect(payload).not_to include("first_published_at")
     end
 
+    it "specifies organistions when the edition is access limited" do
+      edition = build(:edition, :access_limited)
+      allow(edition).to receive(:access_limit_organisation_ids) { %w[org-id] }
+      payload = PublishingApiPayload.new(edition).payload
+      expect(payload["access_limited"]["organisations"]).to eq %w[org-id]
+    end
+
```
theseanything: How do we represent access_limited in the payload when there is no access_limits?

benthorner: Yup. I've added one now. The lack of coverage for `auth_bypass_ids` anywhere distracted me; the new test covers this and implicitly covers the org behaviour without access limits.

Issue #1256, opened 2019-07-17
Our services directory is a mess


benthorner:
As our app grows, an increasing amount of random code is landing in the `app/services` directory. I'm worried it will become impractical to improve on this, as the number of special cases will only increase. Some pain points off the top of my head (definitely not exhaustive):

   - We have a variety of naming styles, which makes it harder to find existing 'services' and to decide on names for new ones
   - We have a variety of interface styles, which makes it harder to understand existing 'services' and decide on interfaces for new ones

Some possible aims (also not exhaustive):

   - We agree what code is a 'service' and what is not
   - We agree on the interface we expect for our services
   - We agree on the naming convention we use for services

Non-aims (but may happen anyway):

   - We categorise the non-services currently in `app/services`
   - We decide on `app/lib` v. `lib/` once and for all
   - We decide if Catrin should give Zane a second chance

## Current Situation

Currently our services directory looks like this:

   - **govspeak_document(.rb)/**
       - 4 relates classes, various methods
   - **requirements/**
      - 15 related classes, various methods
   - **versioning/**
      - 7 related classes/supporting classes, various methods
   - **asset_manager_service.rb**
      - 4 related methods e.g. `delete`
   - **contacts_service.rb**
      - 2 methods: `all_by_organisation`, `by_content_id`
   - **date(time)_parser.rb**
      - 2 methods: `parse`, `issues`
   - **delete_draft_service.rb**
      - 1 method: `delete`
   - **draft_asset_cleanup_service.rb**
      - 1 method: `call`
   - **edition_assertions.rb**
      - 2 methods: `assert_edition_state`, `assert_edition_access`
   - **edition_filter.rb**
      - 2 methods: `editions`, `filter_params`
   - **file_attachment_blob_service.rb**
      - 1 method: `create_blob_revision`
   - **image_blob_service.rb**
      - 1 method: `create_blob_revision`
   - **image_centre_cropper.rb**
      - 1 method: `dimensions`
   - **image_normaliser.rb**
      - 2 methods: `dimensions`, `normalised_file`
   - **linkables_service.rb**
      - 2 methods: `select_options`, `by_content_id`
   - **organisation_service.rb**
      - 2 methods: `by_content_id`, `alternative_format_contact_email`
   - **path_generator_service.rb**
      - 1 method: `path`
   - **preview_asset_service.rb**
      - 3 methods: `upload_assets`, `upload_asset`, `can_preview_asset?`
[...]

PR #1308, opened 2019-08-01
Move initial non-services to lib directory

kevindew:
Following on from actions agreed in https://github.com/alphagov/content-publisher/issues/1256#issuecomment-517197393 this moves classes and tests from `app/services` to the `lib` directory that don't require any renaming or refactoring. 

This also sets up eager loading for code in the `lib` directory


kevindew:
So it turns out that brakeman doesn't scan any of the custom directories in app by default, as I learnt when I moved code from there to lib. This now sets up scanning on all our app directories to avoid these surprises - the alternative consistent approach would be to disable lib so it's ignored like the rest of our app directories, but since we have quite a lot of code in `app/interactors` it feels like we should scan more directories.

As part of this I cleared out our obsolete brakeman ignores and discovered a new one on the access limit SQL. I saw an opportunity here to use some code I wrote the other day for this. This makes the EditionFilter more declarative and consistent with the other ActiveRecord methods used in this class.

kevindew:
> This looks good 👍. One minor inconsistency: I think we agreed the `publishing_api_payload` could be (refactored into) a supporting service. Can we record the future intention for this somewhere, or leave it where it is?

Sure, I've removed this now.

PR #1310, opened 2019-08-01
Add preview on insert contact

alex-ju:
This PR updates the contacts autocomplete to render a preview of the contact associated with the selected a suggestion.

It does this by reusing the fetch to preview govspeak function use in the markdown toolbar (moving it into a shared area of helper functions). There is a potential reuse for this helper when previewing other govspeak entities (.e.g tables, charts).

## Preview
Before
insert-contact-before

After
insert-contact-after

### Alternative approaches
An alternative approach is to try and make use of the `modal-fetch.js` and turn the preview into an extra step within the modal. The problem with it is that we're not able to get the context of the modal from the autocomplete in order to call the function with the current architecture.

[Trello card](https://trello.com/c/8NN6jQiI)


app/assets/javascripts/components/autocomplete.js
```diff
@@ -73,13 +74,41 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
...
-          $insertButton.dataset.modalData = contactSnippetTemplate.replace('#', value)
+          // Prepare snippet for insertion
+          if ($insertButton && contactSnippetTemplate && value) {
+            var contactSnippet = contactSnippetTemplate.replace('#', value)
+            $insertButton.dataset.modalData = contactSnippet
+          }
+
+          // Generate/update contact preview
+          var govspeakPath = document.querySelector('[data-govspeak-path]').dataset.govspeakPath
+          if (contactSnippet && govspeakPath && window.FetchContent) {
```
benthorner: Again, wondering why we're checking the helper module exists. Could that not be the case?

app/assets/javascripts/components/autocomplete.js
```diff
@@ -73,13 +74,41 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
...
+              .then(function (text) {
+                if (!$previewContainer) {
+                  $previewContainer = document.createElement('div')
+                  $previewContainer.classList.add('govuk-inset-text')
+                  $module.appendChild($previewContainer)
+                }
+                $previewContainer.innerHTML = text
+              })
+              .catch(function () {
+                $previewContainer.innerHTML = 'Error previewing contact'
```
benthorner: I wonder if we have a similar issue here to that for govspeak preview: if the fetch takes a little while, the user may be confused since the old snippet/error is still visible until it gets replaced. Would it make sense to use the same 'rendered' technique here that we use for govspeak preview?

app/assets/javascripts/components/autocomplete.js
```diff
@@ -73,13 +74,41 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
...
+              })
+          }
+        } else {
+          // Remove snippet
+          $insertButton.dataset.modalData = null
```
benthorner: Could you help me understand the scenario where this code would run? I can't think of what I could do in the UI that would trigger it.

app/assets/javascripts/application.js
```diff
@@ -14,6 +14,7 @@
 //= require abortcontroller-polyfill/dist/abortcontroller-polyfill-only.js
 //= require url-polyfill/url-polyfill.js
 //= require whatwg-fetch/dist/fetch.umd.js
+//= require helpers/fetch-content.js
```
benthorner: I notice this is the first 'helper', but we also have a 'modal' directory of things that are also 'helpful' in nature. How do you feel about rename just renaming `modal/ -> helpers/`?

app/assets/javascripts/components/autocomplete.js
alex-ju: When emptying the search input

app/assets/javascripts/helpers/fetch-content.js
```diff
@@ -0,0 +1,17 @@
+window.FetchContent = { }
+
+window.FetchContent.govspeak = function (text, path) {
```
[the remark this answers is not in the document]
alex-ju: I was tempted to do so, but the path contains the edition hash, so not very easy to move it into JavaScript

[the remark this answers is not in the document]
alex-ju: > Also, since this is no longer coupled to a component, I think it's reasonable to hard-code the path in here and not have to do the 'data attribute dance'. Still raises the question if the callers are really components anymore, as they rely so much on our app :-|.

I was tempted to hardcode the path, but the path contains the edition hash, so not very easy to move it into JavaScript.

app/assets/javascripts/application.js
alex-ju: Merged them in 0f12071

app/assets/javascripts/components/markdown-editor.js
```diff
@@ -85,37 +85,22 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
...
-      }.bind(this))
-
-    this.toggleElements()
-  }
-
-  MarkdownEditor.prototype.fetchGovspeakPreview = function (text) {
-    var path = this.$module.getAttribute('data-govspeak-path')
-    if (!path) {
-      return window.Promise.reject('Govspeak path not set')
+    if (window.FetchContent) {
```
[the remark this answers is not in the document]
alex-ju: Removed the extra conditions as it seems that we don't check for dependencies to exist in other cases.

app/assets/javascripts/helpers/fetch-content.js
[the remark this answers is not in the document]
alex-ju: Updated as discussed `GovspeakFetch.getBody`

app/assets/javascripts/helpers/govspeak-fetch.js
```diff
@@ -0,0 +1,20 @@
+window.GovspeakFetch = { }
+
+window.GovspeakFetch.getBody = function (text) {
+  // Replace last segment to point to the govspeak preview resource path
```
benthorner: Would it be possible to be a little more generic here by just looking at the first two segments of the URL? Otherwise, it's harder to see that this code will work, depending on which pages it occurs on.

alex-ju: Sure, will update

app/assets/javascripts/helpers/govspeak-fetch.js
```diff
@@ -0,0 +1,20 @@
+window.GovspeakFetch = { }
+
+window.GovspeakFetch.getBody = function (text) {
+  var path = document.location.pathname
+  var segments = path.split('/')
+  var documentSegment = segments[1]
```
benthorner: This looks better - we can be more confident about the URL structure. Still hoping we can be more explicit...how about: `document.location.pathname.match("/documents/([^/]*)/")[1]` to get the edition ID?

We should also throw some kind of textual exception so it's easier for devs to debug this function if it somehow gets called from a non-document URL.

alex-ju: Hm.. it's not really clear to me how the regex helps our case, as after we extract the edition ID, we'll have to prefix it back with the `/documents/` string anyway, then append the `govspeak-preview`, but gave it a go if you think it makes more sense.

benthorner: Agreed the difference is quite subtle, but I can see two reasons why a regex approach would be preferable, unless there are cross-browser issues with it:

   - We want to document the kind of URL we're expecting, so the supported paths are clear by reading the code. At the moment, we're just fishing out the 1st and 2nd segments, which could be anything. Another way to document this would be with a comment.
   - We want to make our code fail in a helpful way. Since we don't check the first segment is '/documents', it would be hard to debug if this code is run with an unexpected `window.location`. Another way would be to check `segments[1] == '/documents'`.

I suggested a regex approach because it should be a more concise way to make the approach robust and clear. Because we're intentionally coupling/hard-coding our JS, we should also make sure it gives a helpful error if that coupling breaks in future.

Happy to talk about this more - I was hoping to catch you on Friday and go through it together.

alex-ju: No cross-browser concerns. I get the reasoning better now, thanks for explaining it. Added a commit as you suggested and I've kicked the CI again as it was failing because of Publishing API.

benthorner:
@alex-ju just documenting what we discussed? Since https://github.com/alphagov/content-publisher/pull/1317 represents a breaking change for this PR, we think it's worth trying a different approach and creating a 'contact preview' component to mirror the 'url preview' we have already.

Pros: less code in the autocomplete, which was becoming more like 'autocomplete_with_contact_preview'; we can do the 'fetch' in the separate JS for the new component; we avoid prematurely refactoring the fetch logic

Cons: takes a bit more time - we could just reinstate the old data attribute, but on the autocomplete component

Lesson: we should aim to do backend work before frontend work, where possible :-|.

alex-ju:
We're going to try a different approach, as per [comment above](https://github.com/alphagov/content-publisher/pull/1310#issuecomment-518240891).

PR #1312, opened 2019-08-01
DRY-up requirements specs with a new matcher

benthorner:

Previously our requirements specs contained repetitious code for
checking that different kinds issue could be rendered in the context of
a form and (potentially) on the summary page, which we do by checking
the associated translations. This DRYs-up the code to check issues can
be rendered in form/summary contexts by replacing it with a couple of
RSpec matchers: 'have_form_item' and 'have_summary_item'

## Examples of failure messages

```
# incorrect issue field/key in code/test
I18n::MissingTranslationData:
  translation missing: en.requirements.backdate_date.in_the_futur.form_message

# checker does not raise an issue
Failure/Error: expect(issues).to have_form_item(:backdate_date, :in_the_future)
       expected [{:text=>"Enter a date from 1 January 1995 onwards", :href=>nil, :target=>nil}] to have item {:text=>"Enter a date in the past", :href=>nil, :target=>nil}

# missing interpolation args in code/test
I18n::MissingInterpolationArgument:
  missing interpolation argument :date in "Enter a date from %{date} onwards" ({:force_i18n_to_raise_missing_interpolation_args=>true} given)
```

### DIFF
--- a/spec/support/requirements_helper.rb
+++ b/spec/support/requirements_helper.rb
@@ -1,0 +1,26 @@
+# frozen_string_literal: true
+
+RSpec::Matchers.define :have_issue do |field, key, context = {}|
+  context[:styles] ||= %i[form]
+
+  match do |issues|
+    issue = issues.find do |actual_issue|
+      actual_issue.field == field && actual_issue.issue_key == key &&
+        actual_issue.context == context.except(:styles)
+    end
+
+    # use the issue message to do a deep check (include context)
+    # this also means we cover all the translations being defined
+    issue && context[:styles].all? do |style|
+      expected = I18n.t!("requirements.#{field}.#{key}.#{style}_message",
+                         context.except(:styles).merge(force_raise: true))
+
+      issue.message(style: style) == expected
+    end
+  end
+
+  failure_message do |actual_issues|
+    issue = Requirements::Issue.new(field, key, context.except(:styles))
+    "expected #{actual_issues.inspect} to have issue #{issue.inspect}"
+  end
+end
--- a/spec/lib/requirements/access_limit_checker_spec.rb
+++ b/spec/lib/requirements/access_limit_checker_spec.rb
@@ -13,9 +13,7 @@ RSpec.describe Requirements::AccessLimitChecker do
     it "returns an issue when the edition has no primary org" do
       edition = build(:edition, :access_limited)
       issues = Requirements::AccessLimitChecker.new(edition, user).pre_update_issues
-
-      form_message = issues.items_for(:access_limit).first[:text]
-      expect(form_message).to eq(I18n.t!("requirements.access_limit.no_primary_org.form_message"))
+      expect(issues).to have_issue(:access_limit, :no_primary_org)
     end
 
     context "when edition is access limited to some orgs" do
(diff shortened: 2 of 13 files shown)


spec/support/requirements_helper.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+%i[form summary].each do |style|
+  RSpec::Matchers.define :"have_#{style}_item" do |field, key, context = {}|
```
kevindew: I'm wondering if we're actually testing the presence of a message rather than an item. The expectation we do is against a message we construct, so there is no way to test the wider properties of the item.

spec/support/requirements_helper.rb
```diff
@@ -0,0 +1,16 @@
...
+    match do |actual_issues|
+      context = context.merge(force_i18n_to_raise_missing_interpolation_args: true)
+      expected_message = I18n.t!("requirements.#{field}.#{key}.#{style}_message", context)
+      expected_message == actual_issues.items_for(field, style: style).first[:text]
```
kevindew: Are there any risks associated with the use of first here? Could it be we do a map of messages and see if expected_message is included?

spec/support/requirements_helper.rb
theseanything: I found the naming for these matcher somewhat confusing. Maybe something more explicit like `have_requirements_form_text`? As we seem to only be checking the translation and 'form/summary' is distinguished at an "item" level.

benthorner: I see what you mean: we're only testing the 'text' key in a hash that can contain other pairs, like 'href'. However, I don't think we actually want to test any of those here, since they're characteristics that are decided by the callers; we could test the Issue class to ensure it supports them, but that's not really in scope here.

benthorner: I agree 'text' would better reflect what we're actually testing. However, I can see it being confusing when written in full e.g. `have_requirements_form_text(:access_limit, :user_not_in_orgs)`, as I would expect the argument to be a simple string.

We could directly test there is an issue by checking the 'field' and 'key' attributes, which reflects the args we're passing in the matcher. This would end up being a generic 'have_issue' matcher, but we could still superficially cover translations existing:

```
# how does this look?
expect(issues).to have_issue(:alt_text, :blank, styles: %i[form summary])
```

If we still want to validate the text, which has some benefit if it contains interpolated args, then it would help to understand better what it is that's confusing for you. Can you explain what it is about the current name that's confusing?

theseanything: Ah yup - I like your suggestion! I think it would also allow us to add future checks for an issue.

I found the original name `have_#{style}_item` confusing because the reference `<style>_item` doesn't map to any concept in the codebase.

benthorner: @kevindew how does `have_issue` sound to you?

kevindew: Sounds good 👍

benthorner:
@kevindew @theseanything this should be ready to review again now. I think the new `have_issue` matcher should address the concerns around interpretation and only the first issue being checked.

@kevindew I'm not sure what you mean about the separate commit, as I'd find it harder to understand the new matcher without seeing it in use. Perhaps a compromise would have been to introduce it and apply in a couple of places?

spec/support/requirements_helper.rb
```diff
@@ -0,0 +1,26 @@
+# frozen_string_literal: true
+
+RSpec::Matchers.define :have_issue do |field, key, context = {}|
+  context[:styles] ||= %i[form]
+
+  match do |actual_issues|
```
theseanything: can we drop the word 'actual' here? (not sure I see the value its adding)

theseanything: (v minor)

benthorner: It might help to understand the convention for RSpec matchers: https://relishapp.com/rspec/rspec-expectations/v/2-4/docs/custom-matchers/define-matcher#define-a-matcher-with-default-messages

The typical custom matcher would use 'expected' and 'actual'. I've been a bit inconsistent with it, and I think the 'actual' part is pretty obvious from the 'match do' block, so I've updated as you suggest.

Can you confirm it looks OK?

PR #1315, opened 2019-08-05
Fix access limit exception in payload

benthorner:
https://trello.com/c/jvJZCHPG/989-present-access-limited-documents-to-publishing-api-and-asset-manager

Looks like a combination of PRs introduced a small bug...

This adds a guard conditional to ensure we only set organisation IDs to
limit access to the Publishing API payload when an access limit exists.


benthorner:
@theseanything I've refactored a little to make the subtle '&&' behaviour more explicit.

PR #1317, opened 2019-08-05
Prepare ContactsController to support requirements

benthorner:
https://trello.com/c/8NN6jQiI/1025-modal-experience-for-contacts

    Previously the insertion of a contact required the routes to be under a
    document scope, which is no longer the case. This moves the routes for
    contacts out of the document scope and adds an '-embed' suffix to be
    consistent with the approach for YouTube videos.
    
    Previously we re-used the 'contacts#index' action to re-render the page
    and show the snippet to insert. This splits the 'index' action into
    'new' and 'create' actions, again to be consistent with the approach for
    video embeds. It would have been painful to try and support requirements
    checking with the previous 'index' action.
    
    This also changes the insertion mechanism to copy the approach we use for
    video embeds, where, in addition to requirements stuff, the controller
    responds with the snippet to insert. As a consequence, the autocomplete
    no longer needs to have knowledge of the markdown template for contacts.


app/controllers/contacts_controller.rb
```diff
@@ -1,13 +1,25 @@
...
+  end
+
+  def insert
+    if rendering_context == "modal"
+      render inline: I18n.t!("contacts.index.contact_markdown",
+                             id: params[:contact_id])
+    else
+      @edition = Edition.find_current(document: params[:document])
+      assert_edition_state(@edition, &:editable?)
+      render "index", layout: rendering_context
```
kevindew: Rather than repeating this code should we just redirect to index with the contact_id?

config/routes.rb
```diff
@@ -38,7 +38,7 @@
     get "/debug" => "debug#index", as: :debug_document
 
     get "/contacts" => "contacts#index"
-    post "/contacts" => "contacts#index"
+    post "/insert-contact" => "contacts#insert"
```
kevindew: This feels like it's be more consistent as `contacts/insert`

I'm wondering whether we intend to use ContactsController for the no-js experience. If not I think we could call this `contacts-embed` to be consistent with `video-embed`. I'm expecting they might be separate controllers as I don't think you'd need the document specific information.

benthorner: `contacts-embed` sounds good to me. I think `contacts/insert` would be somewhat unprecedented in our routes file. The controller does currently handle the no-JS experience, but I'm not sure how this will scale to the new design.

app/controllers/contacts_controller.rb
benthorner: I wondered about that, but while the no-JS design remains the same, we'll need to use the `insert` action to do the requirements aspect of the selection.

benthorner:
@alex-ju following a discussion with @kevindew, I've made some major changes to the way contacts work. I've updated the PR description to reflect this. Sorry for the continued upheaval!

Let me know if you need help understanding what's changed and/or why.

app/controllers/contact_embed_controller.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+class ContactEmbedController < ApplicationController
+  rescue_from GdsApi::BaseError do |e|
+    GovukError.notify(e)
+    render "new_api_down", layout: rendering_context
```
alex-ju: I think this name may be confusing. _Is it using the new API?_

app/controllers/contact_embed_controller.rb
```diff
@@ -0,0 +1,21 @@
...
+  end
+
+  def new
```
alex-ju: _New_ and _create_ are pretty much synonyms to me

benthorner: They are pretty similar, aren't they! For backend devs, they're a [standard Rails boilerplate](https://edgeguides.rubyonrails.org/getting_started.html#getting-up-and-running), so they do have some specific meaning.

app/controllers/contact_embed_controller.rb
benthorner: I hadn't thought of it like that! This is derived from the 'new' action, and we've [got some precedent for this naming](https://github.com/alphagov/content-publisher/blob/master/app/controllers/access_limit_controller.rb#L6) throughout the app.

PR #1324, opened 2019-08-07
Split PublishingApiPayload into supporting classes

benthorner:

Previously we had a single PublishingApiPayload class to generate
payloads for scheduling and draft preview. This splits up the class
into a couple of supporting classes for their respective services,
as part of the work to tidy up the 'app/services/' directory.

One potential issue was the location of the 'PUBLISHING_APP' constant,
which is used by three services. It makes sense for the constant to be
defined close to the 'put content' payload, since the other two services
that use this constant are coupled to the value specified here.

### DIFF
--- a/app/services/delete_draft_service.rb
+++ b/app/services/delete_draft_service.rb
@@ -30,7 +30,7 @@ private
 
   def discard_path_reservations(edition)
     paths = edition.revisions.map(&:base_path).uniq.compact
-    publishing_app = PublishingApiPayload::PUBLISHING_APP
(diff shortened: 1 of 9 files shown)


benthorner:
@kevindew agreed. I was wondering about the `config/application.rb` module, although it would have to be a different name like `PUBLISHING_API_ID` or something. Not so keen on using an env var directly as I doubt that's the intention behind it being defined. Anyway, we can always move the constant in future if it proves to be confusing 👍 .

PR #1325, opened 2019-08-07
Refactor image services

benthorner:

This makes several changes to the code for uploading an image to make it consistent with our approach to services, renaming and moving files as appropriate. The code to normalise an image has been changed so that it's now only used once in the upload pipeline, which is done by introducing a new 'TempImage' abstraction - I'm not set on the name.


lib/image_normaliser.rb
```diff
@@ -0,0 +1,40 @@
...
+  def initialize(raw_file)
+    @raw_file = raw_file
+  end
+
+  def normalise
+    @issues = Requirements::CheckerIssues.new
+    check_image_is_not_animated
+    check_image_is_big_enough
+    issues.any? ? return : normalised_image
+  end
```
kevindew: I think we're heading down a rather strange pattern with this class and date parser classes.

It feels somewhat unintuitive to me that you could do:

```
normaliser = ImageNormaliser.new(file)
if normaliser.issues.empty?
  image = normaliser.normalise # huh it's nil?
end
```

I feel like there's probably an interface that has `issues` and a `value` method (or a better named thing). Where issues is ready after initialize and value raises an exception rather than nil if it's accessed when there are issues.

I guess the categorisation is some form of input_parser?

lib/image_normaliser.rb
```diff
@@ -0,0 +1,40 @@
...
+
+  def initialize(raw_file)
+    @raw_file = raw_file
+  end
+
+  def normalise
+    @issues = Requirements::CheckerIssues.new
+    check_image_is_not_animated
+    check_image_is_big_enough
+    issues.any? ? return : normalised_image
```
kevindew: this seems a bit strange, maybe `normalised_image if issues.empty?`

benthorner: You and @1pretz1 clearly think alike :-). Fixed.

lib/image_normaliser.rb
benthorner: I agree the interface is a bit strange and simply using `value` across all three classes would be fine. But I also think it would be unintuitive for a class like this to have issues as soon as an instance is initialised - in my mental model, it shouldn't have 'done' anything yet.

I reckon we're skirting around another missing abstraction - some kind of `Result` class - that encapsulates the issues and the value after calling `normalise` (or whatever). Very reminiscent of [the maybe type](https://en.wikipedia.org/wiki/Option_type) or similar. Do you think that would address your concerns with the interfaces?

kevindew: > I agree the interface is a bit strange and simply using `value` across all three classes would be fine. But I also think it would be unintuitive for a class like this to have issues as soon as an instance is initialised - in my mental model, it shouldn't have 'done' anything yet.
>
> I reckon we're skirting around another missing abstraction - some kind of `Result` class - that encapsulates the issues and the value after calling `normalise` (or whatever). Very reminiscent of [the maybe type](https://en.wikipedia.org/wiki/Option_type) or similar. Do you think that would address your concerns with the interfaces?

Sure, I see what you mean. You're seeing the `#normalise` call as the point the class does something. In many ways though I think that then has the normalise effectively as an initialize method, particularly if that ends up returning a particular class instance of it's own.

I too was thinking this has parallels with optional types, but was thinking of the class representing that rather than it's output. 

I guess where we're actually caught is that the class is acting somewhere between a service and a value object. To act like a service it should have a single interface, like you suggest, and to act like a value object it should be ready to answer questions from the point of initialisation.

I learn more towards thinking of this as a value object as the extra object feels unnecessary, but I imagine we both agree both scenarios are preferable to current interface.

benthorner: @kevindew just to document where we left off in our discussion on this. We agreed that an `issues`/`value` interface would be preferable to what we have now, but also that these two methods don't make a lot of sense of a verb-type class like `ImageNormaliser`, since it's surprising for an 'actor' to have a value. Instead, we thought of two alternatives:

   - Renaming the class(es) to some kind of noun, like `NormalisedImage`. There's some precedent for doing this with [the new TempImage class](https://github.com/alphagov/content-publisher/blob/05a2243a3eaedc3a79f424f050cecc18111ce65c/lib/image_normaliser/temp_image.rb)
   - Adding a trivial struct class with the suggested interface, which is returned when calling `normalise`

The first approach has the advantage of minimising the number of objects and method calls, since it's not necessary to call a `normalise` method in order to get the result; however, it does depend on finding a name that makes sense for these classes.

The second approach has the advantage of minimal changes to the current code, at the expense of introducing a struct-type object to hold the result, which in turn requires a `normalise` method call to return it.

We agreed it's worth trying the first approach.

PR #1326, opened 2019-08-07
Ensure lib classes are reloaded when they change

benthorner:

Note this only applies in dev.


benthorner:
@kevindew yeah - I meant to link to this: https://stackoverflow.com/questions/19773266/confusing-about-autoload-paths-vs-eager-load-paths-in-rails-4

PR #1327, opened 2019-08-07
Refactor date/time parsers to use new interfaces

benthorner:

This got missed when the original changes were made...

This refactors the date/time parsers to take advantage the new interface
for Requirements::CheckerIssues. This also refactors the associated
specs to make use the the new matcher for requirements.

### DIFF
--- a/lib/date_parser.rb
+++ b/lib/date_parser.rb
@@ -1,31 +1,27 @@
 # frozen_string_literal: true
 
 class DateParser
+  attr_reader :raw_date, :issue_prefix, :issues
+
   def initialize(date:, issue_prefix:)
     @raw_date = date.to_h
     @issue_prefix = issue_prefix
-  end
-
-  def issues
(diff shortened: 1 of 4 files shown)


lib/date_parser.rb
```diff
@@ -1,31 +1,27 @@
...
+    @issues = Requirements::CheckerIssues.new
   end
 
   def parse
-    @issue_items = []
-
+    @issues = Requirements::CheckerIssues.new
     check_date_is_valid
-    issue_items.any? ? return : parsed_date
+    issues.any? ? return : parsed_date
```
1pretz1: I'm aware you didn't write most of this line but could this just be changed to - `parsed_date if issues.empty?`?

benthorner: Good point. The return is more explicit, but your suggestion is more idiomatic, so I've changed it.

1pretz1: 👍

PR #1330, opened 2019-08-07
Refactor preview services

benthorner:

This refactors the 'try' aspect of the existing PreviewService into a new 'Failsafe' version of the class, working towards our aim of each service having a single responsibility. Doing this creates some 'room' to solve a similar multiple-responsibility issue on the PreviewAssetService.

### DIFF
--- a/app/interactors/access_limit/update_interactor.rb
+++ b/app/interactors/access_limit/update_interactor.rb
@@ -72,6 +72,6 @@ private
   end
 
   def update_preview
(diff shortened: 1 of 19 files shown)


benthorner:
@kevindew agreed it's a bit nuanced. The things that made this seem like the path of least resistance for me were: the functionality is somewhat different; and the tests become less complex. If we do move the cleanup service out, then I agree the difference will be petty minimal!

PR #1331, opened 2019-08-08
Make asset cleanup a non-time-sensitive job

benthorner:

Previously we had a DraftAssetCleanupService that worked out which
image/attachment revisions were deleted, compared to the preceding
revision, and updated their assets accordingly in Asset Manager and
in the local DB. Because this service only refers to any preceding
revision, it has to be run every time any change is made to ensure
it always has the opportunity to detect any deletions; this subtle
behaviour makes it brittle and prone to missing deletions in future.

This rewrites the service as a job that operates over all editions
and revisions, which means it no longer needs to be run after every
change. Although we could make the job specific to a document, the
longterm intention is for this to run periodically as a scheduled
job, so adding this behaviour doesn't seem necessary.

Things we could also consider as part of this or future PRs:

   - Making the job specific to a document if that makes more sense
   - Adding in sidekiq-scheduler to take it out of the PreviewService
   - Using it to [replace a bit of code in the PublishAssetService](https://github.com/alphagov/content-publisher/blob/master/app/services/publish_asset_service.rb#L35)


app/jobs/asset_cleanup_job.rb
```diff
@@ -0,0 +1,40 @@
...
+    FileAttachment::Asset
+      .where.not(id: active_file_attachment_asset_ids)
+      .where.not(state: "absent")
+      .each { |asset| delete_asset(asset) }
+
+    Image::Asset
+      .where.not(id: active_image_asset_ids)
+      .where.not(state: "absent")
+      .each { |asset| delete_asset(asset) }
+  end
```
kevindew: I'm not too sure about all the id extracting and passing around. Unbounded that can crash PostgreSQL once the number gets sufficiently high (I think maybe 2^16) and is potentially quite inefficient - this problem caused a lot of confusion when it broke Email Alert API.

I have a proposal that could help us avoid that, by effectively moving all the pre-extractions to be coverd

```
asset_exists_for_active_editions = Edition.where(current: %w[true live])
                                                                     .joins(revision: { image_revisions: :blob_revision })
                                                                     .where("image_assets.blob_revision_id = image_blob_revisions.id")
                                                                     .arel
                                                                     .exists

Image::Asset
  .where.not(state: :absent)
  .where.not(asset_exists_for_active_editions)
  .find_each { |asset| delete_asset(asset) }
```

This effectively moves the pre-loading of information into a subquery.

kevindew:
I've done some work updating this to have single queries and to run automatically, should be given for another review.

spec/jobs/asset_cleanup_job_spec.rb
```diff
@@ -0,0 +1,175 @@
+# frozen_string_literal: true
+
+RSpec.describe AssetCleanupJob do
+  describe "#call" do
```
1pretz1: Shouldn't this be `describe "#perform"`?

app/jobs/asset_cleanup_job.rb
```diff
@@ -2,33 +2,39 @@
...
-      .where.not(id: active_file_attachment_asset_ids)
-      .where.not(state: "absent")
-      .each { |asset| delete_asset(asset) }
+  def clean_image_assets
+    image_asset_used =
+      Edition.where(current: true)
+             .or(Edition.where(live: true))
+             .joins(revision: { image_revisions: :blob_revision })
+             .where("image_assets.blob_revision_id = image_blob_revisions.id")
+             .arel
```
1pretz1: I don't really understand this line and the one below, could you explain if thats okay?

app/jobs/asset_cleanup_job.rb
```diff
@@ -2,33 +2,39 @@
...
-      .where("revisions.id" => active_revision_ids)
-      .pluck("image_assets.id")
+private
 
-    FileAttachment::Asset
-      .where.not(id: active_file_attachment_asset_ids)
-      .where.not(state: "absent")
-      .each { |asset| delete_asset(asset) }
+  def clean_image_assets
+    image_asset_used =
```
1pretz1: Should this be `image_assets_used` instead of `image_asset_used` as this is potentially concerning multiple assets?

app/jobs/asset_cleanup_job.rb
```diff
@@ -2,33 +2,39 @@
...
-      .where.not(state: "absent")
-      .each { |asset| delete_asset(asset) }
+    Image::Asset.where.not(state: :absent)
+                .where.not(image_asset_used)
+                .find_each { |asset| delete_asset(asset) }
   end
 
-private
+  def clean_file_attachment_assets
+    file_attachment_asset_used =
```
1pretz1: Same comment as before around pluralising `assets`

spec/jobs/asset_cleanup_job_spec.rb
kevindew: Good catch 👍

app/jobs/asset_cleanup_job.rb
kevindew: No problem. [Arel](https://www.rubydoc.info/github/rails/arel) is the underlying SQL generation library that Rails uses.

`.exists` is a method that will turn a SELECT query into one embedded in a EXISTS condition. For example: 

```
irb(main):010:0> Edition.where(id: 404).arel.to_sql
=> "SELECT \"editions\".* FROM \"editions\" WHERE \"editions\".\"id\" = $1"
irb(main):011:0> Edition.where(id: 404).arel.exists.to_sql
=> "EXISTS (SELECT \"editions\".* FROM \"editions\" WHERE \"editions\".\"id\" = $1)"
```

You used to be able to run `exists` directly on an ActiveRecord relation but this is now deprecated:

```
irb(main):012:0> Edition.where(id: 404).exists.to_sql
DEPRECATION WARNING: Delegating exists to arel is deprecated and will be removed in Rails 6.0. (called from irb_binding at (irb):12)
=> "EXISTS (SELECT \"editions\".* FROM \"editions\" WHERE \"editions\".\"id\" = $1)"
```

So anyway this gets us around to what this query is doing. We're built up a sub query that checks whether an image_asset is associated with a live or current edition. We're then negating that sub query to check whether there are any image_assets that don't have that association, so that we know it's safe to delete them.

app/jobs/asset_cleanup_job.rb
kevindew: This is looking for whether a single asset is used, it is used on each DB record.

kevindew:
Thanks @1pretz1, sorry I hadn't noticed your review sooner. Think I've addressed all the comments.

app/jobs/asset_cleanup_job.rb
1pretz1: Thanks for the explanation, it has been really helpful! I wonder why they're deprecating the exists call on a relation as it seems weird to have to have to call to arel before using it.

app/jobs/asset_cleanup_job.rb
1pretz1: Ah doh, nm then.

PR #1333, opened 2019-08-08
Fix image workflow bug

1pretz1:
See commits for further information.

### DIFF
--- a/app/interactors/images/update_crop_interactor.rb
+++ b/app/interactors/images/update_crop_interactor.rb
@@ -47,7 +47,8 @@ private
   def update_edition
     updater = Versioning::RevisionUpdater.new(edition.revision, user)
 
-    updater.update_image(image_revision, false)
+    selected = (context.edition.lead_image_revision == context.image_revision)
+    updater.update_image(image_revision, selected)
 
     context.fail! unless updater.changed?
 
--- a/spec/features/editing_images/edit_image_spec.rb
+++ b/spec/features/editing_images/edit_image_spec.rb
@@ -2,7 +2,7 @@
 
 RSpec.feature "Edit image", js: true do
   scenario "lead image" do
-    given_there_is_an_edition_with_images
+    given_there_is_an_edition_with_a_lead_image
 
     when_i_visit_the_images_page
     and_i_edit_the_image_crop
@@ -10,7 +10,9 @@ RSpec.feature "Edit image", js: true do
 
     when_i_edit_the_image_metadata
     then_i_see_the_image_is_updated
+    and_i_see_the_lead_image_is_updated
     and_the_preview_creation_succeeded
+    and_i_see_the_document_history_has_been_updated
   end
 
   scenario "inline image" do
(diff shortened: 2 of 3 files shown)


spec/features/editing_images/edit_image_spec.rb
```diff
@@ -2,14 +2,15 @@
...
-    given_there_is_an_edition_with_images
+    given_there_is_an_edition_with_a_lead_image
 
     when_i_visit_the_images_page
     and_i_edit_the_image_crop
     then_the_image_crop_is_updated
 
     when_i_edit_the_image_metadata
     then_i_see_the_image_is_updated
+    and_the_image_is_still_lead_image
```
benthorner: Good idea to have this check. I wonder if we can make it a visual check - we try to avoid DB checks in the features, since the user can't see the DB - by copying/modifying the preview step and using a 'within' block to check the same thing. Something like this:

```
def and_i_see_the_lead_image_is_updated
  within("something around the lead image") { and_j_see_the_image_is_updated }
end
```

1pretz1: Yep that makes sense.

1pretz1:
@benthorner I've amended the feature as per your request and i've also changed the implementation of the bug fix as the original approach prevented lead images from being removed.

spec/features/editing_images/edit_image_spec.rb
```diff
@@ -2,15 +2,17 @@
...
   scenario "lead image" do
-    given_there_is_an_edition_with_images
+    given_there_is_an_edition_with_a_lead_image
 
     when_i_visit_the_images_page
     and_i_edit_the_image_crop
     then_the_image_crop_is_updated
 
     when_i_edit_the_image_metadata
     then_i_see_the_image_is_updated
```
benthorner: I don't think we need this step anymore.

spec/features/editing_images/edit_image_spec.rb
```diff
@@ -2,15 +2,17 @@
...
+    and_i_see_the_lead_image_is_updated
     and_the_preview_creation_succeeded
+    and_i_see_the_document_history_has_been_updated
```
benthorner: Minor, but we usually avoid 'i_see' after the first one - it's implied.

app/interactors/images/update_crop_interactor.rb
```diff
@@ -47,7 +47,8 @@ def crop_params
   def update_edition
     updater = Versioning::RevisionUpdater.new(edition.revision, user)
 
-    updater.update_image(image_revision, false)
+    selected = (context.edition.lead_image_revision == context.image_revision)
```
benthorner: I'm still not sure we should do this, as the code in the updater is still faulty. What we're really doing here is working around the fault. I had a little go at making it better, and I think we could do

```
def currently_lead?(image_revision)
  revision.lead_image_revision&.image_id == image_revision.image_id
end
```

The `remove_image` method probably shouldn't be calling `next_lead_image` and could just use the above one to work out what the next lead image should be (either nil, or the current one).

1pretz1: It's not really working around the fault as the updater use's two different methods to remove a lead image. Either using `remove_image` if you click on the remove link or by using `update_image` if a user uncheck's the checkbox on the image metadata page. 

Weirdly that's why the code in the updater was put together like that, so changing the `next_lead_image` method to return the image_revision if it's currently lead actually breaks the checkbox way of deselecting a lead image -.-

benthorner: Oh noes. I see what you mean. I hate that checkbox sometimes! Anyway, there's still a smell here that the crop interactor is having to conjure up something that it shouldn't have to care about. This suggests there's something wrong with the updater interface, so I've had a go at spiking a slightly different one that fixes this: https://github.com/alphagov/content-publisher/compare/ben-updater-interface?expand=1.

All that's left would be to update the unit specs to cover the new method. How does it look to you?

1pretz1:
Closing this PR as splitting it into two, one for the frontend change (https://github.com/alphagov/content-publisher/pull/1338) and one for the bug fix (https://github.com/alphagov/content-publisher/pull/1337)

PR #1335, opened 2019-08-08
Introduce a new ApplicationService class

benthorner:

This introduces a new ApplicationService class that all top-level
classes in 'app/services' should inherit from, enforcing a common
interface among our services. Enforcing this interface requires a
variety of small code changes around the app, so that all service
calls are made like 'ServiceClass.call(args)'.

ScheduleService is the only class that still doesn't follow this
interface, which we intend to address in future work.

Sorry in advance that this is quite a large PR/commit, which is 
partly because we don't have time to do lots of smaller PRs and 
having a single commit will make it easier to address major comments. 
It does have the advantage of showing the big picture :-).

### DIFF
--- a/app/interactors/file_attachments/update_interactor.rb
+++ b/app/interactors/file_attachments/update_interactor.rb
@@ -65,7 +65,7 @@ private
   end
 
   def update_preview
-    FailsafePreviewService.new(edition).create_preview
+    FailsafePreviewService.call(edition)
   end
 
   def attachment_params
@@ -73,7 +73,6 @@ private
   end
 
   def blob_revision(file)
-    FileAttachmentBlobService.new(edition.revision, user)
-                             .create_blob_revision(file, replacing: file_attachment_revision)
+    FileAttachmentBlobService.call(edition.revision, user, file, replacing: file_attachment_revision)
   end
 end
--- a/app/services/application_service.rb
+++ b/app/services/application_service.rb
@@ -1,0 +1,9 @@
+# frozen_string_literal: true
+
+class ApplicationService
+  private_class_method :new
+
+  def self.call(*args)
+    new(*args).call
+  end
+end
--- a/app/services/failsafe_preview_service.rb
+++ b/app/services/failsafe_preview_service.rb
@@ -1,18 +1,16 @@
 # frozen_string_literal: true
 
-class FailsafePreviewService
-  attr_reader :edition
-
+class FailsafePreviewService < ApplicationService
   def initialize(edition)
     @edition = edition
   end
 
-  def create_preview
+  def call
     if has_issues?
-      DraftAssetCleanupService.new.call(edition)
+      DraftAssetCleanupService.call(edition)
       edition.update!(revision_synced: false)
     else
-      PreviewService.new(edition).create_preview
+      PreviewService.call(edition)
     end
   rescue GdsApi::BaseError => e
     edition.update!(revision_synced: false)
@@ -20,6 +18,8 @@ class FailsafePreviewService
   end
 
 private
+
+  attr_reader :edition
 
   def has_issues?
     Requirements::EditionChecker.new(edition).pre_preview_issues.any?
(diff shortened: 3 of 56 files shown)


app/services/file_attachment_blob_service.rb
```diff
@@ -1,41 +1,44 @@
...
+      number_of_pages: number_of_pages,
       created_by: user,
       asset: FileAttachment::Asset.new,
     )
   end
 
 private
 
-  attr_reader :revision, :user
+  attr_reader :revision, :user, :file, :replacing
```
kevindew: It seems to be a more common convention in our services classes to have these public. We should probably go one way or the other consistently. Personally I'd vote private.

app/services/failsafe_preview_service.rb
```diff
@@ -1,18 +1,18 @@
...
-class FailsafePreviewService
+class FailsafePreviewService < ApplicationService
   attr_reader :edition
 
   def initialize(edition)
     @edition = edition
   end
 
-  def create_preview
+  def call
```
kevindew: Ah you can see how this change here predicates the need to update the class name.

app/services/file_attachment_blob_service.rb
kevindew: Although the further down I go in this review it might be private that is more common.

app/services/path_generator_service.rb
```diff
@@ -1,33 +1,35 @@
...
+
+  def find_a_unique_path(prefix, slug)
+    (1..max_repeated_titles).each do |appended_count|
       base_path = create_path(prefix, slug, appended_count)
-      return base_path unless document_exists_with_path?(base_path, document)
+      return base_path unless path_in_use?(base_path)
     end
-    raise(ErrorGeneratingPath, "Already >#{@max_repeated_titles} paths with same title.")
+
+    raise "Already >#{max_repeated_titles} paths with same title."
```
kevindew: I guess the fact that this can switch to a RuntimeError means that nothing rescues this error 🙈

app/services/remove_service.rb
```diff
@@ -1,35 +1,51 @@
...
-        details: removal,
-      )
+      check_removeable
+      unpublish_edition
+      update_edition_status
+      create_timeline_entry
     end
 
-    delete_assets(edition.assets)
+    delete_assets
```
kevindew: Although I'm not sure this commit was the right place to do this refactoring I do like how this service is wrote and hope this can be the conventional way our call methods look 👍

app/interactors/file_attachments/update_interactor.rb
```diff
@@ -46,16 +46,13 @@ def update_file_attachment
     revision_attributes = attachment_params.slice(:title)
     revision_attributes[:blob_revision] = blob_revision(attachment_params[:file]) if attachment_params[:file]
     updater.assign(revision_attributes)
-
```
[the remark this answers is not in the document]
benthorner: Sorry - I got carried away. I've removed these changes.

app/services/application_service.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+class ApplicationService
+  private_class_method :new
+
+  def self.call(*args)
+    send(:new, *args).call
```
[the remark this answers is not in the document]
benthorner: So you can - how strange that it was failing for me when I tried it. Changed.

app/services/path_generator_service.rb
benthorner: Indeed. I thought this would be more consistent with how e do this elsewhere.

app/services/remove_service.rb
benthorner: Agreed. If we had more time, I'd have preferred to make these changes as a separate commit/PR, since this seems to be the main opportunity to align with the approach we have elsewhere.

benthorner:
@kevindew this should be ready for review now. I've also made the `attr_reader`s private, as discussed.

PR #1337, opened 2019-08-09
Fix lead image bug

1pretz1:
See commits for further information.

### DIFF
--- a/spec/lib/versioning/revision_updater/image_spec.rb
+++ b/spec/lib/versioning/revision_updater/image_spec.rb
@@ -61,45 +61,61 @@ RSpec.describe Versioning::RevisionUpdater::Image do
       expect(updater.removed_lead_image?).to be_falsey
     end
 
-    it "preserves the given image as the lead if selected" do
+    it "replaces the lead image revision with the updated revision" do
+      updated_image = create :image_revision, image_id: image_revision.image_id
       revision = create :revision, image_revisions: [image_revision],
                                    lead_image_revision: image_revision
 
-      updated_image = create :image_revision, image_id: image_revision.image_id
-      updater = Versioning::RevisionUpdater.new(revision, user)
-      updater.update_image(updated_image, true)
-
-      next_revision = updater.next_revision
-      expect(next_revision.lead_image_revision).to eq updated_image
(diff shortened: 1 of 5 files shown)


spec/lib/versioning/revision_updater/image_spec.rb
```diff
@@ -61,13 +61,27 @@
       expect(updater.removed_lead_image?).to be_falsey
     end
 
+    it "preserves the lead image if its the current image revision" do
```
benthorner: We don't normally combine tests for multiple methods in the same `describe` block - we want to be clear about which method we're testing. I think we can have two groups of test here: one for `#update_image` and a separate group for `#update_lead_image`.

1pretz1: Sure

spec/lib/versioning/revision_updater/image_spec.rb
```diff
@@ -61,45 +61,61 @@
...
-    it "preserves the given image as the lead if selected" do
+    it "preserves the lead image if its the current image revision" do
+      image_revision = create :image_revision
       revision = create :revision, image_revisions: [image_revision],
                                    lead_image_revision: image_revision
 
-      updated_image = create :image_revision, image_id: image_revision.image_id
       updater = Versioning::RevisionUpdater.new(revision, user)
-      updater.update_image(updated_image, true)
+      updater.update_image(image_revision)
```
benthorner: We should really be using a new image revision here, [like we do in some of the other specs](https://github.com/alphagov/content-publisher/pull/1337/files#diff-721754467f81ba105d69e284b97be012L80). Otherwise we're not actually updating anything; the point of this particular spec is to check the code senses that a new revision of the lead image exists and switches the `lead_image_revision` to use it. So the text for this spec should actually be "replaces" or "updates", instead of "preserves".

But...we should also have a companion spec to check it "preserves" the `lead_image_revision` when it's for some other, unrelated image to the one we're trying to update.

1pretz1: Okay dokay, done! & I believe the spec above is the 'companion spec' you're talking about for the `update_image` method.

PR #1355, opened 2019-09-05
Update to govuk_publishing_components 20.5

alex-ju:
This PR updates Content Publisher to govuk_publishing_components 20.5, which includes the usage of GOV.UK Frontend 3.0 and the migration of the summary component to the govuk_publishing_components gem.

~~This is currently failing with `No route matches [GET] "/assets/fonts/light-f38ad40456-v1.woff2"` error.~~ Fixed by [updating miller-columns-element to govuk-frontend 3.2](https://github.com/alphagov/miller-columns-element/pull/23) too.

<table>
<tr><td valign="top">

index-before

</td valign="top"><td>

index-after

</td></tr>

<tr><td valign="top">

summary-before

</td valign="top"><td>

summary-after

</td></tr>

<tr><td valign="top">

edit-before

</td valign="top"><td>

edit-after

</td></tr>

<tr><td valign="top">

modal-before

</td valign="top"><td>

moda-after

</td></tr>

</table>

[Trello card](https://trello.com/c/z6RKO2Jx)


app/assets/stylesheets/components/_markdown-editor.scss
```diff
@@ -27,7 +27,7 @@ $app-toolbar-button-size: 50px;
 .app-c-markdown-editor__preview-toggle {
   width: 100%;
   border-bottom: $govuk-border-width-form-element solid $govuk-border-colour;
-  background-color: govuk-colour("grey-3");
```
benthorner: I'm assuming `grey-3` is now equivalent to `grey-4`.

app/assets/stylesheets/core/_link.scss
```diff
@@ -3,13 +3,14 @@
   color: $govuk-error-colour;
 
   &:link,
-  &:visited {
+  &:visited,
```
benthorner: This looks like we're trying to show the error colour less often?

app/assets/stylesheets/components/_markdown-editor.scss
```diff
@@ -7,13 +7,15 @@ $app-toolbar-button-size: 50px;
...
       outline: none;
+      box-shadow: none;
     }
   }
 }
 
 .app-c-markdown-editor__container--focused {
   outline: $govuk-focus-width solid $govuk-focus-colour;
   outline-offset: 0;
+  box-shadow: inset 0 0 0 2px;
```
benthorner: From this I'm getting that we're using box-shadow instead of outline?

app/assets/stylesheets/core/_link.scss
```diff
@@ -10,7 +10,7 @@
   }
 
   &:focus {
-    color: $govuk-text-colour;
+    color: $govuk-focus-text-colour;
```
benthorner: I'm assuming this is a slightly different red from the error colour?

app/assets/stylesheets/core/_link.scss
```diff
@@ -1,6 +1,7 @@
 // TODO: support this modifier in govuk-frontend
```
benthorner: I'm guessing this didn't happen :-(.

app/assets/stylesheets/core/_link.scss
```diff
@@ -1,6 +1,7 @@
...
 .govuk-link.app-link--destructive {
   color: $govuk-error-colour;
+  @include govuk-font(19);
```
benthorner: I'm surprised this wasn't an issue when the button is used in other contexts i.e. on the images index page. How come it's the right size over there?

app/assets/javascripts/components/toolbar-dropdown.js
```diff
@@ -13,12 +13,17 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
     var $buttons = this.$module.querySelectorAll('.app-c-toolbar-dropdown__button')
     var $links = this.$module.querySelectorAll('.app-c-toolbar-dropdown__link')
 
+    if (!this.hasDetailsSupport()) {
```
benthorner: Just to clarify this against the commit message: do you mean that "since we've dropped govuk-frontend we now need to do the polyfill ourselves"?

app/assets/javascripts/components/toolbar-dropdown.js
```diff
@@ -32,15 +37,31 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
...
+  ToolbarDropdown.prototype.toggleDropdownContainer = function (event) {
+    if (!this.$module.hasAttribute('open')) {
+      this.$module.setAttribute('open', 'open')
+    } else {
+      this.$module.removeAttribute('open')
     }
   }
 
+  ToolbarDropdown.prototype.hasDetailsSupport = function () {
+    var el = document.createElement('DETAILS')
```
benthorner: I wonder if this is always returning false? Looking at the docs for the element, I can't see why it would contain an `open` attribute immediately after creation. Also, should we be cleaning it up?

benthorner: As an afterthought, I'm wondering if this could be separated as a 'true' polyfill? I can see it's tied in to the behaviour here, so the motivation is mainly to simplify the change.

app/assets/stylesheets/components/_toolbar-dropdown.scss
```diff
@@ -7,6 +7,16 @@ $app-dropdown-arrow-spacing: $app-dropdown-arrow-size / 2;
   min-width: 120px;
   background: inherit;
   text-align: left;
+
+  .app-c-toolbar-dropdown__container {
```
benthorner: Perhaps this is related to the potential bug I've asked about in the polyfill method: would these styles conflict with the browser behaviour if the details element is actually supported?

benthorner:
@alex-ju forgot to mention that I didn't get round to testing the modal experience with these changes - I think we should do that before this goes goes in too, as it's surprised us in the past.

alex-ju:
> @alex-ju forgot to mention that I didn't get round to testing the modal experience with these changes - I think we should do that before this goes goes in too, as it's surprised us in the past.

Good call. I will update the screen captures with the modal experience and reply to your comments.
Thanks for making time to review this!

app/assets/stylesheets/components/_markdown-editor.scss
alex-ju: Yes, now both `grey-3` and `grey-4` are replaced by `light-grey` which makes them the same.

app/assets/stylesheets/core/_link.scss
alex-ju: I would say 'more often'. We now show the same shade for `:link`, `:visited`, `:hover` and `:active`, the only state when it's different is `:focus`

app/assets/stylesheets/components/_markdown-editor.scss
alex-ju: Yes, that is right.

app/assets/stylesheets/core/_link.scss
alex-ju: Nope, this is actually a shade of black in order to generate the focus state consistently regardless of the type of link.

app/assets/stylesheets/core/_link.scss
alex-ju: Nope, not yet. I have a note in my to do list.

app/assets/stylesheets/core/_link.scss
alex-ju: This is only for destructive buttons which are not widely used, maybe that's why we haven't got them in a context that didn't have a font-size set.

app/assets/javascripts/components/toolbar-dropdown.js
alex-ju: Yes, now govuk-frontend doesn't enforce the polyfill for all details elements (it's being applied only for the ones with a `data-module` attached to them - which is now generated by the details component in govuk_publishing_components). This helps us to have a custom script for the toolbar-dropdown which doesn't clash with the previous polyfill.

app/assets/javascripts/components/toolbar-dropdown.js
alex-ju: Happy to separate this into it's own PR to give more room for explanation and code documentation.

app/assets/stylesheets/components/_toolbar-dropdown.scss
alex-ju: They shouldn't conflict, but again happy to extract this into its own piece of work.

alex-ju:
> * In development, the height (or something) about the top navigation bar has changed, so the content is now less vertically centred. Is that intentional?

Yes, it seems like the alignment within the header component changed slightly. It is how the component looks like at the moment in both the Design System and Publishing Components - I will raise it with the team to find out why.

> * The new focus style is very yellow!

Yes, it is. You'll get used to it the more you'll see it - so I've been told.

> * The new error text (e.g. for requirements issues) is brighter. I'd say that's harder to read (on a white background). Is that intentional?

It meets the contrast requirements. Again part of the colour scheme update that's being rolled out with govuk-frontend 3.0

> * We now seem to have a focus around our file picker (see below), which is accurate but not very pretty. Should we consider disabling it?

It's an accessibility enhancement. Not pretty indeed, we shouldn't disable.

> * We seem to have lost the underline from the links in the footer (see below), but maintained it in most other places. Is that a footer-specific change?

Yes, that's a footer specific change.

> * We seem to have a few font size inconsistencies around some of the buttons and some links I saw on the topics page (see below, again)

I will investigate and fix those, possibly other elements relying on style leakage from other components.

> * The green 'Content has been published' banner-flash-thing on the publish confirmation page is now a somewhat darker green (see below). That OK?

Yes, part of the colour scheme update.

You can find more about the visual changes in 3.0 in the [govuk-frontend changelog](https://github.com/alphagov/govuk-frontend/blob/master/CHANGELOG.md#300-breaking-release).

app/assets/stylesheets/core/_link.scss
benthorner: Oh yeah - sorry, I was misreading the '&'

app/assets/stylesheets/core/_link.scss
benthorner: Oops - I see now!

app/assets/javascripts/components/toolbar-dropdown.js
alex-ju: Sorry, I didn't mean to evade your question but to make things simpler. Assuming you're asking about the result of `GOVUK.Modules.ToolbarDropdown.prototype.hasDetailsSupport()` (not the one of `document.createElement('DETAILS')` where the comment was added) the result is `true` for browsers with support and `false` otherwise. Tested in all the browsers we support (including IE11 where it returns false).

benthorner: @alex-ju thanks for stopping by to remind me how JS works! To answer my own question: `createElement` doesn't attach things to the DOM, so the object will get cleaned up.

benthorner: As discussed, we think we should add a comment here to explain the unintuitive behaviour of `createElement` and how it auto-populates attributes on whatever it returns. This should clarify whether or not we expect to see the (nil valued) `open` attribute.

alex-ju:
@benthorner I've updated the `toolbar-dropdown.js` to simplify the logic (as you suggested) and added a comment about the object returned by `document.createElement('DETAILS')` depending on browser support. Let me know if that makes sense to you and if there are any other blockers on this PR.

benthorner:
@alex-ju thanks for looking into all those comments. I've found just a few more issues around the red style for 'delete' actions, and then I think we're ready to go...

Screen Shot 2019-09-23 at 18 18 37
Screen Shot 2019-09-23 at 18 18 45
Screen Shot 2019-09-23 at 18 19 01

alex-ju:
> @alex-ju thanks for looking into all those comments. I've found just a few more issues around the red style for 'delete' actions, and then I think we're ready to go...

Good spot @benthorner. I've fixed them now by adding a new commit.

PR #1389, opened 2019-10-08
Replace contextual guidance component

alex-ju:
This updates `govuk_publishing_components` to the latest version (21.5.0) which contains the `contextual-guidance` component. As a follow up work we're going to replace the app-level component with the gem-level one.

While moving the component to the gem we also did a [refactor to make it isolated](https://github.com/alphagov/govuk_publishing_components/pull/1156) (a thing we had in the Content Publisher backlog). This refactor was done in tandem with this PR to ensure compatibility.

[Trello card](https://trello.com/c/Ju8oaE5y/101-move-contextual-guidance-component-into-the-shared-library)


app/views/documents/fields/_govspeak_input.html.erb
```diff
@@ -10,8 +10,8 @@
 <%= render "govuk_publishing_components/components/contextual_guidance", {
   id: field.id,
   html_for: "#{field.id}-field",
-  title: edition.document_type.guidance_for(field.id).title,
-  content: document_contents_guidance
+  title: edition.document_type.guidance_for(field.id) ? edition.document_type.guidance_for(field.id).title : nil,
+  content: edition.document_type.guidance_for(field.id) ? document_contents_guidance : nil
```
ChrisBAshton: Instead of inline ternaries for every field, I'd be tempted to create the hash in one go:

```erb
<%
  field_contextual_guidance = nil
  if edition.document_type.guidance_for(field.id)
    field_contextual_guidance = {
      id: field.id,
      html_for: "...etc"
    }
  end
```

Then it's just:

```erb
<%= render "govuk_publishing_components/components/contextual_guidance", field_contextual_guidance do %>
  // render markdown editor
<% end %>
```

If `id` and `html_for` are required - even when no contextual guidance is rendered - I'd suggest removing that requirement from the component (ideally), or else just switching to a merge strategy:

```erb
<%
  field_contextual_guidance = { id: field.id, html_for: "...etc" }
  if edition.document_type.guidance_for(field.id)
    field_contextual_guidance.merge!({
      title: etc...
    })
  end
```

ChrisBAshton: I usually run `git rebase --ignore-date master` to reset the time on each commit and have them appear in the right order.

alex-ju:
> Still think ideally the component itself should be changed to allow just a `do` block with no `id`/`html_for` params, and then only raise an exception if say an `id` is passed without a `html_for`.

`id` is not required, but the `html_for` is, as we have no way from within the component to set the id of the input (since we don't know what component will be used within the block). I think the components should do as much as possible to prevent misconfiguration, for example calling the guidance without an html_for means that the JavaScript associated with this component will fail to show the guidance - that was the reasoning behind making it mandatory. The oddity within this case is that in tests, we have scenarios where the guidance doesn't exist - which is not a realistic scenario as we have it always.

Tests are passing now, I had a typo in a variable name on my last update.

PR #1393, opened 2019-10-10
Make 'sass' an explicit dependency

benthorner:
This ensures 'sass' is an explicit dependency in production, since it's
required for 'rake assets:precompile'. I think this used to be provided
via govuk_frontend_toolkit.


kevindew:
@benthorner Thanks for digging into this. Do we any idea why sass is being required? The only thing that has sass dependency seems to be scss_lint :-/

benthorner:
Hmm...I thought it would be necessary, but then I agree that doesn't make sense because it should be a dependency as part of the asset pipeline. We have sassc now, after all.

I wonder if herein lies the answer to the double-compilation mystery...seems worth digging in to before this goes in 👍.

benthorner:
@kevindew ah, I see. Anyway, I've amended the commit to explain why we need this...

> A new Rails app would have the 'sass-rails' gem, so I've used that instead of 'sass' directly.

PR #1407, opened 2019-10-28
Ensure all action links have the same font size

alex-ju:
Ensure all links in the action block are rendered with the same font-size - regardless of being wrapped in a paragraph with `govuk-body` applied to it or not.

[Trello card](https://trello.com/c/oo6GrYo7)


alex-ju:
Yep, I checked all the links in the actions box

PR #1409, opened 2019-10-29
Use a bearer token when retrieving export from Whitehall

brucebolt:
Since the Whitehall document export API endpoint is behind authentication (with user requiring the 'Export data' permission), we need to use a bearer token in order to be able to login as an API user with the correct permissions.

The bearer tokens are already configured in govuk-secrets.

Trello card: https://trello.com/c/RyVOGpMX

### DIFF
--- a/lib/tasks/import.rake
+++ b/lib/tasks/import.rake
@@ -1,11 +1,19 @@
 # frozen_string_literal: true
 
+require "gds_api/json_client"
+
 namespace :import do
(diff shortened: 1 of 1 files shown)


brucebolt:
Yes, it successfully migrated https://content-publisher.integration.publishing.service.gov.uk/documents/2c99e96e-f17e-4ce9-bd1e-4e9786e17087:en#edit-document

PR #1412, opened 2019-10-30
Flag imported documents as imported

brucebolt:
The changes here allow GOV.UK developers to see if a document (and which of it's revisions) was imported to Content Publisher, and if so, from which application.

For documents, we are setting the origin application name.  For revisions, we only set a boolean, since the origin application can be determined from the parent document.

Additionally, the Whitehall import process has been updated to set the relevant values for those imported documents.

Trello card: https://trello.com/c/C6UBegEP

### DIFF
--- a/app/models/document.rb
+++ b/app/models/document.rb
@@ -27,7 +27,7 @@ class Document < ApplicationRecord
 
   has_many :timeline_entries
 
-  enum imported_from: { whitehall: "whitehall" }, _prefix: :imported_from
+  enum imported_from: { whitehall: "whitehall" }, _prefix: true
 
   delegate :topics, to: :document_topics
 
--- a/app/models/document.rb
+++ b/app/models/document.rb
@@ -27,6 +27,8 @@ class Document < ApplicationRecord
 
   has_many :timeline_entries
 
+  enum imported_from: { whitehall: "whitehall" }, _prefix: :imported_from
+
   delegate :topics, to: :document_topics
 
   scope :with_current_edition, -> do
--- a/db/migrate/20191030093601_add_imported_to_revision.rb
+++ b/db/migrate/20191030093601_add_imported_to_revision.rb
@@ -1,0 +1,7 @@
+# frozen_string_literal: true
+
+class AddImportedToRevision < ActiveRecord::Migration[5.2]
+  def change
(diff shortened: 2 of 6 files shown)


db/migrate/20191030093601_add_imported_to_revision.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class AddImportedToRevision < ActiveRecord::Migration[5.2]
+  def change
+    add_column :revisions, :imported, :boolean, default: false
```
kevindew: This query could be slow to run - hopefully it's not as I don't think Content Publisher db is that big yet but worth checking on integration as it's locker https://medium.com/braintree-product-technology/postgresql-at-scale-database-schema-changes-without-downtime-20d3749ed680#e4a5

db/migrate/20191030085853_add_import_source_to_document.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class AddImportSourceToDocument < ActiveRecord::Migration[5.2]
+  def change
+    add_column :documents, :import_source, :string
```
kevindew: I feel like this should be flagged as an enum on a document to indicate what is supported, as per https://github.com/alphagov/content-publisher/blob/master/app/models/status.rb#L18-L27

If we change this to be named `imported_from` we could use the [_prefix](https://api.rubyonrails.org/v5.2.3/classes/ActiveRecord/Enum.html) option we can then get a boolean method on the document of `imported_from_whitehall?` for free

db/migrate/20191030093601_add_imported_to_revision.rb
brucebolt: This migration completed in 0.0977s on integration.  I don't think we need to worry about the locking with this duration, especially if the deployment is done outside normal publishing hours.

app/models/document.rb
```diff
@@ -27,6 +27,8 @@ class Document < ApplicationRecord
 
   has_many :timeline_entries
 
+  enum imported_from: { whitehall: "whitehall" }, _prefix: :imported_from
```
kevindew: minor, you can achieve the same with `_prefix: true`

PR #1414, opened 2019-10-30
Import all users from Whitehall document export

brucebolt:
Editors of documents in Whitehall may never have logged into Content Publisher, so we will not have a row in the users table for them.  This change creates a user entry for every person who has ever edited the Whitehall document that is imported.

For security, we give these users no permissions in the Content Publisher database.  If they log into the application in the future, the permissions will get updated based on the values in Sign On.

Users who have logged into Content Publisher are skipped.

This code is branched from https://github.com/alphagov/content-publisher/pull/1416, so should not be merged until that PR is also merged.  It is also dependent on a users array being implemented in the Whitehall document export.

Trello card: https://trello.com/c/JPerZkxp

### DIFF
--- a/lib/tasks/whitehall_importer.rb
+++ b/lib/tasks/whitehall_importer.rb
@@ -16,6 +16,7 @@ module Tasks
     def import
       ActiveRecord::Base.transaction do
         document = create_or_update_document
+        create_users(whitehall_document["users"])
 
         whitehall_document["editions"].each_with_index do |edition, edition_number|
           edition["translations"].each do |translation|
@@ -46,6 +47,15 @@ module Tasks
     end
 
   private
+
+    def create_users(users)
+      users.each do |user|
+        next if User.exists?(uid: user["uid"])
+
+        user_keys = %w[uid name email organisation_slug organisation_content_id]
+        User.create!(user.slice(*user_keys).merge("permissions" => []))
+      end
+    end
 
     def most_recent_edition
       whitehall_document["editions"].max_by { |e| e["created_at"] }
--- a/spec/tasks/whitehall_importer_spec.rb
+++ b/spec/tasks/whitehall_importer_spec.rb
@@ -57,6 +57,16 @@ RSpec.describe Tasks::WhitehallImporter do
           ],
         },
       ],
+      "users" => [
+        {
+          "id" => 1,
+          "name" => "A Person",
+          "uid" => "36d5154e-d3b7-4e3e-aad8-32a50fc9430e",
+          "email" => "a-publisher@department.gov.uk",
+          "organisation_slug" => "a-government-department",
+          "organisation_content_id" => "01892f23-b069-43f5-8404-d082f8dffcb9",
+        },
+      ],
     }
   end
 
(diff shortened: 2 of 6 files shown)


spec/tasks/whitehall_importer_spec.rb
```diff
@@ -162,6 +183,16 @@
...
+        "users" => [
+          {
+            "id" => 1,
+            "name" => "A Person",
+            "uid" => "36d5154e-d3b7-4e3e-aad8-32a50fc9430e",
+            "email" => "a-publisher@department.gov.uk",
+            "organisation_slug" => "a-government-department",
+            "organisation_content_id" => "01892f23-b069-43f5-8404-d082f8dffcb9",
+          },
+        ],
```
ChrisBAshton: This fixture data is repeated 3 times. Can we create a new shared `fixtures` file/folder that can be shared between all the tests?

spec/tasks/whitehall_importer_spec.rb
```diff
@@ -45,6 +55,17 @@
...
+  it "adds users who have never logged into Content Publisher" do
+    importer = Tasks::WhitehallImporter.new(123, import_data)
+    importer.import
+
+    expect(User.last.uid).to eq "36d5154e-d3b7-4e3e-aad8-32a50fc9430e"
+    expect(User.last.name).to eq "A Person"
+    expect(User.last.email).to eq "a-publisher@department.gov.uk"
+    expect(User.last.organisation_slug).to eq "a-government-department"
+    expect(User.last.organisation_content_id).to eq "01892f23-b069-43f5-8404-d082f8dffcb9"
+  end
```
ChrisBAshton: Is it worth adding a test that users who _have_ logged into Content Publisher already don't get re-added?

lib/tasks/whitehall_importer.rb
```diff
@@ -47,17 +48,32 @@ def log_error(error)
...
+      users.each do |user|
+        next if User.exists?(uid: user["uid"])
+
+        User.create!(
+          uid: user["uid"],
+          name: user["name"],
+          permissions: [],
+          email: user["email"],
+          organisation_slug: user["organisation_slug"],
+          organisation_content_id: user["organisation_content_id"],
```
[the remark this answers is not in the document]
kevindew: We don't quite want all the data (as we have the Whitehall id) but we can use your approach with a slice. E.g: `user.slice(*%w[uid name email organisation_slug organisation_content_id]).merge("permissions" => [])`

kevindew:
Do we want to saving user attribution (docs / revisions they created) or is that a separate story? I'm assuming it's separate but thought I'd check.

brucebolt:
@kevindew I'd say separate.  I've hacked a few things in locally (just to see how this approach worked in practice), but I think it'll need writing up properly with acceptance criteria, etc.

kevindew:
> @kevindew I'd say separate. I've hacked a few things in locally (just to see how this approach worked in practice), but I think it'll need writing up properly with acceptance criteria, etc.

Great - I'll make sure we have one for the next sprint. It feels a bit of a weird thing anyway as it crosses the line between audit and presentation data.

lib/tasks/whitehall_importer.rb
```diff
@@ -92,26 +92,26 @@ def create_edition(document, translation, whitehall_edition, edition_number)
...
         "submitted_for_review"
       end
     end
 
     def live?(whitehall_edition)
-      whitehall_edition["edition"]["state"] == "published"
+      whitehall_edition["state"] == "published"
     end
 
     def embed_contacts(body, contacts)
```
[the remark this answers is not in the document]
brucebolt: This was updated in #1416

spec/tasks/whitehall_importer_spec.rb
```diff
@@ -76,6 +86,24 @@
...
+
+  it "does not add users who have logged into Content Publisher" do
+    importer = Tasks::WhitehallImporter.new(123, import_data)
+    User.create!(uid: "36d5154e-d3b7-4e3e-aad8-32a50fc9430e")
+
+    expect { importer.import }.not_to(change { User.count })
+  end
```
ChrisBAshton: This is great 👍  mergeable. But I wonder if it's worth testing that it doesn't overwrite users who have logged into Content Publisher, in case their name/email has changed? The code doesn't overwrite but may be an idea to enforce that with a test.

brucebolt: It won't matter if their data is overwritten, since it'll get updated from Signon next time they login

PR #1415, opened 2019-10-30
Tag imported documents

leenagupte:
Trello: https://trello.com/c/pHYtNC5z

## What's changed
Tags imported documents to:
- their Whitehall publishing organisation
- ministers and government appointments
- topical events
- world locations

Documents that have more than one lead publishing organisation are rejected as the other lead organisations wouldn't be able to update the document once it's been imported to Content Publisher.

## Note for reviewer
The structure of the import from Whitehall is based on: https://github.com/alphagov/whitehall/pull/5093

### DIFF
--- a/lib/tasks/whitehall_importer.rb
+++ b/lib/tasks/whitehall_importer.rb
@@ -85,6 +85,7 @@ module Tasks
             "organisations" => supporting_organisations(whitehall_edition["organisations"]),
             "role_appointments" => tags(whitehall_edition["role_appointments"]),
             "topical_events" => tags(whitehall_edition["topical_events"]),
+            "world_locations" => tags(whitehall_edition["world_locations"]),
           },
         ),
         created_at: whitehall_edition["created_at"],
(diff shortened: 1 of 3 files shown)


lib/tasks/whitehall_importer.rb
```diff
@@ -121,5 +129,45 @@ def embed_contacts(body, contacts)
...
+      unless primary_publishing_organisations.any?
+        raise AbortImportError, "Lead organisation missing"
+      end
+
+      if primary_publishing_organisations.count > 1
+        raise AbortImportError, "Cannot have more than one lead organisation"
+      end
+
+      primary_publishing_organisation = primary_publishing_organisations.find do |organisation|
+        organisation["lead_ordering"] == 1
```
kevindew: Sorry to be a pain but I wonder if we definitively know that the list will be ordered starting from 1?

I think it might be slightly better to do 
`primary_organisations = organisations.select { |o| o["lead"] }.sort { |o| o["lead_ordering"] }` and then we can use first item there.

brucebolt: Good point.  Have updated the PR.

PR #1420, opened 2019-11-04
Import user attribution from Whitehall

brucebolt:
So far, we have imported users and documents from Whitehall, but not linked the users to the documents they have edited.  When importing a document from Whitehall, we now attribute the following to Content Publisher users:
- document creator
- edition creator
- edition last editor

Trello card: https://trello.com/c/jOaNlx3c

### DIFF
--- a/lib/tasks/whitehall_importer.rb
+++ b/lib/tasks/whitehall_importer.rb
@@ -2,7 +2,7 @@
 
 module Tasks
   class WhitehallImporter
-    attr_reader :whitehall_document_id, :whitehall_document, :whitehall_import
+    attr_reader :whitehall_document_id, :whitehall_document, :whitehall_import, :user_ids
 
     SUPPORTED_WHITEHALL_STATES = %w(draft published rejected submitted superseded).freeze
     SUPPORTED_LOCALES = %w(en).freeze
@@ -11,6 +11,7 @@ module Tasks
       @whitehall_document_id = whitehall_document_id
       @whitehall_document = whitehall_document
       @whitehall_import = store_json_blob
+      @user_ids = {}
     end
 
     def import
@@ -50,10 +51,9 @@ module Tasks
 
     def create_users(users)
       users.each do |user|
-        next if User.exists?(uid: user["uid"])
-
         user_keys = %w[uid name email organisation_slug organisation_content_id]
-        User.create!(user.slice(*user_keys).merge("permissions" => []))
+        content_publisher_user = User.create_with(user.slice(*user_keys).merge("permissions" => [])).find_or_create_by!(uid: user["uid"])
+        user_ids[user["id"]] = content_publisher_user["id"]
       end
     end
 
--- a/lib/tasks/whitehall_importer.rb
+++ b/lib/tasks/whitehall_importer.rb
@@ -78,6 +78,7 @@ module Tasks
 
     def create_edition(document, translation, whitehall_edition, edition_number)
       first_author = whitehall_edition["history"].select { |h| h["event"] == "create" }.first
+      last_author = whitehall_edition["history"].last
 
       revision = Revision.create!(
         document: document,
@@ -121,6 +122,7 @@ module Tasks
         created_at: whitehall_edition["created_at"],
         updated_at: whitehall_edition["updated_at"],
         created_by_id: user_ids[first_author["whodunnit"]],
+        last_edited_by_id: user_ids[last_author["whodunnit"]],
       )
     end
 
--- a/lib/tasks/whitehall_importer.rb
+++ b/lib/tasks/whitehall_importer.rb
@@ -77,6 +77,8 @@ module Tasks
     end
 
     def create_edition(document, translation, whitehall_edition, edition_number)
+      first_author = whitehall_edition["history"].select { |h| h["event"] == "create" }.first
+
       revision = Revision.create!(
         document: document,
         number: document.next_revision_number,
@@ -118,6 +120,7 @@ module Tasks
         live: live?(whitehall_edition),
         created_at: whitehall_edition["created_at"],
         updated_at: whitehall_edition["updated_at"],
+        created_by_id: user_ids[first_author["whodunnit"]],
       )
     end
 
--- a/lib/tasks/whitehall_importer.rb
+++ b/lib/tasks/whitehall_importer.rb
@@ -16,8 +16,8 @@ module Tasks
 
     def import
       ActiveRecord::Base.transaction do
-        document = create_or_update_document
         create_users(whitehall_document["users"])
+        document = create_or_update_document
 
         whitehall_document["editions"].each_with_index do |edition, edition_number|
           edition["translations"].each do |translation|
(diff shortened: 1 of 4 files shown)


lib/tasks/whitehall_importer.rb
```diff
@@ -50,10 +51,13 @@ def log_error(error)
 
     def create_users(users)
       users.each do |user|
-        next if User.exists?(uid: user["uid"])
-
         user_keys = %w[uid name email organisation_slug organisation_content_id]
-        User.create!(user.slice(*user_keys).merge("permissions" => []))
+        content_publisher_user = User.create_with(user.slice(*user_keys).merge("permissions" => [])).find_or_create_by!(uid: user["uid"])
```
ChrisBAshton: I found this a bit tricky to parse. I wonder if more verbose would be better?

```ruby
uid = user["uid"]
content_publisher_user = User.exists?(uid) ? User.find(uid) : create_user(user_keys)
```

lib/tasks/whitehall_importer.rb
```diff
@@ -66,12 +66,16 @@ def most_recent_edition
...
+      first_edition = whitehall_document["editions"].first
+      first_author = first_edition["history"].select { |h| h["event"] == "create" }.first
+
+      Document.find_or_create_by!(
         content_id: whitehall_document["content_id"],
         locale: "en",
         document_type_id: "news_story", ## To be updated once Whitehall exports this value
         created_at: whitehall_document["created_at"],
         updated_at: whitehall_document["updated_at"],
+        created_by_id: user_map.select { |u| u["whitehall_user_id"] == first_author["whodunnit"] }.first["content_publisher_user_id"],
```
ChrisBAshton: This would benefit from being extracted to a method, in the next two commits too:

```ruby
created_by_id: whitehall_uid_to_signon_uid(first_author["whodunnit"])
```

lib/tasks/whitehall_importer.rb
```diff
@@ -11,12 +11,13 @@ def initialize(whitehall_document_id, whitehall_document)
       @whitehall_document_id = whitehall_document_id
       @whitehall_document = whitehall_document
       @whitehall_import = store_json_blob
+      @user_map = []
```
kevindew: I find the map part of this variable name a bit confusing as I'd anticipate a map to be an associative array/hash object rather than an array. Might be better to just call it users?

lib/tasks/whitehall_importer.rb
```diff
@@ -50,10 +51,13 @@ def log_error(error)
...
-        next if User.exists?(uid: user["uid"])
-
         user_keys = %w[uid name email organisation_slug organisation_content_id]
-        User.create!(user.slice(*user_keys).merge("permissions" => []))
+        content_publisher_user = User.create_with(user.slice(*user_keys).merge("permissions" => [])).find_or_create_by!(uid: user["uid"])
+        user_map.push(
+          "uid" => user["uid"],
+          "whitehall_user_id" => user["id"],
+          "content_publisher_user_id" => content_publisher_user["id"],
+        )
```
kevindew: I wonder if it'd be easier to traverse this as a hash of whitehall user_ids to content publisher user ids? I assume we don't really use the uid part after this method. That might give some clues to adjust naming too.

lib/tasks/whitehall_importer.rb
kevindew: Just to expand on this if you used a hash of whitehall user ids to content publisher user ids this could be down to:`user_ids[first_author["whodunnit"]`

lib/tasks/whitehall_importer.rb
```diff
@@ -62,17 +66,24 @@ def most_recent_edition
...
         created_at: whitehall_document["created_at"],
         updated_at: whitehall_document["updated_at"],
+        created_by_id: user_map.select { |u| u["whitehall_user_id"] == first_author["whodunnit"] }.first["content_publisher_user_id"],
         imported_from: "whitehall",
       )
     end
 
     def create_edition(document, translation, whitehall_edition, edition_number)
+      first_author = whitehall_edition["history"].select { |h| h["event"] == "create" }.first
+      last_author = whitehall_edition["history"].last
```
kevindew: Do we know if they're always an editor? Looking up events in Whitehall we get 3 used ones so I think this is probably safe.

```
irb(main):003:0> Version.distinct.pluck(:event)
=> ["create", "update", "destroy"]
```

lib/tasks/whitehall_importer.rb
brucebolt: This would have to be written as follows, which seems lengthier

```
content_publisher_user = User.exists?(uid: user["uid"]) ? User.find_by(uid: user["uid"]) : User.create!(user.slice(*user_keys).merge("permissions" => []))
```

Additionally, if the user exists we'd be executing two `SELECT` queries, instead of just the one.

lib/tasks/whitehall_importer.rb
```diff
@@ -2,7 +2,7 @@
 
 module Tasks
   class WhitehallImporter
-    attr_reader :whitehall_document_id, :whitehall_document, :whitehall_import
+    attr_reader :whitehall_document_id, :whitehall_document, :whitehall_import, :user_map
```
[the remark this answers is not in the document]
brucebolt: It doesn't yet feel like there's enough items here to put them on multiple lines (can't see any other cases of `attr_reader` being multilines in Content Publisher).

lib/tasks/whitehall_importer.rb
brucebolt: Indeed, I've updated the PR.  I was planning to use the UIDs at some other point, but turns out they weren't needed for anything.

lib/tasks/whitehall_importer.rb
ChrisBAshton: I think the `User.exists?(uid)` syntax would work ok (tried it in app console on integration)... does it not work locally?

Good point about the multiple `SELECT` queries. But I think this would work via index rather than select, so should be very fast.

ChrisBAshton: (Not a blocker, mind - do feel free to merge following Kevin's "✅")

brucebolt: This matches on the primary key (i.e. `id` for the users table).  It seems there's some magic going on where the first integer section of the string is being used to run the query, e.g.

```
irb(main):368:0> User.exists?("7a2917c0-bbf6-0135-c82a-1234")
  User Exists (1.6ms)  SELECT  1 AS one FROM "users" WHERE "users"."id" = $1 LIMIT $2  [["id", 7], ["LIMIT", 1]]
=> true
irb(main):369:0> User.exists?("917c0-bbf6-0135-c82a-1234")
  User Exists (1.0ms)  SELECT  1 AS one FROM "users" WHERE "users"."id" = $1 LIMIT $2  [["id", 917], ["LIMIT", 1]]
=> false
```

By coincidence I have a user with the id of 7, so the first one returns true when the uid doesn't actually exist.

ChrisBAshton: Understood.

kevindew: That'll be Rails type coercing into an integer. If you run `to_i` on a string it'll stop when it hits a alphanumeric character (though I be there are other complications)

PR #1421, opened 2019-11-04
Pass document_type_id through to metadata_revisions

JonathanHallam:
As part of allowing content publisher to change the document types of documents we need to have the document type id saved in metadata revisions. This duplicates thatf information across metadata revisions and document. This will be removed from Document after later schema changes.

Relies on - https://github.com/alphagov/content-publisher/pull/1419/files
Trello - https://trello.com/b/mn2MqH3M/publishing-workflow-doing-q4-2019-to

Co-authored-by: Aga Dufrat <agnieszka.dufrat@digital.cabinet-office.gov.uk>

### DIFF
--- a/app/models/edition.rb
+++ b/app/models/edition.rb
@@ -92,8 +92,13 @@ class Edition < ApplicationRecord
       .find_by!(find_by)
   end
 
-  def self.create_initial(document, user = nil, tags = {})
-    revision = Revision.create_initial(document, user, tags)
+  def self.create_initial(document:, document_type_id:, user: nil, tags: {})
+    revision = Revision.create_initial(
+      document: document,
+      user: user,
+      tags: tags,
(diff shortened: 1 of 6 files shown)


app/models/edition.rb
```diff
@@ -92,8 +92,8 @@ class Edition < ApplicationRecord
       .find_by!(find_by)
   end
 
-  def self.create_initial(document, user = nil, tags = {})
-    revision = Revision.create_initial(document, user, tags)
+  def self.create_initial(document, user = nil, tags = {}, document_type_id = "")
```
kevindew: I don't think we'd want this to default to an empty string. I think we expect this to actually have a value so may want to make it a required parameter.

As we've now got 4 params we may well want to make this keyword arguments.

JonathanHallam: done!

PR #1426, opened 2019-11-05
[WIP] Import images

ChrisBAshton:
This PR imports image data from Whitehall. It does not do the following (links to their Trello cards):

1) [deal with attachments](https://trello.com/c/4FF2oprK/1182-import-attachments-into-content-publisher)
2) [change Govspeak markdown to use content publisher image syntax](https://trello.com/c/ubTM9ylI/1176-use-imported-images-and-attachments-in-content-publisher)

- [x] Should create the necessary ActiveRecord image references from an edition successfully
- [x] Should cope with all of the following: jpg, jpeg, gif, png
- [x] Should abort if an image is not found
- [x] Should cope with empty caption
- [x] Should rename 2nd image if two images share the same filename
- [x] Should import image(s) added across multiple editions. We've agreed for now we're going to blindly import images even if they're duplicated across multiple editions, just treating them as new image revisions every time.
- [x] Abort import when SVG
- [x] Abort if main image is not 960x640

------

Trello card: https://trello.com/c/jzV3scso/1093-import-images-and-attachments-to-content-publisher

### DIFF
--- a/app/services/image_blob_service.rb
+++ b/app/services/image_blob_service.rb
@@ -3,16 +3,16 @@
 require "mini_magick"
 
 class ImageBlobService < ApplicationService
-  def initialize(revision, user, temp_image)
-    @revision = revision
-    @user = user
+  def initialize(temp_image:, filename:, user: nil)
     @temp_image = temp_image
+    @filename = filename
+    @user = user
   end
 
   def call
     blob = ActiveStorage::Blob.create_after_upload!(
       io: temp_image.file,
-      filename: unique_filename,
+      filename: filename,
       content_type: temp_image.mime_type,
     )
 
@@ -24,21 +24,14 @@ class ImageBlobService < ApplicationService
       crop_y: centre_crop[:y],
       crop_width: centre_crop[:width],
       crop_height: centre_crop[:height],
-      filename: unique_filename,
+      filename: filename,
       created_by: user,
     )
   end
 
 private
 
-  attr_reader :revision, :user, :temp_image
-
-  def unique_filename
-    filenames = revision.image_revisions.map(&:filename)
-
-    @unique_filename ||= UniqueFilenameService
-      .call(filenames, temp_image.original_filename)
-  end
+  attr_reader :temp_image, :filename, :user
 
   def centre_crop
     @centre_crop ||= CentreCrop.new(temp_image.width,
--- a/lib/tasks/whitehall_importer.rb
+++ b/lib/tasks/whitehall_importer.rb
@@ -198,7 +198,7 @@ module Tasks
 
     def create_image_revisions(whitehall_images)
       whitehall_images_added = []
-      add_filenames_to_images_hash(whitehall_images).map do |whitehall_image|
+      validate_images_and_add_fields(whitehall_images).map do |whitehall_image|
         image_revision = Image::Revision.create!(
           image: Image.create!(
             created_at: whitehall_image["created_at"],
@@ -219,9 +219,13 @@ module Tasks
       end
     end
 
-    def add_filenames_to_images_hash(whitehall_images)
+    def validate_images_and_add_fields(whitehall_images)
       whitehall_images.map do |memo|
         memo["filename"] = File.basename(URI.parse(memo["url"]).path)
+        if memo["filename"].end_with?(".svg")
+          raise AbortImportError, "SVG detected: #{memo['url']}"
+        end
+
         memo
       end
     end
--- a/lib/tasks/whitehall_importer.rb
+++ b/lib/tasks/whitehall_importer.rb
@@ -232,8 +232,14 @@ module Tasks
 
     def create_blob_from_image(whitehall_image, images_added_so_far)
       raw_file = URI.parse(whitehall_image["url"]).open
+      image = ImageNormaliser.new(raw_file).normalise
+
+      unless image.width == 960 && image.height == 640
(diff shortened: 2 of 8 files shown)


Gemfile
```diff
@@ -32,6 +32,7 @@ gem "with_advisory_lock", "~> 4"
 group :development do
   gem "brakeman", "~> 4"
   gem "listen", "~> 3"
+  gem "pry", "~> 0.12"
```
kevindew: We've already got [byebug](https://github.com/deivid-rodriguez/byebug) for this a few lines down in the dev/test section.

app/services/image_blob_service.rb
```diff
@@ -31,9 +32,11 @@ def call
 
 private
 
-  attr_reader :revision, :user, :temp_image
+  attr_reader :temp_image, :revision, :user, :hardcoded_filename
 
   def unique_filename
+    return hardcoded_filename if hardcoded_filename.present?
```
kevindew: We still need this filename to be unique so we should run this through the unique filename service

app/services/image_blob_service.rb
```diff
@@ -3,10 +3,11 @@
 require "mini_magick"
 
 class ImageBlobService < ApplicationService
-  def initialize(revision, user, temp_image)
+  def initialize(temp_image:, revision: nil, user: nil, hardcoded_filename: nil)
+    @temp_image = temp_image
     @revision = revision
     @user = user
-    @temp_image = temp_image
+    @hardcoded_filename = hardcoded_filename
```
kevindew: If I understand right we're using this input as a way to skip the unique file service which I think we'll actually need. So I think we can consider removing this.

lib/tasks/whitehall_importer.rb
```diff
@@ -126,6 +127,37 @@ def create_edition(document, translation, whitehall_edition, edition_number)
...
+            credit: whitehall_image["credit"],
+            created_at: whitehall_image["created_at"],
+          ),
+          blob_revision: blob_revision,
+        )
+      end
+    end
+
+    def create_blob_from_image(image_url)
+      unique_filename = File.basename(URI.parse(image_url).path)
```
kevindew: I'm not sure there's anything about this that is unique?

lib/tasks/whitehall_importer.rb
```diff
@@ -126,6 +127,37 @@ def create_edition(document, translation, whitehall_edition, edition_number)
...
+      raw_file = URI.parse(image_url).open
+      temp_image = ImageNormaliser.new(RawFile.new(raw_file, unique_filename)).normalise
```
kevindew: I think these things could be handled by the nested class you've created

lib/tasks/whitehall_importer.rb
```diff
@@ -188,3 +220,13 @@ def tags(associations)
 
   class AbortImportError < RuntimeError; end
 end
+
+class RawFile
```
kevindew: I'm not sure RawFile is the right name for this object, perhaps AssetManagerFile ?

What I'd suggest is that we actually give this an input of the Asset Manager URL and then use this class the responsibility of converting that to a file object. We could also use that URL for the filename.

Gemfile
ChrisBAshton: Ahh nice! Any reason why we're using byebug instead of pry? I haven't come across it before.

(Will be removing this commit though - pointless having two debug gems!).

lib/tasks/whitehall_importer.rb
ChrisBAshton: I've now done away with this extra class, by configuring ImageBlobService to be more easily callable and saving having to stub `revision` and `original_filename` in hacky ways.

app/services/image_blob_service.rb
ChrisBAshton: Have been thinking about this, and not sure it's a good idea to run through the UniqueFilenameService, as this can end up changing the filename if duplicates are detected. And then what? The Govspeak body would be out of date, referencing images by filenames that no longer exist.

For the same reason, I don't think we should parameterize the filenames (e.g. `Foo--1.png` => `foo-1.png`), which happens when we call the UniqueFilenameService regardless of the uniqueness of the filename.

I'd be inclined to check for duplicate image filenames manually and abort the import when that happens. This leaves me with two concerns:

1) What happens if an editor attempts to edit an image that originated from Whitehall? (Presumably if there is buggy behaviour they could work around it by deleting the image and reuploading).
2) Are we in any way tied to the `/media/hash-id/filename.extension` structure (as opposed to `/government/uploads/system/uploads/image_data/file/file-id/filename.extension`)? The [`asset_manager_id` method](https://github.com/alphagov/content-publisher/blob/5acbe8293917ea0b90a96c23b8768e708ebbd2b9/app/models/image/asset.rb#L23-L27)) suggests we might be, but as the portion of the URL preceding the filename acts as an identifier in either case, it may not be an issue.

app/services/image_blob_service.rb
```diff
@@ -3,11 +3,12 @@
 require "mini_magick"
 
 class ImageBlobService < ApplicationService
-  def initialize(temp_image:, proposed_filename:, existing_filenames: [], user: nil)
+  def initialize(temp_image:, proposed_filename:, existing_filenames: [], user: nil, parameterize_name: true)
```
ChrisBAshton: TODO make a separate service for the filename logic. Silly for it to be all in this service.

ChrisBAshton:
Superseded by https://github.com/alphagov/content-publisher/pull/1448.

PR #1433, opened 2019-11-07
Replace govuk-lint with rubocop-govuk and scss_lint-govuk

theseanything:
As govuk-lint is being deprecated, this PR replaces usage of govuk-lint with rubocop, rubocop-govuk, scss_lint, scss_lint-govuk.


theseanything:
> Have `rubocop-govuk` and `scss_lint-govuk` been run? I'm wondering if they highlight any new violations that need to be fixed.

It should have run as a part of CI, I also ran it before I did the commit. It shouldn't create any new violations as it has the same rules as gout-lint.

PR #1437, opened 2019-11-11
Switch to use document_type_id from MetadataRevisions

AgaDufrat:
We are moving document_type_id column from Documents table to MetadataRevisions table so we can change the type of the documents.

This switches the app to use data from the new column in the MetadataRevisions table.
In this PR we:
- Move `document_type` method from Document to MetadataRevision
- Edition delegates document_type_id to Revisions and further to MetadataRevision
- Update factories
- Update tests that rely on document type id
- Update WhitehallImporter
- Ignore `document_type_id` in Document model 

[Trello card](https://trello.com/c/kwnmsIe9/1095-permit-content-publisher-to-have-documents-that-change-type)

Co-authored by @JonathanHallam, with help from @leenagupte

### DIFF
--- a/app/models/document.rb
+++ b/app/models/document.rb
@@ -31,6 +31,8 @@ class Document < ApplicationRecord
 
   delegate :topics, to: :document_topics
 
+  self.ignored_columns = %w(document_type_id)
+
   scope :with_current_edition, -> do
     join_tables = { current_edition: %i[revision status] }
     joins(join_tables).includes(join_tables)
--- a/app/models/document.rb
+++ b/app/models/document.rb
@@ -7,7 +7,7 @@
 #
 # This model is mutable
 class Document < ApplicationRecord
-  attr_readonly :content_id, :locale, :document_type_id
+  attr_readonly :content_id, :locale
 
   belongs_to :created_by, class_name: "User", optional: true
 
@@ -57,7 +57,6 @@ class Document < ApplicationRecord
     transaction do
       document = create!(content_id: content_id,
                          locale: locale,
-                         document_type_id: document_type_id,
                          created_by: user)
 
       document.tap do |d|
@@ -81,10 +80,6 @@ class Document < ApplicationRecord
 
   def to_param
     content_id + ":" + locale
-  end
-
-  def document_type
-    DocumentType.find(document_type_id)
   end
 
   def document_topics
--- a/lib/requirements/topic_checker.rb
+++ b/lib/requirements/topic_checker.rb
@@ -12,7 +12,7 @@ module Requirements
       issues = CheckerIssues.new
 
       begin
-        if document.document_type.topics && document.topics.none?
+        if document.current_edition.document_type.topics && document.topics.none?
           issues << Issue.new(:topics, :none)
         end
       rescue GdsApi::BaseError => e
--- a/lib/tasks/whitehall_importer.rb
+++ b/lib/tasks/whitehall_importer.rb
@@ -76,7 +76,6 @@ module Tasks
       Document.find_or_create_by!(
         content_id: whitehall_document["content_id"],
         locale: "en",
-        document_type_id: "news_story", ## TODO: This is a placeholder, to be removed after document type is moved to revision
         created_at: whitehall_document["created_at"],
         updated_at: whitehall_document["updated_at"],
         created_by_id: user_ids[first_author["whodunnit"]],
@@ -106,7 +105,7 @@ module Tasks
         metadata_revision: MetadataRevision.new(
           update_type: whitehall_edition["minor_change"] ? "minor" : "major",
           change_note: whitehall_edition["change_note"],
-          #document_type_id: whitehall_edition[document_type_key], # TODO: uncomment when document type is moved to revision
+          document_type_id: whitehall_edition[document_type_key],
         ),
         tags_revision: TagsRevision.new(
           tags: {
--- a/spec/factories/document_factory.rb
+++ b/spec/factories/document_factory.rb
@@ -4,7 +4,6 @@ FactoryBot.define do
   factory :document do
     content_id { SecureRandom.uuid }
     locale { I18n.available_locales.sample }
-    document_type_id { build(:document_type, path_prefix: "/prefix").id }
     association :created_by, factory: :user
 
     trait :with_live_edition do
--- a/spec/formats/news_story_spec.rb
+++ b/spec/formats/news_story_spec.rb
@@ -29,7 +29,7 @@ RSpec.feature "Create a news story", format: true do
     fill_in "revision[summary]", with: "A great summary"
 
     document = Document.first
-    base_path = document.document_type.path_prefix + "/a-great-title"
+    base_path = Edition.last.document_type.path_prefix + "/a-great-title"
     stub_publishing_api_has_lookups(base_path => document.content_id)
 
     click_on "Save"
@@ -39,7 +39,7 @@ RSpec.feature "Create a news story", format: true do
   def and_i_add_some_tags
     stub_publishing_api_has_links(role_appointment_links)
 
-    expect(Document.last.document_type.tags.count).to eq(5)
+    expect(Edition.last.document_type.tags.count).to eq(5)
     stub_publishing_api_has_linkables([linkable], document_type: "topical_event")
     stub_publishing_api_has_linkables([linkable], document_type: "world_location")
     stub_publishing_api_has_linkables([linkable], document_type: "organisation")
--- a/spec/formats/press_release_spec.rb
+++ b/spec/formats/press_release_spec.rb
@@ -29,7 +29,7 @@ RSpec.feature "Create a press release", format: true do
     fill_in "revision[summary]", with: "A great summary"
 
     document = Document.first
-    base_path = document.document_type.path_prefix + "/a-great-title"
+    base_path = Edition.last.document_type.path_prefix + "/a-great-title"
     stub_publishing_api_has_lookups(base_path => document.content_id)
 
     click_on "Save"
@@ -39,7 +39,7 @@ RSpec.feature "Create a press release", format: true do
   def and_i_add_some_tags
     stub_publishing_api_has_links(role_appointment_links)
 
-    expect(Document.last.document_type.tags.count).to eq(5)
+    expect(Edition.last.document_type.tags.count).to eq(5)
     stub_publishing_api_has_linkables([linkable], document_type: "topical_event")
     stub_publishing_api_has_linkables([linkable], document_type: "world_location")
     stub_publishing_api_has_linkables([linkable], document_type: "organisation")
--- a/spec/lib/requirements/topic_checker_spec.rb
+++ b/spec/lib/requirements/topic_checker_spec.rb
@@ -12,7 +12,7 @@ RSpec.describe Requirements::TopicChecker do
 
     context "when the Publishing API is available" do
       let(:document_type) { build :document_type, topics: true }
-      let(:document) { build :document, document_type_id: document_type.id }
+      let(:document) { build :document }
 
       before do
         stub_publishing_api_has_links(
@@ -32,7 +32,7 @@ RSpec.describe Requirements::TopicChecker do
 
     context "when the Publishing API is down" do
       let(:document_type) { build :document_type, topics: true }
-      let(:document) { build :document, document_type_id: document_type.id }
+      let(:document) { build :document }
 
       before do
         stub_publishing_api_isnt_available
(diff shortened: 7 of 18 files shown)


spec/lib/requirements/topic_checker_spec.rb
```diff
@@ -5,18 +5,20 @@
 
   describe "#pre_publish_issues" do
     it "returns no issues if there are none" do
-      edition = build :edition, :publishable
-      issues = Requirements::TopicChecker.new(edition.document).pre_publish_issues
+      document = create(:document, :with_current_edition)
+      issues = Requirements::TopicChecker.new(document).pre_publish_issues
```
leenagupte: I'm curious, why did this test need to change?

spec/models/edition_spec.rb
```diff
@@ -27,12 +27,14 @@
 
   describe ".create_initial" do
     let(:document) { build(:document) }
+    let(:document_type_id) { build(:document_type, path_prefix: "/prefix").id }
     let(:user) { build(:user) }
 
     it "creates a current edition" do
+      # byebug
```
leenagupte: Byebug!

JonathanHallam: ha, oops, my mistake!

spec/lib/requirements/topic_checker_spec.rb
AgaDufrat: It's because the document, which we need to pass to TopicChecker didn't have an edition associated with it and we were getting the following error:
```
Failure/Error: if document.editions.last.document_type.topics && document.topics.none?    
     NoMethodError:
       undefined method `document_type' for nil:NilClass
```

leenagupte: Would changing `build :edition` to `create` work? I think the edition factory creates a document if one doesn't exist, and it would match the set-up in the other tests.

app/models/document.rb
```diff
@@ -48,14 +48,13 @@ def self.find_by_param(content_id_and_locale)
   end
 
   def self.create_initial(content_id: SecureRandom.uuid,
-                          document_type_id:,
+                          document_type_id: nil,
```
1pretz1: Why does this now default to `nil`? It's my understanding that a document type is enforced to be set when creating a new document via https://github.com/alphagov/content-publisher/blob/master/app/interactors/new_document/create_interactor.rb#L27

spec/lib/requirements/topic_checker_spec.rb
JonathanHallam: this does work. Have adjusted :)

lib/requirements/topic_checker.rb
```diff
@@ -12,7 +12,7 @@ def pre_publish_issues(rescue_api_errors: true)
       issues = CheckerIssues.new
 
       begin
-        if document.document_type.topics && document.topics.none?
+        if document.editions.last.document_type.topics && document.topics.none?
```
1pretz1: Probably want to use `current_edition` here, so `document.current_edition.document_type.topics`.

spec/models/edition_spec.rb
```diff
@@ -27,12 +27,13 @@
 
   describe ".create_initial" do
     let(:document) { build(:document) }
+    let(:document_type_id) { build(:document_type, path_prefix: "/prefix").id }
```
1pretz1: Do we need `path_prefix: "/prefix"` specified for the document_type for these tests?

lib/requirements/topic_checker.rb
JonathanHallam: did not know this was a thing. Nice one, cheers

app/models/document.rb
JonathanHallam: So this is because the next PR will be a migration to remove it. What your link does (i think) is to ensure that the user selects a document type. This is still used but is passed to Editions down the line to MetadataRevisions rather than being held in Document. There will be a few small PR's to take this stuff out later today but for the time being this is really just a proof that everything works and we don't use or set `document.document_type` at the moment.

1pretz1: Yup I see that, but this still shouldn't default to nil here as it should never be nil. This method is only used from the `NewDocument::CreateInteractor` when creating a new document which will always have a document type, and if not the validation would fail before this code was even ran if that makes sense?

JonathanHallam: So does the `NewDocument::CreateInteractor` not just work with the params? Because if so then we may have to make some changes to it and the edition equivalent and remove it from the document one but as it stands this can be `nil` because as of this pr it's not being set.

spec/models/edition_spec.rb
JonathanHallam: yeah, stuff breaks without it

app/models/document.rb
```diff
@@ -7,7 +7,7 @@
 #
 # This model is mutable
 class Document < ApplicationRecord
-  attr_readonly :content_id, :locale, :document_type_id
+  attr_readonly :content_id, :locale
```
ChrisBAshton: I think this third commit should be fixed up to the first commit.

spec/factories/document_factory.rb
```diff
@@ -4,7 +4,6 @@
   factory :document do
     content_id { SecureRandom.uuid }
     locale { I18n.available_locales.sample }
-    document_type_id { build(:document_type, path_prefix: "/prefix").id }
```
ChrisBAshton: Tbh I think this commit could be fixed up to the first one too. Each commit should be in a state where its tests pass; I suspect the first few commits don't and are only now beginning to pass with these test tweaks.

I don't think it would be too much code, if that's what you're trying to avoid here, and in a case where you're mostly moving things from one class to another I think it would be clearer to have it all in one place 🤔

lib/requirements/topic_checker.rb
ChrisBAshton: Looks like this has slipped back into commit fa92eac41c9bc1b8fc8edfee7b59f2d1c65c95ba

spec/lib/requirements/topic_checker_spec.rb
```diff
@@ -25,26 +27,28 @@
...
       end
     end
 
     context "when the Publishing API is down" do
-      let(:document_type) { build :document_type, topics: true }
-      let(:document) { build :document }
+      let(:document_type) { create :document_type, topics: true }
+      let(:metadata_revision) { create :metadata_revision, document_type_id: document_type.id }
+      let(:revision) { create :revision, metadata_revision: metadata_revision }
+      let(:edition) { create :edition, revision: revision }
```
ChrisBAshton: Could these `let`s be moved to just after the `describe` block on line 6, and reused instead of duplicated by lines 14-17? The first test should continue to pass since you're defining an issueless `edition` 'inline'.

lib/tasks/whitehall_importer.rb
```diff
@@ -106,7 +105,7 @@ def create_edition(document, translation, whitehall_edition, edition_number)
         metadata_revision: MetadataRevision.new(
           update_type: whitehall_edition["minor_change"] ? "minor" : "major",
           change_note: whitehall_edition["change_note"],
-          #document_type_id: whitehall_edition[document_type_key], # TODO: uncomment when document type is moved to revision
```
ChrisBAshton: 🎉

spec/models/edition_spec.rb
```diff
@@ -27,12 +27,13 @@
...
     let(:user) { build(:user) }
 
     it "creates a current edition" do
       edition = Edition.create_initial(
         document: document,
-        document_type_id: document.document_type_id,
+        document_type_id: document_type_id,
```
ChrisBAshton: This last commit should be fixed up with the first ones too.

app/models/document.rb
```diff
@@ -31,6 +31,8 @@ class Document < ApplicationRecord
 
   delegate :topics, to: :document_topics
 
+  self.ignored_columns = %w(document_type_id)
```
ChrisBAshton: Nice! I don't quite see the benefit of it here though - why are we hiding the column? Why are we waiting until the next PR to remove it? Maybe a bit more in the commit message here would be useful.

JonathanHallam: Heya, have adjusted the commit message to better reflect why this is going on. Fully expect this code to be removed but for the time being it gives us a bit of peace of mind that our changes won't break stuff. Have kept it as a separate commit to make sure that we remember why this is happening / that it can be removed in the future.

app/models/document.rb
JonathanHallam: 👍

spec/models/edition_spec.rb
JonathanHallam: 👍

app/models/document.rb
ChrisBAshton: Thanks 👍

lib/requirements/topic_checker.rb
JonathanHallam: Have removed this for a 2nd time!

spec/formats/news_story_spec.rb
```diff
@@ -29,7 +29,7 @@ def and_i_fill_in_the_form_fields
     fill_in "revision[summary]", with: "A great summary"
 
     document = Document.first
-    base_path = document.document_type.path_prefix + "/a-great-title"
+    base_path = document.current_edition.document_type.path_prefix + "/a-great-title"
```
kevindew: You could probably switch this to be `Edition.first` and not need to look up document.

spec/formats/news_story_spec.rb
```diff
@@ -39,7 +39,7 @@ def and_i_fill_in_the_form_fields
   def and_i_add_some_tags
     stub_publishing_api_has_links(role_appointment_links)
 
-    expect(Document.last.document_type.tags.count).to eq(5)
+    expect(Document.last.current_edition.document_type.tags.count).to eq(5)
```
kevindew: Same comment about Document could be replaced with Edition. A good thing to sort here would be that we look up the first Document earlier in this test and last one here. These are predicated on the expectation that this is the only document in the db so we should probably stick with one. I'd suggest using `#last`.

spec/formats/press_release_spec.rb
```diff
@@ -39,7 +39,7 @@ def and_i_fill_in_the_form_fields
   def and_i_add_some_tags
     stub_publishing_api_has_links(role_appointment_links)
 
-    expect(Document.last.document_type.tags.count).to eq(5)
+    expect(Document.last.current_edition.document_type.tags.count).to eq(5)
```
kevindew: Same comments in this file as per news story

spec/lib/requirements/topic_checker_spec.rb
```diff
@@ -4,19 +4,21 @@
...
     end
 
     context "when the Publishing API is available" do
-      let(:document_type) { build :document_type, topics: true }
-      let(:document) { build :document, document_type_id: document_type.id }
-
       before do
         stub_publishing_api_has_links(
-          "content_id" => document.content_id,
+          "content_id" => edition.document.content_id,
```
kevindew:
```suggestion
          "content_id" => edition.content_id,
```

spec/lib/requirements/topic_checker_spec.rb
```diff
@@ -25,26 +27,23 @@
       end
 
       it "returns an issue if there are no topics" do
-        issues = Requirements::TopicChecker.new(document).pre_publish_issues
+        issues = Requirements::TopicChecker.new(edition.document).pre_publish_issues
```
kevindew: Not for this PR but I wonder if we this class should take an edition as an input rather than a document.

spec/formats/news_story_spec.rb
```diff
@@ -29,7 +29,7 @@ def and_i_fill_in_the_form_fields
     fill_in "revision[summary]", with: "A great summary"
 
     document = Document.first
-    base_path = document.document_type.path_prefix + "/a-great-title"
+    base_path = Edition.first.document_type.path_prefix + "/a-great-title"
```
kevindew:
```suggestion
    base_path = Edition.last.document_type.path_prefix + "/a-great-title"
```

spec/formats/press_release_spec.rb
```diff
@@ -29,7 +29,7 @@ def and_i_fill_in_the_form_fields
     fill_in "revision[summary]", with: "A great summary"
 
     document = Document.first
-    base_path = document.document_type.path_prefix + "/a-great-title"
+    base_path = Edition.first.document_type.path_prefix + "/a-great-title"
```
kevindew:
```suggestion
    base_path = Edition.last.document_type.path_prefix + "/a-great-title"
```

spec/formats/news_story_spec.rb
AgaDufrat: I misunderstood your first comment 🤦‍♀

kevindew: @AgaDufrat No problem - I can see how I could have given a mixed message there.

PR #1438, opened 2019-11-11
Import documents a withdrawn status into Content Publisher

leenagupte:
Trello: https://trello.com/c/8U7JPIHT

## What's changed and why?

Update `WhitehallImporter` to support documents that have a status of `withdrawn`.

The public explanation and the internal reason why the document was withdrawn have also been imported.

As Content Publisher supports undoing a withdrawal of content which involves using the published status to determine what to revert the content to (published or published_but_needs_2i_review), the published status details for the document have also been imported.

## Results from testing on integration

[Link to Whitehall export data](https://whitehall-admin.integration.publishing.service.gov.uk/government/admin/export/document/353919)

### Imported document in Content Publisher

```
#<Document id: 623, content_id: "2817f9a0-7c4a-4b48-a272-3c28251e79fe", locale: "en", document_type_id: "news_story", created_at: "2017-03-13 17:39:03", updated_at: "2019-10-29 10:28:38", created_by_id: 86, first_published_at: nil, imported_from: "whitehall">

#<Edition id: 886, number: 1, last_edited_at: "2019-11-11 12:55:48", created_at: "2017-03-13 17:39:03", updated_at: "2019-11-11 12:55:48", document_id: 623, created_by_id: 86, current: true, live: false, last_edited_by_id: 126, status_id: 2677, revision_id: 5332, revision_synced: true, access_limit_id: nil>

#<Revision id: 5332, created_by_id: nil, created_at: "2017-03-13 17:39:03", document_id: 623, lead_image_revision_id: nil, content_revision_id: 2970, metadata_revision_id: 1458, tags_revision_id: 819, preceded_by_id: nil, number: 1, imported: true>

Status.where(revision_at_creation_id: 5332)
=> #<ActiveRecord::Relation [#<Status id: 2677, state: "withdrawn", revision_at_creation_id: 5332, edition_id: 886, created_by_id: 126, created_at: "2019-10-29 10:28:38", updated_at: "2019-11-11 12:55:48", details_type: "Withdrawal", details_id: 3>, #<Status id: 2676, state: "published", revision_at_creation_id: 5332, edition_id: 886, created_by_id: 86, created_at: "2017-03-14 08:41:56", updated_at: "2019-11-11 12:55:48", details_type: nil, details_id: nil>]>

#<Withdrawal id: 3, public_explanation: "The [Dundee to London flight route was extended fo...", created_at: "2019-11-11 12:55:48", published_status_id: 2676, withdrawn_at: "2019-10-29 10:28:38">
```

Screenshot 2019-11-13 at 17 16 09

### DIFF
--- a/lib/tasks/whitehall_importer.rb
+++ b/lib/tasks/whitehall_importer.rb
@@ -165,6 +165,10 @@ module Tasks
 
       raise AbortImportError, "Cannot set a withdrawn status without a revision history event" unless revision_history_event
 
+      if whitehall_edition["unpublishing"].blank?
+        raise AbortImportError, "Cannot create withdrawn status without an unpublishing"
+      end
+
       edition.status = Status.new(
         state: whitehall_edition["state"],
         revision_at_creation: edition.revision,
--- a/lib/tasks/whitehall_importer.rb
+++ b/lib/tasks/whitehall_importer.rb
@@ -168,7 +168,11 @@ module Tasks
         revision_at_creation: edition.revision,
         created_by_id: user_ids[event["whodunnit"]],
         created_at: event["created_at"],
-        details: state_details(whitehall_edition, edition),
+        details: Withdrawal.new(
+          published_status: edition.status,
+          public_explanation: whitehall_edition["unpublishing"]["explanation"],
+          withdrawn_at: whitehall_edition["unpublishing"]["created_at"],
+        ),
       )
 
       edition.save!
@@ -189,16 +193,6 @@ module Tasks
       raise AbortImportError, "Edition is missing a #{state} event" unless event
 
       event
-    end
-
-    def state_details(whitehall_edition, edition)
-      return unless whitehall_edition["state"] == "withdrawn"
-
-      Withdrawal.new(
-        published_status: edition.status,
-        public_explanation: whitehall_edition["unpublishing"]["explanation"],
-        withdrawn_at: whitehall_edition["unpublishing"]["created_at"],
-      )
     end
 
     def state(whitehall_edition)
--- a/lib/tasks/whitehall_importer.rb
+++ b/lib/tasks/whitehall_importer.rb
@@ -256,7 +256,11 @@ module Tasks
 
       associations.map { |association| association["content_id"] }
     end
-  end
 
-  class AbortImportError < RuntimeError; end
+    class AbortImportError < RuntimeError
+      def initialize(message)
+        super(message)
+      end
+    end
+  end
 end
--- a/lib/tasks/whitehall_importer.rb
+++ b/lib/tasks/whitehall_importer.rb
@@ -144,19 +144,33 @@ module Tasks
     end
 
     def initial_status(whitehall_edition, revision)
+      revision_history_event = if whitehall_edition["state"] == "withdrawn"
+                                 published_author_history_event(whitehall_edition)
+                               else
+                                 whitehall_edition["revision_history"].last
+                               end
+
       Status.new(
         state: state(whitehall_edition),
         revision_at_creation: revision,
+        created_by_id: user_ids[revision_history_event["whodunnit"]],
       )
     end
 
     def set_withdrawn_status(whitehall_edition, edition)
+      revision_history_event = whitehall_edition["revision_history"].last
+
       edition.status = Status.new(
         state: whitehall_edition["state"],
         revision_at_creation: edition.revision,
+        created_by_id: user_ids[revision_history_event["whodunnit"]],
       )
 
       edition.save!
+    end
+
+    def published_author_history_event(whitehall_edition)
+      whitehall_edition["revision_history"].select { |h| h["state"] == "published" }.last
     end
 
     def state(whitehall_edition)
--- a/lib/tasks/whitehall_importer.rb
+++ b/lib/tasks/whitehall_importer.rb
@@ -156,6 +156,7 @@ module Tasks
         state: state(whitehall_edition),
         revision_at_creation: revision,
         created_by_id: user_ids[revision_history_event["whodunnit"]],
+        created_at: revision_history_event["created_at"],
       )
     end
 
@@ -168,6 +169,7 @@ module Tasks
         state: whitehall_edition["state"],
         revision_at_creation: edition.revision,
         created_by_id: user_ids[revision_history_event["whodunnit"]],
+        created_at: revision_history_event["created_at"],
       )
 
       edition.save!
--- a/lib/tasks/whitehall_importer.rb
+++ b/lib/tasks/whitehall_importer.rb
@@ -150,6 +150,8 @@ module Tasks
                                  current_author_history_event(whitehall_edition)
                                end
 
+      raise AbortImportError, "Cannot create an initial status without a revision history event" unless revision_history_event
+
       Status.new(
         state: state(whitehall_edition),
         revision_at_creation: revision,
@@ -159,6 +161,8 @@ module Tasks
 
     def set_withdrawn_status(whitehall_edition, edition)
       revision_history_event = current_author_history_event(whitehall_edition)
+
+      raise AbortImportError, "Cannot set a withdrawn status without a revision history event" unless revision_history_event
 
       edition.status = Status.new(
         state: whitehall_edition["state"],
--- a/lib/tasks/whitehall_importer.rb
+++ b/lib/tasks/whitehall_importer.rb
@@ -131,16 +131,20 @@ module Tasks
         number: edition_number,
         revision_synced: true,
         revision: revision,
-        status: Status.new(
-          state: state(whitehall_edition),
-          revision_at_creation: revision,
-        ),
+        status: initial_status(whitehall_edition, revision),
         current: whitehall_edition["id"] == most_recent_edition["id"],
         live: live?(whitehall_edition),
         created_at: whitehall_edition["created_at"],
         updated_at: whitehall_edition["updated_at"],
         created_by_id: user_ids[first_author["whodunnit"]],
         last_edited_by_id: user_ids[last_author["whodunnit"]],
+      )
+    end
+
+    def initial_status(whitehall_edition, revision)
+      Status.new(
+        state: state(whitehall_edition),
+        revision_at_creation: revision,
       )
     end
 
--- a/lib/tasks/whitehall_importer.rb
+++ b/lib/tasks/whitehall_importer.rb
@@ -21,7 +21,6 @@ module Tasks
       speech_type
     ].freeze
 
-
     def initialize(whitehall_document_id, whitehall_document)
       @whitehall_document_id = whitehall_document_id
       @whitehall_document = whitehall_document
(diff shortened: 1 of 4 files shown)


lib/tasks/whitehall_importer.rb
```diff
@@ -149,13 +151,25 @@ def initial_status(whitehall_edition, revision)
...
+    def set_current_status(whitehall_edition, edition)
+      edition.status = Status.new(
+        state: whitehall_edition["state"],
+        revision_at_creation: edition.revision,
+      )
+
+      edition.save!
+    end
+
+    def set_published_state_first?(state)
```
1pretz1: I know you were struggling with the name of this, how about just calling a spade a spade e.g `withdrawn?`. I think it's clear enough from the code by having `initial_status` and `set_current_status` that if the edition is withdrawn there is two statuses?

lib/tasks/whitehall_importer.rb
```diff
@@ -164,6 +174,10 @@ def set_published_state_first?(state)
       state == "withdrawn"
     end
 
+    def published_author(whitehall_edition)
+      whitehall_edition["revision_history"].select { |h| h["state"] == "published" }.first
```
1pretz1: I think you'd want `.last` instead of `.first` here as an edition can be published and the unpublished multiple times on the same edition.

lib/tasks/whitehall_importer.rb
```diff
@@ -145,16 +145,26 @@ def create_edition(document, translation, whitehall_edition, edition_number)
     end
 
     def initial_status(whitehall_edition, revision)
+      author = if set_published_state_first?(whitehall_edition["state"])
+                 published_author(whitehall_edition)
+               else
+                 whitehall_edition["revision_history"].last
```
1pretz1: I've done the same in my unpublishing PR using the last revision_history entry to identify the user who created the status, however this could be inaccurate as a document can be edited after it's had it's state changed by a different publisher. Not sure how much it matters though!

lib/tasks/whitehall_importer.rb
```diff
@@ -182,6 +184,12 @@ def published_author(whitehall_edition)
       whitehall_edition["revision_history"].select { |h| h["state"] == "published" }.first
     end
 
+    def valid_state?(whitehall_edition)
+      return false if whitehall_edition["state"] == "withdrawn" && !whitehall_edition["unpublishing"]
```
1pretz1: Shouldn't this just be `whitehall_edition["state"] == "withdrawn" && whitehall_edition["unpublishing"]` as a withdrawn edition should always have an unpublishing object? 

The logic here currently does this I think (took a while to wrap my head around 😆):
* Return false if withdrawn && non_unpublishing
* Return true if not_withdrawn && unpublishing || not_withdrawn && non_unpublishing || withdrawn && unpublishing

If this is correct it could be wrote like this: `!(whitehall_edition["state"] == "withdrawn" && !whitehall_edition["unpublishing"])`

lib/tasks/whitehall_importer.rb
```diff
@@ -190,6 +191,16 @@ def valid_state?(whitehall_edition)
       true
     end
 
+    def state_details(whitehall_edition, edition)
+      return nil unless whitehall_edition["state"] == "withdrawn"
```
1pretz1: Don't need the nil here as it's implicitly returned - `return unless whitehall_edition["state"] == "withdrawn"`. Although I'm not sure you even need the return as you're raising an error before this code gets run checking for an unpublishing object and a state of withdrawn.

lib/tasks/whitehall_importer.rb
leenagupte: This test will be used for other states too, e.g. "published", which should just fall through to the `true`, hence the slightly odd logic.

lib/tasks/whitehall_importer.rb
```diff
@@ -120,29 +127,85 @@ def create_edition(document, translation, whitehall_edition, edition_number)
...
+        status: initial_status(whitehall_edition, revision),
         current: whitehall_edition["id"] == most_recent_edition["id"],
         live: live?(whitehall_edition),
         created_at: whitehall_edition["created_at"],
         updated_at: whitehall_edition["updated_at"],
         created_by_id: user_ids[first_author["whodunnit"]],
         last_edited_by_id: user_ids[last_author["whodunnit"]],
       )
+
+      set_current_status(whitehall_edition, edition) if set_published_state_first?(whitehall_edition["state"])
```
brucebolt: It looks like this is setting the first and last status of an edition, but since an edition can be published/withdrawn multiple times, should we be storing every status change to capture the events between first and last?

leenagupte: I think I only need to store the most recent withdrawal details for the edition. I'll change this to get the `last` revision history.

lib/tasks/whitehall_importer.rb
leenagupte: I thought there might be more statuses that need to know about the last published state, but I can see this it not the case, so I'll rename, or perhaps remove, this method.

lib/tasks/whitehall_importer.rb
leenagupte: Hmmm. That's a good point. I'll change this get the `revision_history` for the edition state.

lib/tasks/whitehall_importer.rb
```diff
@@ -120,37 +127,93 @@ def create_edition(document, translation, whitehall_edition, edition_number)
...
+        status: initial_status(whitehall_edition, revision),
         current: whitehall_edition["id"] == most_recent_edition["id"],
         live: live?(whitehall_edition),
         created_at: whitehall_edition["created_at"],
         updated_at: whitehall_edition["updated_at"],
         created_by_id: user_ids[first_author["whodunnit"]],
         last_edited_by_id: user_ids[last_author["whodunnit"]],
       )
+
+      set_current_status(whitehall_edition, edition) if whitehall_edition["state"] == "withdrawn"
```
kevindew: I wonder if we can give this a clearer name than current_status? maybe we should go with set_withdrawn_status to make it more explicit. Without changes it's a bit confusing why a withdrawn edition gets a current_status but nothing else does.

lib/tasks/whitehall_importer.rb
```diff
@@ -120,37 +127,93 @@ def create_edition(document, translation, whitehall_edition, edition_number)
...
+    end
+
+    def initial_status(whitehall_edition, revision)
+      author = if whitehall_edition["state"] == "withdrawn"
+                 published_author(whitehall_edition)
+               else
+                 current_author(whitehall_edition)
+               end
```
kevindew: I feel like author might be a bit misleading as a variable name here as you'd expect that to be a user object whereas this returns a history entry.

lib/tasks/whitehall_importer.rb
```diff
@@ -120,37 +127,93 @@ def create_edition(document, translation, whitehall_edition, edition_number)
...
+
+      raise AbortImportError unless author
```
kevindew: What's the reason for aborting here? we probably should have some text that explains why as an argument into the exception.

I'm a little confused as the user could still not be existent on line 159

lib/tasks/whitehall_importer.rb
```diff
@@ -120,37 +127,93 @@ def create_edition(document, translation, whitehall_edition, edition_number)
...
+        revision_at_creation: revision,
+        created_by_id: user_ids[author["whodunnit"]],
+        created_at: author["created_at"],
+      )
+    end
+
+    def set_current_status(whitehall_edition, edition)
+      author = current_author(whitehall_edition)
+
+      raise AbortImportError unless valid_state?(whitehall_edition)
```
kevindew: I think this should also have a message to explain why it's aborted. I'm a bit confused why this doesn't check author where the other method does?

lib/tasks/whitehall_importer.rb
```diff
@@ -120,37 +127,93 @@ def create_edition(document, translation, whitehall_edition, edition_number)
...
+
+    def published_author(whitehall_edition)
+      whitehall_edition["revision_history"].select { |h| h["state"] == "published" }.last
+    end
+
+    def valid_state?(whitehall_edition)
+      return false if whitehall_edition["state"] == "withdrawn" && !whitehall_edition["unpublishing"]
+
+      true
+    end
```
kevindew: I find this method a bit strange as it's entirely withdrawn centric rather than actually covering whether a state is valid. Though I think if you go with my suggestion of `set_withdrawn_status` method you could move that into that and not need this specific method.

spec/tasks/whitehall_importer_spec.rb
```diff
@@ -118,6 +138,24 @@
     expect { importer.import }.to raise_error(Tasks::AbortImportError)
   end
 
+  it "raises AbortImportError when revision history is missing for state" do
+    import_data["editions"][0]["state"] = "published"
+    importer = Tasks::WhitehallImporter.new(123, import_data)
+
+    expect { importer.import }.to raise_error(Tasks::AbortImportError)
```
kevindew: It looks like this class isn't namespaced within WhitehallImporter. We should have it as `Tasks::WhitehallImporter::AbortImportError` - I must have missed that the first time this was setup. Otherwise there is nothing to tie this to `WhitehallImporter`

lib/tasks/whitehall_importer.rb
```diff
@@ -120,37 +127,93 @@ def create_edition(document, translation, whitehall_edition, edition_number)
...
+      when "published", "withdrawn"
         whitehall_edition["force_published"] ? "published_but_needs_2i" : "published"
       else
         "submitted_for_review"
       end
     end
 
     def live?(whitehall_edition)
-      whitehall_edition["state"] == "published"
+      whitehall_edition["state"].in?(%w(published withdrawn))
```
kevindew: 👍

spec/fixtures/whitehall_export_with_one_withdrawn_edition.json
```diff
@@ -0,0 +1,112 @@
...
+        }
+      ],
+      "unpublishing": {
+        "id": 1,
+        "explanation": "User facing explanation",
+        "alternative_url": "",
+        "created_at": "2019-11-07T16:12:52.000+00:00",
+        "updated_at": "2019-11-07T16:12:52.000+00:00",
+        "redirect": false,
+        "unpublishing_reason": "Published in error"
```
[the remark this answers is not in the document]
leenagupte: I'll change it to "No longer current government policy/activity"

lib/tasks/whitehall_importer.rb
leenagupte: I copied the name from here: https://github.com/alphagov/content-publisher/blob/master/lib/tasks/whitehall_importer.rb#L87 but I'm happy to change it.

lib/tasks/whitehall_importer.rb
leenagupte: If there isn't an author you'd never get as far as calling this method. So but this point there will always be at least one author, it's just that there might only be one and current_author might be the same as the published_author if there's weird data in the export.

lib/tasks/whitehall_importer.rb
1pretz1: Sorry to chime in but an edition can't be published and withdrawn multiple time's in whitehall without creating a new edition each time it's published (unwithdrawn) ? An edition can however be unpublished (published in error or consolidated) and published multiple times on the same edition but not withdrawn to my understanding.

lib/tasks/whitehall_importer.rb
1pretz1: I've altered this to revision_history_event, and the author method name's to have a suffix of `_history_event` which makes the intent of these methods a bit clearer.

lib/tasks/whitehall_importer.rb
1pretz1: Added a message and check for an unpublishing.

kevindew:
Thanks Pete, haven't had chance for a few review but looks like most bits are covered. One thing I notice isn't is that we still have AbortImportErrors without a message. 

Is there a way we can adjust the class to error without a message? I feel a bit like a broken record pointing it out so might be worth making it a pre-requisite of the class.

1pretz1:
Thanks @kevindew, have added a message to the raises that were missing them and adding an extension to the AbortImportErrors class as per your request! 😄

kevindew:
@leenagupte @1pretz1 I added a couple of quick commits to this to clean up some areas that still seemed somewhat confused to me. Let me know if you're happy with them and then I think this is ready to merge.

lib/tasks/whitehall_importer.rb
```diff
@@ -169,7 +169,11 @@ def set_withdrawn_status(whitehall_edition, edition)
         revision_at_creation: edition.revision,
         created_by_id: user_ids[event["whodunnit"]],
         created_at: event["created_at"],
-        details: state_details(whitehall_edition, edition),
+        details: Withdrawal.new(
```
leenagupte: In @1pretz1's [PR](https://github.com/alphagov/content-publisher/pull/1439/commits/2badfcc1c6a2425259a2bc52e0236cdeb03d3ba3#diff-713c586411d3e8a35468b374fe29c4a8R205) a `Removal` is added to details, so we may want to leave this as a separate method.

leenagupte:
@kevindew I had one comment in the last commit, but the new state history methods look good.

lib/tasks/whitehall_importer.rb
kevindew: I don't think we should look to making that a big method that handles all of those bits, it's only going to get complex.

PR #1439, opened 2019-11-12
Import unpublishing

1pretz1:
We want to have the ability to import a document that has been removed from GOV.UK. A complication of this is that Whitehall sets unpublished documents to a draft status. So we need to determine whether a user has modified the draft at all for us to know whether it should be an edition with "removed", "draft" or "submitted_for_2i" status.

There is also some quite bad data rot in whitehall to watch out for, like there is a case where a published edition has got an unpublished object attached to it. Some withdrawn documents have got the wrong unpublishing reason attached due to the unpublishing reason ids changing in the code at some point. Also there are some recent documents that are superseded but have an unpublishing object (don't know how you do this through the UI) 😢. 

More info within the commits ->

Trello:
https://trello.com/c/nDK6MZLW/1172-import-an-unpublished-status-into-content-publisher

Importer ran with different scenarios below:

## Removed unpublishing

Screen Shot 2019-11-12 at 12 23 30
Screen Shot 2019-11-12 at 12 23 16

## Removed and edited unpublishing

Screen Shot 2019-11-12 at 12 33 42
Screen Shot 2019-11-12 at 12 34 42

### DIFF
--- a/lib/tasks/whitehall_importer.rb
+++ b/lib/tasks/whitehall_importer.rb
@@ -187,9 +187,12 @@ module Tasks
     end
 
     def status(whitehall_edition, revision)
+      author_id = whitehall_edition["revision_history"].last["whodunnit"]
+
       Status.new(
         state: state(whitehall_edition),
         revision_at_creation: revision,
+        created_by_id: user_ids[author_id],
         details: details(whitehall_edition),
       )
     end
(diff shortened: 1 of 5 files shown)


lib/tasks/whitehall_importer.rb
```diff
@@ -180,10 +180,7 @@ def create_edition(document, translation, whitehall_edition, edition_number)
         number: edition_number,
         revision_synced: true,
         revision: revision,
-        status: Status.new(
-          state: state(whitehall_edition),
-          revision_at_creation: revision,
-        ),
+        status: status(whitehall_edition, revision),
```
brucebolt: Probably worth liaising with @leenagupte on this change, since she's introduced a method here with a different name: https://github.com/alphagov/content-publisher/pull/1438/files#diff-713c586411d3e8a35468b374fe29c4a8R135

PR #1445, opened 2019-11-15
Remove document_type_id from Document

JonathanHallam:
Trello - https://trello.com/c/kwnmsIe9/1095-permit-content-publisher-to-have-documents-that-change-type

This is among the last pieces of work in a stream set to move document_type_id from Documents to MetadataRevisions. It will allow us to change document types on documents within content-publisher. This PR removes document_type_id from Documents now that all of its functionality has been moves to MetadataRevisions.

### DIFF
--- a/db/migrate/20191115103058_remove_document_type_id_from_documents.rb
+++ b/db/migrate/20191115103058_remove_document_type_id_from_documents.rb
@@ -1,0 +1,7 @@
+# frozen_string_literal: true
+
+class RemoveDocumentTypeIdFromDocuments < ActiveRecord::Migration[5.2]
+  def change
+    remove_column :documents, :document_type_id, :string
+  end
+end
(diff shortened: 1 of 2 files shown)


db/migrate/20191115103058_remove_document_type_id_from_documents.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+class RemoveDocumentTypeIdFromDocuments < ActiveRecord::Migration[5.2]
+  def up
```
brucebolt: The `up`/`down` methods could be replaced by a single `change` method, e.g.

```
def change
  remove_column :documents, :document_type_id, :string
end
```

Since the column type is defined in the `remove_column`, the migration is reversible.

JonathanHallam: ah nice, hadn't realised that. Will update, thanks :)

PR #1448, opened 2019-11-15
Import images from Whitehall

ChrisBAshton:
This PR imports image data from Whitehall. It does not do the following (links to their Trello cards):

1) [deal with attachments](https://trello.com/c/4FF2oprK/1182-import-attachments-into-content-publisher)
2) [change Govspeak markdown to use content publisher image syntax](https://trello.com/c/ubTM9ylI/1176-use-imported-images-and-attachments-in-content-publisher)

- [x] Should create the necessary ActiveRecord image references from an edition successfully
- [x] Should cope with all of the following: jpg, jpeg, gif, png
- [x] Should abort if an image is not found
- [x] Should cope with empty caption
- [x] Should rename 2nd image if two images share the same filename
- [x] Should import image(s) added across multiple editions. We've agreed for now we're going to blindly import images even if they're duplicated across multiple editions, just treating them as new image revisions every time.
- [x] Abort import when SVG
- [x] Abort if main image is smaller than 960x640

------

Trello card: https://trello.com/c/jzV3scso/1093-import-images-and-attachments-to-content-publisher

### DIFF
--- a/spec/lib/whitehall_importer/create_image_revision_spec.rb
+++ b/spec/lib/whitehall_importer/create_image_revision_spec.rb
@@ -28,5 +28,26 @@ RSpec.describe WhitehallImporter::CreateImageRevision do
         )
       end
     end
+
+    context "Image is wrong type" do
+      let(:whitehall_image) do
+        build(:whitehall_export_image, filename: "vector.svg", fixture_file: "coffee.svg")
+      end
+
+      it "should raise a WhitehallImporter::AbortImportError" do
+        expect { described_class.call(whitehall_image) }.to raise_error(
+          WhitehallImporter::AbortImportError,
+          I18n.t!("requirements.image_upload.unsupported_type.form_message"),
+        )
+      end
+
+      it "should be passed through the ImageUploadChecker class" do
+        begin
+          expect(Requirements::ImageUploadChecker).to receive(:new).and_call_original
+          described_class.call(whitehall_image)
+        rescue WhitehallImporter::AbortImportError
+        end
+      end
+    end
   end
 end
--- a/spec/lib/whitehall_importer/create_image_revision_spec.rb
+++ b/spec/lib/whitehall_importer/create_image_revision_spec.rb
@@ -1,0 +1,16 @@
+# frozen_string_literal: true
+
+RSpec.describe WhitehallImporter::CreateImageRevision do
+  describe "#call" do
+    let(:whitehall_image) { build(:whitehall_export_image) }
+
+    it "should create an Image::Revision when a valid image is provided" do
+      image_revision = nil
+      expect { image_revision = described_class.call(whitehall_image) }
+        .to change { Image::Revision.count }.by(1)
+      expect(image_revision.caption).to eq(whitehall_image["caption"])
+      expect(image_revision.alt_text).to eq(whitehall_image["alt_text"])
+      expect(image_revision.filename).to eq("valid-image.jpg")
+    end
+  end
+end
--- a/spec/lib/whitehall_importer/create_image_revision_spec.rb
+++ b/spec/lib/whitehall_importer/create_image_revision_spec.rb
@@ -12,5 +12,21 @@ RSpec.describe WhitehallImporter::CreateImageRevision do
       expect(image_revision.alt_text).to eq(whitehall_image["alt_text"])
       expect(image_revision.filename).to eq("valid-image.jpg")
     end
+
+    context "Image is not available" do
+      let(:image_url) { "https://assets.publishing.service.gov.uk/government/uploads/404ing-image.jpg" }
+      let(:whitehall_image) do
+        whitehall_image = build(:whitehall_export_image, url: image_url)
+        stub_request(:get, image_url).to_return(status: 404)
+        whitehall_image
+      end
+
+      it "should raise a WhitehallImporter::AbortImportError" do
+        expect { described_class.call(whitehall_image) }.to raise_error(
+          WhitehallImporter::AbortImportError,
+          "Image does not exist: #{image_url}",
+        )
+      end
+    end
   end
 end
--- a/spec/lib/whitehall_importer/create_image_revision_spec.rb
+++ b/spec/lib/whitehall_importer/create_image_revision_spec.rb
@@ -56,5 +56,16 @@ RSpec.describe WhitehallImporter::CreateImageRevision do
         )
       end
     end
+
+    context "Original image filename is URL-unfriendly" do
+      let(:whitehall_image) do
+        build(:whitehall_export_image, filename: "Whitehall--Asset_-image.jpg")
+      end
+
+      it "should rename the file to something URL-friendly" do
+        described_class.call(whitehall_image)
+        expect(Image::BlobRevision.last.filename).to eq("whitehall-asset_-image.jpg")
+      end
+    end
   end
 end
--- a/spec/lib/whitehall_importer/create_image_revision_spec.rb
+++ b/spec/lib/whitehall_importer/create_image_revision_spec.rb
@@ -34,19 +34,26 @@ RSpec.describe WhitehallImporter::CreateImageRevision do
         build(:whitehall_export_image, filename: "vector.svg", fixture_file: "coffee.svg")
       end
 
-      it "should raise a WhitehallImporter::AbortImportError" do
+      it "should pass through ImageUploadChecker and raise a WhitehallImporter::AbortImportError" do
+        expect(Requirements::ImageUploadChecker).to receive(:new).and_call_original
         expect { described_class.call(whitehall_image) }.to raise_error(
           WhitehallImporter::AbortImportError,
           I18n.t!("requirements.image_upload.unsupported_type.form_message"),
         )
       end
+    end
 
-      it "should be passed through the ImageUploadChecker class" do
-        begin
-          expect(Requirements::ImageUploadChecker).to receive(:new).and_call_original
-          described_class.call(whitehall_image)
-        rescue WhitehallImporter::AbortImportError
-        end
+    context "Image is too small" do
+      let(:whitehall_image) do
+        build(:whitehall_export_image, fixture_file: "100x100.jpg")
+      end
+
+      it "should pass through ImageNormaliser and raise a WhitehallImporter::AbortImportError" do
+        expect(ImageNormaliser).to receive(:new).and_call_original
+        expect { described_class.call(whitehall_image) }.to raise_error(
+          WhitehallImporter::AbortImportError,
+          I18n.t!("requirements.image_upload.too_small.form_message", width: 960, height: 640),
+        )
       end
     end
   end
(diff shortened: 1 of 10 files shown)


spec/tasks/whitehall_importer_spec/create_image_revision_spec.rb
```diff
@@ -0,0 +1,105 @@
...
+
+  context "Image is wrong type" do
+    let(:img_url) { "https://assets.publishing.service.gov.uk/government/uploads/vector.svg" }
+    let(:img_file) { "coffee.svg" }
+
+    it "should pass through ImageUploadChecker and raise an AbortImageError" do
+      expect(Requirements::ImageUploadChecker).to receive(:new).and_call_original
+      expect { Tasks::WhitehallImporter::CreateImageRevision.call(whitehall_image) }.to raise_error(
+        AbortImageError,
+        "Select a jpg, png or gif image file",
```
kevindew: It'd be more consistent with the existing test setup to use: `I18n.t!("requirements.image_upload.unsupported_type.form_message")` as we have a consistency for not using strings from the locale file directly.

spec/tasks/whitehall_importer_spec/create_image_revision_spec.rb
```diff
@@ -0,0 +1,105 @@
...
+  end
+
+  context "Image is too small" do
+    let(:img_file) { "100x100.jpg" }
+
+    it "should pass through ImageNormaliser and raise an AbortImageError" do
+      expect(ImageNormaliser).to receive(:new).and_call_original
+      expect { Tasks::WhitehallImporter::CreateImageRevision.call(whitehall_image) }.to raise_error(
+        AbortImageError,
+        "Select an image at least 960 pixels wide by 640 pixels high",
```
kevindew: Similar comment to ☝️

spec/tasks/whitehall_importer_spec.rb
```diff
@@ -5,6 +5,13 @@
 
   let(:import_data) { whitehall_export_with_one_edition }
 
+  before do
+    binary_image = File.open(File.join(fixtures_path, "files", "960x640.jpg"), "rb").read
```
kevindew: It looks like we can simplify this (and similar) to `file_fixture("960x640.jpg").read` - [`file_fixture`](https://relishapp.com/rspec/rspec-rails/v/3-8/docs/file-fixture) returns a path

spec/tasks/whitehall_importer_spec.rb
```diff
@@ -5,6 +5,13 @@
...
+    stub_request(:get, "https://assets.publishing.service.gov.uk/government/uploads/valid-image.jpg").to_return(status: 200, body: binary_image)
+    stub_request(:get, "https://assets.publishing.service.gov.uk/government/uploads/another-valid-image.jpg").to_return(status: 200, body: binary_image)
+    stub_request(:get, "https://assets.publishing.service.gov.uk/government/uploads/foo/valid-image.jpg").to_return(status: 200, body: binary_image)
+  end
```
kevindew: This block seems to be quite impactful to performance running before all the tests. I saw it increase from 1 second to 17 seconds execution time. Maybe we can just do this for an image test?

ChrisBAshton: On it 💪

1pretz1:
Taking over this card from Chris while he's away. I've made the changes @kevindew requested but will hold off on merging until https://trello.com/c/vNd3wt2Y/1186-refactor-whitehallimporter is complete due to it blocking a few cards and would want to get it out as soon as possible.

Chris has wrote this in a way which conforms to the new subclassing approach which will make it super easy to incorporate into the refactored code.

spec/tasks/whitehall_importer_spec.rb
kevindew: Just marked this as unresolved as I'd expect we need to move this code into the images context.

kevindew:
Thanks @1pretz1 sounds good, we'll give this another look once refactor is through.

spec/lib/whitehall_importer/create_image_revision_spec.rb
```diff
@@ -51,4 +51,28 @@
...
+    let(:svg_file) { File.open(File.join(fixtures_path, "files", "coffee.svg"), "rb").read }
+
+    before do
+      stub_request(:get, img_url).to_return(status: 200, body: svg_file)
+    end
+
+    it "should raise an AbortImageError" do
+      expect { Tasks::WhitehallImporter::CreateImageRevision.call(whitehall_image) }.to raise_error(
+        AbortImageError,
+        "Select a jpg, png or gif image file",
```
ChrisBAshton: TODO fixup ` I18n.t!("requirements.image_upload.unsupported_type.form_message"),`

spec/lib/whitehall_importer/create_image_revision_spec.rb
```diff
@@ -39,7 +45,7 @@
     end
 
     it "should raise an AbortImageError" do
-      expect { Tasks::WhitehallImporter::CreateImageRevision.call(whitehall_image) }.to raise_error(
```
ChrisBAshton: TODO fixup

spec/lib/whitehall_importer/create_image_revision_spec.rb
```diff
@@ -0,0 +1,110 @@
...
+    context "A valid image is provided" do
+      before do
+        stub_request(:get, img_url).to_return(status: 200, body: file_fixture("960x640.jpg"))
+      end
+
+      it "should create an Image::MetadataRevision that preserves the alt text and caption" do
+        described_class.call(whitehall_image)
+        expect(Image::MetadataRevision.count).to eq(1)
+        expect(Image::MetadataRevision.first.caption).to eq(whitehall_image["caption"])
+        expect(Image::MetadataRevision.first.alt_text).to eq(whitehall_image["alt_text"])
```
kevindew: Minor and just my personal opinion. I personally avoid tests that expect the db to empty as these are with `.count` and `.first` as these end up being coupled to the db state which can make the tests fragile.

My order of preference with these is:

1) ideally test code either changes or returns something, then no risks at all about initial state or any concurrency.
2) test last item in the db, which is vulnerable to concurrency but doesn't require initial state. 

For these tests I think your most idiomatic situation would be testing the image revision that is returned rather than looking one up from the db.

spec/lib/whitehall_importer/create_revision_spec.rb
```diff
@@ -151,4 +193,19 @@
...
+    stub_request(:get, image_url).to_return(status: 200, body: @binary_image)
+    {
+      "id": 194072,
+      "alt_text": "Alt text for image",
+      "caption": "This is a caption",
+      "created_at": "2018-11-30T05:08:56.000+00:00",
+      "updated_at": "2018-11-30T05:19:53.000+00:00",
+      "variants": {},
+      "url": image_url,
+    }.with_indifferent_access
```
kevindew: I feel that this would be more consistent with the recent code if it was built using a factory bot factory.

That factory could also set up the webmock stubbing for the file.

kevindew: We also could maybe use the block argument of `to_return` so it only reads the file if the mock is called rather than needing to call it each time the mock is set up (I'm assuming this is why we've set up the instance variable)

spec/lib/whitehall_importer/create_image_revision_spec.rb
```diff
@@ -0,0 +1,110 @@
...
+    let(:whitehall_image) do
+      { "url": img_url, "caption": "Caption", "alt_text": "Alt text" }.with_indifferent_access
+    end
+
+    before do
+      stub_request(:get, img_url).to_return(
+        status: 200,
+        body: file_fixture(img_file).read,
+      )
+    end
```
kevindew: I've found this a bit surprising as we then have a similar `stub_request` in the before of the next context.

Though I've made a suggestion in the `create_image` method of the other test file that could resolve this.

spec/lib/whitehall_importer/create_image_revision_spec.rb
```diff
@@ -0,0 +1,110 @@
...
+      end
+
+      it "should not create any revision instances" do
+        begin
+          described_class.call(whitehall_image)
+        rescue WhitehallImporter::AbortImportError
+          expect(Image::MetadataRevision.count).to eq(0)
+          expect(Image::BlobRevision.count).to eq(0)
+          expect(Image::Revision.count).to eq(0)
+        end
```
kevindew: I don't think we're too concerned about whether these are created or not. We expect AbortImportError to break a transaction and that will involve some data being persisted and discarded. So although I can't see a situation where we'd expect these to actually exist I also don't think we have reason to care about whether any were or not, just the exception being raised should be sufficient.

ChrisBAshton: Agreed that this test can be removed, as really we would want one for every bit of the import that creates something, not just the image logic

I do think it's worth having a higher level test that ensures we're wrapping the entire import in a Transaction - one for a small future PR?

spec/lib/whitehall_importer/create_image_revision_spec.rb
```diff
@@ -0,0 +1,110 @@
...
+        described_class.call(whitehall_image)
+        expect(Image::BlobRevision.count).to eq(1)
+        expect(Image::BlobRevision.first.filename).to eq("valid-image.jpg")
+      end
+
+      it "should create an Image::Revision that references the Metadata and Blob revisions" do
+        described_class.call(whitehall_image)
+        expect(Image::Revision.count).to eq(1)
+        expect(Image::Revision.first.blob_revision).to eq(Image::BlobRevision.first)
+        expect(Image::Revision.first.metadata_revision).to eq(Image::MetadataRevision.first)
```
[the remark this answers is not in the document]
ChrisBAshton: Agreed - wasn't quite sure how to test the creation of multiple classes, thought it was better safe than sorry. But your suggested test is much nicer and still gives reasonable confidence the import has worked 👍 will replace.

spec/lib/whitehall_importer/create_revision_spec.rb
```diff
@@ -31,6 +31,48 @@
...
+        )
+        revision = described_class.call(document, whitehall_edition)
+
+        expect(Image.count).to eq(2)
+        expect(Image::BlobRevision.count).to eq(2)
+        expect(Image::MetadataRevision.count).to eq(2)
+        expect(Image::Revision.count).to eq(2)
+        expect(revision.image_revisions.first.blob_revision.filename).to eq("foo.jpg")
+        expect(revision.image_revisions.last.blob_revision.filename).to eq("bar.jpg")
+      end
```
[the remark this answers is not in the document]
ChrisBAshton: Happy to remove on the basis that we keep the `it "ensures that every image filename is unique"` test (which I think is worth keeping as it tests the implementation of `create_image_revisions` in create_revision.rb). That test indirectly tests that multiple images are imported 👍

spec/lib/whitehall_importer/create_revision_spec.rb
ChrisBAshton: My first instinct was to try to make a factory for it, but as it is representing JSON from the Whitehall import rather than representing an instance of a model in Content Publisher, it didn't feel appropriate.

I'll take another look at factories now I know it's a sound idea!

spec/lib/whitehall_importer/create_image_revision_spec.rb
ChrisBAshton: Have removed references to `.count` and `.first` 👍

PR #1450, opened 2019-11-18
Import AccessLimit from Whitehall

AgaDufrat:
Whitehall sets access limiting by setting `access_limited: true` on Edition. Whereas in Content Publisher, we have AccessLimit model which points to an edition.

When importing AccessLimit:
- created_at is set to the time the edition was last edited
- revision is the last one for this content
- there is no user attribution
- limit_type is set to "tagged_organisations" as this is the only option in Whitehall

The only statuses considered for access_limit are draft, rejected, submitted. In both apps access_limit is cleared once the document is published ([Content Publisher](https://github.com/alphagov/content-publisher/blob/master/app/services/publish_service.rb#L50 ), [Whitehall](https://github.com/alphagov/whitehall/blob/13ccb069d3df45371c7c7390c292968d052c7ea5/app/services/edition_publisher.rb#L32))

[Trello card](https://trello.com/b/mn2MqH3M/publishing-workflow-doing-q4-2019-to)

### DIFF
--- a/lib/tasks/whitehall_importer.rb
+++ b/lib/tasks/whitehall_importer.rb
@@ -138,6 +138,7 @@ module Tasks
         last_edited_by_id: user_ids[last_event["whodunnit"]],
       )
 
+      set_access_limit(whitehall_edition["created_at"], edition, revision) if whitehall_edition["access_limited"]
       set_withdrawn_status(whitehall_edition, edition) if whitehall_edition["state"] == "withdrawn"
     end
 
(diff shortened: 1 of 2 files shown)


lib/tasks/whitehall_importer.rb
```diff
@@ -138,6 +138,8 @@ def create_edition(document, translation, whitehall_edition, edition_number)
         last_edited_by_id: user_ids[last_event["whodunnit"]],
       )
 
+      create_access_limit(whitehall_edition["created_at"], edition, revision) if whitehall_edition["access_limit"]
```
brucebolt:
```suggestion
      create_access_limit(whitehall_edition["created_at"], edition, revision) if whitehall_edition["access_limited"]
```

PR #1453, opened 2019-11-19
Refactor whitehall importer

leenagupte:
Trello: https://trello.com/c/vNd3wt2Y

## What's changed and why?
The WhitehallImporter in content-publisher has gotten too large which was making it hard to read, test and maintain, the tipping point was adding the withdrawal logic around importing. It has been decided to split this class into sub classes to make the code more maintainable.


lib/whitehall_importer/create_edition.rb
```diff
@@ -0,0 +1,112 @@
...
+
+  def create_withdrawn_edition
+    edition = create_edition("published")
+    set_withdrawn_status(edition)
+
+    edition
+  end
+
+  def create_edition(initial_state = whitehall_edition["state"])
+    create_event = create_history_event
```
kevindew: Maybe rename the `create_history_event` method, since it sounds like it's creating a history event. But again not necessary for initial merge.

lib/whitehall_importer/create_revision.rb
```diff
@@ -0,0 +1,99 @@
+# frozen_string_literal: true
+
+class WhitehallImporter::CreateRevision
+  attr_reader :document, :whitehall_edition, :translation
+
+  SUPPORTED_DOCUMENT_TYPES = %w(news_story press_release).freeze
+  DOCUMENT_SUB_TYPES = %w[
+      news_article_type
```
kevindew: Indentation looks to be double here.

lib/whitehall_importer/create_edition.rb
```diff
@@ -0,0 +1,112 @@
...
+  end
+
+  def create_withdrawn_edition
+    edition = create_edition("published")
+    set_withdrawn_status(edition)
+
+    edition
+  end
+
+  def create_edition(initial_state = whitehall_edition["state"])
```
kevindew: Do we need the terminology of initial_state?

Maybe it actually makes sense to inject a built status into it?

kevindew: To inject status you'd also need to have created revision first, but I also think that might be a good thing for simplifying this method.

lib/whitehall_importer/create_edition.rb
```diff
@@ -0,0 +1,112 @@
...
+    Edition.create!(
+      document: document,
+      number: edition_number,
+      revision_synced: true,
+      revision: revision,
+      status: WhitehallImporter::CreateStatus.new(
+        revision, whitehall_edition, user_ids, status: initial_state
+      ).call,
+      current: whitehall_edition["id"] == most_recent_edition["id"],
+      live: live?,
```
kevindew: might want to inline this method

lib/whitehall_importer/create_status.rb
```diff
@@ -0,0 +1,79 @@
...
+    when "published"
+      whitehall_edition["force_published"] ? "published_but_needs_2i" : "published"
+    when "withdrawn" then "withdrawn"
+    else
+      "submitted_for_review"
+    end
+  end
+
+  def details
+    if edition && status == "withdrawn"
```
kevindew: This should probably be a guard clause

lib/whitehall_importer/create_status.rb
```diff
@@ -0,0 +1,79 @@
...
+      created_by_id: user_ids[event["whodunnit"]],
+      created_at: event["created_at"],
+      details: details,
+    )
+  end
+
+private
+
+  def check_supported_state
+    raise WhitehallImporter::AbortImportError, "Edition has an unsupported state" unless valid_state?
```
kevindew: You shouldn't need the WhitehallImporter prefix on these.

lib/whitehall_importer/create_status.rb
```diff
@@ -0,0 +1,79 @@
...
+    when "superseded" then "superseded"
+    when "published"
+      whitehall_edition["force_published"] ? "published_but_needs_2i" : "published"
+    when "withdrawn" then "withdrawn"
+    else
+      "submitted_for_review"
+    end
+  end
+
+  def details
```
kevindew: This method makes me a little concerned that it could build up a lot of logic in it, it might be an idea that is uses a case statement and then calls a method of withdrawn_details to detract from adding to it.

leenagupte: That was my thought too, but it felt a bit odd to do
```
def details
  withdrawal_details
end
```

lib/whitehall_importer/create_status.rb
leenagupte: I think this will need to be re-thought when the unpublishing details are added as I'm not sure that will need to use `edition`.  I think this conditional should probably become a guard clause in the future `withdrawal_details` method.

PR #1456, opened 2019-11-19
Import governments from Whitehall

kevindew:
Trello: https://trello.com/c/WKUoYhxb/1189-spike-into-defining-the-history-mode-feature

This hardcodes government data as config for Content Publisher to have
access to this data. This is intended to be a short-term approach until
governments are exposed to the Publishing API and Content Publisher
consumes them. The effects of duplicating this data is that any changes
to governments will require the config for Content Publisher to be
re-generated.

I decided to pull the data for this directly from the Whitehall database
rather than using the API at https://www.gov.uk/api/governments. The API
seems to only use slugs as identifying information and these slugs seem
to change when governments change to represent the end date, this didn't
strike me as a reliable field we could use and felt the underlying
Whitehall id would be better.

A quirk I also found in the Whitehall data was that it wasn't fully
consistent that the end_date of the previous government is the day
before the start date of the next one. For example:

```
{
  "id": "https://www.gov.uk/api/governments/2015-conservative-government",
  "title": "2015 Conservative government",
  "slug": "2015-conservative-government",
  "details": {
    "start_date": "2015-05-08",
    "end_date": null
  }
},
{
  "id": "https://www.gov.uk/api/governments/2010-to-2015-conservative-and-liberal-democrat-coalition-government",
  "title": "2010 to 2015 Conservative and Liberal Democrat coalition government",
  "slug": "2010-to-2015-conservative-and-liberal-democrat-coalition-government",
  "details": {
    "start_date": "2010-05-12",
    "end_date": "2015-05-08"
  }
},
{
  "id": "https://www.gov.uk/api/governments/2005-to-2010-labour-government",
  "title": "2005 to 2010 Labour government",
  "slug": "2005-to-2010-labour-government",
  "details": {
    "start_date": "2005-05-06",
    "end_date": "2010-05-11"
  }
}
```

Thus I've coded this to be somewhat resilient to this oddity.

### DIFF
--- a/app/models/government.rb
+++ b/app/models/government.rb
@@ -26,9 +26,7 @@ class Government
     id == other.id
   end
 
-  def eql?(other)
-    self == other
-  end
+  alias_method :eql?, :==
 
   def covers?(date)
     return false if date < start_date
--- a/app/models/government.rb
+++ b/app/models/government.rb
@@ -1,0 +1,39 @@
+# frozen_string_literal: true
+
+class Government
+  include ActiveModel::Model
+
+  def self.find(id)
+    all.find { |government| government.id == id }
+  end
+
+  def self.for_date(date)
+    all.find { |government| government.covers?(date) }
+  end
+
+  def self.current
+    for_date(Date.current)
+  end
+
+  def self.all
+    @all ||= YAML.load_file(Rails.root.join("config", "governments.yml"))
+                 .map { |hash| new(hash) }
+  end
+
+  attr_accessor :id, :slug, :name, :start_date, :end_date
+
+  def ==(other)
+    id == other.id
+  end
+
+  def eql?(other)
+    self == other
+  end
+
+  def covers?(date)
+    return false if date < start_date
+    return false if end_date && date > end_date
+
+    true
+  end
+end
--- a/app/models/government.rb
+++ b/app/models/government.rb
@@ -32,7 +32,10 @@ class Government
 
   def covers?(date)
     return false if date < start_date
-    return false if end_date && date > end_date
+    # Most end dates in Whitehall are the last date of a government so we
+    # treat the date as going up to 23:59:59 on the day by appending 1 day to
+    # the date
+    return false if end_date && date >= (end_date + 1)
 
     true
   end
--- a/config/governments.yml
+++ b/config/governments.yml
@@ -1,0 +1,285 @@
+# This list of governments is generated from Whitehall by running the following
+# code in the Rails console:
+# `puts Government.order(start_date: :desc).map { |g| g.as_json(only: %i[id slug name start_date end_date]) }.to_yaml`
+# This is included until government data is in the Publishing API and will
+# need to be updating each time governments change.
+- id: 56
+  slug: 2015-conservative-government
+  name: 2015 Conservative government
+  start_date: 2015-05-08
+  end_date:
+- id: 1
+  slug: 2010-to-2015-conservative-and-liberal-democrat-coalition-government
+  name: 2010 to 2015 Conservative and Liberal Democrat coalition government
+  start_date: 2010-05-12
+  end_date: 2015-05-08
+- id: 2
+  slug: 2005-to-2010-labour-government
+  name: 2005 to 2010 Labour government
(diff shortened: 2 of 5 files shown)


app/models/government.rb
```diff
@@ -0,0 +1,42 @@
...
+  end
+
+  attr_accessor :id, :slug, :name, :start_date, :end_date
+
+  def ==(other)
+    id == other.id
+  end
+
+  def eql?(other)
+    self == other
```
ChrisBAshton: I don't suppose we can do `alias_method :==, :eql?` here? (I suspect it triggers a syntax error, but if it works it'll be a cleaner solution!).

config/governments.yml
```diff
@@ -0,0 +1,285 @@
...
+# code in the Rails console:
+# `puts Government.order(start_date: :desc).map { |g| g.as_json(only: %i[id slug name start_date end_date]) }.to_yaml`
+# This is included until government data is in the Publishing API and will
+# need to be updating each time governments change.
+- id: 56
+  slug: 2015-conservative-government
+  name: 2015 Conservative government
+  start_date: 2015-05-08
+  end_date:
+- id: 1
```
ChrisBAshton: These IDs don't correspond to the government dates - something looks a bit off here?

spec/models/government_spec.rb
```diff
@@ -0,0 +1,125 @@
+# frozen_string_literal: true
+
+RSpec.describe Government do
+  describe ".all" do
+    it "returns a collection of governments" do
+      expect(Government.all).to all be_a(Government)
```
ChrisBAshton: Perhaps assert not_empty here?

spec/models/government_spec.rb
```diff
@@ -0,0 +1,125 @@
...
+      end
+
+      it "returns true on the start date" do
+        expect(government.covers?(Date.parse("2019-11-18"))).to be true
+      end
+
+      it "returns true after the start date" do
+        expect(government.covers?(Date.parse("2019-11-20"))).to be true
+      end
+    end
```
ChrisBAshton: I think the quirk you describe in your commit message (`A quirk I also found in the Whitehall data was that it wasn't fully...`) is being tested in the block above? I'd suggest encapsulating this code and explanation in its own commit. Keep the first commit simpler and then deal with the long tail of logic afterwards.

app/models/government.rb
```diff
@@ -0,0 +1,42 @@
...
+    all.find { |government| government.covers?(date) }
+  end
+
+  def self.current
+    for_date(Date.current)
+  end
+
+  def self.all
+    @all ||= YAML.load_file(Rails.root.join("config", "governments.yml"))
+                 .map { |hash| new(hash) }
```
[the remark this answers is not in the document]
kevindew: Sure, the `new` is the equivalent of calling `Government.new` since we're in the scope of the class.

So what this code is doing is looping through each object in the YAML file and creating an instance of Government for each one.

config/governments.yml
kevindew: I assume what happened is that the created every government from newest to oldest sequentially between 2010 and 2015. Then when the 2015 government was created it was given the newest id.

app/models/government.rb
kevindew: I'll see if that works, I just robbed this from a class we already had that was similar.

kevindew:
@ChrisBAshton Thanks for the review and suggestions. I've implemented them now.

app/models/government.rb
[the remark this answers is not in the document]
ChrisBAshton: Ah ok - I would normally just explicitly write `Government.new`, which removes any ambiguity, but use of `new` on its own is probably just a Ruby idiom I haven't come across before. I appreciate it means one less thing to update if you ever have to rename the class.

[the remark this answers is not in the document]
kevindew: Think of it in the same way that when we call instance methods we don't prefix them with `self` in the same way we prefix with `this` in Javascript.

In this situation we're in the context of Government so self == Government. So this method could be `self.new`, `Government.new` or `new`. Convention in ruby would dictate not using a superfluous prefix.

PR #1458, opened 2019-11-19
Determine whether edition is political is not

kevindew:
Trello: https://trello.com/c/WKUoYhxb/1189-spike-into-defining-the-history-mode-feature

This is a port of the logic from Whitehall that is used to determine
whether an edition is political [1]. This is simplified as Content
Publisher is currently only working with news and press releases and
will need to be expanded with the introduction of more document types
(or logic abstracted to those document types).

Whitehall checks whether content is associated with a minister and a
political organisation. Content Publisher however associates with role
appointments and isn't aware whether these are ministers or not. We
expect that nearly all associations will be ministers so until we have
the ability to determine between these just a check of a role
appointment should suffice.

[1]: https://github.com/alphagov/whitehall/blob/9afd13104412f70d0b739cada49dd1c5629a94b2/lib/political_content_identifier.rb

### DIFF
--- a/config/political_organisations.yml
+++ b/config/political_organisations.yml
@@ -1,0 +1,66 @@
+# This list of content_ids is generated from Whitehall by running the following
+# code in the Rails console:
+# `puts Organisation.where(political: true).order(:name).map { |o| "- '#{o.content_id}' # #{o.name}" }.join("\n")`
+# This is hardcoded in Content Publisher as this data is currently not
+# represented in the Publishing API. This list will need to be updated if
+# the political status of organisations are edited and in the long term
+# should call the Publishing API
+#
+- 'a0f338c5-e94c-42f8-9c26-b9c2eb6850d3' # Accelerated Access Review
+- '25aacd68-dc0c-4041-9c3a-df59a5357c23' # Attorney General's Office
+- 'bb371e6d-6c48-495a-bc64-494856345e99' # Better Regulation Delivery Office
+- '96ae61d6-c2a1-48cb-8e67-da9d105ae381' # Cabinet Office
+- '2bde479a-97f2-42b5-986a-287a623c2a1c' # Department for Business, Energy & Industrial Strategy
+- 'a3773e86-b0b1-4c96-b1f4-5fff487cca39' # Department for Business, Enterprise and Regulatory Reform
+- '569a9ee5-c195-4b7f-b9dc-edc17a09113f' # Department for Business, Innovation & Skills
+- 'aaebe7e8-f73a-4f53-a20d-1a1caca4e289' # Department for Children, Schools and Families
+- 'e7c374b2-35c8-4b7d-a6d5-7f2e9f8f3556' # Department for Constitutional Affairs
+- 'fd62c5a4-714d-47fe-a612-595d1739251c' # Department for Digital, Culture, Media & Sport
+- 'ebd15ade-73b2-4eaf-b1c3-43034a42eb37' # Department for Education
+- '77c1621f-a392-4393-9d8c-9969cd98c1e7' # Department for Education and Skills
+- 'de4e9dc6-cca4-43af-a594-682023b84d6c' # Department for Environment, Food & Rural Affairs
+- 'f323e83c-868b-4bcb-b6e2-a8f9bb40397e' # Department for Exiting the European Union
+- '04b788e1-8d6a-412e-a15a-1c1390328e47' # Department for Innovation, Universities and Skills
+- 'db994552-7644-404d-a770-a2fe659c661f' # Department for International Development
+- '082092f1-656c-470e-b024-229dc6e750e9' # Department for International Trade
+- '4c717efc-f47b-478e-a76d-ce1ae0af1946' # Department for Transport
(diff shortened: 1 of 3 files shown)


spec/lib/political_edition_identifier_spec.rb
```diff
@@ -0,0 +1,77 @@
+# frozen_string_literal: true
+
+RSpec.describe PoliticalEditionIdentifier do
+  describe ".policical_organisation_ids" do
+    it "returns an array of content_ids" do
+      uuid_regex = /^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$/
+      expect(PoliticalEditionIdentifier.political_organisation_ids)
+        .to all match(uuid_regex)
```
ChrisBAshton: May be worth asserting the length of the `political_organisation_ids` array here - we're already hardcoding the YAML file of orgs, so may as well hard code the number of orgs we expect to see.

My concern is that `political_organisation_ids` could return `[]` and all the tests would still currently pass.

spec/lib/political_edition_identifier_spec.rb
```diff
@@ -0,0 +1,77 @@
...
+      role_appointments_field = build(:tag_field,
+                                      type: "multi_tag",
+                                      id: "role_appointments")
+
+      build(:document_type, tags: [primary_organisation_field,
+                                   organisations_field,
+                                   role_appointments_field])
+    end
+
+    context "when a edition is associated with a role appointment" do
```
ChrisBAshton:
```suggestion
    context "when an edition is associated with a role appointment" do
```

spec/lib/political_edition_identifier_spec.rb
```diff
@@ -0,0 +1,77 @@
...
+      it "returns true" do
```
ChrisBAshton: I remember a code review I had from Leena, suggesting that if a `context` only has one `it` block inside it, then it should probably just be an `it` block. Something like:

```ruby
it "returns true when an edition is associated with a role appointment"
```

I quite like the insinuation of using the `context` syntax - that in a specific context, X is expected to happen - but find it more useful as a wrapping block for setting `let` instance variables. In this case, we only actually begin to capture the context when we're already inside the `it` block. I do think removing the `context` keyword would allow each test to feel more self-contained.

spec/lib/political_edition_identifier_spec.rb
```diff
@@ -0,0 +1,77 @@
...
+        edition = build(:edition,
+                        document_type_id: document_type.id,
+                        tags: { role_appointments: [SecureRandom.uuid] })
+        expect(PoliticalEditionIdentifier.new(edition).political?).to be true
+      end
+    end
+
+    context "when an editions primary publishing organisation is political" do
```
ChrisBAshton:
```suggestion
    context "when an edition's primary publishing organisation is political" do
```

spec/lib/political_edition_identifier_spec.rb
```diff
@@ -0,0 +1,77 @@
...
+          .and_return([political_organisation_id])
+
+        edition = build(:edition,
+                        document_type_id: document_type.id,
+                        tags: { primary_publishing_organisation: [political_organisation_id] })
+        expect(PoliticalEditionIdentifier.new(edition).political?).to be true
+      end
+    end
+
+    context "when an editions supporting organisation is political" do
```
ChrisBAshton:
```suggestion
    context "when an edition's supporting organisation is political" do
```

spec/lib/political_edition_identifier_spec.rb
```diff
@@ -0,0 +1,77 @@
...
+        edition = build(:edition,
+                        document_type_id: document_type.id,
+                        tags: { primary_publishing_organisation: [SecureRandom.uuid],
+                                organisations: [political_organisation_id] })
+
+        expect(PoliticalEditionIdentifier.new(edition).political?).to be true
+      end
+    end
+
+    context "when a edition is not associated with a role appointment or a political organisation" do
```
ChrisBAshton:
```suggestion
    context "when an edition is not associated with a role appointment or a political organisation" do
```

config/political_organisations.yml
```diff
@@ -0,0 +1,66 @@
+# This list of content_ids is generated from Whitehall by running the following
+# code in the Rails console:
+# `puts Organisation.where(political: true).map { |o| "- '#{o.content_id}' # #{o.name}" }.join("\n")`
```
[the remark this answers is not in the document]
kevindew: That's a good idea - I'll do that

spec/lib/political_edition_identifier_spec.rb
kevindew: I don't want to assert the length as I don't want the test to fail if the length of the list changes. However you raise a good point about the empty approach, so I'll add an assertion that it's not empty.

spec/lib/political_edition_identifier_spec.rb
kevindew: Yeah that seems a good shout for this block of code since we don't make use of any lets

kevindew:
Thanks for the review @ChrisBAshton I've implemented your suggestions

PR #1467, opened 2019-11-22
Fix URL preview spec due to lack of blur event

benthorner:
Previously we tested the URL preview of a title by clicking the 'body'
field as a way to generate a blur event. This replaces the click method
with an explicit 'tab', as the former doesn't seem to work anymore.


benthorner:
Maybe it's just me...

kevindew:
Oh awesome, thanks Ben. I had been wondering if this was just me when I started seeing it yesterday.

PR #1468, opened 2019-11-25
README: Remove old, dev VM 'how to run the app' instructions

issyl0:
- These instructions referenced the deprecated GOV.UK Development VM, `bowler`, \*and\* `./startup.sh`.
- The new standard is [GOV.UK Docker](https://github.com/alphagov/govuk-docker).

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,6 @@ A unified publishing application for content on GOV.UK
   * Edition - A revision that is in the Publishing API
   * Document - All revisions of a piece of content in a particular locale
 
-
 ## Technical documentation
 
 This is a Ruby on Rails application.
@@ -23,23 +22,13 @@ This is a Ruby on Rails application.
 
 ### Running the application
 
-The first time you run this application you'll want to set you
-[permissions](#permissions).
-
-Then, if you are running on the [GOV.UK development VM][dev-vm]:
+The first time you run this application for development, enable `debug` and `pre_release_features` permissions:
 
 ```
-cd /var/govuk/govuk-puppet/development-vm
-bowl content-publisher
+bundle exec rake development_permissions
 ```
 
-You should be able to access the app at: http://content-publisher.dev.gov.uk/
-
-Otherwise, on your machine:
-
-```
-./startup.sh
-```
+To enable them for your GOV.UK account add them to your account in [Signon](https://github.com/alphagov/signon).
 
 ### Running the test suite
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -21,26 +21,6 @@ This is a Ruby on Rails application.
 - [yarn][] - package manager for JavaScripts
 - [imagemagick][] - image manipulation library
 
-### Running the application
-
-The first time you run this application you'll want to set you
-[permissions](#permissions).
-
-Then, if you are running on the [GOV.UK development VM][dev-vm]:
```
thomasleese: We'll probably want to remove the `dev-vm` link as well.

README.md
```diff
@@ -21,26 +21,6 @@ This is a Ruby on Rails application.
 - [yarn][] - package manager for JavaScripts
 - [imagemagick][] - image manipulation library
 
-### Running the application
-
-The first time you run this application you'll want to set you
```
thomasleese: I suspect this line is still relevant to new people using the app.

issyl0: I kept it. I took the opportunity to rearrange some sections, too - `permissions` didn't need to be linked to.

kevindew:
It'd have been useful to have kept the startup.sh link for own machine and to link people to govuk-docker in place of the dev vm.

issyl0:
We can add it back, but here I did the same as a load of other repos: remove all the things. GOV.UK Docker should be the standard, should it not?

kevindew:
> We can add it back, but here I did the same as a load of other repos: remove all the things. GOV.UK Docker should be the standard, should it not?

We also don't tell anyone how to run it with docker. We now have a section called "Running the application" that doesn't tell you how to run the application. I'd suggest it says the docker approach first then startup.sh, not too fussed if we don't list the startup.sh method but I quite like it myself as is the easiest way to run the app.

issyl0:
> > We can add it back, but here I did the same as a load of other repos: remove all the things. GOV.UK Docker should be the standard, should it not?
> 
> We also don't tell anyone how to run it with docker. We now have a section called "Running the application" that doesn't tell you how to run the application. I'd suggest it says the docker approach first then startup.sh, not too fussed if we don't list the startup.sh method but I quite like it myself as is the easiest way to run the app.

Sure. We want to standardize on one way of doing things, though. And GOV.UK Docker is that way. The dev VM is deprecated. It's confusing for new starters and old timers alike to have three ways of "how to run a Rails app" across every repo which are essentially the same: startup.sh, `bundle exec rails s`, or `bowl app_name`. By not saying any specific way here, we standardize on not giving every app "How To Run A Rails App". GOV.UK devs (new or old) should have GOV.UK Docker in their consciousness as "how to run apps these days" anyway.

PR #1470, opened 2019-11-25
Set political status and government when an edition is updated

kevindew:

This builds on https://github.com/alphagov/content-publisher/pull/1466 and should be reviewed once that is merged in.

This code is used to set the political status and government each time content is updating the Publishing API. This uses the PreviewService and PublishService as places to call this logic. 

It also adds this check into the FailsafePreviewService so that content with issues can still reflect up-to-date political and government statuses.

I wasn't too sure about where to place this code. It feels quite natural to be placed within PreviewService and PublishService since it is data that we're concerned about the value at syncing. However it didn't feel such a natural place on a FailsafePreviewService. Alternatives to this however seemed rather complex and unappealing (lots of duplicated code or callbacks), so I think that maybe in time we'll end up renaming FailsafePreviewService if it acquires other similar concerns.

A late change I made to this was deciding to only set the government when content is published or backdated, previously this used the current time to set the government. I decided against using the current time as it meant that drafts would show incorrect governments whenever the government changed until they were then updated. I opted instead for no government at that point on the basis that no data was better than potentially incorrect data. A consequence of this change is that we need a manual way of setting a government at the point of publishing which is where the concept of a fallback_government comes from.

### DIFF
--- a/app/interactors/editions/create_interactor.rb
+++ b/app/interactors/editions/create_interactor.rb
@@ -15,6 +15,7 @@ class Editions::CreateInteractor < ApplicationInteractor
       create_next_revision
       create_next_edition
       create_timeline_entry
+      preview_next_edition
     end
   end
 
@@ -63,5 +64,9 @@ private
       TimelineEntry.create_for_status_change(entry_type: :new_edition,
                                              status: next_edition.status)
     end
+  end
+
+  def preview_next_edition
+    FailsafePreviewService.call(next_edition)
   end
 end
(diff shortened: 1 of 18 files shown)


kevindew:
Thanks @leenagupte - you've picked on the things I was concerned about with this PR. Originally we were setting government to be the current one on every sync for a draft. Which meant that government would be constantly updated (in case the government changed). I later switched to just not setting a government until point of publishing (which led to the concept of a fallback_government).

I've had a think and tried to re-approach this in a different way. I'll open a PR in a mo.

kevindew:
Superseded by https://github.com/alphagov/content-publisher/pull/1481

PR #1472, opened 2019-11-26
Use imported images and attachments

ChrisBAshton:
Trello card: https://trello.com/c/ubTM9ylI/1176-use-imported-images-and-attachments-in-content-publisher

- [x] Converts Whitehall image syntax to Content Publisher image syntax
- [x] Converts Whitehall attachment syntax to Content Publisher attachment syntax
- [x] Converts Whitehall inline attachment syntax to Content Publisher inline attachment syntax
- [x] Passes `images` to EmbedBodyReferences
- [x] Used first image as the lead image
- [x] ~Passes `attachments` to EmbedBodyReferences~ This will be done in [another Trello card](https://trello.com/c/4FF2oprK/1182-import-attachments-into-content-publisher) and is out of scope for this PR.

### DIFF
--- a/lib/whitehall_importer/create_revision.rb
+++ b/lib/whitehall_importer/create_revision.rb
@@ -25,6 +25,7 @@ module WhitehallImporter
       document_type_key = DOCUMENT_SUB_TYPES.reject { |t| whitehall_edition[t].nil? }.first
       raise AbortImportError, "Edition has an unsupported document type" unless SUPPORTED_DOCUMENT_TYPES.include?(whitehall_edition[document_type_key])
 
+      image_revisions = create_image_revisions(whitehall_edition["images"])
       Revision.create!(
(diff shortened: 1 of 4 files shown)


ChrisBAshton:
Thanks for the review and the context - I agree that's probably not worth coding for, as the examples of images I've seen so far are already on their own line. If it becomes a problem we could fairly easily write up a cleanup script that iterates over documents and adds any missing newlines, or just fix each page manually if there aren't many culprits.

kevindew:
> Thanks for the review and the context - I agree that's probably not worth coding for, as the examples of images I've seen so far are already on their own line. If it becomes a problem we could fairly easily write up a cleanup script that iterates over documents and adds any missing newlines, or just fix each page manually if there aren't many culprits.

Sounds good 👍

PR #1474, opened 2019-11-26
Import Unpublishing

1pretz1:
We want to have the ability to import a document that has been removed from GOV.UK. A complication of this is that Whitehall sets unpublished documents to a draft status. So we need to determine whether a user has modified the draft at all for us to know whether it should be an edition with "removed" status or be a "draft" status with a previous edition of "removed".

This PR starts off with refactoring which is needed to nicely fit the unpublishing workflow in and tidies up some other bits along the way.

## Imported an unedited unpublished edition

```
irb(main):012:0> pp Document.last.editions
[#<Edition:0x00007f49c175f7d0
  id: 923,
  number: 1,
  last_edited_at: Thu, 28 Nov 2019 17:08:53 GMT +00:00,
  created_at: Thu, 28 Nov 2019 17:07:54 GMT +00:00,
  updated_at: Thu, 28 Nov 2019 17:08:14 GMT +00:00,
  document_id: 639,
  created_by_id: 12,
  current: true,
  live: false,
  last_edited_by_id: 12,
  status_id: 2765,
  revision_id: 5452,
  revision_synced: true,
  access_limit_id: nil,
  system_political: false,
  government_id: nil>]

irb(main):013:0> pp Document.last.editions.last.status.state
"removed"
```

## Imported an edited unpublished edition

```
irb(main):006:0> pp Document.last.editions
[#<Edition:0x00007f49c5725540
  id: 921,
  number: 1,
  last_edited_at: Thu, 28 Nov 2019 17:04:06 GMT +00:00,
  created_at: Thu, 28 Nov 2019 16:58:44 GMT +00:00,
  updated_at: Thu, 28 Nov 2019 16:59:52 GMT +00:00,
  document_id: 638,
  created_by_id: 12,
  current: false,
  live: false,
  last_edited_by_id: 12,
  status_id: 2763,
  revision_id: 5451,
  revision_synced: true,
  access_limit_id: nil,
  system_political: false,
  government_id: nil>,
 #<Edition:0x00007f49c5725270
  id: 922,
  number: 2,
  last_edited_at: Thu, 28 Nov 2019 17:04:06 GMT +00:00,
  created_at: Thu, 28 Nov 2019 17:01:57 GMT +00:00,
  updated_at: Thu, 28 Nov 2019 17:01:57 GMT +00:00,
  document_id: 638,
  created_by_id: 12,
  current: true,
  live: false,
  last_edited_by_id: 12,
  status_id: 2764,
  revision_id: 5451,
  revision_synced: true,
  access_limit_id: nil,
  system_political: false,
  government_id: nil>]

irb(main):008:0> pp Document.last.editions.map { |x| x.status.state }
["removed", "draft"]
=> ["removed", "draft"]
```

Trello:
https://trello.com/c/nDK6MZLW/1172-import-an-unpublished-status-into-content-publisher


lib/whitehall_importer/create_edition.rb
```diff
@@ -21,15 +21,61 @@ def call
...
+    end
+
+    def unpublished_edition?
+      whitehall_edition["unpublishing"] && %w[submitted rejected draft].include?(whitehall_edition["state"])
+    end
+
+    def edited_unpublishing?
+      return false unless whitehall_edition["revision_history"].second_to_last
+
+      whitehall_edition["revision_history"].second_to_last["state"] != "published" && unpublished_edition?
```
kevindew: Should this be a question we ask the history object?

lib/whitehall_importer/migrate_state.rb
```diff
@@ -0,0 +1,45 @@
...
+  class MigrateState
+    attr_reader :whitehall_state, :force_published
+
+    SUPPORTED_WHITEHALL_STATES = %w(
+    draft
+    published
+    rejected
+    submitted
+    superseded
+    withdrawn
```
kevindew: Might want to indent this

spec/lib/whitehall_importer/create_edition_spec.rb
```diff
@@ -41,67 +39,231 @@
...
-      edition = described_class.call(document,
-                                     whitehall_document,
-                                     whitehall_edition,
-                                     1,
-                                     user_ids)
+      whitehall_edition = build(
+        :whitehall_export_edition,
+        state: "published",
+        revision_history: [
+          { "event" => "create", "state" => "draft", "whodunnit" => 1 },
```
kevindew: Shall we put these into a factory?

lib/whitehall_importer/create_edition.rb
```diff
@@ -2,15 +2,15 @@
...
   class CreateEdition
-    attr_reader :document, :whitehall_document, :whitehall_edition, :edition_number, :user_ids
+    attr_reader :document, :current, :whitehall_edition, :edition_number, :user_ids
 
     def self.call(*args)
       new(*args).call
     end
 
-    def initialize(document, whitehall_document, whitehall_edition, edition_number, user_ids)
+    def initialize(document:, current:, whitehall_edition:, edition_number:, user_ids:)
```
kevindew: Should we make some of these arguments optional (e.g current: true, edition_number: 1 user_ids: {}) just to make testing less verbose?

spec/lib/whitehall_importer/create_edition_spec.rb
```diff
@@ -41,67 +39,231 @@
...
-        whitehall_edition = build(
-          :whitehall_export_edition,
-          state: "withdrawn",
-          revision_history: [
-            { "event" => "create", "state" => "published", "whodunnit" => 1 },
-            { "event" => "update", "state" => "withdrawn", "whodunnit" => 1 },
-          ],
-          unpublishing: build(:whitehall_export_unpublishing),
-        )
+    it "sets the status to the user that created it and at the time that was done" do
```
kevindew: should sets be attributes?

spec/lib/whitehall_importer/create_edition_spec.rb
```diff
@@ -41,67 +39,231 @@
...
+        build(:whitehall_export_edition,
+              revision_history: [
+                { "event" => "create", "state" => "draft", "whodunnit" => 1, "created_at" => "2017-11-24T15:51:23+00:00" },
+                { "event" => "update", "state" => "published", "whodunnit" => 1, "created_at" => "2018-11-25T15:51:23+00:00" },
+                { "event" => "update", "state" => "draft", "whodunnit" => 1, "created_at" => "2019-11-26T15:51:23+00:00" },
+              ],
+              unpublishing: build(:whitehall_export_unpublishing,
+                                  alternative_url: "https://www.gov.uk/flextension",
+                                  unpublishing_reason: "Consolidated into another GOV.UK page",
+                                  explanation: "Brexit is being delayed again"))
```
kevindew: Maybe want to go with something a little less political here?

spec/lib/whitehall_importer/create_edition_spec.rb
```diff
@@ -41,67 +39,231 @@
...
+                             user_ids: user_ids)
+      end
+
+      it "creates an edition with a status of removed" do
+        expect(edition.removed?).to be_truthy
+      end
+
+      it "creates a removal record" do
+        expect(edition.status.details).to be_an_instance_of(Removal)
+      end
```
kevindew: I don't think you need this test since the next one covers it.

spec/lib/whitehall_importer/create_edition_spec.rb
```diff
@@ -41,67 +39,231 @@
...
+      it "sets the correct removal metadata" do
+        removal = edition.status.details
+        expect(removal.explanatory_note).to eq(whitehall_edition["unpublishing"]["explanation"])
+        expect(removal.alternative_path).to eq(whitehall_edition["unpublishing"]["alternative_url"])
+        expect(removal.redirect).to be_truthy
+      end
+
+      it "sets the correct timestamps on the edition" do
+        expect(edition.created_at).to eq("2017-11-24T15:51:23+00:00")
+        expect(edition.updated_at).to eq("2019-11-26T15:51:23+00:00")
```
kevindew: should we use let variables for this since this is quite removed from the setup?

kevindew: Could potentially check the timestamps in the check for a removed edition?

spec/lib/whitehall_importer/create_edition_spec.rb
```diff
@@ -41,67 +39,231 @@
...
+      let!(:edition) do
+        described_class.call(document: document,
+                             current: true,
+                             whitehall_edition: whitehall_edition,
+                             edition_number: 1,
+                             user_ids: user_ids)
+      end
+
+      it "creates two editions" do
+        expect(Edition.count).to eq(2)
```
kevindew: I'd suggest doing this as expect `{ described_class.call() }.to change { Edition.count }.by(2)`

spec/lib/whitehall_importer/create_edition_spec.rb
```diff
@@ -41,67 +39,231 @@
...
+      it "creates a removal record" do
+        expect(Edition.first.status.details).to be_an_instance_of(Removal)
+      end
+
+      it "sets the correct removal metadata" do
+        removal = Edition.first.status.details
+        expect(removal.explanatory_note).to eq(whitehall_edition["unpublishing"]["explanation"])
+        expect(removal.alternative_path).to eq(whitehall_edition["unpublishing"]["alternative_url"])
+        expect(removal.redirect).to be_truthy
+      end
```
kevindew: On these tests, rather than using Edition.first maybe we want to do document.editions.first?

spec/lib/whitehall_importer/import_spec.rb
```diff
@@ -59,5 +58,35 @@
...
+
+      expect(WhitehallImporter::CreateEdition).to receive(:call).with(
+        hash_including(current: false),
+      ).ordered
+
+      expect(WhitehallImporter::CreateEdition).to receive(:call).with(
+        hash_including(current: true),
+      ).ordered
+
+      described_class.call(import_data)
```
kevindew: If we set up a revision history factory can this be simpler at all?

spec/lib/whitehall_importer/create_edition_spec.rb
1pretz1: Done

lib/whitehall_importer/create_edition.rb
1pretz1: sure

spec/lib/whitehall_importer/import_spec.rb
1pretz1: it certainly can!

spec/lib/whitehall_importer/create_edition_spec.rb
1pretz1: Variables are now being used! 🌵

spec/lib/whitehall_importer/create_edition_spec.rb
1pretz1: Haha yes, you're probably right... changed!

lib/whitehall_importer/create_edition.rb
1pretz1: Moved to EditionHistory 👍

1pretz1:
Ready for review no.2 @kevindew! 😄

spec/factories/whitehall_export/edition_factory.rb
```diff
@@ -21,7 +21,7 @@
     contacts { [] }
     revision_history do
       [
-        { "event" => "create", "state" => state, "whodunnit" => 1 },
+        build(:revision_history_event),
```
kevindew: This looks good, be nice to pass the state into it.

kevindew: could also pass in the created_at value from the edition?

1pretz1: Good suggestion with passing `created_at` in however I don't think passing state in is a good idea, solely because the default revision_history_event mimics the first event created when an edition is freshly drafted. E.g event has the value of 'create' and not 'update'. 

Passing the state in would mean that you could have an event in the revision history that couldn't occur in reality. I hit this problem and decided we probably shouldn't pass the state in now some code realises on the accuracy of the version history. What do you think?

1pretz1:
Tested originally but have retested it again and put the content-publisher db results in the summary above! 🍏

PR #1475, opened 2019-11-26
[Spike] ResyncService

ChrisBAshton:
Trello: https://trello.com/c/kLriQCoa/1195-create-service-to-re-publish-content-to-publishing-api

## TODO

Create a rake task for this too. Can be done in the 'real' card; this is still a spike.

## Aside

Looks like we’re aborting on `scheduled` status at the moment: https://github.com/alphagov/content-publisher/blob/300c54d5fbf6319c9139a1535b1212694b236172/lib/whitehall_importer/migrate_state.rb. Some work will need to be done on the ResyncService if/when schedule support is added.

## Local testing

In one terminal:

```sh
cd ~/govuk/publishing-api
govuk-docker-up
```

In the other:

```
curl http://publishing-api.dev.gov.uk/v2/content/cdaebb8f-02ae-48ef-be0a-b718b49f9a24 \
  -H "Content-Type: application/json" --request PUT \
  --data '{ "base_path": "/foo/bar/baz", "document_type": "press_release", "publishing_app": "content-publisher", "rendering_app": "government-frontend", "routes": [ { "path": "/foo/bar/baz", "type": "exact" } ], "schema_name": "news_article", "title": "test", "update_type": "major", "details": { "body": "this is a body" } }'
```

Can then `curl http://publishing-api.dev.gov.uk/v2/content/cdaebb8f-02ae-48ef-be0a-b718b49f9a24` to retrieve the content.

In another terminal, can `govuk-docker-run bundle exec rails console` in `publishing-api` and run `Document.last` to see the document (`Document.last.editions` to see its contents).

Publish: `curl http://publishing-api.dev.gov.uk/v2/content/cdaebb8f-02ae-48ef-be0a-b718b49f9a24/publish -H "Content-Type: application/json" --request POST --data '{}'`

In the Rails console, `Document.last` has a `stale_lock_version`, currently at `1`. On `Document.last.editions`, we have one edition with a `user_facing_version: 1`.

Re-run the original PUT command (perhaps with an updated `body`) and we now have two editions. The `Document.last.stale_lock_version` is now on 2, and the draft edition has `user_facing_version: 2`, but the published one is still `1`. We can use this value (presumably) as the optional `previous_version` parameter in our requests.

### Testing the republish

As above, I created a local document in publishing API, that had two editions [first: published, second: draft].

I sent a request to `PUT /v2/content/:content_id` with the original payload of the first edition, and a `"update_type": "republish"` merged into that.

My expectation was that it would create a new, live edition, so that we would now have [first: published, second: draft, third: published-with-update_type:"republish"].

What it actually did was overwrite the draft edition so that we have [first: published, second: draft-with-update_type:"republish"]. If we then were to 'sync' the second edition, we'd overwrite again to [first: published, second: draft]. In other words, only the draft edition ever gets synced, the live version remains untouched.

I then tried a `PUT /v2/content/:content_id` (with `"update_type": "republish"`) on a document whose **latest edition is 'published'**. I expected it to remain `published` but with an update_type of `republish`. In fact, it created a new draft. So we need an additional step.

```
when syncing a live edition:
  PUT  /v2/content/:content_id (with update_type: republish)
  # this is not published, it's only overwritten a draft edition, so need to publish
  POST /v2/content/:content_id/publish

when syncing a draft edition:
  PUT  /v2/content/:content_id (with no specific update_type)
```

### Testing migration

Creating a Whitehall document:

```
curl http://publishing-api.dev.gov.uk/v2/content/cdaebb8f-02ae-48ef-be0a-b718b49f9a26 \
  -H "Content-Type: application/json" --request PUT \
  --data '{ "base_path": "/bar", "document_type": "press_release", "publishing_app": "whitehall", "rendering_app": "whitehall-frontend", "routes": [ { "path": "/bar", "type": "exact" } ], "schema_name": "news_article", "title": "test", "update_type": "major", "details": { "body": "Whitehall content" } }' 
```

Publish it:

`curl http://publishing-api.dev.gov.uk/v2/content/cdaebb8f-02ae-48ef-be0a-b718b49f9a26/publish -H "Content-Type: application/json" --request POST --data '{}'`

Reserve the path (from Content Publisher script):

`curl http://publishing-api.dev.gov.uk/paths/bar -H "Content-Type: application/json" --request PUT --data '{ "publishing_app": "content-publisher", "override_existing": true }'`

Sync Content Publisher version (note we've changed the publishing_app and rendering_app):

```
curl http://publishing-api.dev.gov.uk/v2/content/cdaebb8f-02ae-48ef-be0a-b718b49f9a26 \
  -H "Content-Type: application/json" --request PUT \
  --data '{ "base_path": "/bar", "document_type": "press_release", "publishing_app": "content-publisher", "rendering_app": "government-frontend", "routes": [ { "path": "/bar", "type": "exact" } ], "schema_name": "news_article", "title": "test", "update_type": "major", "details": { "body": "Whitehall content" } }' 
```

Publish it:

`curl http://publishing-api.dev.gov.uk/v2/content/cdaebb8f-02ae-48ef-be0a-b718b49f9a26/publish -H "Content-Type: application/json" --request POST --data '{}'`

A new edition is created, with the new publishing/rendering apps. The previous editions remain intact.

### DIFF
--- a/app/services/resync_service.rb
+++ b/app/services/resync_service.rb
@@ -8,6 +8,8 @@ class ResyncService < ApplicationService
   end
 
   def call
+    reserve_base_path(document.current_edition)
+
     if document.live_edition.present?
       resync(document.live_edition)
       return if document.current_edition == document.live_edition
@@ -20,6 +22,17 @@ class ResyncService < ApplicationService
   end
 
 private
+
+  def reserve_base_path(edition)
+    # Reserves a path for a publishing application, in case it has
+    # been imported from Whitehall. There's no harm in calling
+    # this even if we already own the path.
+    GdsApi.publishing_api_v2.put_path(
+      edition.base_path,
+      publishing_app: "content-publisher",
+      override_existing: true,
+    )
+  end
 
   def resync(edition)
     if edition.withdrawn?
--- a/app/services/resync_service.rb
+++ b/app/services/resync_service.rb
@@ -9,11 +9,11 @@ class ResyncService < ApplicationService
 
   def call
     if document.live_edition.present?
-      republish_live_edition
-      update_draft_edition unless document.live_edition == document.current_edition
-    else
-      update_draft_edition
+      resync(document.live_edition)
+      return if document.current_edition == document.live_edition
     end
+
+    resync(document.current_edition)
   rescue GdsApi::BaseError => e
     GovukError.notify(e)
     raise
@@ -21,43 +21,43 @@ class ResyncService < ApplicationService
 
 private
 
-  def republish_live_edition
-    # Now we want to promote the content item and assets from
-    # the draft stack to the live stack. We can't use the
-    # PublishService here as it would promote `current_edition`
-    # to `live_edition`. (We just want to resync the two with
-    # Publishing API, not change their states). It also calls
-    # `publish` rather than `republish`.
-    # And we can't use PreviewService for presenting the doc
-    # to publishing API as there is no way of hooking the
-    # `update_type`/`bulk_publishing` into the payload.
-    # (We could - and possibly should - modify the
-    # PreviewService to allow us to pass this override)
+  def resync(edition)
+    if edition.withdrawn?
+      resync_live_withdrawn(edition)
+    elsif edition.published? || edition.published_but_needs_2i?
+      resync_live(edition)
+    elsif edition.draft? || edition.submitted_for_review?
+      resync_draft(edition)
+    end
+  end
 
-    payload = PreviewService::Payload.new(document.live_edition).payload
+  def resync_live_withdrawn(edition)
+    WithdrawService.call(edition, edition.status.details)
+  end
 
-    # Manually update live edition. Whether the latest edition
-    # is published or draft, this will create a new draft with
-    # an `update_type: "republish"`. We then need to publish it.
+  def resync_live(edition)
+    payload = PreviewService::Payload.new(edition).payload
+
+    # This will create a new draft with an `update_type: "republish"`.
     GdsApi.publishing_api_v2.put_content(
-      document.content_id,
+      edition.document.content_id,
       payload.merge(update_type: "republish", bulk_publishing: true),
     )
 
-    # Manually publish assets to live stack before we publish
-    # the document, otherwise we risk referencing assets that
-    # aren't visible yet.
-    PublishAssetService.call(document.live_edition, nil)
+    # Publish assets to live stack before we publish the document,
+    # otherwise we risk referencing assets that aren't visible yet.
+    PublishAssetService.call(edition, nil)
 
+    # Finally, publish the doc (promote from draft stack to live)
     GdsApi.publishing_api_v2.publish(
-      document.content_id,
+      edition.document.content_id,
       nil, # Sending update_type is deprecated (now in payload)
-      locale: document.locale,
+      locale: edition.document.locale,
     )
   end
 
-  def update_draft_edition
-    # Present to publishing API & Asset Manager draft stack
-    PreviewService.call(document.current_edition)
+  def resync_draft(edition)
+    # Present to publishing API and Asset Manager draft stack
+    PreviewService.call(edition)
   end
 end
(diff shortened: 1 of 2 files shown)


app/services/resync_service.rb
```diff
@@ -0,0 +1,37 @@
...
+      # Publish document to live stack
+      #  (We can't use PublishService directly as it promotes
+      #  `current_edition` to `live_edition`. We just want to
+      #  resync the two with Publishing API, not change their
+      #  states.)
+      GdsApi.publishing_api_v2.publish(
+        document.content_id,
+        nil, # Sending update_type is deprecated (now in payload)
+        locale: document.locale,
+      )
```
kevindew: A thing we'll have to ponder between the preview and publish call is that one of them needs to tell the Publishing API that we're sending a republish.

app/services/resync_service.rb
```diff
@@ -0,0 +1,37 @@
...
+
+      # Publish assets to live stack
+      PublishAssetService.call(document.live_edition, document.live_edition)
+    end
+
+    # Present to publishing API & Asset Manager draft stack
+    PreviewService.call(document.current_edition)
```
kevindew: We'd not need to do this by the way if the live edition is the same as the current edition

app/services/resync_service.rb
```diff
@@ -0,0 +1,37 @@
...
+    document.reload
```
kevindew: we shouldn't need to do this too by the way - though I'm guessing it's an artefact from the work in progress.

app/services/resync_service.rb
ChrisBAshton: Ah, I meant for this to be a republish. Good spot.

app/services/resync_service.rb
ChrisBAshton: Have now added a conditional for this.

app/services/resync_service.rb
kevindew: Just a few notes on that front.

Publishing API has a [/v2/republish](https://github.com/alphagov/publishing-api/blob/master/doc/api.md#post-v2contentcontent_idrepublish) endpoint. This is used to transition unpublished content back to published so cannot be used in this scenario.

We need to set an update_type of "republish" to the Publishing API. This can either be done through PutContent (the preferred way) or through the Publish method (the deprecated way). So if we can set it at the preview point I'd recommend that.

The whole aspect of `update_type: republish` is a bit quirky so don't be too surprised if we hit something a bit thorny.

ChrisBAshton: Ok - have updated the GdsApi helper method we call.

ChrisBAshton: >  through the Publish method (the deprecated way)

Should we edit https://github.com/alphagov/publishing-api/blob/master/doc/api.md#post-v2contentcontent_idrepublish to make it clear this approach is no longer recommended?

ChrisBAshton: Also this doc needs updating I think: https://github.com/alphagov/publishing-api/blob/master/doc/api.md#post-v2contentcontent_idpublish

> update_type (conditionally required, deprecated)
> This is deprecated, clients should instead set the update type when putting the content.

...but if I publish without setting an update type (`curl http://publishing-api.dev.gov.uk/v2/content/cdaebb8f-02ae-48ef-be0a-b718b49f9a23/publish -H "Content-Type: application/json" --request POST --data '{}'`) I get an error:

`{"error":{"code":422,"message":"update_type is required","fields":{"update_type":["is invalid"]}}}`

kevindew: > > through the Publish method (the deprecated way)
> 
> Should we edit https://github.com/alphagov/publishing-api/blob/master/doc/api.md#post-v2contentcontent_idrepublish to make it clear this approach is no longer recommended?

I think we have crossed wires here as this republish method is allowed and fine - it's just not for our needs.

ChrisBAshton:
Work to be completed in https://trello.com/c/cT9yYO4X/1214-create-a-resyncservice.

PR #1477, opened 2019-11-27
Add rake task for creating CSVs of political status by organisation

AgaDufrat:
Several published documents on Content Publisher will go into history mode when a government falls.  We need to let managing editors know which documents are politically sensitive and will go into history mode so they can audit their documents and identify exceptions.

This rake task produces a CSV files with a list of live documents with their political status for each organisation.

It is intended to be removed once the history mode is implemented in the user interface. We will raise an exception as a prompt.

_Tested in integration_

[Trello card](https://trello.com/c/JP693RnO/1188-work-out-which-docs-are-political)

### DIFF
--- a/lib/tasks/csv_report.rake
+++ b/lib/tasks/csv_report.rake
@@ -5,6 +5,8 @@ require "csv"
 namespace :csv_report do
   desc "A CSV report of political documents published by organisations"
   task political_status_by_organisation: :environment do
+    raise "csv_report:political_status_by_organisation rake task should be removed" unless Time.zone.today < Date.new(2020, 1, 1)
+
     live_editions = Edition.where(live: true).joins(revision: :tags_revision)
     org_content_ids = live_editions.pluck(Arel.sql("tags_revisions.tags->'primary_publishing_organisation'->0")).uniq
     links = Linkables.new("organisation")
--- a/lib/tasks/csv_report.rake
+++ b/lib/tasks/csv_report.rake
@@ -1,0 +1,36 @@
+# frozen_string_literal: true
+
+require "csv"
+
+namespace :csv_report do
+  desc "A CSV report of political documents published by organisations"
+  task political_status_by_organisation: :environment do
+    live_editions = Edition.where(live: true).joins(revision: :tags_revision)
+    org_content_ids = live_editions.pluck(Arel.sql("tags_revisions.tags->'primary_publishing_organisation'->0")).uniq
+    links = Linkables.new("organisation")
+
+    organisations = org_content_ids.map { |id| links.by_content_id(id) }.compact
+
+    organisations.each do |organisation|
+      path = Rails.root.join("tmp", "#{organisation['internal_name'].parameterize}-political-status.csv")
+      csv_headers = ["Public URL", "Admin URL", "Title", "Document type", "Political", "Summary"]
+
+      CSV.open(path, "w", headers: csv_headers, write_headers: true) do |csv|
+        editions_published_by_organisation = live_editions.merge(TagsRevision.primary_organisation_is(organisation["content_id"]))
+
+        editions_published_by_organisation.each do |edition|
+          csv << [
+            "https://gov.uk#{edition.base_path}",
+            "https://content-publisher.publishing.service.gov.uk/documents/#{edition.content_id}:#{edition.locale}",
+            edition.title,
+            edition.document_type.id.humanize,
+            PoliticalEditionIdentifier.new(edition).political? ? "Yes" : "No",
+            edition.summary,
+          ]
+        end
+      end
+
+      puts "Report available at #{path}"
+    end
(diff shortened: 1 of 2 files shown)


lib/tasks/csv_report.rake
```diff
@@ -0,0 +1,39 @@
+# frozen_string_literal: true
+
+require "csv"
+
+namespace :csv_report do
+  desc "A CSV report of political documents published by organisations"
+  task political_status_by_organisation: :environment do
+    raise "csv_report:political_status_by_organisation rake task should be removed" unless Time.zone.today < Date.new(2020, 1, 1)
```
ChrisBAshton: This behaviour isn't captured in a test nor in the Trello card. What are its origins?

lib/tasks/csv_report.rake
```diff
@@ -0,0 +1,39 @@
...
+
+    # Get all primary publishing organisations in use
+    org_content_ids = Edition.where(live: true).joins(revision: :tags_revision).pluck(Arel.sql("tags_revisions.tags->'primary_publishing_organisation'->0")).uniq
```
ChrisBAshton: This line would be easier to read if it was split over several lines 😄 Haven't come across Arel before, nice!

lib/tasks/csv_report.rake
```diff
@@ -0,0 +1,39 @@
...
+    links = Linkables.new("organisation")
+
+    organisations = org_content_ids.map { |id| links.by_content_id(id) }.compact
+
+    organisations.each do |organisation|
+      path = Rails.root.join("tmp", "#{organisation['internal_name'].parameterize}-political-status.csv")
+      csv_headers = ["Public URL", "Admin URL", "Title", "Document type", "Political", "Summary"]
+
+      CSV.open(path, "w", headers: csv_headers, write_headers: true) do |csv|
+        editions_published_by_organisation = Edition.where(live: true).joins(revision: :tags_revision).merge(TagsRevision.primary_organisation_is(organisation["content_id"]))
```
ChrisBAshton: `Edition.where(live: true).joins(revision: :tags_revision)` is a repetition of line 11 - I would be tempted to move this out into a variable on line 9: `live_editions = Edition.where(live: true).joins(revision: :tags_revision)`

Can then reuse this value on line 11 and 22, saving a bit of database querying (and shortening the lines, not as much chaining needed).

lib/tasks/csv_report.rake
```diff
@@ -0,0 +1,39 @@
...
+
+        editions_published_by_organisation.each do |edition|
+          csv << [
+            "https://gov.uk#{edition.base_path}",
+            "https://content-publisher.publishing.service.gov.uk/documents/#{edition.content_id}:en",
```
ChrisBAshton: Even though we _know_ they'll all be `en` at the moment, I'd still suggest futureproofing:

```suggestion
            "https://content-publisher.publishing.service.gov.uk/documents/#{edition.content_id}:#{edition.locale}",
```

spec/lib/tasks/csv_report_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+
+    it "ignores draft content" do
+      create(:edition, tags: { primary_publishing_organisation: [organisation1["content_id"]] })
+
+      expect(CSV).not_to receive(:open)
+
+      Rake::Task["csv_report:political_status_by_organisation"].invoke
+    end
+  end
+end
```
ChrisBAshton: The tests look great. 🎉

lib/tasks/csv_report.rake
AgaDufrat: Makes sense to split it.
I wasn't initially going to use Arel but this warning popped up:
```
DEPRECATION WARNING: Dangerous query method (method whose arguments are used
as raw SQL) called with non-attribute argument(s): "tags_revisions.tags
->'primary_publishing_organisation'->0". Non-attribute arguments will be 
disallowed in Rails 6.0. This method should not be called with user-provided
values, such as request parameters or model attributes. Known-safe values can 
be passed by wrapping them in Arel.sql().
```

lib/tasks/csv_report.rake
kevindew: I'll add a bit of backstory...

It's very common in our rake tasks (and even some app features) that are used once or twice then never used again but remain in the code for perpetuity. If we can forsee that being the case it's worth adding the equivalent of dynamite so that when we hit a certain point in time we have to deal with either removing the feature, extending the date or removing the dynamite.

This feature seems something that's unlikely to be needed long term once we have filtering in so seems a candidate for removal.

Though I would suggest it be flagged in it's own commit so the intent is documented.

AgaDufrat: Kevin suggested this approach. I will try to remove the task before it causes any annoyance.

PR #1481, opened 2019-11-28
Alternative way to automatically set political status and government

kevindew:

Alternative implementation to https://github.com/alphagov/content-publisher/pull/1470

This creates a service that is called at the point of editing an edition which is used to change the data, set the user and time it was done and performs any automatic operations that are needed - in this case these are the political and government associations.

As part of this I also created a service to assign a status to an edition which allowed moving a reasonable amount of code out of the Edition model.

As future work it'd be good to remove create_initial from models and that could reduce code in models further.


spec/services/edit_edition_service_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+      revision = build(:revision)
+
+      expect { EditEditionService.call(edition, user, revision: revision) }
+        .to change { edition.revision }.to(revision)
+    end
+
+    it "does not save the edition" do
+      EditEditionService.call(edition, user, {})
+
+      expect(edition).to be_new_record
```
leenagupte: I like the use of `be_new_record` here, but why is it important that the edition isn't saved?

spec/services/edit_edition_service_spec.rb
```diff
@@ -36,5 +36,65 @@
...
+          .with(edition)
+          .and_return(instance_double(PoliticalEditionIdentifier, political?: false))
+
+        expect { EditEditionService.call(edition, user) }
+          .to change { edition.system_political }
+          .to(false)
+      end
+    end
+
+    describe "associates the edition with a government" do
```
leenagupte: These tests are great, but could they also include a test for what happens if a government doesn't exist for the provided date?

app/services/edit_edition_service.rb
```diff
@@ -13,9 +13,26 @@ def call
     edition.assign_attributes(
       attributes.merge(last_edited_by: user, last_edited_at: Time.current),
     )
+
+    determine_political
+    determine_government
```
leenagupte: So all content is going to be associated with a government, regardless of whether it's political or not?

app/services/publish_service.rb
```diff
@@ -29,6 +31,20 @@ def publish_assets(live_edition)
...
 
+  def associate_with_government
+    return if edition.government
+
+    date = edition.backdated_to || edition.document.first_published_at || Time.current
+    government = Government.for_date(date)
+    edition.assign_attributes(government_id: government&.content_id)
+  end
+
+  def sync_current_edition
```
leenagupte: This method seems tightly couple with `associate_with_government`. Could they be merged?

app/services/publish_service.rb
```diff
@@ -29,6 +31,20 @@ def publish_assets(live_edition)
     PublishAssetService.call(edition, live_edition)
   end
 
+  def associate_with_government
+    return if edition.government
+
+    date = edition.backdated_to || edition.document.first_published_at || Time.current
+    government = Government.for_date(date)
+    edition.assign_attributes(government_id: government&.content_id)
```
leenagupte: It feels a bit odd to have almost identical code in two places. I understand why it's being done, but the use of `Time.current` to find the appropriate government is a very subtle difference, and I worry about it being missed by future developers.

Could there be case for creating a `AssociateWithGovernmentService` that allows documents flagged as new at the time of calling to use `Time.current`?

spec/services/publish_service_spec.rb
```diff
@@ -49,6 +52,78 @@
...
+                         first_published_at: first_published_at)
+        expect(Government).to receive(:for_date).with(first_published_at)
+                          .and_return(government)
+
+        expect { PublishService.call(edition, user, with_review: true) }
+          .to change { edition.government_id }
+          .to(government.content_id)
+      end
+
+      it "Associate with current government as a fallback" do
```
leenagupte: Nitpick: This test name should be "associates...".

spec/services/assign_edition_status_service_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+    end
+
+    it "does not save the edition" do
+      AssignEditionStatusService.call(edition, user, :submitted_for_review)
+
+      expect(edition).to be_new_record
+    end
+
+    it "updates last edited" do
+      travel_to(Time.current) do
```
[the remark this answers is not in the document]
kevindew: I'll try switching to freeze_time

app/services/edit_edition_service.rb
```diff
@@ -13,9 +13,26 @@ def call
...
+    identifier = PoliticalEditionIdentifier.new(edition)
+    edition.system_political = identifier.political?
+  end
+
+  def determine_government
+    date = edition.backdated_to || edition.document.first_published_at
+
+    edition.government_id = if date
+                              government = Government.for_date(date)
+                              government.content_id
```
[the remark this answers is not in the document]
kevindew: Shoot - should have done.

spec/services/edit_edition_service_spec.rb
kevindew: Will check.

app/services/edit_edition_service.rb
kevindew: Yes - that's how Whitehall has done it.

app/services/publish_service.rb
kevindew: I'll see if the sync fits into the associate method

spec/services/edit_edition_service_spec.rb
kevindew: Sure, I've updated the commit message.

app/interactors/access_limit/update_interactor.rb
```diff
@@ -31,7 +31,7 @@ def update_access_limit
 
     if LIMIT_TYPES.exclude?(limit_type)
       context.fail! if edition.access_limit.nil?
-      edition.assign_as_edit(user, access_limit: nil)
+      EditEditionService.call(edition, user, access_limit: nil)
```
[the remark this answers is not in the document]
kevindew: I'll pull it out into a different commit.

app/services/assign_edition_status_service.rb
```diff
@@ -0,0 +1,27 @@
+# frozen_string_literal: true
+
+class AssignEditionStatusService < ApplicationService
```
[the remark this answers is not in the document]
kevindew: Possibly, I'm not sure it'll gain much there but definitely free to be used if the scenario fits.

spec/services/publish_service_spec.rb
kevindew: Thanks

app/services/publish_service.rb
kevindew: Sure, I understand your concerns. I wanted to avoid a situation where we have overly specific methods for this e.g. `Edition#date_for_government` or a service that has to deal with the `Time.current` edge case (which is how we got `fallback_government` last time around.

In the end since we only have this logic twice and it's slightly different both times and ultimately quite simple I felt it was ok to duplicate it. I'm happy to address this if we hit a situation where we have it another time.

kevindew:
@leenagupte Thanks for the review all comments should be addressed.

spec/services/assign_edition_status_service_spec.rb
```diff
@@ -21,7 +21,7 @@
     end
 
     it "updates last edited" do
-      travel_to(Time.current) do
+      freeze_time do
```
leenagupte: I like the use of `freeze_time` here, but this change should probably to fixed up to the previous commit.

kevindew: Sorted.

PR #1482, opened 2019-11-28
Abort import task if import fails

kevindew:
This ensures we get an exit code of 1 to indicate an error and outputs
text to explain why it failed.

### DIFF
--- a/spec/lib/tasks/import_spec.rb
+++ b/spec/lib/tasks/import_spec.rb
@@ -5,6 +5,7 @@ RSpec.describe "Import tasks" do
     let(:whitehall_host) { Plek.new.external_url_for("whitehall-admin") }
 
     before do
+      allow($stdout).to receive(:puts)
       Rake::Task["import:whitehall"].reenable
       stub_request(:get, "#{whitehall_host}/government/admin/export/document/123")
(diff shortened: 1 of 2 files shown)


spec/lib/tasks/import_spec.rb
```diff
@@ -13,5 +14,14 @@
     it "creates a document" do
       expect { Rake::Task["import:whitehall"].invoke("123") }.to change { Document.count }.by(1)
     end
+
+    it "aborts if the import fails" do
+      expect(WhitehallImporter::Import).to receive(:call).and_raise("Error importing")
+
+      expect($stdout).to receive(:puts).with("Import failed")
+      expect($stdout).to receive(:puts).with("Error: Error importing")
+      expect { response = Rake::Task["import:whitehall"].invoke("123") }
```
ChrisBAshton:
```suggestion
      expect { Rake::Task["import:whitehall"].invoke("123") }
```

Linter is complaining about useless variable.

kevindew: D'oh thanks!

kevindew:
Thanks Chris, by the way here is an integration run: https://deploy.integration.publishing.service.gov.uk/job/run-rake-task/100722/console

PR #1484, opened 2019-11-29
Edition method for public first published time

kevindew:

This is a method that returns the date that is shown on GOV.UK to show
the first published at time on an edition. It is intended to be used to
calculate the government at a particular time.

### DIFF
--- a/app/services/publish_service.rb
+++ b/app/services/publish_service.rb
@@ -33,8 +33,11 @@ private
   def associate_with_government
     return if edition.government
 
-    date = edition.backdated_to || edition.document.first_published_at || Time.current
-    government = Government.for_date(date)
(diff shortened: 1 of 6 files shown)


app/services/publish_service.rb
```diff
@@ -33,8 +33,11 @@ def publish_assets(live_edition)
   def associate_with_government
     return if edition.government
 
-    date = edition.backdated_to || edition.document.first_published_at || Time.current
-    government = Government.for_date(date)
+    government = if edition.public_first_published_at
+                   Government.for_date(edition.public_first_published_at)
+                 else
+                   Government.current
```
leenagupte: This is much nicer and easier to follow!

PR #1487, opened 2019-11-29
Resync service

ChrisBAshton:
Adds a Resync service for use in syncing Whitehall-imported documents, or mass-updating documents that need to be put into history mode.

Trello card: https://trello.com/c/cT9yYO4X/1214-create-a-resyncservice

### DIFF
--- a/app/services/preview_service.rb
+++ b/app/services/preview_service.rb
@@ -1,9 +1,9 @@
 # frozen_string_literal: true
 
 class PreviewService < ApplicationService
-  def initialize(edition, optional_params = {})
+  def initialize(edition, republish: false)
     @edition = edition
-    @optional_params = optional_params
+    @republish = republish
   end
 
   def call
@@ -16,7 +16,7 @@ class PreviewService < ApplicationService
 
 private
 
-  attr_reader :edition, :optional_params
+  attr_reader :edition, :republish
 
   def put_draft_content
     payload = Payload.new(edition).payload
@@ -28,5 +28,14 @@ private
   def put_draft_assets
     edition.image_revisions.each(&:ensure_assets)
     edition.assets.each { |asset| PreviewAssetService.call(edition, asset) }
+  end
+
+  def optional_params
+    return {} unless republish
+
+    {
+      "update_type" => "republish",
+      "bulk_publishing" => true,
+    }
   end
 end
--- a/app/services/preview_service.rb
+++ b/app/services/preview_service.rb
@@ -19,23 +19,13 @@ private
   attr_reader :edition, :republish
 
   def put_draft_content
-    payload = Payload.new(edition).payload
-    GdsApi.publishing_api_v2.put_content(edition.content_id,
-                                         payload.merge(optional_params))
+    payload = Payload.new(edition, republish: republish).payload
+    GdsApi.publishing_api_v2.put_content(edition.content_id, payload)
     edition.update!(revision_synced: true)
   end
 
   def put_draft_assets
     edition.image_revisions.each(&:ensure_assets)
     edition.assets.each { |asset| PreviewAssetService.call(edition, asset) }
-  end
-
-  def optional_params
-    return {} unless republish
-
-    {
-      "update_type" => "republish",
-      "bulk_publishing" => true,
-    }
   end
 end
--- a/app/services/preview_service.rb
+++ b/app/services/preview_service.rb
@@ -1,8 +1,9 @@
 # frozen_string_literal: true
 
 class PreviewService < ApplicationService
-  def initialize(edition)
+  def initialize(edition, optional_params = {})
     @edition = edition
+    @optional_params = optional_params
   end
 
   def call
@@ -15,11 +16,12 @@ class PreviewService < ApplicationService
 
 private
 
-  attr_reader :edition
+  attr_reader :edition, :optional_params
 
   def put_draft_content
     payload = Payload.new(edition).payload
-    GdsApi.publishing_api_v2.put_content(edition.content_id, payload)
+    GdsApi.publishing_api_v2.put_content(edition.content_id,
+                                         payload.merge(optional_params))
     edition.update!(revision_synced: true)
   end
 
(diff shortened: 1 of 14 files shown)


app/services/preview_service.rb
```diff
@@ -1,8 +1,9 @@
 # frozen_string_literal: true
 
 class PreviewService < ApplicationService
-  def initialize(edition)
+  def initialize(edition, optional_params = {})
```
kevindew: Do we anticipate to need to use these for any purpose other than republish? might be more explicit to have a boolean keyword argument of republish.

app/services/resync_service.rb
```diff
@@ -0,0 +1,86 @@
...
+      alternative_path: removal.alternative_path,
+      locale: edition.locale,
+    )
+  end
+
+  def government_id(edition, document)
+    date = edition.backdated_to || document.first_published_at
+    return unless date
+
+    Government.for_date(date)&.content_id
```
kevindew: We can use the [Edition#first_published_at](https://github.com/alphagov/content-publisher/blob/3460397f837bf11c132a59e577e67abee9e3403c/app/models/edition.rb#L159) method to avoid the conditional here

app/services/resync_service.rb
```diff
@@ -0,0 +1,86 @@
...
+      publish(edition)
+    end
+
+    edition.update!(revision_synced: true)
+  end
+
+  def update_current_edition(edition)
+    edition.update!(
+      revision_synced: false,
+      system_political: PoliticalEditionIdentifier.new(edition).political?,
```
kevindew: We should set government on this too - in case the content was already published or is backdated.

app/services/resync_service.rb
```diff
@@ -0,0 +1,86 @@
...
+  end
+
+private
+
+  def update_live_edition(edition)
+    edition.update!(
+      revision_synced: false,
+      system_political: PoliticalEditionIdentifier.new(edition).political?,
+      government_id: government_id(edition, document),
+    )
```
kevindew: since we are mutating an edition (potentially multiple times) we should lock them so they're not vulnerable to any concurrent changes - we'll also have to run the whole thing within a transaction.

app/services/resync_service.rb
```diff
@@ -0,0 +1,86 @@
...
+  attr_reader :document
+
+  def initialize(document)
+    @document = document
+  end
+
+  def call
+    update_live_edition(document.live_edition) if document.live_edition
+    update_current_edition(document.current_edition) unless
+      document.current_edition == document.live_edition
```
kevindew: I think these method names could be re-considered to capture what they do - as I find the use of `update` a bit confusing. You might be able to have `sync_live_edition` and `sync_draft_edition`

spec/services/resync_service_spec.rb
```diff
@@ -0,0 +1,142 @@
...
+          ResyncService.call(document)
+        end
+      end
+
+      context "when the live edition is unpublished with redirect" do
+        before do
+          document.current_edition.stub(:removed?).and_return(true)
+          document.current_edition.status.details.stub(:redirect?).and_return(true)
+          document.current_edition.status.details.stub(:alternative_path).and_return("/foo/bar")
+          document.current_edition.status.details.stub(:explanatory_note).and_return("Explanation")
```
kevindew: It'd be more consistent with our existing code if we had a factory to set this up rather than stubbing. It doesn't look like we have a `:removed` trait yet on an edition factory but it should be pretty straight forward to set one up and this could be like:

```
let(:edition) do
  create(:edition, :removed, redirect: true, alternative_path: "/foo/bar", expanatory_note: "Explanation")
end
```

then we could call document from the edition

spec/services/resync_service_spec.rb
```diff
@@ -0,0 +1,142 @@
...
+
+    context "when there are both live and current editions" do
+      let(:document) { create(:document, :with_current_and_live_editions) }
+
+      it "updates the system_political value associated with both editions" do
+        expect(document.live_edition.system_political).to be false
+        expect(document.current_edition.system_political).to be false
+        ResyncService.call(document)
+        expect(document.live_edition.system_political).to be true
+        expect(document.current_edition.system_political).to be true
```
kevindew: Might want to refactor this as: 

```
expect { ResyncService.call(document) }
  .to change { document.live_edition.system_political }.to(true)
  .and change { document.current_edition.system_political }.to(true)
```

You also might want to mock the PoliticalEditionIdentifier to return true otherwise I'm not really sure what about this document is supposed to mean that this content is political.

spec/services/resync_service_spec.rb
```diff
@@ -0,0 +1,142 @@
...
+      end
+
+      it "updates the government_id associated with the live edition" do
+        expect(document.live_edition.government_id).to be nil
+        ResyncService.call(document)
+        expect(document.live_edition.government_id).to eq "d4fbc1b9-d47d-4386-af04-ac909f868f92"
```
kevindew: We should do some mocking here as this is the id of the current UK government and this test would start failing when that government changed.

app/services/resync_service.rb
ChrisBAshton: I was going to rename them later when they're actually syncing with publishing API 👍 think I forgot to do that. In the early commits all they do is update the local models.

spec/services/resync_service_spec.rb
```diff
@@ -0,0 +1,146 @@
...
+      allow(PoliticalEditionIdentifier)
+        .to receive(:new)
+        .and_return(instance_double(PoliticalEditionIdentifier, political?: true))
+    end
+
+    context "when there is no live edition" do
+      let(:document) { create(:document, :with_current_edition) }
+
+      it "synchronises the current edition, but does not publish" do
+        expect(PreviewService).to receive(:call).with(document.current_edition).and_call_original
```
kevindew: Do you need the call_original?

leenagupte: Yes, so we can check that the revision has been synced,

ChrisBAshton:
Superseded by #1505.

PR #1490, opened 2019-12-02
Import scheduled Whitehall documents

AgaDufrat:
We want to have the ability to import a document that is scheduled for publication.

In Whitehall you can determine that an edition is scheduled as it has a state of “scheduled” and a field of `scheduled_publication` with the time.
In Content Publisher the scheduled publishing is represented by a status of “scheduled” with a Scheduling object that knows:
- who created the scheduling. This is handled in the existing `build_status` method
- whether the content was reviewed before scheduling. This is set based on the value of `force_published` on the `whitehall_edition`. If this is true, reviewed should is false otherwise this is true.
- the status prior to the scheduling to allow rolling back to that. We determine the status using whitehall document's `revision_history` by checking if the previous one is "submitted". If it isn't we the pre scheduled status is "draft".
- publish time

importing_scheduled

- **After unscheduling**

submitted_unshceduled
draft_unscheduled

[Trello card](https://trello.com/c/nsm9dRgz/1192-import-a-scheduled-whitehall-document-into-content-publisher)

### DIFF
--- a/lib/whitehall_importer/create_edition.rb
+++ b/lib/whitehall_importer/create_edition.rb
@@ -87,7 +87,9 @@ module WhitehallImporter
     end
 
     def set_withdrawn_status(edition)
-      raise AbortImportError, "Cannot create withdrawn status without an unpublishing" if whitehall_edition["unpublishing"].blank?
+      unless whitehall_edition["unpublishing"]
+        raise AbortImportError, "Cannot create withdrawn status without an unpublishing"
+      end
 
       withdrawal = Withdrawal.new(
         published_status: edition.status,
@@ -99,7 +101,9 @@ module WhitehallImporter
     end
 
     def create_scheduled_edition
-      raise AbortImportError, "Cannot create scheduled status without scheduled_publication" if whitehall_edition["scheduled_publication"].blank?
+      unless whitehall_edition["scheduled_publication"]
+        raise AbortImportError, "Cannot create scheduled status without scheduled_publication"
+      end
 
       pre_scheduled_state = history.state_event("submitted") ? "submitted_for_review" : "draft"
       edition = create_edition(status: build_status(pre_scheduled_state),
@@ -114,7 +118,9 @@ module WhitehallImporter
 
     def build_removal
       unpublishing = whitehall_edition["unpublishing"]
-      raise AbortImportError, "Cannot create removal status without an unpublishing" if unpublishing.blank?
+      unless unpublishing
+        raise AbortImportError, "Cannot create removal status without an unpublishing"
+      end
 
       Removal.new(
         explanatory_note: unpublishing["explanation"],
--- a/lib/whitehall_importer/create_edition.rb
+++ b/lib/whitehall_importer/create_edition.rb
@@ -101,7 +101,8 @@ module WhitehallImporter
     def create_scheduled_edition
       raise AbortImportError, "Cannot create scheduled status without scheduled_publication" if whitehall_edition["scheduled_publication"].blank?
 
-      edition = create_edition(status: build_status("submitted_for_review"),
+      pre_scheduled_state = history.state_event("submitted") ? "submitted_for_review" : "draft"
+      edition = create_edition(status: build_status(pre_scheduled_state),
                                current: current,
                                edition_number: edition_number)
       scheduling = Scheduling.new(pre_scheduled_status: edition.status,
--- a/lib/whitehall_importer/create_edition.rb
+++ b/lib/whitehall_importer/create_edition.rb
@@ -99,6 +99,8 @@ module WhitehallImporter
     end
 
     def create_scheduled_edition
+      raise AbortImportError, "Cannot create scheduled status without scheduled_publication" if whitehall_edition["scheduled_publication"].blank?
+
       edition = create_edition(status: build_status("submitted_for_review"),
                                current: current,
                                edition_number: edition_number)
--- a/lib/whitehall_importer/create_edition.rb
+++ b/lib/whitehall_importer/create_edition.rb
@@ -51,7 +51,7 @@ module WhitehallImporter
     end
 
     def split_unpublished_edition
-      unpublishing_event = history.last_unpublishing_event
+      unpublishing_event = history.last_unpublishing_event!
       create_edition(
         status: build_status("removed", build_removal),
         current: false,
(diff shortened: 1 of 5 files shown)


lib/whitehall_importer/create_edition.rb
```diff
@@ -96,6 +98,28 @@ def set_withdrawn_status(edition)
...
+      raise AbortImportError, "Cannot create scheduled status without scheduled_publication" if whitehall_edition["scheduled_publication"].blank?
+
+      scheduling = Scheduling.new(
+        pre_scheduled_status: edition.status,
+        reviewed: !whitehall_edition["force_published"],
+        publish_time: whitehall_edition["scheduled_publication"],
+      )
+
+      edition.update!(status: build_status("scheduled", scheduling))
+    end
```
kevindew: I wonder if these 3 methods will read nicer as a single one, since not that much actually goes on in them and that would save us from a variety of scheduled methods methods only used in this context.

```
def create_scheduled_edition
  raise AbortImportError, "Cannot create scheduled status without scheduled_publication" if whitehall_edition["scheduled_publication"].blank?

  pre_scheduled_state = history.state_event("submitted") ? "submitted_for_review" : "draft"
  edition = create_edition(status: build_status(pre_scheduled_state), 
                           current: current, 
                           edition_number: edition_number)

  scheduling = Scheduling.new(pre_scheduled_status: status,
                              reviewed: !whitehall_edition["force_published"],
                              publish_time: whitehall_edition["scheduled_publication"])
  edition.update!(status: build_status("scheduled", scheduling)
end
```

What do you think?

spec/lib/whitehall_importer/create_edition_spec.rb
```diff
@@ -224,4 +224,130 @@
...
+      expect(edition.statuses.count).to eq(2)
+    end
+
+    it "sets the correct scheduled status" do
+      expect(edition).to be_scheduled
+    end
+
+    it "sets the correct previous status" do
+      expect(edition.statuses.first).to be_submitted_for_review
+    end
```
kevindew: I'm not sure if these 3 things need to be separate tests, we can probably just have one test that captures that the state history is submitted_for_review then scheduled.

spec/lib/whitehall_importer/create_edition_spec.rb
```diff
@@ -224,4 +224,130 @@
...
+
+    it "sets the correct scheduling metadata" do
+      expect(edition.status.details.pre_scheduled_status).to be_submitted_for_review
+      expect(edition.status.details.reviewed).to be true
+      expect(edition.status.details.publish_time).to eq(
+        whitehall_edition["scheduled_publication"],
+      )
+    end
```
kevindew: 👍

spec/lib/whitehall_importer/create_edition_spec.rb
```diff
@@ -224,4 +224,130 @@
...
+          build(:revision_history_event, event: "update", state: "submitted"),
+          build(:revision_history_event, event: "update", state: "scheduled"),
+        ],
+        scheduled_publication: future_date,
+      )
+    end
+
+    let(:edition) do
+      described_class.call(document: document, whitehall_edition: whitehall_edition)
+    end
```
kevindew: We should try avoid the practice we've used here where the code that is being under test is called outside of the test. As the effect of it is that when you're reading the tests you're having to go off to other places to work out where the code that is being tested is called. 

It looks like you followed precedent from withdraw here which unfortunately already had this pattern whereas Pete didn't use that for unpublishing.

lib/whitehall_importer/edition_history.rb
```diff
@@ -6,14 +6,18 @@ def initialize(revision_history)
...
     end
 
-    def state_event(state)
+    def state_event!(state)
       event = revision_history.select { |h| h["state"] == state }.last
 
       raise AbortImportError, "Edition is missing a #{state} event" unless event
 
       event
     end
```
kevindew: I'd suggest adjusting this to be:

```
state_event(state) || raise AbortImportError, "Edition is missing a #{state} event"
```

so we don't need to duplicate the logic.

It's probably worth doing the same bang equivalents on the create_event and next_event methods to for consistency.

spec/lib/whitehall_importer/create_edition_spec.rb
```diff
@@ -292,5 +275,69 @@
         described_class.call(document: document, whitehall_edition: whitehall_edition)
       }.to raise_error(WhitehallImporter::AbortImportError)
     end
+
+    context "when a draft document was force scheduled" do
```
ChrisBAshton: This feels like quite a lot of overhead to be adding when the only difference between the two contexts is the value of the second revision history event. Hopefully more liberal use of RSpec `let` blocks as suggested earlier will provide a path for making this more concise.

lib/whitehall_importer/edition_history.rb
```diff
@@ -14,6 +14,10 @@ def state_event!(state)
...
 
+    def state_event(state)
```
ChrisBAshton: Typo in the commit message where this method is added. `exepction`

lib/whitehall_importer/create_edition.rb
```diff
@@ -99,6 +99,8 @@ def set_withdrawn_status(edition)
     end
 
     def create_scheduled_edition
+      raise AbortImportError, "Cannot create scheduled status without scheduled_publication" if whitehall_edition["scheduled_publication"].blank?
```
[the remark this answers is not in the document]
kevindew: Good shout - that's waaayyy past the 80 char mark.

spec/lib/whitehall_importer/create_edition_spec.rb
```diff
@@ -224,4 +224,56 @@
...
+        state: "scheduled",
+        revision_history: [
+          build(:revision_history_event),
+          build(:revision_history_event, event: "update", state: "submitted"),
+          build(:revision_history_event, event: "update", state: "scheduled"),
+        ],
+        scheduled_publication: future_date,
+      )
+    end
+
```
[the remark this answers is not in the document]
kevindew: `subject` brings with it a share of complexity. There's often disagreement over whether subject refers to what the method being tested returns or to the class itself. There is disagreement about whether to refer to subject by name or not. As this is not an approach that's rooted in the Content Publisher testing I'm keen not to let it leak in here as it'll lead to inconsistency, although I do appreciate the points.

The approach I'm going to take is see if I can cut out that where we're writing the same call over and over.

kevindew:
@ChrisBAshton this should be good for another look now.

PR #1493, opened 2019-12-03
Import attachments

1pretz1:
When importing data from Whitehall into Content Publisher we want to import the images and attachments of all the editions of a document.

Below are some screenshots of testing importing a document with file attachments with this branch on integration:

## Edition has valid file attachment

Screenshot 2019-12-05 at 11 07 14

## Edition has valid small file attachment (2b)
Screenshot 2019-12-05 at 14 44 52

Trello:
https://trello.com/c/4FF2oprK/1182-import-attachments-into-content-publisher


spec/factories/whitehall_export/file_attachment_factory.rb
```diff
@@ -0,0 +1,75 @@
...
+        "id" => 1,
+        "carrierwave_file" => filename,
+        "content_type" => "text/plain",
+        "file_size" => 1057,
+        "number_of_pages" => nil,
+        "created_at" => created_at,
+        "updated_at" => updated_at,
+        "replaced_by_id" => nil,
+        "uploaded_to_asset_manager_at" => updated_at,
+        "present_at_unpublish" => false,
```
ChrisBAshton: This value should be `nil` I think? (based on https://whitehall-admin.integration.publishing.service.gov.uk/government/admin/export/document/165362)

lib/whitehall_importer/create_file_attachment_revision.rb
```diff
@@ -0,0 +1,33 @@
...
+
+  private
+
+    attr_reader :whitehall_file_attachment
+
+    def download_file
+      file = URI.parse(whitehall_file_attachment["url"]).open
+      if file.is_a?(StringIO)
+        # files less than 10 KB return StringIO (we have to manually cast to a tempfile)
+        Tempfile.new.tap { |tmp| File.write(tmp.path, file.string) }
```
ChrisBAshton: Worth adding something to the commit message that this method is copied from another class?
And a note that we should look at extracting it out to its own helper later?

spec/lib/whitehall_importer/create_file_attachment_revision_spec.rb
```diff
@@ -32,5 +32,18 @@
         "File attachment does not exist: #{whitehall_file_attachment['url']}",
       )
     end
+
+    it "for a file attachment with requirements issues" do
+      too_long_title = ("A" * Requirements::FileAttachmentChecker::TITLE_MAX_LENGTH + "A")
```
ChrisBAshton: Think this is a little easier to read:

```suggestion
      too_long_title = (Requirements::FileAttachmentChecker::TITLE_MAX_LENGTH * "A") + "A"
```

spec/lib/whitehall_importer/create_file_attachment_revision_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+      expect(metadata_revision.title).to eq(whitehall_file_attachment["title"])
+    end
+
+    it "creates a FileAttachment::Asset" do
+      expect { described_class.call(whitehall_file_attachment) }
+        .to change { FileAttachment::Asset.count }.by(1)
+    end
+  end
+
+  context "aborts creating a file attachment" do
```
ChrisBAshton: This feels like an odd use for contexts. You're using it to group related tests, but what relates them is the outcome rather than the setup conditions.

I'm not a fan of [shared examples](https://relishapp.com/rspec/rspec-core/v/2-11/docs/example-groups/shared-examples) but think this would be a good candidate for them.

```ruby
  shared_examples "rejected file attachment" do
    it "raises an AbortImportError with an informative error" do
      create_revision = described_class.new(whitehall_file_attachment)
      expect { create_revision.call }.to raise_error(
        WhitehallImporter::AbortImportError,
        error_message,
      )
    end
  end

  context "file attachment URL is invalid" do
    let(:error_message) { "File attachment does not exist: #{whitehall_file_attachment['url']}" }

    before { stub_request(:get, whitehall_file_attachment["url"]).to_return(status: 404) }
    
    it_behaves_like "rejected file attachment"
  end

  context "file attachment does not satisfy requirements" do
    let(:too_long_title) { ("A" * Requirements::FileAttachmentChecker::TITLE_MAX_LENGTH + "A") }
    let(:whitehall_file_attachment) { build(:whitehall_export_file_attachment, title: too_long_title) }
    let(:error_message) do
      I18n.t!("requirements.title.too_long.form_message", max_length: Requirements::FileAttachmentChecker::TITLE_MAX_LENGTH)
    end
    
    it_behaves_like "rejected file attachment"
  end

  context "whitehall attachment is of an unsupported type" do
    let(:whitehall_file_attachment) { build(:whitehall_export_file_attachment, type: "ExternalAttachment") }
    let(:error_message) { "Unsupported file attachment: #{whitehall_file_attachment['type']}" }
    
    it_behaves_like "rejected file attachment"
  end
```

lib/whitehall_importer/create_file_attachment_revision.rb
```diff
@@ -0,0 +1,91 @@
...
+    end
+
+    def download_file
+      file = URI.parse(whitehall_file_attachment["url"]).open
+      if file.is_a?(StringIO)
+        # files less than 10 KB return StringIO (we have to manually cast to a tempfile)
+        Tempfile.new.tap { |tmp| File.write(tmp.path, file.string) }
+      else
+        file
+      end
```
kevindew: Should we move this dealing with tempfiles somewhere else since we've got the same logic in this and for images. Alternatively is there a nicer way where can avoid the need for this StringIO handling? it might even be worth checking if this stuff is needed for attachment as we might be able to just work with a stringio instance within the decorated class.

lib/whitehall_importer/create_file_attachment_revision.rb
```diff
@@ -0,0 +1,91 @@
...
+  end
+
+  class AttachmentFileDecorator < SimpleDelegator
+    attr_reader :original_filename, :content_type
+
+    def initialize(tmp_file, original_filename)
+      super(tmp_file)
+      @original_filename = original_filename
+    end
+  end
```
kevindew: 👍

lib/whitehall_importer/create_file_attachment_revision.rb
```diff
@@ -0,0 +1,91 @@
...
+
+    def abort_on_issue(issues)
+      return if issues.empty?
+
+      raise WhitehallImporter::AbortImportError, issues.first.message(style: "form")
+    end
+  end
+
+  class AttachmentFileDecorator < SimpleDelegator
+    attr_reader :original_filename, :content_type
```
kevindew: might be prudent to explicitly set @content_type to nil to show it's intentional

spec/features/editing_file_attachments/replace_file_attachment_file_spec.rb
```diff
@@ -27,7 +27,7 @@ def when_i_click_to_insert_an_attachment
     visit edit_document_path(@edition.document)
     find("markdown-toolbar details").click
     click_on "Attachment"
-    expect(page).to have_content("0 Bytes")
+    expect(page).to have_content(@attachment_revision.blob_revision.byte_size)
```
kevindew: I'd suggest using a bytes term here since it can be a more realistic end-to-end test when not relying on computed values (since we don't know their validity).

We have a precedent for using size as part of the naming of fixtures which makes this easier. For example 13kb-1-page-attachment.pdf

spec/lib/whitehall_importer/create_file_attachment_revision_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+      expect { metadata_revision = described_class.call(whitehall_file_attachment) }
+        .to change { FileAttachment::MetadataRevision.count }.by(1)
+
+      expect(metadata_revision.title).to eq(whitehall_file_attachment["title"])
+    end
+
+    it "creates a FileAttachment::Asset" do
+      expect { described_class.call(whitehall_file_attachment) }
+        .to change { FileAttachment::Asset.count }.by(1)
+    end
```
kevindew: We can probably test all these with a single test as per: https://github.com/alphagov/content-publisher/blob/920864a5e072918fe2475f1ab5340b2fde8c47d9/spec/lib/whitehall_importer/create_image_revision_spec.rb#L6-L14

From a quick glance I think this test could use being a bit more consistent with that one.

spec/factories/whitehall_export/file_attachment_factory.rb
1pretz1: Hmm, the export I tried (screenshot in the trello card https://trello.com/c/4FF2oprK/1182-import-attachments-into-content-publisher) had a value of `false`, maybe the value can either be false, true or nil.

lib/whitehall_importer/create_file_attachment_revision.rb
1pretz1: Agree with you, I implemented it's own class and then saw Kev's comment so reverted to how it was before just without the StringIO bit as it's actually not needed for file attachments.

spec/lib/whitehall_importer/create_file_attachment_revision_spec.rb
1pretz1: Hmm ruby didn't like that approach and threw and error converting a string to fixnum so I went with `("A" * Requirements::FileAttachmentChecker::TITLE_MAX_LENGTH) + "A"`, don't quite understand why ruby accepted that way round of doing it! 🤔

spec/lib/whitehall_importer/create_file_attachment_revision_spec.rb
1pretz1: Implemented `shared_examples` approach, haven't used this before, interesting!

app/services/file_attachment_blob_service.rb
```diff
@@ -1,21 +1,20 @@
 # frozen_string_literal: true
 
 class FileAttachmentBlobService < ApplicationService
-  def initialize(revision, user, file, replacing: nil)
-    @revision = revision
+  def initialize(temp_file_attachment:, filename:, user: nil)
```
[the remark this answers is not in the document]
1pretz1: Yep I felt the same but was trying to keep the code consistent... done!

lib/whitehall_importer/create_file_attachment_revision.rb
1pretz1: Did this - `Should we move this dealing with tempfiles somewhere else since we've got the same logic in this and for images`

Turns out we don't even need the StringIO bit as you though, tested in integration importing a 2kb -> StringIO file and imported successfully! 👍

lib/whitehall_importer/create_file_attachment_revision.rb
1pretz1: Sure

spec/features/editing_file_attachments/replace_file_attachment_file_spec.rb
1pretz1: Makes sense, I suppose that's at the compromise of having to changing the values manually if the files change.

spec/lib/whitehall_importer/create_file_attachment_revision_spec.rb
1pretz1: Sure, I've merged the tests where it's testing for metadata and blob record being created into whether the revision has been created as a file_attachment_revision must have metadata and blob record to of been created!

1pretz1:
Ready for review number 2 @ChrisBAshton @kevindew 😄

kevindew:
@1pretz1 not sure you meant that last screenshot?

1pretz1:
I did initially just to show it fail for a none 'FileAttachment' whitehall type, but It doesn't really need to be on the PR you're right @kevindew!

spec/features/editing_file_attachments/replace_file_attachment_file_spec.rb
```diff
@@ -27,7 +27,7 @@ def when_i_click_to_insert_an_attachment
     visit edit_document_path(@edition.document)
     find("markdown-toolbar details").click
     click_on "Attachment"
-    expect(page).to have_content("0 Bytes")
+    expect(page).to have_content("74 Bytes")
```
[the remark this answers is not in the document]
1pretz1: Done!

spec/lib/whitehall_importer/create_file_attachment_revision_spec.rb
```diff
@@ -11,7 +11,7 @@
       expect(create_revision.call).to have_requested(:get, whitehall_file_attachment["url"])
     end
 
-    it "creates a FileAttachment::Revision and set correct metadata" do
+    it "creates a FileAttachment::Revision and sets correct metadata" do
```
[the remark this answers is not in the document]
1pretz1: Done

1pretz1:
Fixed typo and the examples are now using `Time.current` 🌮

PR #1495, opened 2019-12-04
Add in history mode status banner

JonathanHallam:
As part of the work to introduce history mode to content publisher we need to let publishers know when a document is in history mode. This introduces a notice that warns publishers when a political document is marked as being from a previous government. This feature is not yet ready for release and as such is hidden behind a feature flag. This will be removed at a later date.

## Before

Screenshot 2019-12-04 at 12 13 29

## After

Screenshot 2019-12-04 at 12 13 57

Trello - https://trello.com/c/rhEtuy5V/1208-show-history-mode-status-on-summary-view

Co-authored-by: Bruce Bolt <bruce.bolt@digital.cabinet-office.gov.uk>

### DIFF
--- a/app/views/documents/show/_historical_notice.html.erb
+++ b/app/views/documents/show/_historical_notice.html.erb
@@ -1,0 +1,12 @@
+<%= render "govuk_publishing_components/components/notice",
+  {
+    title: I18n.t(
+      "documents.show.historical.title",
+      document_type: @edition.document_type.label.downcase
+    ),
+    description_text: I18n.t(
+      "documents.show.historical.description",
+      government_name: @edition.government.name
+    ),
+  }
+%>
--- a/spec/requests/historical_banner_spec.rb
+++ b/spec/requests/historical_banner_spec.rb
@@ -1,0 +1,31 @@
+# frozen_string_literal: true
+
+RSpec.describe "Documents" do
+  describe "GET /documents/:document" do
+    describe "history mode banner" do
+      context "when in history mode" do
+        it "shows the history mode banner" do
+          @edition = create(:edition, :political, :past_government)
+          get document_path(@edition.document)
+          expect(response.body).to include(I18n.t!("documents.show.historical.title", document_type: @edition.document_type.label.downcase))
+          expect(response.body).to include(I18n.t!("documents.show.historical.description", government_name: @edition.government.name))
+        end
+
+        it "won't show the history mode banner when the edition was created under the current government" do
+          @edition = create(:edition, :political, :current_government)
+          get document_path(@edition.document)
+          expect(response.body).not_to include(I18n.t!("documents.show.historical.title", document_type: @edition.document_type.label.downcase))
+          expect(response.body).not_to include(I18n.t!("documents.show.historical.description", government_name: @edition.government.name))
+        end
+      end
+
+      context "when not in history_mode" do
+        it "won't show the history mode banner" do
+          @edition = create(:edition, :not_political)
+          get document_path(@edition.document)
+          expect(response.body).not_to include(I18n.t!("documents.show.historical.title", document_type: @edition.document_type.label.downcase))
+        end
+      end
+    end
+  end
+end
--- a/app/views/documents/show/_summary_tab.html.erb
+++ b/app/views/documents/show/_summary_tab.html.erb
@@ -4,6 +4,10 @@
       <%= render "documents/show/submitted_for_review" %>
     <% end %>
 
+    <% if @edition.history_mode? && current_user.has_permission?(User::PRE_RELEASE_FEATURES_PERMISSION) %>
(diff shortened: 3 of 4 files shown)


app/views/documents/show/_historical_notice.html.erb
```diff
@@ -0,0 +1,6 @@
+<%= render "govuk_publishing_components/components/notice", {
+  title: I18n.t("documents.show.historical.title",
+                document_type: @edition.document_type.label.downcase),
+  description_text: I18n.t("documents.show.historical.description",
```
ChrisBAshton: This alignment seems a bit off to me - I suppose it's to keep the line length below 80 - would it be neater if everything were in its own JSON-style block?

```ruby
{
  document_type: @edition.document_type.label.downcase),
  description_text: I18n.t("documents.show.historical.description",
  government_name: @edition.government.name),
}
```

spec/requests/historical_banner_spec.rb
```diff
@@ -0,0 +1,23 @@
...
+    @edition = create(:edition, :published, :political, :current_government)
+    get document_path(@edition.document)
+    expect(response.body).not_to include(I18n.t!("documents.show.historical.title", document_type: @edition.document_type.label.downcase))
+    expect(response.body).not_to include(I18n.t!("documents.show.historical.description", government_name: @edition.government.name))
+  end
+
+  it "won't show the history mode banner when the edition is not political" do
+    @edition = create(:edition, :published, :not_political)
+    get document_path(@edition.document)
+    expect(response.body).not_to include(I18n.t!("documents.show.historical.title", document_type: @edition.document_type.label.downcase))
```
ChrisBAshton: Just for consistency, this test is not checking `title` _and_ `description` like the "created under the current government" test above. I suggest if it's needed for one, it's needed for the other - and if it's not needed for one, it's not needed for the other. I'd be tempted to move these assertions into a method we can call from both tests.

spec/requests/historical_banner_spec.rb
```diff
@@ -0,0 +1,23 @@
...
+  end
```
ChrisBAshton: Are we missing a test for not showing history mode banner if user does not have the prerelease permission?

kevindew: Our precedent is to avoid testing whether this is off and on as it gets pretty messy and the logic for pre-release features is usually trivial. The test user is assumed to have pre-release permissions.

kevindew:
This is looking good - what are thoughts on using request tests rather than feature for this?

One thing I note is that that request test is set up somewhat like a particular feature, whereas it should be coupled to its route. For example:

```
RSpec.describe "Documents" do
  describe "GET /documents/:document" do
    describe "history mode banner" do
      context "when in history mode" ...
    end
  end
end

```

spec/requests/historical_banner_spec.rb
ChrisBAshton: Ok, cheers! 👍

app/views/documents/show/_historical_notice.html.erb
JonathanHallam: Will adjust, thanks :) and thanks for thinking it was something to do with linting, it's actually an oversight as the linter doesn't check these files. Will go with your interpretation though, makes me look better.

spec/requests/historical_banner_spec.rb
JonathanHallam: So that's done because the description relies on there being a government attached. This test doesn't have a government because it's not political. I think that we should be testing it for at least one test but we physically can't for this one. Mind if I leave this as is?

spec/requests/historical_banner_spec.rb
JonathanHallam: There's prior art on us not doing this. I assume because it will only be removed later. 
https://github.com/alphagov/content-publisher/pull/1251/files

JonathanHallam:
@kevindew Didn't realise the syntax was different, sorry, sorted!

spec/requests/historical_banner_spec.rb
ChrisBAshton: Sure 👍 thanks for explaining!

spec/requests/historical_banner_spec.rb
```diff
@@ -0,0 +1,31 @@
+# frozen_string_literal: true
+
+RSpec.describe "Documents" do
+  describe "GET /documents/:document" do
+    describe "history mode banner" do
+      context "when in history mode" do
+        it "shows the history mode banner" do
+          @edition = create(:edition, :published, :political, :past_government)
```
kevindew: I don't think you need :published here or in the other places where we have it in this test.

JonathanHallam: sorted!

PR #1496, opened 2019-12-04
Add legends to radios

alex-ju:
This PR updates the markup generated for the radios across the application ensuring that each instance has a legend. It doesn't contain any visual changes.

An example of using the main page heading as legend below:

### Before
```
<h1 class="govuk-heading-l">What is the content for?</h1>
...
<fieldset class="govuk-fieldset">
  <div class="govuk-radios" data-module="govuk-radios">
    <div class="gem-c-radio govuk-radios__item">
      <input type="radio" name="supertype" id="radio-0802af77-0" value="news" class="govuk-radios__input" aria-describedby="label-hint-20c762ca" data-gtm="choose-supertype" data-gtm-value="News" aria-controls="conditional-ced84d75" aria-expanded="false">
      <label for="radio-0802af77-0" class="gem-c-label govuk-label govuk-label--s govuk-radios__label">News</label>
    </div>
    ...
  </div>
</fieldset>
```

### After
```
<fieldset class="govuk-fieldset">
  <legend class="govuk-fieldset__legend govuk-fieldset__legend--xl gem-c-title">
    <h1 class="gem-c-title__text">What is the content for?</h1>
  </legend>
  <div class="govuk-radios" data-module="govuk-radios">
    <div class="gem-c-radio govuk-radios__item">
      <input type="radio" name="supertype" id="radio-0802af77-0" value="news" class="govuk-radios__input" aria-describedby="label-hint-20c762ca" data-gtm="choose-supertype" data-gtm-value="News" aria-controls="conditional-ced84d75" aria-expanded="false">
      <label for="radio-0802af77-0" class="gem-c-label govuk-label govuk-label--s govuk-radios__label">News</label>
    </div>
    ...
  </div>
</fieldset>
```

This problem is related to the fact that [the radio component](https://components.publishing.service.gov.uk/component-guide/radio) doesn't require a legend in the first place – I captured this as [an issue in govuk_publishing_components](https://github.com/alphagov/govuk_publishing_components/issues/1215) and will to address it separately. 

[Trello card](https://trello.com/c/QCUNqBIw)

### DIFF
--- a/app/assets/stylesheets/utilities/_overwrites.scss
+++ b/app/assets/stylesheets/utilities/_overwrites.scss
@@ -100,3 +100,13 @@
     }
   }
 }
+
+.govuk-fieldset {
+  .gem-c-title {
+    margin-bottom: 0;
+  }
+
+  .gem-c-title__text {
+    @extend %govuk-heading-l;
+  }
+}
--- a/app/views/access_limit/edit.html.erb
+++ b/app/views/access_limit/edit.html.erb
@@ -1,4 +1,3 @@
-<% content_for :title, t("access_limit.edit.title") %>
 <% content_for :browser_title, t("access_limit.edit.browser_title") %>
 <% content_for :back_link, render_back_link(href: document_path(@edition.document)) %>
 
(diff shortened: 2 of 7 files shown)


alex-ju:
Thank you for the review Chris!

> One typo in the last commit - `ragardless` 

Fixed.

> General feedback on the commit message formatting. (When the commit message gets longer, it's nice to split it out into several lines, manually chopping it up to around 80 characters long per line. This way it scans more easily, both on GitHub.com and in the terminal).

While I do not agree with the hard-wrapping rule (which is at 72 characters), I've updated these commits as it seems to be common practice across the team.

I don't agree with hard wrapping in general as content should be independent of the medium it's being presented on – different people use different clients to read commit messages and these are managing the wrapping. A common scenario is when people hard-wrap at ~80 and the client soft-wrapps at 72 then you end up with orphan words on various lines. It's also possible to set the width of the terminal or use a VIM plugin to do soft-wrapping.

Do you happen to know if we have any guidance in GDS around this?

> One last thing - the "Generate schedule page heading as legend" commit comes after the "Refactor change note markup for radios" - could it come before it instead? That way all the similar commits are closer together.

Swapped them around.

alex-ju:
> I think the GDS Way says we should do it though: https://gds-way.cloudapps.digital/standards/git.html#structure

Thanks, Chris. I will try and challenge that and see what are people's thoughts around it.

PR #1498, opened 2019-12-05
Filter index page based on political and history mode status

ChrisBAshton:
Adds two new filters to the index page:

- `Marked as political` => "", "Yes", "No"
- `In history mode` => "", "Yes", "No"

We did consider having labels for the latter as "Yes - past government", "No - current government", but these don't fully explain what constitutes history mode (e.g. a document published by a past government is not in history mode if it is not classed as political content).

Screenshot:

Trello card: https://trello.com/c/7ZtSXVFx/1211-filter-index-page-based-on-political-and-history-mode-status

### DIFF
--- a/app/views/documents/index/_filters.html.erb
+++ b/app/views/documents/index/_filters.html.erb
@@ -120,6 +120,26 @@
     %>
   </div>
 
+  <div class="govuk-form-group">
+    <%= render "govuk_publishing_components/components/label", {
+      text: t("documents.index.filter.history_mode"),
+      html_for: "history_mode-filter",
+      bold: true
+    } %>
+
+    <%= select_tag "history_mode",
+      options_for_select([["Yes", "yes"], ["No", "no"]], [params[:history_mode]]),
+      include_blank: true,
+      id: "history_mode-filter",
+      class: "govuk-select",
+      data: {
+        gtm: "select-history_mode",
+        "gtm-visibility-tracking": params[:history_mode].present? || nil,
+        "gtm-value": params[:history_mode]
+      }
+    %>
+  </div>
+
   <%= hidden_field_tag "sort", @sort %>
 
   <%= render "govuk_publishing_components/components/button", {
--- a/app/views/documents/index/_filters.html.erb
+++ b/app/views/documents/index/_filters.html.erb
@@ -100,45 +100,47 @@
     %>
   </div>
 
-  <div class="govuk-form-group">
-    <%= render "govuk_publishing_components/components/label", {
-      text: t("documents.index.filter.political"),
-      html_for: "political-filter",
-      bold: true
-    } %>
+  <% if current_user.has_permission?(User::PRE_RELEASE_FEATURES_PERMISSION) %>
+    <div class="govuk-form-group">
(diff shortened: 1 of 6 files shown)


app/views/documents/index/_filters.html.erb
```diff
@@ -100,6 +100,26 @@
...
 
+  <div class="govuk-form-group">
+    <%= render "govuk_publishing_components/components/label", {
+      text: t("documents.index.filter.political"),
+      html_for: "political-filter",
+      bold: true
+    } %>
+
+    <%= select_tag "political",
+      options_for_select([["Yes", "include_political"], ["No", "exclude_political"]], [params[:political]]),
```
kevindew: Might be simpler to have these refer to yes or no rather than having the include_political and exclude_political, not sure why they'd need to differ from the defaults.

lib/edition_filter.rb
```diff
@@ -73,6 +73,12 @@ def filtered_scope(scope)
         end
       when :organisation
         memo.merge(TagsRevision.tagged_organisations_include(value))
+      when :political
+        if value == "include_political"
+          memo.political
+        else
+          memo.political(false)
```
kevindew: Can probably simplify this down to `memo.political(value == "include_political")`

spec/lib/edition_filter_spec.rb
```diff
@@ -83,6 +83,17 @@
       expect(editions).to match_array([edition1, edition2])
     end
 
+    it "filters the editions by political status" do
+      edition1 = create(:edition)
```
kevindew: might be more explicit to tag this as `not_political`

app/views/documents/index/_filters.html.erb
ChrisBAshton: We were basing it off the HTML in the prototype, but agreed, `yes`/`no` would be fine as values 👍

kevindew: That makes sense, we should be cautious of that since the HTML isn't wrote by a developer.

By the way on that note in the design for the history mode option there is text of "Yes - current government" and "No - past government" which don't seem to make sense as these yes and no don't associate with current or past government. So worth considering that as Yes/No too.

ChrisBAshton: Changed. cc'ing @gclssvglx just to keep him in the loop.

Graham, we changed this because a document may not be in history mode for a number of reasons, not just whether it's connected to a past government.

lib/edition_filter.rb
```diff
@@ -73,6 +73,10 @@ def filtered_scope(scope)
         end
       when :organisation
         memo.merge(TagsRevision.tagged_organisations_include(value))
+      when :political
+        memo.political(value == "yes")
+      when :history_mode
+        memo.history_mode(value == "yes")
```
kevindew: Oh this was nice and simple in the end 👍

ChrisBAshton:
Great, thanks! I'll suggest holding off on merging until Monday, and letting @gclssvglx hit the Merge button. Might be a good chance to show him how we deploy to staging and production too.

I'll be WFH but maybe another dev could pair on the deploying?

PR #1504, opened 2019-12-06
Show political status and allow status to be overridden

brucebolt:
Publishers need to be able to see the political status of their documents.  Managing editors need to be able to override the automatically determined political status for their documents.

To this end, this PR adds a "Political content" value to the "Content settings" section of the document summary page.  This also has a link to change the political status, which can only be done when the edition is editable (i.e. it is draft).  Non-managing editors will see a message asking them to contact their managing editor, whilst managing editors will see a form with radio buttons.  Once the status has been overridden, a timeline history entry is added and the preview updated (in case the document has been backdated and a history mode banner needs to be shown).

This sits behind a pre-release feature flag until history mode is fully released.

Trello card: https://trello.com/c/ambyZ2ia

### DIFF
--- a/app/controllers/political_controller.rb
+++ b/app/controllers/political_controller.rb
@@ -1,0 +1,17 @@
+# frozen_string_literal: true
+
+class PoliticalController < ApplicationController
+  before_action :check_permission
+
+  def edit
+    @edition = Edition.find_current(document: params[:document])
+    assert_edition_state(@edition, &:editable?)
+  end
+
+  def check_permission
+    return if current_user.has_permission?(User::MANAGING_EDITOR_PERMISSION)
+
+    @edition = Edition.find_current(document: params[:document])
+    render :non_managing_editor, status: :forbidden
+  end
+end
--- a/app/controllers/political_controller.rb
+++ b/app/controllers/political_controller.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 class PoliticalController < ApplicationController
+  before_action { authorise_user!(User::PRE_RELEASE_FEATURES_PERMISSION) }
   before_action :check_permission
 
   def edit
--- a/app/controllers/political_controller.rb
+++ b/app/controllers/political_controller.rb
@@ -8,6 +8,11 @@ class PoliticalController < ApplicationController
     assert_edition_state(@edition, &:editable?)
   end
 
+  def update
+    result = Political::UpdateInteractor.call(params: params, user: current_user)
+    redirect_to document_path(result.edition.document)
+  end
+
   def check_permission
     return if current_user.has_permission?(User::MANAGING_EDITOR_PERMISSION)
 
--- a/app/interactors/political/update_interactor.rb
+++ b/app/interactors/political/update_interactor.rb
@@ -1,0 +1,46 @@
+# frozen_string_literal: true
+
+class Political::UpdateInteractor < ApplicationInteractor
+  delegate :params,
+           :user,
+           :edition,
+           :next_revision,
+           to: :context
+
+  def call
+    Edition.transaction do
+      find_and_lock_edition
+      create_next_revision
+      create_timeline_entry
+      update_preview
+    end
+  end
+
+private
+
+  def find_and_lock_edition
+    context.edition = Edition.lock.find_current(document: params[:document])
+    assert_edition_state(edition, &:editable?)
+  end
+
+  def create_next_revision
+    updater = Versioning::RevisionUpdater.new(edition.revision, user)
+
+    updater.assign(editor_political: params[:political] == "yes")
+    EditEditionService.call(edition, user, revision: updater.next_revision)
+    edition.save!
+  end
+
+  def create_timeline_entry
+    TimelineEntry.create_for_revision(
+      entry_type: :political_status_changed,
+      revision: edition.revision,
+      edition: edition,
+      created_by: user,
+    )
+  end
+
+  def update_preview
+    FailsafePreviewService.call(edition)
+  end
+end
--- a/app/views/political/non_managing_editor.html.erb
+++ b/app/views/political/non_managing_editor.html.erb
@@ -1,0 +1,2 @@
+<% content_for :back_link, render_back_link(href: document_path(@edition.document)) %>
+<%= render partial: "errors/error", locals: { title: t("political.no_managing_editor_permission.title", title: @edition.title), body_govspeak: t("political.no_managing_editor_permission.body_govspeak") } %>
--- a/spec/features/editing_content_settings/political_spec.rb
+++ b/spec/features/editing_content_settings/political_spec.rb
@@ -1,0 +1,41 @@
+# frozen_string_literal: true
+
+RSpec.feature "History mode" do
+  scenario "not political document" do
+    given_there_is_a_not_political_document
+    when_i_visit_the_summary_page
+    then_i_see_that_the_content_is_not_political
+  end
+
+  scenario "political document" do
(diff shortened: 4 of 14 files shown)


spec/spec_helper.rb
```diff
@@ -35,6 +35,7 @@
   config.include GdsApi::TestHelpers::AssetManager
   config.include GovukSchemas::RSpecMatchers
   config.include AuthenticationHelper, type: :feature
+  config.include AuthenticationHelper, type: :request
```
1pretz1: You'll have to change lines 56 - 58 too: https://github.com/alphagov/content-publisher/pull/1504/commits/b5d4f17064dfbfbd12e8b6cfa1dc7176e05f0210#diff-93830fa29d616f7c87903d08b5b1b29aR56-R58 

Here's a snippet from my branch as I encountered the same thing when creating some requests tests (yours will probs be merged before mine) 😄 -
```
  config.after :each, type: lambda { |s| [:feature, :request].include?(s) } do
    reset_authentication
  end
```

app/controllers/political_controller.rb
```diff
@@ -0,0 +1,29 @@
...
+
+  def update
+    result = Political::UpdateInteractor.call(params: params, user: current_user)
+    redirect_to document_path(result.edition.document)
+  end
+
+  def check_pre_release_permission
+    return if current_user.has_permission?(User::PRE_RELEASE_FEATURES_PERMISSION)
+
+    head :forbidden
```
kevindew: I think rather than needing to do a conditional approach like this we can call have gds-sso handle it - check https://github.com/alphagov/content-publisher/blob/d95207d9118738e11a01382bd4e29da28bae3a01/app/controllers/debug_controller.rb#L4

That'll also mean a user gets served a page explaining too.

app/interactors/political/update_interactor.rb
```diff
@@ -0,0 +1,46 @@
...
+           :next_revision,
+           to: :context
+
+  def call
+    Edition.transaction do
+      find_and_lock_edition
+      create_next_revision
+      create_timeline_entry
+      update_preview
+    end
```
kevindew: Nice one 👍

spec/factories/user_factory.rb
```diff
@@ -7,7 +7,7 @@
     email { "someone@example.com" }
 
     trait :managing_editor do
-      permissions { [User::MANAGING_EDITOR_PERMISSION] }
+      permissions { [User::MANAGING_EDITOR_PERMISSION, User::PRE_RELEASE_FEATURES_PERMISSION] }
```
kevindew: I think @1pretz1 is likely to do something a bit more with permissions so this may change - this is just a heads up.

spec/features/editing_content_settings/political_spec.rb
```diff
@@ -0,0 +1,82 @@
...
+    when_i_visit_the_summary_page
+    then_i_see_that_the_content_is_not_political
+    and_i_do_not_see_the_history_mode_banner
+    when_i_click_to_change_the_status
+    then_i_enable_political_status
+    and_i_see_that_the_content_is_political
+    and_i_do_not_see_the_history_mode_banner
+    when_i_click_to_backdate_the_content
+    and_i_enter_a_date_to_backdate_the_content_to
+    and_i_see_the_history_mode_banner
```
kevindew: Nice - this covers loads of the feature!

spec/features/editing_content_settings/political_spec.rb
```diff
@@ -0,0 +1,82 @@
+# frozen_string_literal: true
+
+RSpec.feature "Political content" do
+  scenario "changing not political document to political and backdating enables history mode" do
```
kevindew: You shouldn't need to name the scenario since we only have one.

spec/features/editing_content_settings/political_spec.rb
```diff
@@ -0,0 +1,82 @@
+# frozen_string_literal: true
+
+RSpec.feature "Political content" do
```
kevindew: I'd suggest naming this history mode since that's a bit more common way of referring to the feature than political.

spec/interactors/political/update_interactor_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+    end
+
+    before { stub_any_publishing_api_put_content }
+
+    it "sets the edition to political" do
+      Political::UpdateInteractor.call(**args)
+      edition.reload
+
+      expect(edition.political?).to be true
+    end
```
kevindew: Should we have an inverse of this test?

spec/requests/political_edit_spec.rb
```diff
@@ -0,0 +1,23 @@
+# frozen_string_literal: true
+
+RSpec.describe "Documents" do
```
kevindew: I'd include this in the documents request spec. We hit a bit of a naming problem that I looked at earlier: https://github.com/alphagov/content-publisher/pull/1507

spec/requests/political_edit_spec.rb
```diff
@@ -0,0 +1,23 @@
...
+    end
+
+    context "with managing editor permission" do
+      it "does not return a forbidden status" do
+        @edition = create(:edition)
+        login_as(create(:user, :managing_editor))
+        get political_path(@edition.document)
+        expect(response.status).to eq(200)
+      end
+    end
```
kevindew: Might not these two to be in context blocks since that can probably be captured in the it description. 

We might want to have the POST endpoint covered too.

Also do we want to check the forbidden one shows the expected text?

spec/features/editing_content_settings/political_spec.rb
kevindew: May want to just check the presence of the timeline entry from this too?

spec/requests/documents_spec.rb
```diff
@@ -28,4 +28,38 @@
...
+      expect(response.body).to include(I18n.t!("political.no_managing_editor_permission.title", title: @edition.title))
+    end
+
+    it "does not return a forbidden status for user with managing editor permission" do
+      @edition = create(:edition)
+      login_as(create(:user, :managing_editor))
+      post political_path(@edition.document)
+      expect(response.status).to_not eq(403)
+    end
+  end
```
kevindew: Much nicer it's 👍

app/views/political/non_managing_editor.html.erb
```diff
@@ -0,0 +1,8 @@
+<% content_for :back_link, render_back_link(href: document_path(@edition.document)) %>
+<% content_for :title, t("political.no_managing_editor_permission.title", title: @edition.title) %>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <%= render_govspeak(t("political.no_managing_editor_permission.body_govspeak")) %>
+  </div>
+</div>
```
kevindew: Not a biggie but this view could be:

```
<%= render partial: "errors/error", locals: t(""political.no_managing_editor_permission") %>
```

spec/features/editing_content_settings/political_spec.rb
```diff
@@ -0,0 +1,89 @@
...
+  def then_i_enable_political_status
+    @request = stub_publishing_api_put_content(@edition.content_id, {})
+    choose(I18n.t!("political.edit.labels.political"))
+    click_on "Save"
+  end
+
+  def and_i_see_the_timeline_entry
+    within("#document-history") do
+      expect(page).to have_content(I18n.t!("documents.history.entry_types.political_status_changed"))
+    end
```
kevindew: Nice 👍

PR #1505, opened 2019-12-06
New service to re-sync documents with publishing-api

leenagupte:
Trello: https://trello.com/c/cT9yYO4X
Replaces PR #1487 

## What's changed and why?
Adds a service to re-sync a document with publishing-api. This will mainly be used in two scenarios:

1. Documents imported from Whitehall will need to be re-presented to publishing-api and have their base_path reservation moved from Whitehall to Content Publisher.

2. All documents will need to be represented to publishing-api to associate them with the correct government, so that if there is a new government, they can successfully transition into history mode.

co-authored with @ChrisBAshton

### DIFF
--- a/app/services/edit_edition_service.rb
+++ b/app/services/edit_edition_service.rb
@@ -28,9 +28,6 @@ private
   end
 
   def associate_with_government
-    edition.government_id = if edition.public_first_published_at
-                              government = Government.for_date(edition.public_first_published_at)
-                              government&.content_id
-                            end
+    edition.government_id = Government.for_date(edition.public_first_published_at)&.content_id
   end
 end
--- a/app/services/resync_service.rb
+++ b/app/services/resync_service.rb
@@ -23,6 +23,7 @@ private
   def sync_live_edition
     live_edition.lock!
     set_political_and_government(live_edition)
+    reserve_path(live_edition.base_path)
     PreviewService.call(live_edition, republish: true)
     PublishAssetService.call(live_edition, nil)
 
@@ -38,6 +39,7 @@ private
   def sync_draft_edition
     current_edition.lock!
     set_political_and_government(current_edition)
+    reserve_path(current_edition.base_path)
     FailsafePreviewService.call(current_edition)
 
     schedule if current_edition.scheduled?
@@ -48,6 +50,14 @@ private
       revision_synced: false,
       system_political: PoliticalEditionIdentifier.new(edition).political?,
       government_id: Government.for_date(edition.public_first_published_at)&.content_id,
+    )
+  end
+
+  def reserve_path(base_path)
+    GdsApi.publishing_api_v2.put_path(
+      base_path,
+      publishing_app: "content-publisher",
+      override_existing: true,
     )
   end
 
--- a/app/services/resync_service.rb
+++ b/app/services/resync_service.rb
@@ -22,6 +22,7 @@ private
 
   def sync_live_edition
     live_edition.lock!
+    set_political_and_government(live_edition)
     PreviewService.call(live_edition, republish: true)
     PublishAssetService.call(live_edition, nil)
 
@@ -36,9 +37,18 @@ private
 
   def sync_draft_edition
     current_edition.lock!
+    set_political_and_government(current_edition)
     FailsafePreviewService.call(current_edition)
 
     schedule if current_edition.scheduled?
+  end
+
+  def set_political_and_government(edition)
+    edition.update!(
+      revision_synced: false,
+      system_political: PoliticalEditionIdentifier.new(edition).political?,
+      government_id: Government.for_date(edition.public_first_published_at)&.content_id,
+    )
   end
 
   def publish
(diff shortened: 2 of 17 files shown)


spec/services/resync_service_spec.rb
```diff
@@ -50,5 +50,37 @@
...
+
+        withdraw_params = {
+          type: "withdrawal",
+          explanation: explanation,
+          locale: edition.locale,
+        }
+
+        expect(GovspeakDocument)
+          .to receive(:new)
+          .with(edition.status.details, edition)
```
kevindew: You can probably change this to:

```
        expect(GovspeakDocument)
          .to receive(:new)
          .with(edition.status.details, edition)
          .and_return(instance_double(GovspeakDocument, payload_html: explanation))
```
and remove the need for the allow on line 65

spec/factories/edition_factory.rb
```diff
@@ -119,6 +119,25 @@
...
 
+    trait :removed do
+      live { true }
+
+      transient do
+        removal { nil }
+      end
+
+      after(:build) do |edition, evaluator|
+        edition.revision = evaluator.association(:revision)
```
kevindew: I don't think we should or need to do this, it seems to be a mistake we have with withdrawn factory too.

spec/services/resync_service_spec.rb
```diff
@@ -82,5 +82,58 @@
...
       end
     end
+
+    context "when the live edition has been removed" do
+      let(:explanation) { "explanation" }
+
+      before do
+        stub_any_publishing_api_unpublish
+        allow(PreviewAssetService).to receive(:call)
+        allow(PublishAssetService).to receive(:call)
```
kevindew: Do these stubs need to be here? not sure if they have an impact.

app/services/edit_edition_service.rb
```diff
@@ -28,9 +28,6 @@ def determine_political
   end
 
   def associate_with_government
-    edition.government_id = if edition.public_first_published_at
-                              government = Government.for_date(edition.public_first_published_at)
-                              government&.content_id
-                            end
+    edition.government_id = Government.for_date(edition.public_first_published_at)&.content_id
```
kevindew: Nice 👍

lib/tasks/resync.rake
```diff
@@ -0,0 +1,21 @@
...
+    document = Document.find_by_param(args.content_id_and_locale)
+
+    raise "No document exists for #{content_id_and_locale}" unless document
+
+    ResyncService.call(document)
+  end
+
+  desc "Resync all documents with the publishing-api e.g. resync:all"
+  task all: :environment do
+    Document.all.each do |document|
```
kevindew: We should swap this to `Document.find_each` so it loads data in batches.

spec/lib/tasks/resync_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+        .with(document)
+
+      Rake::Task["resync:document"].invoke("#{document.content_id}:#{document.locale}")
+    end
+  end
+
+  describe "resync:all" do
+    it "resyncs all documentss" do
+      create(:document)
+      create(:document)
```
kevindew: minor but you can do `create_list(:document, 2)`

spec/services/resync_service_spec.rb
leenagupte: I was getting an because I'd set up the factory incorrectly. Will remove them.

leenagupte:
@kevindew Thanks for the review. I've addressed your comments and fixed up.

leenagupte:
> Shall we add an AC to the trello card about the gds-api-adapters change so we don't forget to do that?

New ticket added here: https://trello.com/c/NC10wudi/1276-add-a-test-helper-to-stub-the-putpath-method-in-gds-api-adapters

app/services/resync_service.rb
```diff
@@ -0,0 +1,42 @@
...
+    end
+  end
+
+private
+
+  def sync_live_edition
+    live_edition.lock!
+    PreviewService.call(live_edition, republish: true)
+    PublishAssetService.call(live_edition, nil)
+    publish
```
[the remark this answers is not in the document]
leenagupte: I don't think the edition is being updated without a bang (!), so I don't think a `save!` is needed. Why do you think it's needed?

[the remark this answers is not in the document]
ChrisBAshton: Ahh, you're right. We're calling `update!`, which calls `save!` under the hood. I was worried we weren't releasing the lock.

app/services/resync_service.rb
```diff
@@ -0,0 +1,106 @@
...
+
+  def redirect_or_remove
+    removal = live_edition.status.details
+    GdsApi.publishing_api_v2.unpublish(
+      live_edition.content_id,
+      type: removal.redirect? ? "redirect" : "gone",
+      explanation: removal.explanatory_note,
+      alternative_path: removal.alternative_path,
+      locale: live_edition.locale,
+      allow_draft: true,
```
kevindew: We probably should have an unpublished_at here as well.

leenagupte: @kevindew I've updated and fixed up

PR #1508, opened 2019-12-06
Set placeholder alt text for images

alex-ju:
Provide the user with a placeholder alt text for freshly uploaded images.

[Trello card](https://trello.com/c/6idOVWgU)

### DIFF
--- a/app/views/images/crop.html.erb
+++ b/app/views/images/crop.html.erb
@@ -25,7 +25,7 @@ content_for :title,t(title_key, title: @edition.title_or_fallback)
     src: url_for(@image_revision.blob),
     no_js_src: url_for(@image_revision.crop_variant('960x640')),
     exact_dimensions: @image_revision.at_exact_dimensions?,
-    alt_text: @image_revision.alt_text,
+    alt_text: @image_revision.alt_text || t("images.edit.alt_text_placeholder"),
     crop_x: @image_revision.crop_x,
     crop_y: @image_revision.crop_y,
     crop_width: @image_revision.crop_width,
--- a/config/locales/en/images/edit.yml
+++ b/config/locales/en/images/edit.yml
@@ -3,6 +3,7 @@ en:
     edit:
       title: "Image details for ‘%{title}’"
       title_modal: "Image details"
+      alt_text_placeholder: Your uploaded image
       form_labels:
         alt_text: Alt text
         caption: Image caption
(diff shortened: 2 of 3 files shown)


app/views/images/crop.html.erb
```diff
@@ -25,7 +25,7 @@ content_for :title,t(title_key, title: @edition.title_or_fallback)
     src: url_for(@image_revision.blob),
     no_js_src: url_for(@image_revision.crop_variant('960x640')),
     exact_dimensions: @image_revision.at_exact_dimensions?,
-    alt_text: @image_revision.alt_text,
+    alt_text: @image_revision.alt_text || t("images.edit.alt_text_placeholder"),
```
ChrisBAshton: I'm not sure this fits the brief of:

> We need to provide the user with a default/placeholder alt text in the markup for freshly uploaded images. This is not to be stored anywhere - only presented until the alt text is provided by the user.

If the user doesn't make any changes to the alt text, it would now by default be persisted as "Your uploaded image". I would interpret this requirement as being the `placeholder` attribute rather than the `value` of the input element.

config/locales/en/images/edit.yml
```diff
@@ -3,6 +3,7 @@ en:
     edit:
       title: "Image details for ‘%{title}’"
       title_modal: "Image details"
+      alt_text_placeholder: Your uploaded image
```
ChrisBAshton: Do we want to make this more of a call to action? e.g. "You must provide an alt value for this image", or "Describe this image to screen readers".

app/views/images/crop.html.erb
alex-ju: I don't think I follow this. How would a user end up with "Your uploaded image" be persisted? We're only using this text when rendering the `` tags - not in the forms.

config/locales/en/images/edit.yml
alex-ju: This content was provided by our content designer. Happy for it to be something else - I would recommend challenging this on the Trello card.

app/views/images/crop.html.erb
ChrisBAshton: Ahhh that makes sense.

config/locales/en/images/edit.yml
ChrisBAshton: As it's not in the form but on the image itself, this is fine. I misunderstood earlier.

PR #1509, opened 2019-12-09
Redirect attachments spike

ChrisBAshton:
Quick spike to figure out bits of https://trello.com/c/DMpTECux/1262-handle-redirecting-and-removing-old-assets-as-part-of-whitehall-migration

Lots still to figure out, but early comments welcome.

### DIFF
--- a/db/migrate/20191213123801_create_whitehall_imported_assets.rb
+++ b/db/migrate/20191213123801_create_whitehall_imported_assets.rb
@@ -1,0 +1,26 @@
+class CreateWhitehallImportedAssets < ActiveRecord::Migration[6.0]
+  def change
+    create_table :whitehall_imported_assets do |t|
+      t.references :whitehall_import,
+                   foreign_key: { to_table: :whitehall_imports,
+                                  on_delete: :restrict },
+                   index: true,
+                   null: false
+
+      # Asset must be one of FileAttachment or Image
+      # @TODO - write a custom validator?
+      t.references :file_attachment_revision,
+                   foreign_key: { to_table: :file_attachment_revisions,
+                                  on_delete: :cascade }
+      t.references :image_revision,
+                   foreign_key: { to_table: :image_revisions,
+                                  on_delete: :cascade }
+
+      t.string :original_asset_url, null: false
+      t.string :variants, array: true, default: []
+      t.string :state, null: false, default: "not_processed"
+      t.text :error_log
+      t.timestamps
+    end
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2019_11_20_204541) do
+ActiveRecord::Schema.define(version: 2019_12_13_094311) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
@@ -334,6 +334,8 @@ ActiveRecord::Schema.define(version: 2019_11_20_204541) do
     t.text "error_log"
     t.datetime "created_at", null: false
     t.datetime "updated_at", null: false
+    t.bigint "document_id", null: false
+    t.index ["document_id"], name: "index_whitehall_imports_on_document_id"
   end
 
   create_table "withdrawals", force: :cascade do |t|
@@ -403,5 +405,6 @@ ActiveRecord::Schema.define(version: 2019_11_20_204541) do
   add_foreign_key "timeline_entries", "revisions", on_delete: :restrict
   add_foreign_key "timeline_entries", "statuses", on_delete: :restrict
   add_foreign_key "timeline_entries", "users", column: "created_by_id", on_delete: :restrict
+  add_foreign_key "whitehall_imports", "documents", on_delete: :cascade
   add_foreign_key "withdrawals", "statuses", column: "published_status_id", on_delete: :restrict
 end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2019_12_13_094311) do
+ActiveRecord::Schema.define(version: 2019_12_13_123801) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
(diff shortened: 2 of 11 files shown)


db/schema.rb
```diff
@@ -336,6 +337,13 @@
     t.datetime "updated_at", null: false
   end
 
+  create_table "whitehall_imported_assets", force: :cascade do |t|
+    t.bigint "document_id", null: false
+    t.string "asset_manager_url", null: false
+    t.string "new_url"
```
kevindew: You probably won't be able to do this as we won't know the URL until we've synced with Asset Manager - might be better to save a reference to the revision and variant.

db/schema.rb
```diff
@@ -336,6 +337,13 @@
...
+    t.boolean "processed", default: false
```
kevindew: Might want timestamps on this table too.

lib/whitehall_importer.rb
```diff
@@ -10,7 +10,7 @@ def self.import(whitehall_document)
     )
 
     begin
-      Import.call(whitehall_document)
+      Import.call(whitehall_document, record)
```
kevindew: As the record has access to the Whitehall document we may be able to just push that in.

kevindew:
@ChrisBAshton Hmm a few of my comments didn't come through then - not sure what went wrong.

Things I flagged that seem to have got lost are:

- use of deprecate terminology. This isn't a concern for exploring ideas but this word seems to be a new bit of terminology in this migration and, on instinct, doesn't feel right to me. So as this becomes more formed I'd like to see if we can either use existing terminology or have a chat about what we should call things.
- Inheritance on CreateAsset. I think we should start with duplicated code on these before embracing inheritance. Ideally we'd want to avoid inheritance if at all possible because of the negative impact it has on code comprehension and instead a compositional approach should be preferred. Obviously these classes have aspects that may lend themselves to inheritance so maybe that ends up being the right approach but I'd suggest arriving at that destination after exhausting other possibilities rather than starting there.
- Rather than `Document.find(record.created_document_id)` we should use the model relationship to access the object. For example `record.created_document`
- I don't think we cracked the naming `created_document_id`. I actually think we'd be fine going with just document_id here since the whitehall equivalent is prefixed. Alternatively we could go with `content_publisher_document_id` for consistency but that'll be a bit of a mouthful, not necessarily needed and may present other consistency concerns

Now to wonder how github lost my things 🤔 - do feel free to tell me to butt out if I'm adding these things too early

ChrisBAshton:
Your GitHub comments were probably lost as I was rebasing and pushing up changes, most of which correlate with your suggestions already, which is great! 🎉 Early comments are always helpful in a spike such as this, so don't worry about butting in.

Not sure what to call the deprecate method, would be good to decide that soon. Perhaps along the lines of `LockWhitehallDocument`, although I imagine we'll be calling that before we begin the import, and this task is more about tidying up the loose ends after an import. We could use the terminology you use in the migration document - `PerformIrreversibleSteps` - or perhaps something more descriptive such as `MigrateUpstream`?

ChrisBAshton:
Questions:

1. Best way of determining the new asset URL and mapping it to the old one that needs redirecting.
2. Best way of creating the redirect from the old to the new asset URL.
3. How many references are needed in the `whitehall_imported_assets` table (I think we could get more value out of an `import_id` associated with the `whitehall_imports` table than by `document_id` alone).
4. Whether `revision_number` is a good thing to be tracking (can probably just track `live` vs `draft` editions). Let's say we have a document with edition A (first), B (second), C (third - live), D (fourth - draft). And let's say we have an asset in edition A, removed in edition B, a new asset in edition C and no new assets in edition D. I'd say it's out of scope for us to go back into edition A and delete the asset - that should have been Whitehall's responsibility in edition B. I think we should only be looking at the assets in editions C and D.
5. I need to spend a bit of time figuring out how best to create the `whitehall_imported_assets` table vs `WhitehallImportedAsset` class. I've only ever created a Rails migration before, and don't yet have experience of setting up an ORM model from scratch! Looking forward to it though, and will put it in my personal development spreadsheet.
6. Are there any variants of attachments that we actually want to keep? 🤔 all I know of are the PDF thumbnails which I know we want to delete.

I'm in meetings most of the day today, would be great if you could have a quick look at this spike and these questions by tomorrow @kevindew 😊 thanks!

ChrisBAshton:
Sat down with Kevin IRL a couple of days ago and got the answers to my questions:

1. Can't remember exactly, I think the Whitehall asset URL will be linked with a revision (Image/Attachment), and when the document is synced a new revision will be created which has the new asset URLs? We could then get the previous revision to get the old URL and set the redirect.
2. Use `GdsApi.asset_manager.update_asset(live_asset.asset_manager_id, redirect_url: current_asset.file_url)` as per https://github.com/alphagov/content-publisher/blob/7f4a1601cbbf10ef3f61c4c78a94143fb63e555f/app/services/publish_asset_service.rb#L73-L86
3. We will just use an import_id, not a document_id - and a lot of this doesn't need to be explicitly added, Rails adds ID columns automatically. Easier to experiment with migrations rather than edit schema directly.
4. Yes, we'll associate assets with a Content Publisher Revision rather than Whitehall revisions. But probably Image Revisions and Attachment Revisions, rather than Revision. 
5. ...
6. There are no variants we will want to keep.

kevindew:
> 1. Can't remember exactly, I think the Whitehall asset URL will be linked with a revision (Image/Attachment), and when the document is synced a new revision will be created which has the new asset URLs? We could then get the previous revision to get the old URL and set the redirect.

Prior to syncing to live we don't have asset manager urls for Image::Revision and FileAttachment::Revision assets. Once synced we can access them like so:

For a FileAttachment::Revision id (22) : 
```
irb(main):002:0> FileAttachment::Revision.find(22).asset_url
=> "https://assets.publishing.service.gov.uk/media/5d95b627ed915d08f0f326a8/letter-to-wda-holders.pdf"
```

For an Image::Revision id (1304) and variant name ("960") : 

```
irb(main):005:0> Image::Revision.find(1304).asset_url("960")
=> "https://assets.publishing.service.gov.uk/media/5defb9fae5274a71e5049268/food.jpg"
```

ChrisBAshton:
Cool, and to confirm, we could get the previous revision to get the old whitehall asset url?

```
irb(main):002:0> FileAttachment::Revision.find(22).asset_url
=> "https://assets.publishing.service.gov.uk/media/5d95b627ed915d08f0f326a8/letter-to-wda-holders.pdf"

irb(main):002:0> FileAttachment::Revision.find(21).asset_url
=> "https://assets.publishing.service.gov.uk/path/to/old/whitehall/asset/Letter--to__-WDA-holders.pdf"
```

db/migrate/20191213123801_create_whitehall_imported_assets.rb
```diff
@@ -0,0 +1,25 @@
...
+                   null: false
+
+      # Asset must be one of FileAttachment or Image
+      # @TODO - write a custom validator?
+      t.references :file_attachment_revision,
+                   foreign_key: { to_table: :file_attachment_revisions,
+                                  on_delete: :restrict }
+      t.references :image_revision,
+                   foreign_key: { to_table: :image_revisions,
+                                  on_delete: :restrict }
```
ChrisBAshton: Actually these latter two should probably have `on_delete: :cascade` relationships.

ChrisBAshton:
Hi Kevin - I've rewritten the spike commits and outlined a structure I think this can take. The only bit I haven't yet specced out fully is the `asset_exists_in_a_live_edition?` method (see last commit).

Could you have a quick look and make sure this is on the right track? On Monday I plan on closing this spike PR and starting the real one 🤞

ChrisBAshton:
Closing in favour of the 'real' branch: https://github.com/alphagov/content-publisher/pull/1530

But do feel free to leave some comments here which I can incorporate into my new branch.

PR #1511, opened 2019-12-09
History mode permissions

1pretz1:
We need to create a permission that can be granted to GDS Editors that will allow them to make changes to a live edition that is in history mode.

This PR starts to go down the road of request testing which is the new approach we want to take where logic doesn't need an explicit feature test. There is some repetition especially concerning the withdrawn / unwithdrawn permission logic so maybe we could abstract this away to a class to deal with permissions but then again we'd still have to handle the rendering from within the controller.

The two managing editor feature tests for withdraw and unwithdraw flows could be moved into request tests and will be in future work.

GDS editor permission error screen:
Screenshot 2019-12-10 at 15 59 31

TODO:

- [x]  test in integration
- [x]  content review

Trello:
https://trello.com/c/DXonAtIX/1209-permission-to-create-new-edition-withdraw-remove-restricted-when-document-in-history-mode

### DIFF
--- a/app/controllers/editions_controller.rb
+++ b/app/controllers/editions_controller.rb
@@ -1,8 +1,19 @@
 # frozen_string_literal: true
 
 class EditionsController < ApplicationController
+  before_action :check_permissions
+
   def create
     Editions::CreateInteractor.call(params: params, user: current_user)
     redirect_to edit_document_path(params[:document])
+  end
+
+private
+
+  def check_permissions
+    @edition = Edition.find_current(document: params[:document])
+    return if !@edition.history_mode? || current_user.has_permission?(User::MANAGE_LIVE_HISTORY_MODE)
+
+    render "missing_permissions/update_history_mode", status: :forbidden
   end
 end
--- a/app/controllers/unwithdraw_controller.rb
+++ b/app/controllers/unwithdraw_controller.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 class UnwithdrawController < ApplicationController
-  before_action :check_permission
+  before_action :check_permissions
 
   def confirm
     @edition = Edition.find_current(document: params[:document])
@@ -22,10 +22,16 @@ class UnwithdrawController < ApplicationController
 
 private
 
-  def check_permission
-    return if current_user.has_permission?(User::MANAGING_EDITOR_PERMISSION)
-
+  def check_permissions
     @edition = Edition.find_current(document: params[:document])
-    render :non_managing_editor, status: :forbidden
+
+    if !current_user.has_permission?(User::MANAGE_LIVE_HISTORY_MODE) && @edition.history_mode?
+      render "missing_permissions/update_history_mode", status: :forbidden
+      return
+    end
+
+    if !current_user.has_permission?(User::MANAGING_EDITOR_PERMISSION)
+      render :non_managing_editor, status: :forbidden
+    end
   end
 end
--- a/app/controllers/withdraw_controller.rb
+++ b/app/controllers/withdraw_controller.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 class WithdrawController < ApplicationController
-  before_action :check_permission
+  before_action :check_permissions
 
   def new
     @edition = Edition.find_current(document: params[:document])
@@ -35,10 +35,16 @@ class WithdrawController < ApplicationController
 
 private
 
-  def check_permission
-    return if current_user.has_permission?(User::MANAGING_EDITOR_PERMISSION)
-
+  def check_permissions
     @edition = Edition.find_current(document: params[:document])
-    render :non_managing_editor, status: :forbidden
+
+    if !current_user.has_permission?(User::MANAGE_LIVE_HISTORY_MODE) && @edition.history_mode?
+      render "missing_permissions/update_history_mode", status: :forbidden
+      return
+    end
+
+    if !current_user.has_permission?(User::MANAGING_EDITOR_PERMISSION)
+      render :non_managing_editor, status: :forbidden
+    end
   end
 end
--- a/spec/factories/user_factory.rb
+++ b/spec/factories/user_factory.rb
@@ -5,9 +5,16 @@ FactoryBot.define do
     name { "John Smith" }
     uid { SecureRandom.uuid }
     email { "someone@example.com" }
+    transient do
+      managing_editor { false }
+      manage_live_history_mode { false }
+    end
 
-    trait :managing_editor do
-      permissions { [User::MANAGING_EDITOR_PERMISSION] }
+    permissions do
+      [User::PRE_RELEASE_FEATURES_PERMISSION].tap do |p|
+        p << User::MANAGING_EDITOR_PERMISSION if managing_editor
+        p << User::MANAGE_LIVE_HISTORY_MODE if manage_live_history_mode
+      end
     end
   end
 end
--- a/spec/requests/editions_spec.rb
+++ b/spec/requests/editions_spec.rb
@@ -9,5 +9,26 @@ RSpec.describe "Editions" do
       expect { post create_edition_path(edition.document) }
         .to change { Edition.where(document_id: edition.document.id).count }.by(1)
     end
+
+    context "when the edition is in history mode" do
+      let(:edition) { create(:edition, :published, :political, :past_government) }
+
+      it "lets users holding the manage_live_history_mode permisssion create a new edition" do
+        user = create(:user, manage_live_history_mode: true)
+        login_as(user)
+        stub_publishing_api_put_content(edition.content_id, {})
+
+        expect { post create_edition_path(edition.document) }
+          .to change { Edition.where(document_id: edition.document.id).count }.by(1)
+      end
+
+      it "prevents users without the permission creating a new edition" do
+        edition = create(:edition, :published, :political, :past_government)
+        post create_edition_path(edition.document)
+
+        expect(response.body).to include(I18n.t!("missing_permissions.update_history_mode.title", title: edition.title))
+        expect(response.status).to eq(403)
+      end
+    end
   end
 end
--- a/spec/requests/editions_spec.rb
+++ b/spec/requests/editions_spec.rb
@@ -1,0 +1,13 @@
+# frozen_string_literal: true
+
+RSpec.describe "Editions" do
+  describe "POST /document/:document/editions" do
+    it "creates a new edition" do
+      edition = create(:edition, :published)
+      stub_publishing_api_put_content(edition.content_id, {})
+
+      expect { post create_edition_path(edition.document) }
+        .to change { Edition.where(document_id: edition.document.id).count }.by(1)
+    end
+  end
+end
--- a/spec/requests/unwithdraw_spec.rb
+++ b/spec/requests/unwithdraw_spec.rb
@@ -1,0 +1,54 @@
+# frozen_string_literal: true
+
+RSpec.describe "Unwithdraw" do
+  let(:edition) { create(:edition, :withdrawn, :political, :past_government) }
+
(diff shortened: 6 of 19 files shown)


spec/requests/editions_spec.rb
```diff
@@ -0,0 +1,14 @@
...
+
+RSpec.describe "Editions" do
+  describe "POST /document/:document/editions" do
+    it "creates a new edition" do
+      edition = create(:edition, :published)
+      stub_publishing_api_put_content(edition.content_id, {})
+      post create_edition_path(edition.document)
+      follow_redirect!
+
+      expect(response.body).to include(I18n.t!("documents.edit.title", title: edition.title))
```
kevindew: This expectation doesn't capture that we've created a new edition - checking the title seems somewhat incidental. Might make more sense to have a 

```
expect { post create_edition_path(edition.document) }
  .to change { Edition.where(document_id: document.id).count }.by(1)
```

app/controllers/editions_controller.rb
```diff
@@ -1,8 +1,19 @@
...
   def create
     Editions::CreateInteractor.call(params: params, user: current_user)
     redirect_to edit_document_path(params[:document])
   end
+
+private
+
+  def check_permissions
+    @edition = Edition.find_current(document: params[:document])
+    return if current_user.can_access_in_history_mode?(@edition) || !@edition.history_mode?
```
kevindew: It seems somewhat unnecessary to have this `can_access_in_history_mode?` method if we also need to check if the edition is in history mode. Might as well be:

```
if @edition.history.mode? && !current_user.has_permission?(User::MANAGE_LIVE_HISTORY_MODE)
  render :non_manage_live_history_mode, status: :forbidden
end
```

app/views/editions/non_manage_live_history_mode.html.erb
```diff
@@ -0,0 +1,8 @@
+<% content_for :back_link, render_back_link(href: document_path(@edition.document)) %>
+<% content_for :title, t("editions.non_manage_live_history_mode_permission.title", title: @edition.title) %>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <%= render_govspeak(File.read("app/views/editions/non_manage_live_history_mode.govspeak.md")) %>
+  </div>
+</div>
```
kevindew: I wonder if we can make use of the errors partial for this to save duplication.

e.g.
```
<%= render partial: "errors/error",
  locals: {
    title: t("editions.non_manage_live_history_mode_permission.title", title: @edition.title),
    body_govspeak: File.read("app/views/editions/non_manage_live_history_mode.govspeak.md"),
  } %>
```

config/locales/en/editions/editions.yml
```diff
@@ -0,0 +1,4 @@
+en:
+  editions:
+    non_manage_live_history_mode_permission:
```
kevindew: If you use "permission" in the name here you should also use it in the view file names for consistency.

spec/requests/editions_spec.rb
```diff
@@ -10,5 +10,28 @@
...
+      let(:edition) { create(:edition, :published, :political, :past_government) }
+
+      it "lets users holding the manage_live_history_mode permisssion create a new edition" do
+        user = create(:user, manage_live_history_mode: true)
+        login_as(user)
+        stub_publishing_api_put_content(edition.content_id, {})
+        post create_edition_path(edition.document)
+        follow_redirect!
+
+        expect(response.body).to include(I18n.t!("documents.edit.title", title: edition.title))
```
kevindew: This has the same issue of not really testing what it says it does.

spec/requests/editions_spec.rb
```diff
@@ -10,5 +10,28 @@
...
+      end
+
+      it "prevents users without the permission creating a new edition" do
+        edition = create(:edition, :published, :political, :past_government)
+        stub_publishing_api_put_content(edition.content_id, {})
+        post create_edition_path(edition.document)
+
+        expect(response.body).to include(I18n.t!("editions.non_manage_live_history_mode_permission.title", title: edition.title))
+        expect(response.status).to eq(403)
```
kevindew: Looks good 👍

spec/requests/editions_spec.rb
```diff
@@ -10,5 +10,28 @@
...
+        stub_publishing_api_put_content(edition.content_id, {})
+        post create_edition_path(edition.document)
+        follow_redirect!
+
+        expect(response.body).to include(I18n.t!("documents.edit.title", title: edition.title))
+      end
+
+      it "prevents users without the permission creating a new edition" do
+        edition = create(:edition, :published, :political, :past_government)
+        stub_publishing_api_put_content(edition.content_id, {})
```
kevindew: May want to put a `before { stub_any_publishing_api_put_content }` so we don't need this in each test.

app/controllers/withdraw_controller.rb
```diff
@@ -35,10 +35,16 @@ def create
...
 private
 
-  def check_permission
-    return if current_user.has_permission?(User::MANAGING_EDITOR_PERMISSION)
-
+  def check_permissions
     @edition = Edition.find_current(document: params[:document])
-    render :non_managing_editor, status: :forbidden
+
+    unless current_user.can_access_in_history_mode?(@edition) && current_user.has_permission?(User::MANAGING_EDITOR_PERMISSION)
```
kevindew: I've found myself a bit confused by this conditional not sure where the managing editor part kicks in. Might need you to explain this to me.

app/views/unwithdraw/non_manage_live_history_mode.govspeak.md
```diff
@@ -0,0 +1,6 @@
+Only a GDS admin who is also a managing editor can remove the withdrawn banner and republish this content.
+
+If you think this content needs to be updated, or should not be in history mode, you need
+[to make a request to GDS in GOV.UK support](https://support.publishing.service.gov.uk/content_advice_request/new) (link opens in a new window).
+
+Read the [full guidance on creating content](https://www.gov.uk/guidance/content-design) (link opens in a new window).
```
kevindew: Since these are relatively short we could put them in the translations file which might make them feel like there is less stuff to deal with as part of this.

spec/features/workflow/unwithdraw_without_managing_editor_permissions_spec.rb
```diff
@@ -23,7 +23,7 @@ def and_i_click_on_undo_withdrawal
 
   def then_i_see_a_message_to_ask_my_managing_editor_to_unwithdraw_content
     expect(page).to have_content(
-      I18n.t!("unwithdraw.no_managing_editor_permission.title"),
+      I18n.t!("unwithdraw.no_manage_live_history_mode_permission.title", title: @withdrawn_edition.title),
```
kevindew: It doesn't feel quite right that this test needs to be changed - maybe I've missed something?

spec/requests/unwithdraw_spec.rb
```diff
@@ -0,0 +1,56 @@
+# frozen_string_literal: true
+
+RSpec.describe "Unwithdraw" do
+  let(:edition) { create(:edition, :withdrawn, :political, :past_government) }
+
+  describe "POST /documents/:document/unwithdraw" do
+    context "when the edition is in history mode" do
+      context "and the user is a managing_editor" do
```
kevindew: Since we don't have a separate context can this be within the previous context?

spec/factories/user_factory.rb
```diff
@@ -5,9 +5,15 @@
...
+      managing_editor { false }
+      manage_live_history_mode { false }
+    end
 
-    trait :managing_editor do
-      permissions { [User::MANAGING_EDITOR_PERMISSION] }
+    after(:build) do |user, options|
+      user.permissions << User::MANAGING_EDITOR_PERMISSION if options.managing_editor
+      user.permissions << User::MANAGE_LIVE_HISTORY_MODE if options.manage_live_history_mode
+      user.permissions << User::PRE_RELEASE_FEATURES_PERMISSION
```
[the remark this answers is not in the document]
1pretz1: Sure!

spec/spec_helper.rb
```diff
@@ -36,6 +36,7 @@
   config.include GdsApi::TestHelpers::AssetManager
   config.include GovukSchemas::RSpecMatchers
   config.include AuthenticationHelper, type: :feature
+  config.include AuthenticationHelper, type: :request
```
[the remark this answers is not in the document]
1pretz1: Yep you can also use a block of code here... done!

spec/spec_helper.rb
```diff
@@ -53,7 +54,7 @@
     ActionMailer::Base.deliveries.clear
   end
 
-  config.after :each, type: :feature do
+  config.after :each, type: lambda { |s| %i[feature request].include?(s) } do
```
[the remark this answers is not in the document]
1pretz1: stabby it is 🔪

spec/requests/editions_spec.rb
1pretz1: Good catch!

app/views/editions/non_manage_live_history_mode.html.erb
1pretz1: I've decided to keep as is due to all controllers using the same template meaning less clutter.

config/locales/en/editions/editions.yml
1pretz1: This no longer applies as per using one template

spec/requests/editions_spec.rb
1pretz1: Updated the test!

app/controllers/editions_controller.rb
1pretz1: Removed the method from the user model 👍

app/controllers/withdraw_controller.rb
1pretz1: Discussed offline and now sorted!

app/views/unwithdraw/non_manage_live_history_mode.govspeak.md
1pretz1: This is now one template so I think is fine as is!

spec/features/workflow/unwithdraw_without_managing_editor_permissions_spec.rb
1pretz1: You're correct, reverted back!

spec/requests/unwithdraw_spec.rb
1pretz1: Context removed!

1pretz1:
Made all the changes you've requested @kevindew 😄

app/controllers/unwithdraw_controller.rb
```diff
@@ -22,10 +22,16 @@ def unwithdraw
...
-    render :non_managing_editor, status: :forbidden
+
+    if !current_user.has_permission?(User::MANAGING_EDITOR_PERMISSION) && !@edition.history_mode?
+      render :non_managing_editor, status: :forbidden
+      return
+    end
+
+    if !current_user.has_permission?(User::MANAGE_LIVE_HISTORY_MODE) && @edition.history_mode?
+      render "missing_permissions/update_history_mode", status: :forbidden
+    end
```
kevindew: I feel like this would be simpler if we did:

```
    if !current_user.has_permission?(User::MANAGE_LIVE_HISTORY_MODE) && @edition.history_mode?
      render "missing_permissions/update_history_mode", status: :forbidden
      return
    end

    if !current_user.has_permission?(User::MANAGING_EDITOR_PERMISSION)
      render :non_managing_editor, status: :forbidden
    end
```

1pretz1: Ahhhhh, you're correct!

PR #1520, opened 2019-12-12
Change public political naming

kevindew:
Trello https://trello.com/c/LHR7GodT/1234-release-history-mode

The terminology used in the application has diverged since the initial implementation of the feature. This changes the public routes where we show the word "political" to instead reference "gets history mode" as is used in the UI.

Fixing this doesn't really resolve any huge problems but my logic is that this will reduce scatterings of confusion where things don't reflect naming - and since this feature isn't released yet it is still relatively pliable. 

Obviously these could diverge again and we could end up with the naming not matching again, and likely we wouldn't fix it then as it would cause bc breaks for analytics and such. But since we're not in that situation and naming changes are very infrequent this seems a good course for now.


kevindew:
> I'm not sure i'm overly keen on 'gets history mode' in general as I think it looks a bit icky in places but I acknowledge it's a tricky concept to name! But yay for consistency! 🎉

Yup, I know what you mean. I'm happy that it's a big improvement on the ambiguity of "political", so yay for iteration.

PR #1523, opened 2019-12-12
Add request tests for withdraw and unwithdraw workflows

1pretz1:
This PR replaces the none 'happy' path withdraw and unwithdraw features with their respective request tests. A reduction of 5 files and 81 lines of code seems a pretty nice win.

Below are rough times specs took to run using unix's time command, e.g `time bundle exec rspec spec/requests/withdraw_spec.rb`

### Withdraw tests run speed - master:
* Request test: 4.605s
* api_down, requirements, managing_editor features: 5.119s

Total: ~9.724s

### Withdraw tests run speed - (this branch):
* Request test (including api_down, requirements, managing_editor tests): 5.161s

Total: ~5.161s

Although the request test took about the same time as the three features, it encompasses way more tests and improves test coverage.

Trello:
https://trello.com/c/G3YtOdaW/1247-replace-withdraw-and-unwithdraw-features-with-request-tests


spec/requests/withdraw_spec.rb
```diff
@@ -1,57 +1,81 @@
...
+      withdrawal = published_edition.reload.status.details
+      expect(response.body).to include(
+        I18n.t!("documents.show.withdrawn.title",
+                document_type: published_edition.document_type.label.downcase,
+                withdrawn_date: withdrawal.created_at.strftime("%-d %B %Y")),
+      )
+    end
+
     context "when the edition is in history mode" do
       it "lets users holding manage_live_history_mode permisssion withdraw the edition" do
```
ChrisBAshton: Not your code, but noticed a typo here (`permisssion`). Happens in a few places in this file, could you do a mass find-and-replace in a separate commit please?

EDIT: looks like it also applies in `unwithdraw_spec.rb` - maybe do the find-and-replace on the whole repo!

spec/requests/withdraw_spec.rb
```diff
@@ -31,6 +31,13 @@
       expect(response.body).to include(I18n.t!("withdraw.new.flashes.publishing_api_error.title"))
     end
 
+    it "prevents users without managing_editor permission withdraw the edition" do
```
ChrisBAshton:
```suggestion
    it "prevents users without managing_editor permission from withdrawing the edition" do
```

spec/requests/withdraw_spec.rb
```diff
@@ -21,6 +21,16 @@
...
     end
 
+    it "returns an error when publishing-api is down" do
+      publishing_api_isnt_available
+      login_as(managing_editor)
+
+      post withdraw_path(published_edition.document), params: { public_explanation: "Just cos" }
+      follow_redirect!
+
+      expect(response.body).to include(I18n.t!("withdraw.new.flashes.publishing_api_error.title"))
```
ChrisBAshton: Are we missing a status code assertion here?

```ruby
      expect(response.status).to eq(502)
```

spec/requests/unwithdraw_spec.rb
```diff
@@ -1,52 +1,71 @@
 # frozen_string_literal: true
 
 RSpec.describe "Unwithdraw" do
-  let(:edition) { create(:edition, :withdrawn, :political, :past_government) }
+  let(:managing_editor) { create(:user, managing_editor: true) }
+  let(:withdrawn_edition) { create(:edition, :withdrawn) }
+  let(:history_mode_edition) { create(:edition, :withdrawn, :political, :past_government) }
```
ChrisBAshton: I don't think this variable fully explains what it's actually creating:

```suggestion
  let(:withdrawn_edition_in_history_mode) { create(:edition, :withdrawn, :political, :past_government) }
```

spec/requests/unwithdraw_spec.rb
```diff
@@ -1,52 +1,71 @@
...
+      login_as(managing_editor)
+
+      post unwithdraw_path(withdrawn_edition.document)
+      follow_redirect!
+
+      expect(response.body).to include(I18n.t!("documents.history.entry_types.unwithdrawn"))
+    end
+
     context "when the edition is in history mode" do
       it "lets users holding manage_live_history_mode permisssion unwithdraw the edition" do
```
ChrisBAshton:
```suggestion
      it "lets managing editors holding manage_live_history_mode permission unwithdraw the edition" do
```

spec/requests/unwithdraw_spec.rb
```diff
@@ -1,52 +1,71 @@
 # frozen_string_literal: true
 
 RSpec.describe "Unwithdraw" do
-  let(:edition) { create(:edition, :withdrawn, :political, :past_government) }
+  let(:managing_editor) { create(:user, managing_editor: true) }
```
ChrisBAshton: I'd be tempted to have a `let(:user_permissions) { { managing_editor: true } }` at the top of the file, and change this line to:

```suggestion
  let(:managing_editor) { create(:user, user_permissions) }
```

That way we can nicely set a `let(:user_permissions) { { managing_editor: true, manage_live_history_mode: true } }` on line 22 and make that test consistent with the rest of these.

spec/requests/unwithdraw_spec.rb
```diff
@@ -26,6 +26,13 @@
       expect(response.body).to include(I18n.t!("user_facing_states.withdrawn.name"))
     end
 
+    it "prevents users without managing_editor permission unwithdraw the edition" do
```
ChrisBAshton:
```suggestion
    it "prevents users without managing_editor permission from unwithdrawing the edition" do
```

spec/requests/unwithdraw_spec.rb
```diff
@@ -59,6 +66,13 @@
       expect(response.body).to include(I18n.t!("documents.show.unwithdraw.title"))
     end
 
+    it "prevents users without managing_editor permission to access unwithdraw page" do
```
ChrisBAshton:
```suggestion
    it "prevents users without managing_editor permission from accessing unwithdraw page" do
```

spec/requests/unwithdraw_spec.rb
```diff
@@ -1,52 +1,95 @@
...
+    end
+
+    it "returns an error when publishing-api is down" do
+      publishing_api_isnt_available
+      login_as(managing_editor)
+
+      post unwithdraw_path(withdrawn_edition.document)
+      follow_redirect!
+
+      expect(response.body).to include(I18n.t!("user_facing_states.withdrawn.name"))
```
kevindew: I think this should be some error text.

spec/requests/unwithdraw_spec.rb
kevindew: That does slip into a world where we need every change of user to be in a different context. I'll let Pete see how he feels about that but does seem to add some extra need for context blocks and more variables outside of test.

spec/requests/withdraw_spec.rb
1pretz1: I got too eager with the s's 🤦‍♂

spec/requests/withdraw_spec.rb
1pretz1: Hmm not really because the request is following a redirect so before the redirect it will be a `302` and after the redirect the response status will be `200` (successfully rendered error page) .

spec/requests/unwithdraw_spec.rb
1pretz1: sure

spec/requests/unwithdraw_spec.rb
1pretz1: I'd have to agree with Kevin on this one, nice suggestion though!

spec/requests/unwithdraw_spec.rb
1pretz1: Good catch!

1pretz1:
So if made some changes to the spec file's to try and adhere to @kevindew & @ChrisBAshton suggestions. My english was / is ridiculously poor on this PR, back to school for me! 😆 

I spent a bit of time trying to format the tests in a nice way to use before blocks for logging in and stubbing but this just ended up with messy code with additional contexts and not really improving the situation -> https://github.com/alphagov/content-publisher/compare/wip-request-pattern?expand=1 (see last two commits for two approaches).

I ended up going with defining lets for each request url, this brings these variables closer to the actual tests using them and will mean that as the spec file grows we don't have heaps of lets at the top of the page even though this comes with the cost of some code duplication.

Following changes are:
* test what happens when an incorrect status is used to withdraw / unwithdraw
* make tests more rspec'y - use http_status_code(:forbidden) instead of eq(403) for example 
* some other tweaks

PR #1526, opened 2019-12-13
Make navigation landmarks unique

alex-ju:
Landmarks must have a unique role or role/label. Given that we always have the main navigation on the page, any additional navigation landmark must be accompanied by a label to make it unique. `aria-label` is being appended to the landmark type (e.g. navigation) when provided – as seen in the example below.

**Note:** The updated copy is pending a review from a content designer.

**Example:** Using VoiceOver (⌘+F5 to enable VoiceOver, Ctrl+⎇+U to switch to landmarks and arrows to switch mode)
Screenshot 2019-12-13 at 11 52 39

[Trello card](https://trello.com/c/oCRMPDT2)

### DIFF
--- a/app/views/components/_side_navigation.html.erb
+++ b/app/views/components/_side_navigation.html.erb
@@ -3,7 +3,7 @@
   items ||= []
 %>
 
-<%= tag.nav class: "app-c-side-navigation", id: id, role: 'navigation' do %>
+<%= tag.nav class: "app-c-side-navigation", id: id, role: 'navigation', aria: { label: aria_label } do %>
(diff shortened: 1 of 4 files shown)


JonathanHallam:
So having had a bit of a google it's my understanding that the use of aria-labels is for helping screen reader users navigate in places where you don't want to have a visible label. If this is what we're doing here would you mind adding something of that effect to the commit messages? Just aware that at the moment they only make mention of uniqueness which could be done with regular label tags. Think this would make it clearer further down the line given that aria-labels aren't too common.

PR #1530, opened 2019-12-16
Redirect Whitehall images and attachments

ChrisBAshton:
This PR creates a MigrateAssets service which is called from the WhitehallImporter `sync`/`import_and_sync` methods. It:

* Creates a record for each imported asset (image or file attachment)
* On successful import, it will either redirect the Whitehall asset to the Content Publisher one or delete it if it only exists on the draft stack
* For each variant of the asset, the variant is either deleted (for file attachments, or for draft images) or redirected (live image variants are redirected to the canonical Content Publisher image).

This work comes out of a spike in #1509.

Todo:
- [ ] Better variants
- [ ] Error logging

Trello card: https://trello.com/c/DMpTECux/1232-handle-redirecting-and-removing-old-assets-as-part-of-whitehall-migration

### DIFF
--- a/app/models/whitehall_imported_asset.rb
+++ b/app/models/whitehall_imported_asset.rb
@@ -1,0 +1,6 @@
+# frozen_string_literal: true
+
+# Represents the raw import of an asset from Whitehall Publisher and
+# the import status of the asset into Content Publisher
+class WhitehallImportedAsset < ApplicationRecord
+end
--- a/app/models/whitehall_imported_asset.rb
+++ b/app/models/whitehall_imported_asset.rb
@@ -18,6 +18,12 @@ class WhitehallImportedAsset < ApplicationRecord
     end
   end
 
+  def asset_manager_id
+    url_array = original_asset_url.to_s.split("/")
+    # https://github.com/alphagov/asset-manager#create-an-asset
+    url_array[url_array.length - 2]
+  end
+
 private
 
   def associated_with_only_image_or_file_attachment
--- a/app/models/whitehall_imported_asset.rb
+++ b/app/models/whitehall_imported_asset.rb
@@ -5,5 +5,7 @@
 class WhitehallImportedAsset < ApplicationRecord
   belongs_to :whitehall_import
 
+  # belongs to one of these, not both
   belongs_to :image_revision, class_name: "Image::Revision", optional: true
+  belongs_to :file_attachment_revision, class_name: "FileAttachment::Revision", optional: true
 end
--- a/app/models/whitehall_imported_asset.rb
+++ b/app/models/whitehall_imported_asset.rb
@@ -10,6 +10,14 @@ class WhitehallImportedAsset < ApplicationRecord
 
   validate :associated_with_only_image_or_file_attachment
 
+  def revision
+    if image_revision.present?
+      image_revision
+    elsif file_attachment_revision.present?
+      file_attachment_revision
+    end
+  end
+
 private
 
   def associated_with_only_image_or_file_attachment
--- a/app/models/whitehall_imported_asset.rb
+++ b/app/models/whitehall_imported_asset.rb
@@ -18,6 +18,14 @@ class WhitehallImportedAsset < ApplicationRecord
     end
   end
 
+  def main_asset
+    if image_revision.present?
+      image_revision.asset("960")
+    else
+      file_attachment_revision.asset
+    end
+  end
+
   def asset_manager_id
     url_array = original_asset_url.to_s.split("/")
     # https://github.com/alphagov/asset-manager#create-an-asset
--- a/app/models/whitehall_imported_asset.rb
+++ b/app/models/whitehall_imported_asset.rb
@@ -3,4 +3,7 @@
 # Represents the raw import of an asset from Whitehall Publisher and
 # the import status of the asset into Content Publisher
 class WhitehallImportedAsset < ApplicationRecord
+  belongs_to :whitehall_import
+
+  belongs_to :image_revision, class_name: "Image::Revision", optional: true
 end
--- a/app/models/whitehall_imported_asset.rb
+++ b/app/models/whitehall_imported_asset.rb
@@ -5,7 +5,16 @@
 class WhitehallImportedAsset < ApplicationRecord
   belongs_to :whitehall_import
 
-  # belongs to one of these, not both
   belongs_to :image_revision, class_name: "Image::Revision", optional: true
   belongs_to :file_attachment_revision, class_name: "FileAttachment::Revision", optional: true
+
+  validate :associated_with_only_image_or_file_attachment
+
+private
+
+  def associated_with_only_image_or_file_attachment
+    if image_revision.present? && file_attachment_revision.present?
+      errors.add(:base, "Cannot be associated with both image revision AND file attachment revision")
+    end
+  end
 end
--- a/db/migrate/20191213123801_create_whitehall_imported_assets.rb
+++ b/db/migrate/20191213123801_create_whitehall_imported_assets.rb
@@ -1,0 +1,25 @@
+# frozen_string_literal: true
+
+class CreateWhitehallImportedAssets < ActiveRecord::Migration[6.0]
+  def change
+    create_table :whitehall_imported_assets do |t|
+      t.references :whitehall_import,
+                   foreign_key: { to_table: :whitehall_imports,
+                                  on_delete: :restrict },
+                   index: true,
+                   null: false
+      t.references :file_attachment_revision,
+                   foreign_key: { to_table: :file_attachment_revisions,
+                                  on_delete: :restrict }
+      t.references :image_revision,
+                   foreign_key: { to_table: :image_revisions,
+                                  on_delete: :restrict }
+
+      t.string :original_asset_url, null: false
+      t.json :variants, default: {}, null: false
+      t.string :state, null: false, default: "not_processed"
+      t.text :error_message
+      t.timestamps
+    end
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2019_12_13_094311) do
+ActiveRecord::Schema.define(version: 2019_12_13_123801) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
@@ -326,6 +326,21 @@ ActiveRecord::Schema.define(version: 2019_12_13_094311) do
     t.datetime "updated_at", null: false
   end
 
+  create_table "whitehall_imported_assets", force: :cascade do |t|
+    t.bigint "whitehall_import_id", null: false
+    t.bigint "file_attachment_revision_id"
+    t.bigint "image_revision_id"
+    t.string "original_asset_url", null: false
+    t.json "variants", default: {}, null: false
+    t.string "state", default: "not_processed", null: false
+    t.text "error_message"
+    t.datetime "created_at", precision: 6, null: false
+    t.datetime "updated_at", precision: 6, null: false
+    t.index ["file_attachment_revision_id"], name: "index_whitehall_imported_assets_on_file_attachment_revision_id"
+    t.index ["image_revision_id"], name: "index_whitehall_imported_assets_on_image_revision_id"
+    t.index ["whitehall_import_id"], name: "index_whitehall_imported_assets_on_whitehall_import_id"
+  end
+
   create_table "whitehall_imports", force: :cascade do |t|
     t.bigint "whitehall_document_id", null: false
     t.json "payload", null: false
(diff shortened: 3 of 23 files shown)


spec/factories/whitehall_import_factory.rb
```diff
@@ -0,0 +1,13 @@
...
+FactoryBot.define do
+  factory :whitehall_import do
+    state { "importing" }
+    payload { build(:whitehall_export_document) }
+
+    after(:build) do |record|
+      record.whitehall_document_id = record.payload["id"]
+      record.content_id = record.payload["content_id"]
+    end
+  end
```
kevindew: Nice 👍

spec/factories/whitehall_import_factory.rb
```diff
@@ -0,0 +1,13 @@
...
+
+FactoryBot.define do
+  factory :whitehall_import do
+    state { "importing" }
+    payload { build(:whitehall_export_document) }
+
+    after(:build) do |record|
+      record.whitehall_document_id = record.payload["id"]
+      record.content_id = record.payload["content_id"]
+    end
```
kevindew: You don't actually need these to be in an `after(:build)` you can instead do:

```
    whitehall_document_id { payload["id"] }
    content_id { payload["content_id"] }
```

db/migrate/20191213094311_add_document_ref_to_whitehall_imports.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+class AddDocumentRefToWhitehallImports < ActiveRecord::Migration[6.0]
+  def change
+    add_reference :whitehall_imports,
+                  :document,
+                  foreign_key: { on_delete: :cascade },
```
kevindew: It'd be more consistent with https://github.com/alphagov/content-publisher/blob/master/docs/adr/0013-permitted-deletion.md to use restrict since this feature isn't related to access limited. That builds off https://github.com/alphagov/content-publisher/blob/master/docs/adr/0008-restricted-deletion.md where restrict was just set as the default everywhere so we don't get too bogged down in whether things should be able to delete others.

A later concern that we should log though is that we may need to delete WhitehallImports that pull in access limited data as part of the data sync (what an annoying complication)

db/migrate/20191213094311_add_document_ref_to_whitehall_imports.rb
```diff
@@ -0,0 +1,11 @@
...
+                  null: true,
+                  index: true
```
kevindew: I don't really think we'll need an index on this as I don''t expect we'll be traversing WhitehallImports via document.

db/migrate/20191213123801_create_whitehall_imported_assets.rb
```diff
@@ -0,0 +1,28 @@
...
+                   null: false
+
+      # Asset must be one of FileAttachment or Image
+      # @TODO - write a custom validator?
+      t.references :file_attachment_revision,
+                   foreign_key: { to_table: :file_attachment_revisions,
+                                  on_delete: :cascade }
+      t.references :image_revision,
+                   foreign_key: { to_table: :image_revisions,
+                                  on_delete: :cascade }
```
kevindew: Both these references should also be restricts as well for ADR consistency.

Also at a purely semantic level I don't think we'd want this record deleted automatically as it'd leave the asset in the abyss.

app/models/whitehall_imported_asset.rb
```diff
@@ -0,0 +1,6 @@
+# frozen_string_literal: true
+
+# Represents the raw import of an asset from Whitehall Publisher and
+# the import status of the asset into Content Publisher
+class WhitehallImportedAsset < ApplicationRecord
```
kevindew: Might be an idea to namespace this class under whitehall_import (as we've done with file_attachment and image) which may reduce a proliferation of classes in models.

Saying this I am expecting we're going to have a WhitehallMigration model and I think it'd make sense to have both import and asset namespaced under that.

db/migrate/20191213123801_create_whitehall_imported_assets.rb
```diff
@@ -0,0 +1,28 @@
...
+
+      t.string :original_asset_url, null: false
+      t.string :variants, json: true, default: "{}"
```
kevindew: I think rather than dealing with a collection of asset url and variants it might make sense for each record in this to be an individual asset_url. The thing I'm thinking is if there are multiple assets on a single record then a single state doesn't really make sense as one asset can be successful and another one fail for instance.

I anticipated it to be that we'd have a table like:

asset_url
image_revision
file_attachment_revision
image_revision_variant
state

and then we could have a destination_url method on the model like so:

```
def destination_url
  return file_attachment_revision.asset_url if file_attachment_revision
  return image_revision.asset_url(image_revision_variant) if image_revision
end
```

then can use a string from this to redirect and a nil from it to do a removal.

db/migrate/20191213123801_create_whitehall_imported_assets.rb
```diff
@@ -0,0 +1,28 @@
...
+      t.string :state, null: false, default: "not_processed"
+      t.text :error_log
```
kevindew: Not a biggie, as I know this is consistent with whitehall_import class, but I do wonder if we're starting an usual precedent with this error_log field. Since we don't actually use it to store a log of errors and just a single one. 

Just something to think about given we're repeating it so I wonder if you share my thoughts whether we want to rename this to say "error_message" and then do same to WhitehallImport. What do you think?

db/schema.rb
```diff
@@ -326,6 +326,21 @@
...
+  create_table "whitehall_imported_assets", force: :cascade do |t|
+    t.bigint "whitehall_import_id", null: false
+    t.bigint "file_attachment_revision_id"
+    t.bigint "image_revision_id"
+    t.string "original_asset_url", null: false
+    t.string "variants", default: "{}"
+    t.string "state", default: "not_processed", null: false
+    t.text "error_log"
+    t.datetime "created_at", precision: 6, null: false
+    t.datetime "updated_at", precision: 6, null: false
```
kevindew: Oh these precisions are new, I assume this is something Rails did automatically?

They seem great though as I think that'll mean Ruby and postgres agree on precision of time which might mean this code works in tests:

```
time = Time.current
m = Model.create(time: time)
expect(m.time).to eq(time)
```

app/models/whitehall_imported_asset.rb
```diff
@@ -3,4 +3,12 @@
...
 # the import status of the asset into Content Publisher
 class WhitehallImportedAsset < ApplicationRecord
+  belongs_to :whitehall_import
+
+  belongs_to :image_revision, class_name: "Image::Revision", optional: true
+
+  def variants
+    # Stored as a stringified hash - need to turn it back to JSON
+    JSON.parse(read_attribute("variants").gsub("=>", ":"))
+  end
```
kevindew: We shouldn't need to do any of this ActiveRecord will convert the JSON automatically to a hash. I'm assuming you hit an issue that led to this. 

Just as an example:

```
irb(main):009:0> cr = FactoryBot.create(:content_revision, contents: { "key_1" => "value_1", "key_2" => "value_2" })
   (0.2ms)  BEGIN
  User Create (0.4ms)  INSERT INTO "users" ("name", "email", "uid", "created_at", "updated_at") VALUES ($1, $2, $3, $4, $5) RETURNING "id"  [["name", "John Smith"], ["email", "someone@example.com"], ["uid", "65df8608-5519-4095-bb7f-f9f661e9cb5e"], ["created_at", "2019-12-17 19:39:41.096261"], ["updated_at", "2019-12-17 19:39:41.096261"]]
   (1.1ms)  COMMIT
   (0.1ms)  BEGIN
  ContentRevision Create (0.3ms)  INSERT INTO "content_revisions" ("contents", "created_at", "created_by_id") VALUES ($1, $2, $3) RETURNING "id"  [["contents", "{\"key_1\":\"value_1\",\"key_2\":\"value_2\"}"], ["created_at", "2019-12-17 19:39:41.099551"], ["created_by_id", 51011]]
   (0.2ms)  COMMIT
=> #<ContentRevision id: 15771, title: nil, base_path: nil, summary: nil, contents: {"key_1"=>"value_1", "key_2"=>"value_2"}, created_at: "2019-12-17 19:39:41", created_by_id: 51011>
irb(main):011:0> cr.reload
  ContentRevision Load (0.4ms)  SELECT "content_revisions".* FROM "content_revisions" WHERE "content_revisions"."id" = $1 LIMIT $2  [["id", 15771], ["LIMIT", 1]]
=> #<ContentRevision id: 15771, title: nil, base_path: nil, summary: nil, contents: {"key_1"=>"value_1", "key_2"=>"value_2"}, created_at: "2019-12-17 19:39:41", created_by_id: 51011>
irb(main):012:0> cr.contents
=> {"key_1"=>"value_1", "key_2"=>"value_2"}
irb(main):013:0> cr.contents.class
=> Hash
```

lib/whitehall_importer/create_revision.rb
```diff
@@ -115,13 +115,21 @@ def tags(associations)
 
     def create_image_revisions(images)
       images.reduce([]) do |memo, image|
-        memo << WhitehallImporter::CreateImageRevision.call(record, image, memo.map(&:filename))
+        already_imported = WhitehallImportedAsset.find_by(original_asset_url: image["url"])
+        unless already_imported
+          memo << WhitehallImporter::CreateImageRevision.call(record, image, memo.map(&:filename))
+        end
+        memo
```
kevindew: Oh neat - I see that this will save us adding superfluous images and attachments which is super cool. Not sure if my suggestion of asset record per file complicates that, hopefully there's a way.

Do we use this create_image_revisions as the means of getting the array of all the image revisions of the document? Assuming so, I believe by skipping items we'll be missing them from the array of image revisions so we may need the means to look up ones that already exist. So perhaps this become more of a `find_or_create_image_revisions`?

Minor but this can be a little cleaner as 

```
        already_imported = WhitehallImportedAsset.find_by(original_asset_url: image["url"])
        next memo if already_imported
        memo << WhitehallImporter::CreateImageRevision.call(record, image, memo.map(&:filename))
```

kevindew:
Oh another thing. I think I might have said this to you already, but just in case I haven't. 

The one-to-one mapping between a WhitehallImport and a document isn't going to be sufficient once we have translations since a single Whitehall document will create multiple Content Publisher ones. This isn't going to be a problem for us now since we explicitly don't allow multiple languages and I think changing this mapping will be just one of many changes needed. But in case you weren't aware I thought I'd make sure you knew so it's not a surprise when that problem comes around.

spec/factories/whitehall_import_factory.rb
ChrisBAshton: Ahh that's nicer, thanks!

db/schema.rb
ChrisBAshton: Yep, this came for free with Rails 6 👍

db/migrate/20191213123801_create_whitehall_imported_assets.rb
ChrisBAshton: I agree that the naming of `error_log` suggests a number of entries, and that we only really intend to store one message. `error_message` would be a better name. I'll change it in this table 👍

1pretz1:
Just had a skim through this PR and to add to Kevins point, I've copied your AddDocumentRefToWhitehallImports migration and `belongs_to ...` in the WhitehallDocument model with adjustments from Kevins suggestions which means I can continue with my work with only the migration filename being different between our branches! 

I also noticed you may wan't to rebase against master as `lib/whitehall_importer.rb` for example is missing some recently updated code! 😄

app/models/whitehall_imported_asset.rb
ChrisBAshton: You're right, I was facing issues that forced me to write this - something along the lines of "expected these to match, but they didn't":

```diff
- "{ :foo => \"bar\" }"
+ "{ \"foo\": \"bar\" }"
```

But I've gone back and rebased, and removed this method and the tests are passing! 🤷‍♂

ChrisBAshton: Curious - it just happened again!

`govuk-docker-run bundle exec rake` passes, but if I run `govuk-docker-run bundle exec rspec spec/lib/whitehall_importer/create_image_revision_spec.rb` in isolation, I get this:

```
  2) WhitehallImporter::CreateImageRevision#call Valid image is provided should create a WhitehallImportedAsset record
     Failure/Error: expect(WhitehallImportedAsset.last.variants).to eq(whitehall_image["variants"])
     
       expected: {"s960"=>"https://assets.publishing.service.gov.uk/government/uploads/s960_valid-image.jpg"}
            got: "{\"s960\"=>\"https://assets.publishing.service.gov.uk/government/uploads/s960_valid-image.jpg\"}"
```

ChrisBAshton: For the record, this was down to me using `t.string :variants, json: true, default: "{}"` in the migration. It should have been `t.json :variants, default: {}, null: false`.

1pretz1:
Just letting you know @ChrisBAshton, you may want to do a rebase against master again as -> https://github.com/alphagov/content-publisher/pull/1535 has been merged which includes the code mine and this PR shared but with a few slight differences, apologies! 😭

lib/whitehall_importer/create_revision.rb
ChrisBAshton: Good point. Have now changed it to `find_or_create_image_revisions`.

db/migrate/20191213123801_create_whitehall_imported_assets.rb
ChrisBAshton: Thanks for the idea. I like it.

It took quite a bit of reworking so I've created a new PR: https://github.com/alphagov/content-publisher/pull/1568

I'll close this PR but won't delete the branch, that way I can roll back to this version if needed.

ChrisBAshton:
Closed in favour of #1568.

PR #1535, opened 2019-12-18
Add more import statuses

1pretz1:
Currently the [WhitehallImport](https://github.com/alphagov/content-publisher/blob/fb45dd1ff913e9ffed8ab59ed4799be033f6eb59/app/models/whitehall_import.rb) has 3 statuses (importing, completed and failed) these are now not sufficient for the amount of things that can happen.

Below is the full list of import statuses implemented in this PR:
- importing -> doing the import process
- imported -> importing is complete
- import_aborted -> an AbortImportError is raised during import
- import_failed -> an unexpected error is raised during import
- syncing -> syncing the data with GOV.UK
- sync_failed -> an unexpected error is raised during syncing
- completed -> everything was successful

Related PR (separates out the migration which is also used by https://github.com/alphagov/content-publisher/pull/1530):
https://github.com/alphagov/content-publisher/pull/1534

Trello:
https://trello.com/c/GbgBv1Nz/1233-track-and-implement-more-statuses-in-whitehall-import-aborted-import-failed-syncing-sync-failed-complete

### DIFF
--- a/app/models/whitehall_import.rb
+++ b/app/models/whitehall_import.rb
@@ -6,6 +6,10 @@ class WhitehallImport < ApplicationRecord
   belongs_to :document, optional: true
 
   enum state: { importing: "importing",
-                completed: "completed",
-                failed: "failed" }
+                imported: "imported",
+                import_aborted: "import_aborted",
+                import_failed: "import_failed",
+                syncing: "syncing",
+                sync_failed: "sync failed",
+                completed: "completed" }
 end
--- a/app/models/whitehall_import.rb
+++ b/app/models/whitehall_import.rb
@@ -3,6 +3,8 @@
 # Represents the raw import of a document from Whitehall Publisher and
 # the import status of the document into Content Publisher
 class WhitehallImport < ApplicationRecord
+  belongs_to :document, optional: true
+
   enum state: { importing: "importing",
                 completed: "completed",
                 failed: "failed" }
--- a/db/migrate/20191213094311_add_document_ref_to_whitehall_imports.rb
+++ b/db/migrate/20191213094311_add_document_ref_to_whitehall_imports.rb
@@ -1,0 +1,11 @@
+# frozen_string_literal: true
+
+class AddDocumentRefToWhitehallImports < ActiveRecord::Migration[6.0]
+  def change
+    add_reference :whitehall_imports,
+                  :document,
+                  foreign_key: { on_delete: :restrict },
+                  null: true,
+                  index: false
+  end
+end
--- a/lib/whitehall_importer.rb
+++ b/lib/whitehall_importer.rb
@@ -14,7 +14,7 @@ module WhitehallImporter
       record.update!(state: "completed")
     rescue StandardError => e
       record.update!(error_log: e.message,
-                     state: "failed")
+                     state: "import_failed")
     end
 
     record
--- a/lib/whitehall_importer.rb
+++ b/lib/whitehall_importer.rb
@@ -19,6 +19,8 @@ module WhitehallImporter
   def self.import(whitehall_import)
     document = Import.call(whitehall_import.payload)
     whitehall_import.update!(document: document, state: "imported")
+  rescue AbortImportError => e
+    whitehall_import.update!(error_log: e.inspect, state: "import_aborted")
   rescue StandardError => e
     whitehall_import.update!(error_log: e.inspect, state: "import_failed")
   end
--- a/lib/whitehall_importer.rb
+++ b/lib/whitehall_importer.rb
@@ -1,27 +1,31 @@
 # frozen_string_literal: true
 
 module WhitehallImporter
-  def self.import(whitehall_document)
-    record = WhitehallImport.create!(
+  def self.import_and_sync(whitehall_document)
+    whitehall_import = WhitehallImport.create!(
       whitehall_document_id: whitehall_document["id"],
       payload: whitehall_document,
       content_id: whitehall_document["content_id"],
       state: "importing",
     )
 
-    begin
-      Import.call(whitehall_document)
-      record.update!(state: "completed")
-    rescue StandardError => e
-      record.update!(error_log: e.message,
-                     state: "import_failed")
-    end
+    import(whitehall_import)
 
-    record
+    sync(whitehall_import) if whitehall_import.imported?
+
+    whitehall_import
   end
 
-  def self.sync(document)
-    ResyncService.call(document)
-    ClearLinksetLinks.call(document.content_id)
+  def self.import(whitehall_import)
+    document = Import.call(whitehall_import.payload)
+    whitehall_import.update!(document: document, state: "imported")
+  rescue StandardError => e
+    whitehall_import.update!(error_log: e.inspect, state: "import_failed")
+  end
+
+  def self.sync(whitehall_import)
+    ResyncService.call(whitehall_import.document)
+    ClearLinksetLinks.call(whitehall_import.document.content_id)
+    whitehall_import.update!(state: "completed")
   end
 end
--- a/lib/whitehall_importer.rb
+++ b/lib/whitehall_importer.rb
@@ -17,22 +17,30 @@ module WhitehallImporter
   end
 
   def self.import(whitehall_import)
-    document = Import.call(whitehall_import.payload)
-    whitehall_import.update!(document: document, state: "imported")
-  rescue AbortImportError => e
-    whitehall_import.update!(error_log: e.inspect, state: "import_aborted")
-  rescue StandardError => e
-    whitehall_import.update!(error_log: e.inspect, state: "import_failed")
+    raise "Cannot import with a state of #{whitehall_import.state}" unless whitehall_import.importing?
+
+    begin
+      document = Import.call(whitehall_import.payload)
+      whitehall_import.update!(document: document, state: "imported")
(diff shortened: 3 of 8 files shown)


db/migrate/20191213094311_add_document_ref_to_whitehall_imports.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+class AddDocumentRefToWhitehallImports < ActiveRecord::Migration[6.0]
+  def change
+    add_reference :whitehall_imports,
+                  :document,
+                  foreign_key: { on_delete: :restrict },
+                  null: true
```
kevindew: May want index: false here as I don't expect we'll be using the index.

spec/factories/whitehall_import_factory.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+FactoryBot.define do
+  factory :whitehall_import do
+    state { "importing" }
+    payload { build(:whitehall_export_document) }
+    whitehall_document_id { payload["id"] }
+    content_id { payload["content_id"] }
+    document { create(:document, imported_from: "whitehall") }
```
kevindew: can we do this through a regular association? e.g `association :document, imported_from: "whitehall"` ?

app/models/whitehall_import.rb
```diff
@@ -6,6 +6,10 @@ class WhitehallImport < ApplicationRecord
...
 
   enum state: { importing: "importing",
-                completed: "completed",
-                failed: "failed" }
+                imported: "imported",
+                import_aborted: "import_aborted",
+                import_failed: "import_failed",
+                syncing: "syncing",
+                sync_failed: "sync failed",
+                completed: "completed" }
```
kevindew: 👍

spec/lib/whitehall_importer_spec.rb
```diff
@@ -1,30 +1,53 @@
...
-      whitehall_export = build(:whitehall_export_document)
-      record = WhitehallImporter.import(whitehall_export)
-      expect(record.payload).to eq(whitehall_export)
+    it "doesn't sync if import fails" do
+      allow(WhitehallImporter::Import).to receive(:call)
+        .with(whitehall_export_document)
+        .and_raise(WhitehallImporter::AbortImportError, "Booo, import failed")
+
+      expect(ResyncService).not_to receive(:call)
+      expect(WhitehallImporter::ClearLinksetLinks).not_to receive(:call)
```
kevindew: You can probably simplify this to be that :sync doesn't get called.

lib/whitehall_importer.rb
```diff
@@ -1,27 +1,31 @@
...
+    document = Import.call(whitehall_import.payload)
+    whitehall_import.update!(document: document, state: "imported")
+  rescue StandardError => e
+    whitehall_import.update!(error_log: e.message, state: "import_failed")
   end
 
-  def self.sync(document)
-    ResyncService.call(document)
-    ClearLinksetLinks.call(document.content_id)
+  def self.sync(whitehall_import)
```
kevindew: Should this and `.import` perhaps raise if the import model isn't in the expected state?

lib/whitehall_importer.rb
```diff
@@ -19,6 +19,8 @@ def self.import_and_sync(whitehall_document)
   def self.import(whitehall_import)
     document = Import.call(whitehall_import.payload)
     whitehall_import.update!(document: document, state: "imported")
+  rescue AbortImportError => e
+    whitehall_import.update!(error_log: e.message, state: "import_aborted")
```
kevindew: Nice 👍

spec/lib/whitehall_importer_spec.rb
```diff
@@ -101,5 +101,22 @@
...
+          .and_raise(GdsApi::HTTPTooManyRequests.new(429, message))
+      end
+
+      let(:message) { "Ahhh too many requests" }
+
+      it "marks the import as failed due to sync issues and logs the error" do
+        WhitehallImporter.sync(whitehall_import)
+
+        expect(whitehall_import).to be_sync_failed
+        expect(whitehall_import.error_log).to eq(message)
```
kevindew: Not really a problem for now but I wonder if what we're storing in this error log isn't really sufficient without the class name of the exception 🤔

db/migrate/20191213094311_add_document_ref_to_whitehall_imports.rb
1pretz1: Sure!

spec/factories/whitehall_import_factory.rb
1pretz1: We sure can, done!

spec/lib/whitehall_importer_spec.rb
1pretz1: Good shout!

lib/whitehall_importer.rb
1pretz1: Hmm I don't think these raises are needed due to import being called just after the WhitehallImport record is persisted with a state `importing`, if this for whatever reason failed then the code wouldn't get run anyway! 

As for the sync function, there is a conditional checking whether the thing has been imported before calling the function so again, the sync code would never run if it hasn't been imported. I can see the value if these functions were to be called separately but right now it seems overkill, what do you think?

spec/lib/whitehall_importer_spec.rb
1pretz1: I had a little think about this and think it's okay as we have quite descriptive abort messages and any ruby / rails exception should have a pretty clear message of went wrong without needing the actual class. 

Possibly the error messages from GdsApi may be poor but I'd hope that's not the case.

1pretz1:
Ready for review number 2! @kevindew  💯

spec/lib/whitehall_importer_spec.rb
kevindew: Yeah I think it's fine for abort too as we already know it from the status.

For an unexpected exception I had a little play around and think using inspect might be the most promising approach:

```
[3] pry(#<GdsApi::TestHelpers::PublishingApi::#stub_publishing_api_has_path_reservation_for>)> error.to_s
=> "URL: http://publishing-api.dev.gov.uk/paths/foo\nResponse body:\n{\"error\":{\"code\":422,\"message\":\"Base path /foo is already reserved by foo-publisher\",\"fields\":{\"base_path\":[\"/foo is already reserved by foo-publisher\"]}}}\n\nRequest body:\n{:publishing_app=>\"bar-publisher\"}"
[4] pry(#<GdsApi::TestHelpers::PublishingApi::#stub_publishing_api_has_path_reservation_for>)> error.class.name
=> "GdsApi::HTTPUnprocessableEntity"
[5] pry(#<GdsApi::TestHelpers::PublishingApi::#stub_publishing_api_has_path_reservation_for>)> error.inspect
=> "#<GdsApi::HTTPUnprocessableEntity: URL: http://publishing-api.dev.gov.uk/paths/foo\nResponse body:\n{\"error\":{\"code\":422,\"message\":\"Base path /foo is already reserved by foo-publisher\",\"fields\":{\"base_path\":[\"/foo is already reserved by foo-publisher\"]}}}\n\nRequest body:\n{:publishing_app=>\"bar-publisher\"}>"
```

lib/whitehall_importer.rb
kevindew: I think this depends on whether we consider these methods to be part of the public interface of WhitehallImporter or not.

We can mark them as private with `private_class_method` if they're just private. Though I'd expect we'd at least want sync available in case we want to try retry if a sync fails (Bruce and I have been talking about failures and retries today).

I anticipated we'd want the raises as a sense check since if a whitehall_import is passed to either of these in an unexpected state it causes data inconsistency. For example if you call .import with something that is already imported you'll get it set "import_failed" when that is not the case. Similarly if you pass a whitehall_import that is "importing" into .sync then it'll end up with a sync_failed status despite having never been imported.

1pretz1: True, however we my want to think of the states that should execute the raise if we're thinking about this code being used beyond the scope of this piece of work. E.g for sync, I assume a state of `sync_failed` or `imported` should be allowed to pass through, but any of the rest should be raised as `sync_failed` could've occurred because of a transient publishing-api error.

spec/lib/whitehall_importer_spec.rb
1pretz1: Inspect looks like the best of all worlds so will alter the code accordingly!

lib/whitehall_importer.rb
1pretz1: Have added the raises on the back of your suggestion!

lib/whitehall_importer.rb
```diff
@@ -1,27 +1,42 @@
...
+    import(whitehall_import)
 
-    record
+    sync(whitehall_import) if whitehall_import.imported?
+
+    whitehall_import
+  end
+
+  def self.import(whitehall_import)
+    raise "Cannot import with a state of #{whitehall_import.state}" unless whitehall_import.importing?
```
kevindew: I don't think this exception should be caught by this method (should be a begin / end) - as we don't to change the state of the import on this sort of error.

Same for sync.

1pretz1: Good point!

PR #1538, opened 2019-12-18
Yet more resilience on BulkData

kevindew:
Trello: https://trello.com/c/TQ6GPOHU/1228-spike-into-loading-governments-from-publishing-api

I learnt today that the Publishing API can take a shockingly long amount of time to return data from the editions endpoint. See:

Screenshot 2019-12-18 at 12 57 30

To handle these situations a bit better this adds further logging in for when an error occurs and increases the timeout massively. Hopefully I can have a chat with Platform Health to see if they can look into this as it's very strange that it's taking quite so long.

### DIFF
--- a/app/jobs/populate_bulk_data_job.rb
+++ b/app/jobs/populate_bulk_data_job.rb
@@ -9,5 +9,8 @@ class PopulateBulkDataJob < ApplicationJob
     run_exclusively do
       BulkData::GovernmentRepository.new.populate_cache(older_than: 5.minutes.ago)
     end
+  rescue BulkData::RemoteDataUnavailableError => e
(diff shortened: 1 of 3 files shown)


kevindew:
> Wow, is that in seconds? 47 for one request is -> 🐌Will be interesting to see what's causing the holdup!

From digging in I think it's caused by keyset pagination in Publishing API when it is checking if it is the first page: https://github.com/alphagov/publishing-api/blob/master/app/queries/keyset_pagination.rb#L58-L63 looks like a massive table scan is involved /cc @thomasleese

1pretz1:
Ah ha, that looks like it could be the culprit! If I'm understanding correctly, it's scanning the entire editions table for every page it's returning through pagination?

PR #1549, opened 2019-12-20
Remove use of publishing_api_v2

leenagupte:
## What's changed and why?

`publishing_api` and `publishing_api_v2` were combined in version 63.1.0 of `gds-api-adapters`, so all references to publishing_api_v2 have been removed.

### DIFF
--- a/spec/services/resync_service_spec.rb
+++ b/spec/services/resync_service_spec.rb
@@ -211,12 +211,5 @@ RSpec.describe ResyncService do
         end
       end
     end
-
-    def stub_publishing_api_path_reservation(base_path, params)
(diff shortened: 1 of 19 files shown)


spec/services/resync_service_spec.rb
```diff
@@ -211,12 +211,5 @@
         end
       end
     end
-
-    def stub_publishing_api_path_reservation(base_path, params)
-      endpoint = GdsApi::TestHelpers::PublishingApi::PUBLISHING_API_ENDPOINT
-      url = endpoint + "/paths#{base_path}"
-
-      stub_request(:put, url).with(body: params)
-    end
```
kevindew: Yay 🎉

PR #1568, opened 2019-12-23
Redirect Whitehall images and attachments

ChrisBAshton:
This PR has all the code for tidying up a Whitehall document's assets once the document has been imported into Content Publisher. After import, we can run the new `MigrateAssets` service to:

* delete all Whitehall assets (images/file attachments) that only existed on the draft stack
* redirect all 'master' assets (e.g. the 960 size image, rather than its thumbnails) to the new master version on Content Publisher, if the Whitehall assets were on the live stack
* delete all variants of Whitehall file attachments (whether on live or draft stack), since they'll no longer appear anywhere
* redirect all variants of Whitehall images to the new master image on Content Publisher, since the old images may be hotlinked from elsewhere
* log any assets that fail to migrate, and fail the sync step at the end

### At the import stage
Creates a `WhitehallMigration::AssetImport` record for every asset (Image/FileAttachment) referenced in the Whitehall export. It creates a record for each variant of each asset too. However, it avoids creating duplicate records, so that if an asset exists in multiple editions in a Whitehall export, it will only have one record created for it.

### At the migration stage

We've created a `MigrateAssets` service which is called from the WhitehallImporter `sync`/`import_and_sync` methods. It iterates over each `WhitehallMigration::AssetImport` record, deleting the asset if its 'master' asset is not live, or deleting it if it is a 'variant' file attachment (e.g. thumbnail). Otherwise, it redirects the Whitehall asset to the main asset on Content Publisher.

**NOTE: if something goes wrong in the import_and_sync step, such that the master asset was not successfully synchronised with asset manager and does not have the 'live' state, there is a risk we will delete Whitehall assets and have no assets at all. However, in a dev discussion before Christmas we decided that the migration code should make the assumption that everything was imported correctly.**

## Background information

This work comes out of a spike in #1509, and further investigative work in #1530.

Trello card: https://trello.com/c/DMpTECux/1232-handle-redirecting-and-removing-old-assets-as-part-of-whitehall-migration

### DIFF
--- a/app/models/whitehall_migration/asset_import.rb
+++ b/app/models/whitehall_migration/asset_import.rb
@@ -14,6 +14,18 @@ class WhitehallMigration::AssetImport < ApplicationRecord
 
   validate :associated_with_only_image_or_file_attachment
 
+  def content_publisher_asset
+    if image_revision.present?
+      if variant.nil? || variant == "s960"
+        image_revision.asset("960")
+      elsif variant == "s300"
+        image_revision.asset("300")
+      end
+    elsif variant.nil?
+      file_attachment_revision.asset
+    end
+  end
+
   def whitehall_asset_id
     url_array = original_asset_url.to_s.split("/")
     # https://github.com/alphagov/asset-manager#create-an-asset
--- a/app/models/whitehall_migration/asset_import.rb
+++ b/app/models/whitehall_migration/asset_import.rb
@@ -1,0 +1,10 @@
+# frozen_string_literal: true
+
+# Represents the raw import of an asset from Whitehall Publisher and
+# the import status of the asset into Content Publisher
+class WhitehallMigration::AssetImport < ApplicationRecord
+  enum state: { redirected: "redirected",
+                removed: "removed",
+                migration_failed: "migration_failed",
+                pending: "pending" }
+end
--- a/app/models/whitehall_migration/asset_import.rb
+++ b/app/models/whitehall_migration/asset_import.rb
@@ -3,6 +3,10 @@
 # Represents the raw import of an asset from Whitehall Publisher and
 # the import status of the asset into Content Publisher
 class WhitehallMigration::AssetImport < ApplicationRecord
+  belongs_to :document_import
+  belongs_to :image_revision, class_name: "Image::Revision", optional: true
+  belongs_to :file_attachment_revision, class_name: "FileAttachment::Revision", optional: true
+
   enum state: { redirected: "redirected",
                 removed: "removed",
                 migration_failed: "migration_failed",
--- a/app/models/whitehall_migration/asset_import.rb
+++ b/app/models/whitehall_migration/asset_import.rb
@@ -14,6 +14,12 @@ class WhitehallMigration::AssetImport < ApplicationRecord
 
   validate :associated_with_only_image_or_file_attachment
 
+  def whitehall_asset_id
+    url_array = original_asset_url.to_s.split("/")
+    # https://github.com/alphagov/asset-manager#create-an-asset
+    url_array[url_array.length - 2]
+  end
+
 private
 
   def associated_with_only_image_or_file_attachment
--- a/app/models/whitehall_migration/asset_import.rb
+++ b/app/models/whitehall_migration/asset_import.rb
@@ -11,4 +11,14 @@ class WhitehallMigration::AssetImport < ApplicationRecord
                 removed: "removed",
                 migration_failed: "migration_failed",
                 pending: "pending" }
+
+  validate :associated_with_only_image_or_file_attachment
+
+private
+
+  def associated_with_only_image_or_file_attachment
+    if image_revision.present? && file_attachment_revision.present?
+      errors.add(:base, "Cannot be associated with both image revision AND file attachment revision")
+    end
+  end
 end
--- a/app/models/whitehall_migration/document_import.rb
+++ b/app/models/whitehall_migration/document_import.rb
@@ -4,6 +4,7 @@
 # the import status of the document into Content Publisher
 class WhitehallMigration::DocumentImport < ApplicationRecord
   belongs_to :document, optional: true
+  has_many :assets, class_name: "WhitehallMigration::AssetImport"
 
   enum state: { pending: "pending",
                 importing: "importing",
--- a/app/models/whitehall_migration/document_import.rb
+++ b/app/models/whitehall_migration/document_import.rb
@@ -14,4 +14,8 @@ class WhitehallMigration::DocumentImport < ApplicationRecord
                 syncing: "syncing",
                 sync_failed: "sync failed",
                 completed: "completed" }
+
+  def migratable_assets
+    assets.select { |a| a.pending? || a.migration_failed? }
+  end
 end
--- a/lib/whitehall_importer/create_file_attachment_revision.rb
+++ b/lib/whitehall_importer/create_file_attachment_revision.rb
@@ -19,13 +19,15 @@ module WhitehallImporter
       check_file_requirements(decorated_file)
 
       blob_revision = create_blob_revision(decorated_file)
-      FileAttachment::Revision.create!(
+      revision = FileAttachment::Revision.create!(
         blob_revision: blob_revision,
         file_attachment: FileAttachment.create!,
         metadata_revision: FileAttachment::MetadataRevision.create!(
           title: whitehall_file_attachment["title"],
         ),
       )
+      record_assets(revision)
+      revision
     end
 
   private
@@ -64,6 +66,22 @@ module WhitehallImporter
       return if whitehall_file_attachment["type"] == "FileAttachment"
 
       raise WhitehallImporter::AbortImportError, "Unsupported file attachment: #{whitehall_file_attachment['type']}"
+    end
+
+    def record_assets(revision)
+      WhitehallMigration::AssetImport.create!(
+        document_import: document_import,
+        file_attachment_revision: revision,
+        original_asset_url: whitehall_file_attachment["url"],
+      )
+      whitehall_file_attachment["variants"].each do |variant, metadata|
+        WhitehallMigration::AssetImport.create!(
+          document_import: document_import,
+          file_attachment_revision: revision,
+          original_asset_url: metadata["url"],
+          variant: variant,
+        )
+      end
     end
 
     def abort_on_issue(issues)
--- a/lib/whitehall_importer/create_file_attachment_revision.rb
+++ b/lib/whitehall_importer/create_file_attachment_revision.rb
@@ -6,7 +6,8 @@ module WhitehallImporter
       new(*args).call
     end
 
-    def initialize(whitehall_file_attachment, existing_filenames = [])
+    def initialize(document_import, whitehall_file_attachment, existing_filenames = [])
+      @document_import = document_import
       @whitehall_file_attachment = whitehall_file_attachment
       @existing_filenames = existing_filenames
     end
@@ -29,7 +30,7 @@ module WhitehallImporter
 
   private
 
-    attr_reader :whitehall_file_attachment, :existing_filenames
+    attr_reader :document_import, :whitehall_file_attachment, :existing_filenames
 
     def download_file
       URI.parse(whitehall_file_attachment["url"]).open
--- a/lib/whitehall_importer/create_image_revision.rb
+++ b/lib/whitehall_importer/create_image_revision.rb
@@ -17,7 +17,7 @@ module WhitehallImporter
     def call
       temp_image = normalise_image(download_file)
       blob_revision = create_blob_revision(temp_image)
-      Image::Revision.create!(
+      revision = Image::Revision.create!(
         image: Image.new,
         metadata_revision: Image::MetadataRevision.new(
           caption: whitehall_image["caption"],
@@ -25,6 +25,8 @@ module WhitehallImporter
         ),
         blob_revision: blob_revision,
       )
+      record_assets(revision)
+      revision
     end
 
   private
@@ -59,6 +61,22 @@ module WhitehallImporter
       abort_on_issue(normaliser.issues)
 
       image
+    end
+
+    def record_assets(revision)
+      WhitehallMigration::AssetImport.create!(
+        document_import: document_import,
+        image_revision: revision,
+        original_asset_url: whitehall_image["url"],
+      )
+      whitehall_image["variants"].each do |variant, url|
+        WhitehallMigration::AssetImport.create!(
+          document_import: document_import,
+          image_revision: revision,
+          original_asset_url: url,
+          variant: variant,
+        )
+      end
     end
 
     def abort_on_issue(issues)
--- a/lib/whitehall_importer/create_image_revision.rb
+++ b/lib/whitehall_importer/create_image_revision.rb
@@ -2,13 +2,14 @@
 
 module WhitehallImporter
   class CreateImageRevision
-    attr_reader :whitehall_image, :filenames
+    attr_reader :document_import, :whitehall_image, :filenames
 
     def self.call(*args)
       new(*args).call
     end
 
-    def initialize(whitehall_image, filenames = [])
+    def initialize(document_import, whitehall_image, filenames = [])
+      @document_import = document_import
       @whitehall_image = whitehall_image
       @filenames = filenames
     end
--- a/lib/whitehall_importer/create_revision.rb
+++ b/lib/whitehall_importer/create_revision.rb
@@ -27,8 +27,8 @@ module WhitehallImporter
       document_type_key = DOCUMENT_SUB_TYPES.reject { |t| whitehall_edition[t].nil? }.first
       raise AbortImportError, "Edition has an unsupported document type" unless SUPPORTED_DOCUMENT_TYPES.include?(whitehall_edition[document_type_key])
 
-      file_attachment_revisions = create_file_attachment_revisions(whitehall_edition["attachments"])
-      image_revisions = create_image_revisions(whitehall_edition["images"])
+      file_attachment_revisions = find_or_create_file_attachment_revisions(whitehall_edition["attachments"])
+      image_revisions = find_or_create_image_revisions(whitehall_edition["images"])
       Revision.create!(
         document: document,
         number: document.next_revision_number,
@@ -118,15 +118,29 @@ module WhitehallImporter
       associations.map { |association| association["content_id"] }
     end
 
-    def create_image_revisions(images)
+    def find_or_create_image_revisions(images)
       images.reduce([]) do |memo, image|
-        memo << WhitehallImporter::CreateImageRevision.call(document_import, image, memo.map(&:filename))
+        already_imported = WhitehallMigration::AssetImport.find_by(original_asset_url: image["url"])
+        revision = if already_imported
+                     already_imported.image_revision
+                   else
+                     WhitehallImporter::CreateImageRevision
+                       .call(document_import, image, memo.map(&:filename))
+                   end
+        memo << revision
       end
     end
 
-    def create_file_attachment_revisions(file_attachments)
+    def find_or_create_file_attachment_revisions(file_attachments)
       file_attachments.reduce([]) do |memo, file_attachment|
-        memo << WhitehallImporter::CreateFileAttachmentRevision.call(document_import, file_attachment, memo.map(&:filename))
+        already_imported = WhitehallMigration::AssetImport.find_by(original_asset_url: file_attachment["url"])
+        revision = if already_imported
+                     already_imported.file_attachment_revision
+                   else
+                     WhitehallImporter::CreateFileAttachmentRevision
+                       .call(document_import, file_attachment, memo.map(&:filename))
+                   end
+        memo << revision
       end
     end
 
--- a/lib/whitehall_importer/create_revision.rb
+++ b/lib/whitehall_importer/create_revision.rb
@@ -118,13 +118,13 @@ module WhitehallImporter
 
     def create_image_revisions(images)
       images.reduce([]) do |memo, image|
-        memo << WhitehallImporter::CreateImageRevision.call(image, memo.map(&:filename))
+        memo << WhitehallImporter::CreateImageRevision.call(nil, image, memo.map(&:filename))
       end
     end
 
     def create_file_attachment_revisions(file_attachments)
       file_attachments.reduce([]) do |memo, file_attachment|
-        memo << WhitehallImporter::CreateFileAttachmentRevision.call(file_attachment, memo.map(&:filename))
+        memo << WhitehallImporter::CreateFileAttachmentRevision.call(nil, file_attachment, memo.map(&:filename))
       end
     end
 
--- a/lib/whitehall_importer/create_revision.rb
+++ b/lib/whitehall_importer/create_revision.rb
@@ -2,7 +2,9 @@
 
 module WhitehallImporter
   class CreateRevision
-    attr_reader :document, :whitehall_edition
+    attr_reader :document_import, :whitehall_edition
+
+    delegate :document, to: :document_import
 
     SUPPORTED_DOCUMENT_TYPES = %w(news_story press_release).freeze
     DOCUMENT_SUB_TYPES = %w[
@@ -16,8 +18,8 @@ module WhitehallImporter
       new(*args).call
     end
 
-    def initialize(document, whitehall_edition)
-      @document = document
+    def initialize(document_import, whitehall_edition)
+      @document_import = document_import
       @whitehall_edition = whitehall_edition
     end
 
@@ -118,13 +120,13 @@ module WhitehallImporter
 
     def create_image_revisions(images)
       images.reduce([]) do |memo, image|
-        memo << WhitehallImporter::CreateImageRevision.call(nil, image, memo.map(&:filename))
+        memo << WhitehallImporter::CreateImageRevision.call(document_import, image, memo.map(&:filename))
       end
     end
 
     def create_file_attachment_revisions(file_attachments)
       file_attachments.reduce([]) do |memo, file_attachment|
-        memo << WhitehallImporter::CreateFileAttachmentRevision.call(nil, file_attachment, memo.map(&:filename))
+        memo << WhitehallImporter::CreateFileAttachmentRevision.call(document_import, file_attachment, memo.map(&:filename))
       end
     end
 
--- a/lib/whitehall_importer/import.rb
+++ b/lib/whitehall_importer/import.rb
@@ -2,14 +2,15 @@
 
 module WhitehallImporter
   class Import
-    attr_reader :whitehall_document
+    attr_reader :document_import, :whitehall_document
 
     def self.call(*args)
       new(*args).call
     end
 
-    def initialize(whitehall_document)
-      @whitehall_document = whitehall_document
+    def initialize(document_import)
+      @document_import = document_import
+      @whitehall_document = document_import.payload
     end
 
     def call
--- a/lib/whitehall_importer/import.rb
+++ b/lib/whitehall_importer/import.rb
@@ -17,6 +17,7 @@ module WhitehallImporter
       ActiveRecord::Base.transaction do
         user_ids = create_users(whitehall_document["users"])
         document = create_document(user_ids)
+        document_import.update!(document: document)
 
         whitehall_document["editions"].each_with_index do |edition, edition_number|
           CreateEdition.call(
--- a/lib/whitehall_importer/import.rb
+++ b/lib/whitehall_importer/import.rb
@@ -16,12 +16,11 @@ module WhitehallImporter
     def call
       ActiveRecord::Base.transaction do
         user_ids = create_users(whitehall_document["users"])
-        document = create_document(user_ids)
-        document_import.update!(document: document)
+        document_import.update!(document: create_document(user_ids))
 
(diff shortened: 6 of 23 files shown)


app/models/whitehall_migration/asset_import.rb
```diff
@@ -9,6 +9,12 @@ class WhitehallMigration::AssetImport < ApplicationRecord
 
   validate :associated_with_only_image_or_file_attachment
 
+  def asset_manager_id
```
kevindew: This isn't actually an asset manager id - this is the id whitehall assigns for the asset. Might be better named as whitehall_asset_id

app/models/whitehall_migration/document_import.rb
```diff
@@ -4,6 +4,8 @@
 # the import status of the document into Content Publisher
 class WhitehallMigration::DocumentImport < ApplicationRecord
   belongs_to :document, optional: true
+  has_many :asset_imports
+  alias_attribute :assets, :asset_imports
```
kevindew: Might be better to name the association assets e.g. `has_many :assets, class: "WhitehallMigration::AssetImport", foreign_key: :asset_import_id`

lib/whitehall_importer/migrate_assets.rb
```diff
@@ -15,7 +15,12 @@ def initialize(whitehall_import)
     def call
       whitehall_import.assets.each do |whitehall_asset|
         whitehall_asset.update!(state: "processing")
-        # TODO
+        if whitehall_asset.main_asset.state == "live"
```
kevindew: This can be `whitehall_asset.main_asset.live?`

lib/whitehall_importer/migrate_assets.rb
```diff
@@ -0,0 +1,35 @@
...
+            )
+          else
+            GdsApi.asset_manager.delete_asset(whitehall_asset.asset_manager_id)
+          end
+          whitehall_asset.update!(state: "processed")
+        rescue GdsApi::BaseError, StandardError => e
+          whitehall_asset.update!(state: "migration_failed", error_message: e.inspect)
+        end
+      end
+    end
```
kevindew: I think there's a bit too much going on in this method and could used being separated out.

app/models/whitehall_migration/asset_import.rb
```diff
@@ -0,0 +1,33 @@
...
+
+  validate :associated_with_only_image_or_file_attachment
+
+  def main_asset
+    if image_revision.present?
+      image_revision.asset("960")
+    else
+      file_attachment_revision.asset
+    end
+  end
```
kevindew: I wonder if this method should be content_publisher_asset and be able to determine what based on the variant.

For example if we have a variant of original (or nil) we return `image_revision.asset("960")` or `file_attachment_revision.asset` then after that we have things like s960 and similar where they may return an image_revision asset. Then things like thumbnail or unsupported Content Publisher sizes where we'd return nil.

app/models/whitehall_migration/asset_import.rb
```diff
@@ -0,0 +1,33 @@
...
+
+# Represents the raw import of an asset from Whitehall Publisher and
+# the import status of the asset into Content Publisher
+class WhitehallMigration::AssetImport < ApplicationRecord
+  belongs_to :document_import
+  belongs_to :image_revision, class_name: "Image::Revision", optional: true
+  belongs_to :file_attachment_revision, class_name: "FileAttachment::Revision", optional: true
+
+  validate :associated_with_only_image_or_file_attachment
+
```
kevindew: We probably want an enum of states in here that this object can have.

I think you've got not_processed, processing and processed. Is that right? It might be better to start with pending for consistency with DocumentImport. Then I'd suggest having end states of either removed or redirected rather than processed that capture the outcome.

lib/whitehall_importer/create_revision.rb
```diff
@@ -111,15 +113,23 @@ def tags(associations)
...
 
-    def create_image_revisions(images)
+    def find_or_create_image_revisions(images)
       images.reduce([]) do |memo, image|
-        memo << WhitehallImporter::CreateImageRevision.call(image, memo.map(&:filename))
+        memo << if (already_imported = WhitehallMigration::AssetImport.find_by(original_asset_url: image["url"]))
+                  already_imported.image_revision
+                else
+                  WhitehallImporter::CreateImageRevision.call(record, image, memo.map(&:filename))
+                end
```
kevindew: There's quite a lot going on in this assignment it looks like it might be working more like a map than a reduce unless I'm missing something.

lib/whitehall_importer/migrate_assets.rb
```diff
@@ -0,0 +1,35 @@
...
+    def self.call(*args)
+      new(*args).call
+    end
+
+    def initialize(whitehall_import)
+      @whitehall_import = whitehall_import
+    end
+
+    def call
+      whitehall_import.assets.reject { |asset| asset.state == "processed" }.each do |whitehall_asset|
```
kevindew: Might be better to have this as a scope rather than a reject method? also I'd imagine we only want things that are not_processed as (unlikely as it could be) wouldn't want something that is processing.

spec/factories/whitehall_migration_asset_import_factory.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
+
+FactoryBot.define do
+  factory :whitehall_migration_asset_import, class: WhitehallMigration::AssetImport do
+    document_import { association :whitehall_migration_document_import }
+    state { "not_processed" }
+
+    # default to being an image, as validation requires these values to exist
+    original_asset_url { "https://asset-manager.gov.uk/blah/847150/foo.jpg" }
+    image_revision { build(:image_revision) }
```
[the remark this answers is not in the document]
ChrisBAshton: I did try that and it wouldn't work for me 😞 may have to pair on that one!

Only snag with each trait erasing the other's attributes is that I write a validator test that explicitly passes both `:image` and `:file_attachment` traits to make sure the validation error is raised. I suppose I could create a new factory that creates both, just seems a bit odd to have a new factory that will only be used by one test.

app/models/whitehall_migration/document_import.rb
ChrisBAshton: Thanks, was searching for the right syntax to use!

spec/factories/whitehall_migration_asset_import_factory.rb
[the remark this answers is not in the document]
kevindew: Sure, you might want to test that by just creating an instance of the class without a factory? E.g. `AssetImport.new(image_revision: image, file_attachment_revision: file)`

lib/whitehall_importer/create_revision.rb
ChrisBAshton: Ahh, I just went in to refactor this and then remembered why it's a `reduce`. It's to keep track of which filenames have been used, so that we avoid having any duplicate filenames (this is passed in the `memo.map(&:filename)` part).

kevindew: Oh yes shoot sorry that rings a bell. 

Maybe we can do something like:

```
    def find_or_create_file_attachment_revisions(file_attachments)
      file_attachments.reduce([]) do |memo, file_attachment|
        already_imported = WhitehallMigration::AssetImport.find_by(original_asset_url: file_attachment["url"])
        
        revision = if already imported
                     already_imported.file_attachment_revision
                   else
                     WhitehallImporter::CreateFileAttachmentRevision
                       .call(record, file_attachment,memo.map(&:filename))
                   end
        memo << revision
      end
    end
```

app/models/whitehall_migration/document_import.rb
ChrisBAshton: Hmmm this is causing some failures - I had to change `class: "WhitehallMigration::AssetImport"` to `class_name: "WhitehallMigration::AssetImport"`, but was getting some transaction PostgreSQL errors like this:

```
     ActiveRecord::StatementInvalid:
       PG::InFailedSqlTransaction: ERROR:  current transaction is aborted, commands ignored until end of transaction block
     # ./lib/whitehall_importer.rb:45:in `rescue in sync'
     # ./lib/whitehall_importer.rb:36:in `sync'
     # ./lib/whitehall_importer.rb:14:in `import_and_sync'
     # ./spec/lib/whitehall_importer_spec.rb:21:in `block (3 levels) in <top (required)>'
     # ------------------
     # --- Caused by: ---
     # PG::UndefinedColumn:
     #   ERROR:  column whitehall_migration_asset_imports.asset_import_id does not exist
     #   LINE 1: ....* FROM "whitehall_migration_asset_imports" WHERE "whitehall...
     #                                                                ^
     #   ./lib/whitehall_importer/migrate_assets.rb:16:in `reject'
```

Also tried renaming `foreign_key: :whitehall_migration_asset_import_id` but that column doesn't exist either.

kevindew: Sure, maybe come grab me about that if you're still about the office and we can work it out.

spec/factories/whitehall_migration_asset_import_factory.rb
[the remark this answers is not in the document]
ChrisBAshton: For the record, I was getting:

```
Failures:

  1) WhitehallImporter::MigrateAssets.call should take a WhitehallMigration::DocumentImport record as an argument
     Failure/Error: expect { described_class.call(whitehall_import) }.not_to raise_error
     
       expected no Exception, got #<NoMethodError: undefined method `image=' for #<WhitehallMigration::AssetImport:0x000055fd7eff9a20>> with backtrace:
         # ./spec/lib/whitehall_importer/migrate_assets_spec.rb:8:in `block (3 levels) in <top (required)>'
         # ./spec/lib/whitehall_importer/migrate_assets_spec.rb:17:in `block (3 levels) in <top (required)>'
         # ./spec/lib/whitehall_importer/migrate_assets_spec.rb:24:in `block (4 levels) in <top (required)>'
         # ./spec/lib/whitehall_importer/migrate_assets_spec.rb:24:in `block (3 levels) in <top (required)>'
     # ./spec/lib/whitehall_importer/migrate_assets_spec.rb:24:in `block (3 levels) in <top (required)>'
```

But if I rename the trait to `for_image`, like in the examples, it works 🎉  must be some name conflict, as `Image` is a thing. Happy to go through and change this everywhere in the PR (as it's cleaner) unless you have any objections.

app/models/whitehall_migration/document_import.rb
ChrisBAshton: For the record, `foreign_key: :asset_import_id` wasn't needed. Omitting it made it work 🎉

app/models/whitehall_migration/asset_import.rb
ChrisBAshton: So you're proposing something like this?

```ruby
def content_publisher_asset
  if variant.nil?
    image_revision.asset("960") || file_attachment_revision.asset
  elsif file_attachment_revision.present?
    # No content publisher equivalents for file attachment variants
    nil
  else
    case image_revision.variant
    when "s960", "s712", "s630"
      image_revision.asset("960")
    when "s465", "s300", "s216"
      image_revision.asset("300")
    else
      # unsupported file size
      nil
    end
  end
end
```

...and in the migrate assets script, we could then simplify to something like:

```diff
if whitehall_asset.main_asset.live? &&
-    (whitehall_asset.variant.nil? || whitehall_asset.image_revision.present?)
+    (whitehall_asset.content_publisher_asset.present?)
  GdsApi.asset_manager.update_asset(
    whitehall_asset.whitehall_asset_id,
-    redirect_url: whitehall_asset.main_asset.file_url,
+    redirect_url: whitehall_asset.content_publisher_asset,
  )
else
  GdsApi.asset_manager.delete_asset(whitehall_asset.whitehall_asset_id)
end
```

kevindew: Yup 👍

lib/whitehall_importer/migrate_assets.rb
ChrisBAshton: Having difficulty getting this working with associations - see the "WIP: add a 'pending' scope" commit. Would appreciate any pointers! Agree the code will look much nicer when this is done 👍

ChrisBAshton: For the record, you provided some ideas here: https://github.com/alphagov/content-publisher/commit/0a2f64502c1a7e277fe0efad744176d8c0b2b989

> All but the method with the select will fail the tests as the scope and
association based approaches will make a fresh db query and that will
mean any mocks that are set won't have an affect.
> It also needs to use a create approach for the tests as with build any
methods that require fresh queries (scope or different has_many) won't
return any results.
> We already get pre-defined scopes for each of the enum values, which is
how I was able to do a scope of just `pending.or(migration_failed)`

I've ended up abandoning the scopes idea and have opted for a `migratable_assets` method on the `DocumentImport` class - a variation on what you have provided in your spike - which returns a subset of its `assets`.

lib/whitehall_importer/create_file_attachment_revision.rb
```diff
@@ -18,18 +19,19 @@ def call
...
       blob_revision = create_blob_revision(decorated_file)
-      FileAttachment::Revision.create!(
+      revision = FileAttachment::Revision.create!(
         blob_revision: blob_revision,
         file_attachment: FileAttachment.create!,
         metadata_revision: FileAttachment::MetadataRevision.create!(
           title: whitehall_file_attachment["title"],
         ),
       )
+      record_assets(revision)
```
kevindew: Minor but It seems a little strange that this record_assets returns a revision. 

You could do `revision.tap { |r| record_assets(r) }` instead? or just a:
````
record_assets(revision)
revision
```

You might find it's more intuitive to call this method with a tap maybe? revision

lib/whitehall_importer/create_image_revision.rb
```diff
@@ -2,28 +2,30 @@
...
-      Image::Revision.create!(
+      revision = Image::Revision.create!(
         image: Image.new,
         metadata_revision: Image::MetadataRevision.new(
           caption: whitehall_image["caption"],
           alt_text: whitehall_image["alt_text"],
         ),
         blob_revision: blob_revision,
       )
+      record_assets(revision)
```
kevindew: Similar point about return (accidentally hit single comment rather than review previously)

lib/whitehall_importer/create_revision.rb
```diff
@@ -116,15 +118,27 @@ def tags(associations)
...
 
-    def create_file_attachment_revisions(file_attachments)
+    def find_or_create_file_attachment_revisions(file_attachments)
       file_attachments.reduce([]) do |memo, file_attachment|
-        memo << WhitehallImporter::CreateFileAttachmentRevision.call(file_attachment, memo.map(&:filename))
+        already_imported = WhitehallMigration::AssetImport.find_by(original_asset_url: file_attachment["url"])
+        revision = if already_imported
+                     already_imported.file_attachment_revision
+                   else
+                     WhitehallImporter::CreateFileAttachmentRevision.call(document_import, file_attachment, memo.map(&:filename))
```
kevindew: Minor but this line looks pretty long would this work?

```
                     WhitehallImporter::CreateFileAttachmentRevision
                       .call(document_import, file_attachment, memo.map(&:filename))
```

lib/whitehall_importer/import.rb
```diff
@@ -2,32 +2,33 @@
...
+            document_import: document_import,
             current: current?(edition),
             whitehall_edition: edition,
             edition_number: edition_number + 1,
             user_ids: user_ids,
           )
         end
 
-        document
+        document_import.document
```
kevindew: Do we use the return from this anywhere? I wonder if we need to return anything since all we doing is mutating the document_input model.

lib/whitehall_importer/migrate_assets.rb
```diff
@@ -0,0 +1,49 @@
...
+          whitehall_asset.update!(state: "migration_failed", error_message: e.inspect)
+        end
+      end
+    end
+
+  private
+
+    def migrate_asset(whitehall_asset)
+      whitehall_asset.update!(state: "processing")
+      if whitehall_asset.content_publisher_asset.present? && whitehall_asset.content_publisher_asset.live?
```
kevindew: You can probably reduce the two conditions to: `whitehall_asset.content_publisher_asset.&live?`

spec/lib/whitehall_importer/create_file_attachment_revision_spec.rb
```diff
@@ -4,26 +4,42 @@
...
+      expect { revision = described_class.call(document_import, whitehall_file_attachment) }
+        .to change { asset_import.where(variant: nil).count }.by(1)
+        .and change { asset_import.where(variant: "thumbnail").count }.by(1)
+        .and change { asset_import.count }.by(2)
+      expect(asset_import.find_by(variant: nil).file_attachment_revision).to eq(revision)
+      expect(asset_import.find_by(variant: "thumbnail").file_attachment_revision).to eq(revision)
+      expect(asset_import.find_by(variant: nil).original_asset_url)
+        .to eq(whitehall_file_attachment["url"])
+      expect(asset_import.find_by(variant: "thumbnail").original_asset_url)
+        .to eq(whitehall_file_attachment["variants"]["thumbnail"]["url"])
```
kevindew: We've hit quite a lot of assertions in this test I wonder if we can reduce them to make the test more focused. Maybe something like:

```
revision = described_class.call(document_import, whitehall_file_attachment)

expect(document_import.assets.size).to eq(2)
expect(document_import.assets.map(&:attributes)).to contain_exactly(
  a_hash_including(variant: nil,
                   file_attachment_revision_id: revision.id,
                   original_asset_url: whitehall_file_attachment["url"]),
  a_hash_including(variant: "thumbnail,
                   file_attachment_revision_id: revision.id,
                   original_asset_url: whitehall_file_attachment["variants"]["thumbnail"]["url"]),
)
```

spec/lib/whitehall_importer/create_image_revision_spec.rb
```diff
@@ -3,14 +3,32 @@
...
+        expect { image_revision = described_class.call(document_import, whitehall_image) }
+          .to change { asset_import.where(variant: nil).count }.by(1)
+          .and change { asset_import.where(variant: "s960").count }.by(1)
+          .and change { asset_import.count }.by(2)
+        expect(asset_import.find_by(variant: nil).image_revision).to eq(image_revision)
+        expect(asset_import.find_by(variant: "s960").image_revision).to eq(image_revision)
+        expect(asset_import.find_by(variant: nil).original_asset_url)
+          .to eq(whitehall_image["url"])
+        expect(asset_import.find_by(variant: "s960").original_asset_url)
+          .to eq(whitehall_image["variants"]["s960"])
```
kevindew: Similar point about amount of assertions.

spec/lib/whitehall_importer/create_revision_spec.rb
```diff
@@ -69,12 +70,21 @@
...
         expect(revision.image_revisions.last.blob_revision.filename).to eq("image-1.jpg")
       end
 
+      it "skips any image it has encountered before" do
+        image = build(:whitehall_export_image, filename: "image.jpg")
+        revision1 = described_class.call(document_import, build(:whitehall_export_edition, images: [image]))
+        revision2 = described_class.call(document_import, build(:whitehall_export_edition, images: [image]))
+        expect(revision1.image_revisions.count).to eq(1)
+        expect(revision2.image_revisions.count).to eq(1)
+        expect(revision1.image_revisions.first).to eq(revision2.image_revisions.first)
```
kevindew: Can we assert here that `expect(revision1.image_revisions).to eq(revision2.image_revisions)`?

spec/lib/whitehall_importer/create_revision_spec.rb
```diff
@@ -107,14 +117,23 @@
...
         expect(revision.file_attachment_revisions.last.blob_revision.filename).to eq("attach-1.txt")
       end
+
+      it "skips any attachment it has encountered before" do
+        attachment = build(:whitehall_export_file_attachment, filename: "attach.txt")
+        revision1 = described_class.call(document_import, build(:whitehall_export_edition, attachments: [attachment]))
+        revision2 = described_class.call(document_import, build(:whitehall_export_edition, attachments: [attachment]))
+        expect(revision1.file_attachment_revisions.count).to eq(1)
+        expect(revision2.file_attachment_revisions.count).to eq(1)
+        expect(revision1.file_attachment_revisions.first).to eq(revision2.file_attachment_revisions.first)
```
kevindew: Similar point about whether there's a different assertion we can make that these are effectively equal.

spec/lib/whitehall_importer/migrate_assets_spec.rb
```diff
@@ -0,0 +1,143 @@
...
+    end
+
+    it "should take a WhitehallMigration::DocumentImport record as an argument" do
+      expect { described_class.call(build(:whitehall_migration_document_import)) }.not_to raise_error
+    end
+
+    it "should mark each asset as processing before its final state" do
+      whitehall_import = build(:whitehall_migration_document_import, assets: [asset])
+      expect(asset).to receive(:update!).with(state: "processing").once.ordered
+      expect(asset).to receive(:update!).with(hash_including(:state)).once.ordered
```
kevindew: I found this test a bit strange as we've got to get in pretty deep with the mocking and not test anything resultant. It made me wonder if we actually need this "processing" state - I'm not sure it really does anything for us?

spec/lib/whitehall_importer/migrate_assets_spec.rb
```diff
@@ -0,0 +1,143 @@
...
+      described_class.call(whitehall_import)
+    end
+
+    it "should skip migrating any assets that have already been processed" do
+      asset = create(:whitehall_migration_asset_import, state: "removed")
+      whitehall_import = build(:whitehall_migration_document_import, assets: [asset])
+      expect(asset).not_to receive(:update!)
```
kevindew: I wonder if here we should be asserting more on the side-effects. Namely that there aren't any asset manager requests?

spec/lib/whitehall_importer/migrate_assets_spec.rb
```diff
@@ -0,0 +1,143 @@
...
+      expect(asset.error_message).to include("Some problem with GDS API")
+    end
+
+    it "should log standard errors and put into an aborted state" do
+      whitehall_import = build(:whitehall_migration_document_import, assets: [asset])
+      allow(asset).to receive(:whitehall_asset_id).and_raise(StandardError.new("Some error"))
+      described_class.call(whitehall_import)
+      expect(asset.state).to eq("migration_failed")
+      expect(asset.error_message).to include("Some error")
+    end
```
kevindew: I'm not sure we need both of these tests. I can see in the code we catch both but since we don't treat `GdsApi::BaseError` any differently to standard error it seems unnecessary to catch GDS Api specially?

spec/lib/whitehall_importer/migrate_assets_spec.rb
```diff
@@ -0,0 +1,143 @@
...
+
+    it "should redirect live attachments to their content publisher equivalents" do
+      asset = create(:whitehall_migration_asset_import,
+                     :for_file_attachment,
+                     original_asset_url: original_asset_url)
+      whitehall_import = build(:whitehall_migration_document_import, assets: [asset])
+      new_file_url = "https://asset-manager.gov.uk/NEW/847151/foo.pdf"
+      allow(asset.file_attachment_revision.asset).to receive(:live?).and_return(true)
+      allow(asset.file_attachment_revision.asset).to receive(:file_url).and_return(new_file_url)
```
kevindew: Can we create the file_attachment_revision with these attributes rather than needing to mock them?

Similar point on the tests below.

lib/whitehall_importer/migrate_assets.rb
```diff
@@ -0,0 +1,49 @@
...
+    def initialize(whitehall_import)
+      @whitehall_import = whitehall_import
+    end
+
+    def call
+      whitehall_import.migratable_assets.each do |whitehall_asset|
+        begin
+          migrate_asset(whitehall_asset)
+        rescue GdsApi::BaseError, StandardError => e
+          whitehall_asset.update!(state: "migration_failed", error_message: e.inspect)
```
kevindew: Is our decision that we will continue migrating even if assets fail? 

Something we should consider as a follow up story for this or at least for thought is what the process is if migrating assets fails. If we just carry on how does anyone know of the problem? if we don't should we abort the whole sync step.

My instinct is probably that we should run migrate assets last and let it mark the sync as failed for the whole import.

spec/lib/whitehall_importer/migrate_assets_spec.rb
ChrisBAshton: I see it as more likely that a GdsApi::BaseError could be thrown rather than a StandardError, but wanted to catch all errors (just in case) so we can fail gracefully.
My assumption was that if GdsApi::BaseError is not handled, our migration would crash rather than delegate the handling to the StandardError handler, so I've explicitly handled both. In hindsight I'm wondering if this is the case, and whether only capturing `StandardError` would actually be enough to capture custom errors too? If so, I agree we should remove the GdsApi::BaseError handler.

lib/whitehall_importer/import.rb
ChrisBAshton: There are 4 tests in import_spec.rb that fail if I remove this line. I agree it (and the tests) can probably be removed, but feels out of scope for this PR.

spec/lib/whitehall_importer/migrate_assets_spec.rb
ChrisBAshton: I would argue that it's worth setting the `processing` state so that, were we to attempt to run the migration task twice in quick succession, this wouldn't be part of the `migratable_assets` on the second run and we'd avoid attempting to migrate the same asset twice.

If it's worth setting the `processing` state, it's worth testing that it's set. Happy to remove though, as the thing it's protecting against is quite edge-casey.

spec/lib/whitehall_importer/migrate_assets_spec.rb
kevindew: GdsApi::BaseError inherits from StandardError so if we catch StandardError it's the same as catching GdsApi::BaseError.

catching StandardError is basically saying catch all errors, which is something we probably want to do if we want to log that a problem occurred rather than just crash.

spec/lib/whitehall_importer/migrate_assets_spec.rb
kevindew: I don't think the consequences are too severe (are they actually negative? I guess it'd just do the same operation) were it to happen in quick succession so I think it's better to remove it and avoid needing quite deep tests for it.

lib/whitehall_importer/import.rb
kevindew: Sure, we can put together a list of things to do post merge?

spec/lib/whitehall_importer/migrate_assets_spec.rb
ChrisBAshton: Thanks for confirming. I'll simplify this commit 👍

spec/lib/whitehall_importer/migrate_assets_spec.rb
ChrisBAshton: In that case do we want to remove the "processing" state altogether? It sounds like we have no real need for it - I wouldn't just want to remove the test but leave the application code in. Shall I go back through the commits and remove it from the enum, etc? (Happy to do so)

lib/whitehall_importer/migrate_assets.rb
ChrisBAshton: Looking at WhitehallImporter and its calls to ResyncService and ClearLinksetLinks, it does appear that we handle all errors at the highest level, rather than in each called service - so to follow that convention, you're right that we should abort inside MigrateAssets and mark the entire sync as failed.

Would we still want to log the particular asset that it failed on? I'm thinking something like this inside MigrateAssets:

```ruby
rescue StandardError => e
  whitehall_asset.update!(state: "migration_failed", error_message: e.inspect)
  raise StandardError("Failed migrating Whitehall asset #{whitehall_asset.id}: #{e.inspect}")
end
```

spec/lib/whitehall_importer/migrate_assets_spec.rb
kevindew: That sounds a bit of a pain since there are 19 commits so I don't mind you adding a commit at the end which removes it - whatever you prefer.

lib/whitehall_importer/migrate_assets.rb
kevindew: Yes we'd want to do it on a particular asset.

I actually think that in this case we might want to try do every asset and abort afterwards on the whole thing (on the logic that aborting at 1 asset failing and a remaining 10 wrong is worse than 1 out of 10 failing). It might be worth us grabbing a few mins tomorrow to chat about this as it quite nuanced and there are consistency concerns, as you rightly pointed out.

ChrisBAshton: I've had a go at implementing in the final commit 👍

kevindew:
@ChrisBAshton I had a play around to see if I could remove the need for factories and managed it but got a little carried away with changes. Hopefully that is helpful for you rather than annoying. It's up in a single commit here: https://github.com/alphagov/content-publisher/commit/4fb020099e49b6bf1308a9d494685bb2a53c0283

ChrisBAshton:
Thanks @kevindew , I've incorporated your suggestions.

PR #1569, opened 2019-12-23
Bind contextual guidance to inputs

alex-ju:
Using `aria-described-by` we now bind the guidance content to each input so it's being appended to the input description and provide additional information to screen readers' users.

Example for the document title using VoiceOver
Screenshot 2019-12-23 at 18 03 20

[Trello card](https://trello.com/c/LQIMhHs0)

### DIFF
--- a/app/views/withdraw/new.html.erb
+++ b/app/views/withdraw/new.html.erb
@@ -19,10 +19,11 @@
   <% end %>
 
   <%= render "govuk_publishing_components/components/contextual_guidance", {
-    id: "withdrawal-public-explanation-guidance",
+    id: "withdrawal-public-explanation",
     html_for: "withdrawal_public_explanation",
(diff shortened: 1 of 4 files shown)


app/views/withdraw/new.html.erb
```diff
@@ -19,10 +19,11 @@
   <% end %>
 
   <%= render "govuk_publishing_components/components/contextual_guidance", {
-    id: "withdrawal-public-explanation-guidance",
+    id: "withdrawal-public-explanation",
```
leenagupte: Why has the id changed?

alex-ju: I thought would be good to align this with all the other ones so they consistently have the `guidance_id` with a `-guidance` suffix and the component to use a more generic name.

PR #1572, opened 2019-12-24
Update organisation import rake task to get list of documents from Whitehall

brucebolt:
In #1565, we created a rake task (`import:whitehall_migration`) that takes organisation and document type as arguments, then creates a `WhitehallMigration` model.  The changes here build on that work to:
* get a list of documents to import from Whitehall's export API (and deal with the paginated response from Whitehall)
* allow `WhitehallMigration::DocumentImport` to have a status of 'pending' (i.e. ready to import, but no import has been started yet)
* remove not-null requirement for two fields on `WhitehallMigration::DocumentImport` (since we will not have that data whilst the import is in a pending state)
* create a pending `WhitehallMigration::DocumentImport` for each document due to be imported
* create the `WhitehallMigration` and `WhitehallMigration::DocumentImport` models within a single transaction (per run of the `import:whitehall_migration` rake task) so we won't get partial data stored if the import fails for any reason

Later work will then take these pending `WhitehallMigration::DocumentImport` models and queue a job for each document.

Trello card: https://trello.com/c/4OFU8HHs

### DIFF
--- a/lib/tasks/import.rake
+++ b/lib/tasks/import.rake
@@ -3,9 +3,13 @@
 require "gds_api/whitehall_export"
 
 namespace :import do
-  desc "Import all documents matching an organisation and document type from Whitehall Publisher, e.g. import:whitehall_migration[organisation-slug, document-type]"
+  desc "Import all documents matching an organisation and document type from Whitehall Publisher, e.g. import:whitehall_migration[\"cabinet-office\",\"NewsArticle\"]"
   task :whitehall_migration, %i[organisation_slug document_type] => :environment do |_, args|
-    whitehall_migration = WhitehallImporter.create_migration(args.organisation_slug, args.document_type)
+    organisation_content_id = GdsApi.publishing_api.lookup_content_id(
+      base_path: "/government/organisations/#{args.organisation_slug}",
+    )
+
+    whitehall_migration = WhitehallImporter.create_migration(organisation_content_id, args.document_type)
 
     documents_to_import = WhitehallMigration::DocumentImport.where(whitehall_migration_id: whitehall_migration.id).count
     puts "Identified #{documents_to_import} documents to import"
--- a/lib/tasks/import.rake
+++ b/lib/tasks/import.rake
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-require "gds_api/json_client"
+require "gds_api/whitehall_export"
 
 namespace :import do
   desc "Import all documents matching an organisation and document type from Whitehall Publisher, e.g. import:whitehall_migration[organisation-slug, document-type]"
@@ -12,16 +12,8 @@ namespace :import do
 
   desc "Import a single document from Whitehall Publisher using Whitehall's internal document ID e.g. import:whitehall_document[123]"
   task :whitehall_document, [:document_id] => :environment do |_, args|
-    host = Plek.new.external_url_for("whitehall-admin")
-    endpoint = "#{host}/government/admin/export/document/#{args.document_id}"
-    options = {
-      "bearer_token": ENV["WHITEHALL_BEARER_TOKEN"] || "example",
-    }
-    client = GdsApi::JsonClient.new(options)
-    response = client.get_json(endpoint)
-    whitehall_export = response.to_hash
-
-    whitehall_import = WhitehallImporter.import_and_sync(whitehall_export)
+    whitehall_export = GdsApi.whitehall_export.document_export(args.document_id)
+    whitehall_import = WhitehallImporter.import_and_sync(whitehall_export.to_hash)
 
     unless whitehall_import.completed?
       puts whitehall_import.state.humanize
--- a/lib/tasks/import.rake
+++ b/lib/tasks/import.rake
@@ -5,9 +5,10 @@ require "gds_api/whitehall_export"
 namespace :import do
   desc "Import all documents matching an organisation and document type from Whitehall Publisher, e.g. import:whitehall_migration[organisation-slug, document-type]"
   task :whitehall_migration, %i[organisation_slug document_type] => :environment do |_, args|
-    WhitehallMigration.create!(organisation_slug: args.organisation_slug,
-                               document_type: args.document_type,
-                               start_time: Time.zone.now)
+    whitehall_migration = WhitehallImporter.create_migration(args.organisation_slug, args.document_type)
+
+    documents_to_import = WhitehallMigration::DocumentImport.where(whitehall_migration_id: whitehall_migration.id).count
+    puts "Identified #{documents_to_import} documents to import"
   end
 
   desc "Import a single document from Whitehall Publisher using Whitehall's internal document ID e.g. import:whitehall_document[123]"
--- a/lib/whitehall_importer.rb
+++ b/lib/whitehall_importer.rb
@@ -1,6 +1,28 @@
 # frozen_string_literal: true
 
 module WhitehallImporter
+  def self.create_migration(organisation_slug, document_type)
+    ActiveRecord::Base.transaction do
+      whitehall_migration = WhitehallMigration.create!(organisation_slug: organisation_slug,
+                                 document_type: document_type,
+                                 start_time: Time.current)
+
+      whitehall_export = GdsApi.whitehall_export.document_list(organisation_slug, document_type)
+
+      whitehall_export.each do |page|
+        page["documents"].each do |document|
+          WhitehallMigration::DocumentImport.create!(
+            whitehall_document_id: document["document_id"],
+            whitehall_migration_id: whitehall_migration.id,
+            state: "pending",
+          )
+        end
+      end
+
+      whitehall_migration
+    end
+  end
+
   def self.import_and_sync(whitehall_document)
     whitehall_import = WhitehallMigration::DocumentImport.create!(
       whitehall_document_id: whitehall_document["id"],
--- a/lib/whitehall_importer.rb
+++ b/lib/whitehall_importer.rb
@@ -1,13 +1,13 @@
 # frozen_string_literal: true
 
 module WhitehallImporter
-  def self.create_migration(organisation_slug, document_type)
+  def self.create_migration(organisation_content_id, document_type)
     ActiveRecord::Base.transaction do
-      whitehall_migration = WhitehallMigration.create!(organisation_slug: organisation_slug,
+      whitehall_migration = WhitehallMigration.create!(organisation_content_id: organisation_content_id,
                                  document_type: document_type,
                                  start_time: Time.current)
 
-      whitehall_export = GdsApi.whitehall_export.document_list(organisation_slug, document_type)
+      whitehall_export = GdsApi.whitehall_export.document_list(organisation_content_id, document_type)
 
       whitehall_export.each do |page|
         page["documents"].each do |document|
--- a/spec/factories/whitehall_export/index_document_factory.rb
+++ b/spec/factories/whitehall_export/index_document_factory.rb
@@ -1,0 +1,19 @@
+# frozen_string_literal: true
+
+FactoryBot.define do
+  factory :whitehall_export_index_document, class: Hash do
+    skip_create
+
+    sequence(:document_id)
+
+    document_information {
+      {
+        locales: %w(en),
+        sub_types: %w(news_story),
+        lead_organisations: %w(96ae61d6-c2a1-48cb-8e67-da9d105ae381),
+      }.stringify_keys
+    }
+
+    initialize_with { attributes.stringify_keys }
+  end
+end
--- a/spec/factories/whitehall_export/index_factory.rb
+++ b/spec/factories/whitehall_export/index_factory.rb
@@ -1,0 +1,14 @@
+# frozen_string_literal: true
+
+FactoryBot.define do
+  factory :whitehall_export_index, class: Hash do
+    skip_create
+
+    documents { [build(:whitehall_export_index_document)] }
+
+    page_count { documents.size }
+    sequence(:page_number, 1)
+
+    initialize_with { attributes.stringify_keys }
+  end
+end
--- a/spec/lib/tasks/import_spec.rb
+++ b/spec/lib/tasks/import_spec.rb
@@ -7,12 +7,15 @@ RSpec.describe "Import tasks" do
     before do
       allow($stdout).to receive(:puts)
       Rake::Task["import:whitehall_migration"].reenable
+      stub_publishing_api_has_lookups(
+        "/government/organisations/cabinet-office" => "96ae61d6-c2a1-48cb-8e67-da9d105ae381",
+      )
       allow(WhitehallImporter).to receive(:create_migration).and_return(whitehall_migration_document_import)
     end
 
     it "calls WhitehallImport::create_migration with correct arguments" do
       Rake::Task["import:whitehall_migration"].invoke("cabinet-office", "NewsArticle")
-      expect(WhitehallImporter).to have_received(:create_migration).with("cabinet-office", "NewsArticle")
+      expect(WhitehallImporter).to have_received(:create_migration).with("96ae61d6-c2a1-48cb-8e67-da9d105ae381", "NewsArticle")
     end
   end
 
--- a/spec/lib/tasks/import_spec.rb
+++ b/spec/lib/tasks/import_spec.rb
@@ -2,17 +2,17 @@
 
 RSpec.describe "Import tasks" do
   describe "import:whitehall_migration" do
+    let(:whitehall_migration_document_import) { build(:whitehall_migration_document_import) }
+
     before do
+      allow($stdout).to receive(:puts)
       Rake::Task["import:whitehall_migration"].reenable
+      allow(WhitehallImporter).to receive(:create_migration).and_return(whitehall_migration_document_import)
     end
 
-    it "creates a WhitehallMigration" do
-      freeze_time do
-        expect { Rake::Task["import:whitehall_migration"].invoke("cabinet-office", "news_story") }.to change { WhitehallMigration.count }.by(1)
-        expect(WhitehallMigration.last.organisation_slug).to eq("cabinet-office")
-        expect(WhitehallMigration.last.document_type).to eq("news_story")
-        expect(WhitehallMigration.last.start_time).to eq(Time.current)
-      end
+    it "calls WhitehallImport::create_migration with correct arguments" do
+      Rake::Task["import:whitehall_migration"].invoke("cabinet-office", "NewsArticle")
+      expect(WhitehallImporter).to have_received(:create_migration).with("cabinet-office", "NewsArticle")
     end
   end
 
--- a/spec/lib/whitehall_importer_spec.rb
+++ b/spec/lib/whitehall_importer_spec.rb
@@ -1,6 +1,32 @@
 # frozen_string_literal: true
 
 RSpec.describe WhitehallImporter do
+  describe ".create_migration" do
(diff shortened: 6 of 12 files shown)


lib/tasks/import.rake
```diff
@@ -3,10 +3,10 @@
 require "gds_api/json_client"
 
 namespace :import do
-  desc "Import all documents matching an organisation and document type from Whitehall Publisher, e.g. import:whitehall_migration[organisation-slug, document-type]"
-  task :whitehall_migration, %i[organisation_slug document_type] => :environment do |_, args|
+  desc "Import all documents matching an organisation and document type from Whitehall Publisher, e.g. import:whitehall_migration[organisation-content-id, document-type]"
+  task :whitehall_migration, %i[organisation_content_id document_type] => :environment do |_, args|
```
leenagupte: I wonder if rather than asking users to call the rake task with a content-id, which isn't very friendly, we could change slug to base_path, and do a lookup to get the content-id? 
See: https://github.com/alphagov/gds-api-adapters/blob/master/lib/gds_api/publishing_api.rb#L77

lib/tasks/import.rake
```diff
@@ -5,9 +5,27 @@ require "gds_api/json_client"
...
-                               document_type: args.document_type,
-                               start_time: Time.zone.now)
+    ActiveRecord::Base.transaction do
+      WhitehallMigration.create!(organisation_slug: args.organisation_slug,
+                                 document_type: args.document_type,
+                                 start_time: Time.zone.now)
+
+      items_per_page = 100
+
+      1.step do |page_number|
```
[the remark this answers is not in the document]
brucebolt: `step` provides an index (which I'm using as the page number) which we wouldn't get with `loop`.  I guess `until` would also do a similar job, but I'd need to increment the page number manually in each iteration.

I think although it's not written, I'm effectively testing it doesn't loop any more than twice since those are the only HTTP requests that are stubbed (a third request would have `page_number=3`.  But I'll see if there's some way I could explicitly test this.

lib/tasks/import.rake
```diff
@@ -5,9 +5,27 @@ require "gds_api/json_client"
...
+        params = {
+          lead_organisation: args.organisation_slug,
+          type: args.document_type,
+          page_number: page_number,
+          page_count: items_per_page,
+        }
+        endpoint = "/government/admin/export/document?#{params.to_query}"
+        whitehall_export = WhitehallImporter.whitehall_get_json(endpoint)
+        WhitehallImporter.load_index_page(whitehall_export)
+        break if whitehall_export["page_count"] < items_per_page
```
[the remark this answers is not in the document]
brucebolt: I changed this to use `until` then the linter complained:

```
Lint/Loop: Use Kernel#loop with break rather than begin/end/until(or while). (https://rubystyle.guide#loop-with-break)
```

spec/lib/tasks/import_spec.rb
```diff
@@ -2,18 +2,32 @@
...
+        expect { Rake::Task["import:whitehall_migration"].invoke("cabinet-office", "NewsArticle") }.to change { WhitehallMigration.count }.by(1)
         expect(WhitehallMigration.last.organisation_slug).to eq("cabinet-office")
-        expect(WhitehallMigration.last.document_type).to eq("news_story")
+        expect(WhitehallMigration.last.document_type).to eq("NewsArticle")
         expect(WhitehallMigration.last.start_time).to eq(Time.current)
       end
     end
+
+    it "creates a pending WhitehallMigration::DocumentImport for each listed item" do
+      expect { Rake::Task["import:whitehall_migration"].invoke("cabinet-office", "NewsArticle") }.to change { WhitehallMigration::DocumentImport.count }.by(113)
```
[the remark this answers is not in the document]
brucebolt: I've reduced this down to a maximum of five, and allowed the number of documents per page to be specified when the rake task is invoked.

lib/whitehall_importer.rb
```diff
@@ -1,6 +1,15 @@
 # frozen_string_literal: true
 
 module WhitehallImporter
+  def self.import_multiple_documents(whitehall_export)
+    whitehall_export["documents"].each do |document|
+      WhitehallMigration::DocumentImport.create!(
+        whitehall_document_id: document["document_id"],
+        state: "pending",
+      )
```
kevindew: This feels a bit strange to me as it's coupled to the API response that we do in the rake task so feels somewhat like a helper method for that task than a method in it's own right.

brucebolt:
I think everything has been addressed (it's quite a substantial rewrite, sorry!)

The only thing I'm not too sure about is dealing with the pagination.  The current solution is functional, but isn't particularly elegant (especially trying to overcome this styleguide principle: https://rubystyle.guide/#loop-with-break).  Any other way of dealing with this felt like I was massively overcomplicating the problem.  Maybe something for me to look at in the new year.

lib/tasks/import.rake
```diff
@@ -3,23 +3,42 @@
...
+        with_drafts: true,
+        exclude_document_types: [],
+        exclude_unpublishing_types: [],
+      )
+
+      WhitehallMigration.create!(organisation_content_id: organisation_content_id,
+                                 document_type: args.document_type,
+                                 start_time: Time.zone.now)
+
+      items_per_page = args.items_per_page || 100
```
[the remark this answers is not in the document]
brucebolt: I'd made this value modifiable so we don't need to include 100 documents in the stubbed response to test the pagination

lib/whitehall_adapter.rb
```diff
@@ -0,0 +1,28 @@
+# frozen_string_literal: true
+
+class WhitehallAdapter
```
kevindew: It'd be good if this was more consistent with a GDS API Adapter. Perhaps subclassing `GdsApi::Base` and putting it in `lib/gds_api` would help bring consistency. 

It's also worth considering is this better in GDS API Adapters overall?

lib/whitehall_adapter.rb
```diff
@@ -0,0 +1,28 @@
...
+  def self.get_document_list(args = {})
```
kevindew: I'd imagine it'd be more consistent with GDS API Adapters and Ruby idioms to drop the `get_` from the endpoint methods.

lib/whitehall_importer.rb
```diff
@@ -1,6 +1,35 @@
...
 
 module WhitehallImporter
+  def self.create_migration(organisation_content_id, document_type, items_per_page = 100)
+    whitehall_migration = WhitehallMigration.create!(organisation_content_id: organisation_content_id,
+                               document_type: document_type,
+                               start_time: Time.current)
+
+    page_number = 1
+    loop do
+      puts "Loading page #{page_number}"
```
kevindew: We should reserve any puts to be done in the rake task.

spec/lib/whitehall_importer_spec.rb
```diff
@@ -1,6 +1,51 @@
 # frozen_string_literal: true
 
 RSpec.describe WhitehallImporter do
+  describe ".create_migration" do
+    let(:whitehall_host) { Plek.new.external_url_for("whitehall-admin") }
+    let(:whitehall_export_page_1) {
```
kevindew: Convention is to use do and end for a multiline block.

spec/factories/whitehall_export/index_factory.rb
```diff
@@ -0,0 +1,18 @@
...
+  factory :whitehall_export_index, class: Hash do
+    skip_create
+
+    documents do
+      [
+        build(:whitehall_export_index_document),
+      ]
+    end
+
+    page_count { 1 }
```
kevindew: Is the the number of documents? you could use documents.size if so.

spec/factories/whitehall_export/index_document_factory.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+FactoryBot.define do
+  factory :whitehall_export_index_document, class: Hash do
+    skip_create
+
+    document_id { "123" }
```
kevindew: You could use a sequence for this: https://github.com/thoughtbot/factory_bot/blob/master/GETTING_STARTED.md#sequences that'd make it easier for initialising a few. As you could do build_list(:whitehall_export_index_document, 5) to create 5 rather than having to repeat that 5 times.

Any reason why it's a string?

lib/whitehall_adapter.rb
brucebolt: I've created a `GdsApi::Whitehall` adapter, with the relevant methods within `content-publisher`.  I haven't moved it to gds-api-adapters since it feels that the only use for accessing Whitehall is from this application.

lib/tasks/import.rake
```diff
@@ -1,25 +1,23 @@
...
-                               document_type: args.document_type,
-                               start_time: Time.zone.now)
+  desc "Import all documents matching an organisation and document type from Whitehall Publisher, e.g. import:whitehall_migration[\"cabinet-office\",\"NewsArticle\"]"
+  task :whitehall_migration, %i[organisation_slug document_type items_per_page] => :environment do |_, args|
+    ActiveRecord::Base.transaction do
+      organisation_content_id = GdsApi.publishing_api.lookup_content_id(
+        base_path: "/government/organisations/#{args.organisation_slug}",
+      )
+
+      whitehall_migration = WhitehallImporter.create_migration(organisation_content_id, args.document_type, args.items_per_page)
```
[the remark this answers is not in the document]
brucebolt: I've moved the pagination into the adapter, so a Enumerable is returned.

lib/whitehall_importer.rb
brucebolt: Sure, removed.  This was leftover from moving the code out of the rake task.

lib/whitehall_adapter.rb
brucebolt: Done.  Renamed to `document_list` and `document_export.

spec/factories/whitehall_export/index_factory.rb
brucebolt: Sure, updated to use `documents.size`.

spec/factories/whitehall_export/index_document_factory.rb
brucebolt: I've updated both the `document_id` (in this factory) and `page_number` (in the `whitehall_export_index` factory) to both use sequence.

It's also now an integer, being a string was an oversight (since I'd wrongly assumed this was the content ID, rather than the Whitehall ID).

spec/lib/whitehall_importer_spec.rb
brucebolt: This is no longer an issue in my latest changes.

lib/tasks/import.rake
```diff
@@ -1,26 +1,23 @@
...
   task :whitehall_migration, %i[organisation_slug document_type] => :environment do |_, args|
-    WhitehallMigration.create!(organisation_slug: args.organisation_slug,
-                               document_type: args.document_type,
-                               start_time: Time.zone.now)
+    ActiveRecord::Base.transaction do
+      organisation_content_id = GdsApi.publishing_api.lookup_content_id(
+        base_path: "/government/organisations/#{args.organisation_slug}",
+      )
+
+      whitehall_migration = WhitehallImporter.create_migration(organisation_content_id, args.document_type)
```
kevindew: Shall we move the transaction into this method?

spec/factories/whitehall_export/index_factory.rb
```diff
@@ -0,0 +1,18 @@
...
+
+FactoryBot.define do
+  factory :whitehall_export_index, class: Hash do
+    skip_create
+
+    documents do
+      [
+        build(:whitehall_export_index_document),
+      ]
+    end
```
kevindew: Minor: might want to put this on one line?

spec/lib/gds_api/whitehall_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+
+RSpec.describe GdsApi::Whitehall do
+  describe "document_list" do
+    let(:whitehall_adapter) { GdsApi::Whitehall.new(Plek.find("whitehall-admin")) }
+    let(:whitehall_export_page_1) { build(:whitehall_export_index, documents: build_list(:whitehall_export_index_document, 100)) }
+    let(:whitehall_export_page_2) { build(:whitehall_export_index, documents: build_list(:whitehall_export_index_document, 10)) }
+
+    before do
+      stub_request(:get, "https://whitehall-admin.test.gov.uk/government/admin/export/document?lead_organisation=123&type=NewsArticle&page_count=100&page_number=1")
+        .to_return(status: 200, body: whitehall_export_page_1.to_json)
```
kevindew: We might be able to set these as `let` and then add them as an argument into `have_requested(page_1_request)`

spec/lib/gds_api/whitehall_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+      stub_request(:get, "https://whitehall-admin.test.gov.uk/government/admin/export/document?lead_organisation=123&type=NewsArticle&page_count=100&page_number=2")
+        .to_return(status: 200, body: whitehall_export_page_2.to_json)
+    end
+
+    it "makes a GET request to whitehall for correct number of pages" do
```
kevindew: Minor, should we tweak this to say it allows you to iterate through a list of documents?

lib/gds_api/whitehall.rb
```diff
@@ -0,0 +1,36 @@
...
+        params[:page_number] += 1
+        next_link = response["page_count"] == params[:page_count] ? document_list_url(params) : false
+      end
+    end
+  end
+
+  def document_export(document_id)
+    path = "#{endpoint}/government/admin/export/document/#{document_id}"
+    response = get_json(path)
+    response.to_hash
```
kevindew: I think by default we don't convert GDS API Adapters responses into hashes, so might be cool to just return the response?

lib/gds_api/whitehall.rb
```diff
@@ -0,0 +1,36 @@
...
+      lead_organisation: organisation_content_id,
+      type: document_type,
+      page_number: 1,
+      page_count: 100,
+    }
+    Enumerator.new do |yielder|
+      next_link = document_list_url(params)
+      while next_link
+        yielder.yield begin
+          response = get_json(next_link).to_hash
```
kevindew: Similar comment about hash here

spec/lib/gds_api/whitehall_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+      whitehall_document_list = whitehall_adapter.document_list("123", "NewsArticle")
+      expect(whitehall_document_list.next).to have_requested(:get, "https://whitehall-admin.test.gov.uk/government/admin/export/document?lead_organisation=123&type=NewsArticle&page_count=100&page_number=1")
+      expect(whitehall_document_list.next).to have_requested(:get, "https://whitehall-admin.test.gov.uk/government/admin/export/document?lead_organisation=123&type=NewsArticle&page_count=100&page_number=2")
+      expect { whitehall_document_list.next }.to raise_error(StopIteration)
+    end
+
+    it "returns a Hash version of the JSON response for each page" do
+      whitehall_document_list = whitehall_adapter.document_list("123", "NewsArticle")
+      expect(whitehall_document_list.next).to eq(whitehall_export_page_1)
+      expect(whitehall_document_list.next).to eq(whitehall_export_page_2)
```
kevindew: Not sure we need this test if we don't apply any logic to the response.

spec/lib/whitehall_importer_spec.rb
```diff
@@ -1,6 +1,33 @@
...
+        expect { WhitehallImporter.create_migration("123", "NewsArticle") }.to change { WhitehallMigration.count }.by(1)
+        expect(WhitehallMigration.last.organisation_content_id).to eq("123")
+        expect(WhitehallMigration.last.document_type).to eq("NewsArticle")
+        expect(WhitehallMigration.last.start_time).to eq(Time.current)
+      end
+    end
+
+    it "creates a pending WhitehallMigration::DocumentImport for each listed item" do
+      expect { WhitehallImporter.create_migration("123", "NewsArticle") }.to change { WhitehallMigration::DocumentImport.count }.by(110)
+      expect(WhitehallMigration::DocumentImport.all.pluck(:state).uniq).to eq(%w(pending))
```
kevindew: I don't think you need the `all` here

spec/lib/whitehall_importer_spec.rb
```diff
@@ -1,6 +1,33 @@
...
+      freeze_time do
+        expect { WhitehallImporter.create_migration("123", "NewsArticle") }.to change { WhitehallMigration.count }.by(1)
+        expect(WhitehallMigration.last.organisation_content_id).to eq("123")
+        expect(WhitehallMigration.last.document_type).to eq("NewsArticle")
+        expect(WhitehallMigration.last.start_time).to eq(Time.current)
+      end
+    end
+
+    it "creates a pending WhitehallMigration::DocumentImport for each listed item" do
+      expect { WhitehallImporter.create_migration("123", "NewsArticle") }.to change { WhitehallMigration::DocumentImport.count }.by(110)
```
kevindew: Can probably do this: `WhitehallMigration::DocumentImport.pending.count`

spec/factories/whitehall_export/index_factory.rb
brucebolt: Sure, updated to be on one line

spec/lib/gds_api/whitehall_spec.rb
brucebolt: Updated the name to `iterates through the correct number of pages`

lib/gds_api/whitehall.rb
brucebolt: I've removed the `to_hash`, which means the related test can also be removed

spec/lib/gds_api/whitehall_spec.rb
brucebolt: Removed to `to_hash`, so this test is no longer required

spec/lib/whitehall_importer_spec.rb
brucebolt: Updated  to use `pending`, which removes the following line

spec/lib/whitehall_importer_spec.rb
brucebolt: Combined both expects into a single line

spec/lib/gds_api/whitehall_spec.rb
brucebolt: I wasn't able to do this, but did manage to reduce the length of the URL slightly to make this easier to read

lib/tasks/import.rake

lib/whitehall_importer.rb
```diff
@@ -1,6 +1,28 @@
...
 module WhitehallImporter
+  def self.create_migration(organisation_content_id, document_type)
+    ActiveRecord::Base.transaction do
+      whitehall_migration = WhitehallMigration.create!(organisation_content_id: organisation_content_id,
+                                 document_type: document_type,
+                                 start_time: Time.current)
+
+      whitehall_export = GdsApi.whitehall_export.document_list(organisation_content_id, document_type)
+
+      whitehall_export.each_entry do |page|
```
kevindew: From googling about what #each_entry is I don't think we need to use it instead of each: https://stackoverflow.com/questions/19840059/is-there-a-difference-between-enumerableeach-and-enumerableeach-entry-in-ruby

brucebolt: Good spot, I've updated this to use `each`

PR #1575, opened 2019-12-27
Create timeline entry for when document is imported

AgaDufrat:
### User story
As a publisher
I want to see the history of a document
So that I can see what has changed and why it has changed over time.

We want to indicate when a document was imported into Content Publisher so that
publishers can understand any variations in how the history events are
presented.

[Trello card](https://trello.com/c/5Rfe68rk/1279-create-timeline-entry-for-when-its-imported)

_Tested in integration:_

with_notice_component


AgaDufrat:
If the styling is fairly straightforward I will add a commit here and ask Alex to have a look.

PR #1577, opened 2019-12-27
SPIKE:  Migration integrity check

leenagupte:
Trello: https://trello.com/c/ANPGlgjY

## What do we need to do and why?

SPIKE to determine whether is possible to do any type of integrity checking before resyncing Whitehall documents in publishing-api (the point of no return) and officially transferring ownership of them to Content Publisher.

## Findings
Attributes to match on:
- base_path
- title
- description
- document_type
- schema_name
- body text
- image alt_text
- image caption
- organisations
- primary_publishing_organisation

Of the `links`, we should only try matching on "organisations" and "primary_publishing_organisation". It's not possible to match on "taxons" as they are only added to the payload if they have been updated. Nor can we match on "governments" as they are only set at the point of publishing-api being updated (after any integrity check would run).

## Proposal

- Write this as a class (WhitehallImporter::IntegrityChecker) that returns bool and an array of problems
- Change class to not need to know about publish
- Remove Levenshtein check
- Include unpublishing checks
- Add check to within WhitehallImporter::Import to execute before transaction completes
- On issues we need to raise an AbortImportError - can we join array of issues for that


kevindew:
Since https://github.com/alphagov/content-publisher/pull/1602 was merged it seems reasonable to close this.

PR #1586, opened 2019-12-30
Add missing data lost during import from Whitehall to Content Publisher

brucebolt:
We were missing some data when importing a document from Whitehall to Content Publisher.

This PR adds the following:
- Imported editions are marked with a `backdated_to` timestamp where appropriate (when the edition was backdated in Whitehall, determined by comparing the first published timestamp with the time of the first publish event)
- Import the `first_published_at` date for the document (taken from the first publish event on the first edition of a document; cannot be overridden by backdating)
- Editions have a `last_edited_at` timestamp (most recent edit event from the Whitehall `revision_history`)
- Imported editions are not marked as `revision_synced` (i.e. not marked as synced with Publishing API, which has not been completed at the import stage)

Trello card: https://trello.com/c/1RIiRBUF

### DIFF
--- a/lib/whitehall_importer/import.rb
+++ b/lib/whitehall_importer/import.rb
@@ -61,8 +61,18 @@ module WhitehallImporter
         created_at: whitehall_document["created_at"],
         updated_at: whitehall_document["updated_at"],
         created_by_id: user_ids[event["whodunnit"]],
+        first_published_at: first_publish_date,
         imported_from: "whitehall",
       )
+    end
+
+    def first_publish_date
+      first_publish_event = first_edition_history.first_state_event("published") || {}
+      first_publish_event["created_at"]
+    end
+
+    def first_edition_history
+      EditionHistory.new(whitehall_document["editions"].first["revision_history"])
     end
   end
 end
--- a/lib/whitehall_importer/import.rb
+++ b/lib/whitehall_importer/import.rb
@@ -52,8 +52,7 @@ module WhitehallImporter
         raise AbortImportError, "Document with content_id #{content_id} already exists"
       end
 
-      event = whitehall_document["editions"].first["revision_history"].select { |h| h["event"] == "create" }.first
-      raise AbortImportError, "First edition is missing a create event" unless event
+      event = first_edition_history.create_event!
 
       Document.create!(
         content_id: content_id,
--- a/spec/lib/whitehall_importer/create_edition_spec.rb
+++ b/spec/lib/whitehall_importer/create_edition_spec.rb
@@ -57,6 +57,13 @@ RSpec.describe WhitehallImporter::CreateEdition do
       expect(edition).to be_live
     end
 
+    it "does not mark the edition as synced with Publishing API" do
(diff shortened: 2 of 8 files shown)


spec/lib/whitehall_importer/create_edition_spec.rb
```diff
@@ -173,6 +180,7 @@
 
         expect(edition.created_at).to eq(created_at)
         expect(edition.updated_at).to eq(updated_at)
+        expect(edition.last_edited_at).to eq(whitehall_edition.dig("revision_history", -1, "created_at"))
```
kevindew: This dig is a bit cryptic. I think we're expecting it to be the same as the updated_at value that we set it in a let? might be a bit cleaner to compare to that.

spec/lib/whitehall_importer/create_revision_spec.rb
```diff
@@ -31,6 +31,23 @@
       expect(revision.base_path).to eq("/a-path")
     end
 
+    it "sets backdated_to when edition has been backdated" do
+      whitehall_edition = build(:whitehall_export_edition,
+                                first_published_at: "2015-10-03 12:34:00 +0000")
+
+      revision = described_class.call(document, whitehall_edition)
+
+      expect(revision.backdated_to).to eq("2015-10-03 12:34:00 +0000")
```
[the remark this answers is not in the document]
brucebolt: I've updated this test to use an actual timestamp

brucebolt:
I've removed the remaining `.fetch(foo, nil)` and updated a line of code to use the `EditionHistory` class.

We can review whether a `state_events` method would be more suitable if/when we find a need for a non-first or last event.

lib/whitehall_importer/import.rb
```diff
@@ -61,8 +61,22 @@ def create_document(user_ids)
...
+      first_publish_event["created_at"]
+    end
+
+    def first_edition
+      whitehall_document["editions"].first
+    end
+
+    def history(edition)
+      @history ||= EditionHistory.new(edition["revision_history"])
+    end
```
kevindew: You can't use a memo with an argument as it'll be ignored on subsequent calls.

Since you're only accessing first_edition for history you might just want to have a first_edition_history method and drop the first_edition one?

brucebolt: Good spot!  I've updated to have a `first_edition_history` method instead

lib/whitehall_importer/import.rb
```diff
@@ -52,7 +52,7 @@ def create_document(user_ids)
         raise AbortImportError, "Document with content_id #{content_id} already exists"
       end
 
-      event = whitehall_document["editions"].first["revision_history"].select { |h| h["event"] == "create" }.first
+      event = first_edition_history.create_event
```
kevindew: Argh sorry to be a pain - I didn't notice this before. But if we use the bang equivalent here we can drop the need for the raise below this.

brucebolt: I've pushed a change to reflect this

PR #1588, opened 2019-12-30
Consistently handle API errors in ScheduleController

kevindew:
This resolves an inconsistency we had where API errors are handled in ScheduleController rather than the corresponding interactor. This also adds in tests for the interactors.

### DIFF
--- a/spec/interactors/schedule/create_interactor_spec.rb
+++ b/spec/interactors/schedule/create_interactor_spec.rb
@@ -1,0 +1,81 @@
+# frozen_string_literal: true
+
+RSpec.describe Schedule::CreateInteractor do
+  describe ".call" do
+    before { stub_any_publishing_api_put_intent }
+    let(:edition) { create(:edition, :schedulable) }
+    let(:user) { build(:user) }
+
+    def build_params(document: edition.document, review_status: "reviewed")
+      ActionController::Parameters.new(document: document.to_param,
+                                       review_status: review_status)
+    end
+
+    it "succeeds with default paramaters" do
+      result = Schedule::CreateInteractor.call(params: build_params, user: user)
+      expect(result).to be_success
+    end
+
+    it "delegates to ScheduleService to schedule the edition for publishing" do
+      publish_time = 3.days.from_now.beginning_of_day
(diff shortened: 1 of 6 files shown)


spec/interactors/schedule/create_interactor_spec.rb
```diff
@@ -0,0 +1,81 @@
...
+
+    it "raises an error when the edition hasn't got a proposed publish time" do
+      edition = create(:edition, proposed_publish_time: nil)
+      params = build_params(document: edition.document)
+      expect { Schedule::CreateInteractor.call(params: params, user: user) }
+        .to raise_error(EditionAssertions::StateError)
+    end
+
+    it "raises an error when the edition isn't publishable" do
+      edition = create(:edition, proposed_publish_time: Time.zone.tomorrow)
```
leenagupte: It's not obvious from the test set-up why the edition isn't publishable. Is there a missing attribute I should be aware of?

spec/interactors/schedule/create_interactor_spec.rb
```diff
@@ -0,0 +1,81 @@
...
+      expect { Schedule::CreateInteractor.call(params: build_params, user: user) }
+        .to change { TimelineEntry.where(entry_type: :scheduled).count }
+        .by(1)
+    end
+
+    it "returns a success with edition" do
+      result = Schedule::CreateInteractor.call(params: build_params, user: user)
+
+      expect(result).to be_success
+      expect(result.edition).to eq(edition)
```
leenagupte: Should this test also check `publish_time`, as the UpdateInteractor test does?

kevindew: Good catch, it's more that we shouldn't need to test publish_time since it's only needed internally to the class. I've amended this to be that we just test that both succeed by default.

spec/interactors/schedule/create_interactor_spec.rb
kevindew: Sure, I can see how that is the case. I've added a trait non_publishable to make this explicit.

kevindew:
Thanks for the review @leenagupte amends have been made.

PR #1591, opened 2019-12-31
Move Document history from Tab to Secondary navigation and add pagination

AgaDufrat:
Previously Document's history was shown on a single page as part of a tabbed interface. With the migration of Whitehall content into Content Publisher, we anticipate that this will bring documents with longer histories than Content Publisher has dealt with into the mix and may mean we hit the issues of a very long or slow to load page.

This PR:
- **Uses Secondary navigation instead of tabs**

<table>
<tr><td>

content-publisher dev gov uk_documents_88929f8d-2f2c-45a0-8938-d303c390fcaf_en(iPad Pro) (1)
</td><td>

content-publisher dev gov uk_documents_88929f8d-2f2c-45a0-8938-d303c390fcaf_en(iPad Pro)
</td></tr>
</table>

- **Adds pagination**
pagination

Trello:
https://trello.com/c/TsRTLbiW/1276-move-document-history-from-tab-to-secondary-navigation-and-add-pagination


config/routes.rb
```diff
@@ -14,6 +14,7 @@
     get "" => "documents#show", as: :document
     patch "" => "documents#update"
     delete "" => "documents#destroy"
+    get "/document-history" => "documents#history", as: :history_path
```
kevindew: I think `/history` would be a more appropriate path here since the full url will be `/documents/:document/history`

app/views/documents/show.html.erb
```diff
@@ -2,22 +2,25 @@
...
-      tab_data_attributes: { gtm: "history-tab" }
+      href: history_path_url,
+      data_attributes: {
+        gtm: "history-tab"
+      }
     }
   ]
 } %>
+
+<%= render("documents/show/summary_tab") %>
```
kevindew: I'd expect we'd more the summary_tab view into this one.

app/views/documents/show.html.erb
```diff
@@ -2,22 +2,25 @@
...
-      title: t("documents.show.tabs.history"),
-      content: render("documents/show/history_tab"),
-      tab_data_attributes: { gtm: "history-tab" }
+      href: history_path_url,
+      data_attributes: {
+        gtm: "history-tab"
+      }
     }
   ]
 } %>
```
kevindew: I expect this component will end up in a partial where we specify a parameter of which one is current.

app/controllers/documents_controller.rb
```diff
@@ -23,6 +23,10 @@ def show
     @edition = Edition.find_current(document: params[:document])
   end
 
+  def history
+    @edition = Edition.find_current(document: params[:document])
```
kevindew: I suspect here that we're also going to want to pull out timeline entries rather have that done in the view (it'll be essential for pagination as well as just a nice separation of concerns improvement)

https://github.com/alphagov/content-publisher/blob/bb263e3b8438f074f17a3f49e486405fdf16f108/app/views/documents/show/_history_tab.html.erb#L5

app/views/documents/history.html.erb
```diff
@@ -0,0 +1,26 @@
...
+      href: history_path_url,
+      current: true,
+      data_attributes: {
+        gtm: "history-tab"
+      }
+    }
+  ]
+} %>
+
+<%= render("documents/show/history_tab") %>
```
kevindew: Same comment as show where I expect this code will be part of this view

app/controllers/documents_controller.rb
```diff
@@ -23,6 +23,16 @@ def show
...
+    @timeline_entries = @edition.document.timeline_entries
+                                  .includes(:created_by, :details)
+                                  .order(created_at: :desc)
+                                  .includes(:edition)
+                                  .page(params.fetch(:page, 1))
+                                  .per(25)
```
kevindew: Indentation seems a little off here. I also wonder if we want to query TimelineEntry directly for example via `TimelineEntry.where(document: @edition.document)` and then do the chain?

config/routes.rb
```diff
@@ -14,6 +14,7 @@
     get "" => "documents#show", as: :document
     patch "" => "documents#update"
     delete "" => "documents#destroy"
+    get "/history" => "documents#history", as: :history_path
```
kevindew: You don't want the path suffix from this - just `as: :history` or `as: :document_history` if just history is too generic.

app/views/documents/_secondary_navigation.html.erb
```diff
@@ -0,0 +1,21 @@
...
+      href: document_path,
+      current: current_page?(document_path),
+      data_attributes: {
+        gtm: "summary-tab"
+      }
+    },
+    {
+      label: t("documents.secondary_navigation.history"),
+      href: history_path_url,
+      current: current_page?(history_path_url),
```
kevindew: The use of history_path_url seems a mistake here. As discussed this seems to be because we named the route history_path rather than just history.

app/views/documents/_secondary_navigation.html.erb
```diff
@@ -0,0 +1,21 @@
+<%= render "components/secondary_navigation", {
+  aria_label: "Document",
+  items: [
+    {
+      label: t("documents.secondary_navigation.summary"),
```
kevindew: I'm not sure these should be in the translations file based on: https://github.com/alphagov/content-publisher/blob/29f45e570e7c9c0ab965c629735aaacc07969f68/docs/adr/0004-editing-microcopy.md - have a read and see what your stance is. I don't mind too much which way we go just that we're consistent with the doc and app.

app/views/documents/_secondary_navigation.html.erb
```diff
@@ -0,0 +1,21 @@
...
+      href: document_path,
+      current: current_page?(document_path),
+      data_attributes: {
+        gtm: "summary-tab"
```
kevindew: We should see about changing these gtm attributes to reflect that these are no longer tabs. Maybe summary-nav and history-nav ?

app/views/documents/history.html.erb
```diff
@@ -67,6 +73,16 @@
...
       </div>
     <% end %>
+
+    <% pages = { } %>
+
+    <% pages[:previous_page] = { url: history_path_url(page: @timeline_entries.prev_page),
+                                 title: "Previous page" } if @timeline_entries.prev_page %>
+
+    <% pages[:next_page] = { url: history_path_url(page: @timeline_entries.next_page),
+                             title: "Next page" } if @timeline_entries.next_page %>
```
kevindew: We can probably put this into a single block of embedded ruby:

<%
pages = {}
pages[:previous_page] = { url: history_path_url(page: @timeline_entries.prev_page),
                                 title: "Previous page" } if @timeline_entries.prev_page
pages[:next_page] = { url: history_path_url(page: @timeline_entries.next_page),
                             title: "Next page" } if @timeline_entries.next_page
%>

app/views/documents/history.html.erb
```diff
@@ -1,8 +1,14 @@
+<% content_for :back_link, render_back_link(href: documents_path) %>
```
kevindew: This should go back to the summary page

app/views/documents/show.html.erb
```diff
@@ -2,22 +2,62 @@
...
-    }
-  ]
-} %>
+<%= render("documents/secondary_navigation") %>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <% if flash[:submitted_for_review] %>
+      <%= render "documents/show/submitted_for_review" %>
+    <% end %>
```
kevindew: This will need merging with Emma's PR of: https://github.com/alphagov/content-publisher/pull/1609

1pretz1: Rebasing only broke 100 tests! 😆

1pretz1:
It's ready for another review @kevindew! 😄 

I've left some of the spec improvements in but have taken out the maybe more subjective stuff, I'm sorry as I know you said that doing this would be harder to review but I think I went too far down doing it to take everything out again. I'll know for next time! I've tried to remedy this by making consistent small changes with detailed commit messages.

spec/features/editing_content_settings/history_mode_spec.rb
```diff
@@ -67,12 +65,12 @@ def then_i_enable_political_status
...
   def and_i_see_the_timeline_entry
-    within("#document-history") do
-      expect(page).to have_content(I18n.t!("documents.history.entry_types.political_status_changed"))
-    end
+    click_on "Document history"
+    expect(page).to have_content(I18n.t!("documents.history.entry_types.political_status_changed"))
   end
 
   def when_i_click_to_backdate_the_content
+    visit document_path(@edition.document)
```
kevindew: Can we click the "Document summary" link rather than visit a url directly? - the latter is discouraged in feature tests

spec/features/editing_images/choose_lead_image_spec.rb
```diff
@@ -24,14 +26,15 @@ def given_there_is_an_edition_with_images
     @edition = create(:edition,
                       document_type_id: document_type.id,
                       image_revisions: [@image_revision])
+
+    @publishing_api_request = stub_publishing_api_put_content(@edition.content_id, {})
```
kevindew: I wonder if we want this to be just a stub_any_publishing_api_put_content since we assert the specific request in the "preview_creation_succeeded" step so looks like we're double checking this.

spec/features/editing_images/remove_lead_image_spec.rb
```diff
@@ -26,20 +28,20 @@ def given_there_is_an_edition_with_a_lead_image
     @edition = create(:edition,
                       document_type_id: document_type.id,
                       lead_image_revision: @image_revision)
+
+    @publishing_api_request = stub_publishing_api_put_content(@edition.content_id, {})
```
kevindew: Similar point here - by the way I think a convention we've got is setting up the stub near the point of the request which we may be starting to break here. Without doing a bit of an audit I'm not sure how prevalent it is.

spec/features/editing_tags/edit_tags_spec.rb
```diff
@@ -62,14 +63,17 @@ def then_i_can_see_the_tags
...
       expect(page).to have_content I18n.t!("documents.history.entry_types.updated_tags")
     end
   end
 
   def and_the_preview_creation_succeeded
-    expect(@request).to have_been_requested
+    expect(@put_content_request).to have_been_requested
     expect(page).to have_content(I18n.t!("user_facing_states.draft.name"))
 
     expect(a_request(:put, /content/).with { |req|
```
kevindew: Hmm this seems to be another double one - am I missing something and that these are testing two different things? I'm now a bit confused since they're quite prevalent.

spec/features/scheduling/scheduled_publishing_failed_spec.rb
```diff
@@ -17,8 +17,8 @@
 
     when_i_visit_the_summary_page
     then_i_can_see_the_publishing_failed
-    and_there_is_a_history_entry
     and_i_can_edit_the_edition
+    and_there_is_a_history_entry
```
kevindew: Should we update this to timeline_entry which I think is the terminology used on the other tests.

spec/features/scheduling/scheduled_publishing_failed_spec.rb
```diff
@@ -58,13 +58,13 @@ def then_the_editors_are_notified_of_a_publishing_failure
     expected_subject = I18n.t("scheduled_publish_mailer.failure_email.subject",
                               title: @edition.title)
 
-    expect(message.to).to include(current_user.email)
+    expect(message.to).to have_content(current_user.email)
```
kevindew: This seems a bit of a strange place to use have_content since this is a plain string rather than HTML

spec/features/scheduling/scheduled_publishing_spec.rb
```diff
@@ -46,12 +47,15 @@ def then_i_see_the_edition_is_scheduled
     expect(page).to have_content(I18n.t!("schedule.scheduled.title"))
     visit document_path(@edition.document)
 
+    expect(page).to have_content(I18n.t!("user_facing_states.scheduled.name"))
+    expect(page).to have_content("Scheduled to publish at 9:00am on 22 June 2019")
+  end
+
+  def and_there_is_a_scheduled_history_event
+    click_on "Document history"
     within(".app-timeline-entry") do
```
kevindew: I assume we don't need this within now this is on a new page?

spec/features/scheduling/scheduled_publishing_spec.rb
```diff
@@ -77,14 +81,14 @@ def and_the_editors_are_notified
     expected_subject = I18n.t("scheduled_publish_mailer.success_email.subject.published",
                               title: @edition.title)
 
-    expect(message.to).to include(current_user.email)
+    expect(message.to).to have_content(current_user.email)
```
kevindew: similar point about have_content on plain string

spec/features/withdrawing/edit_withdrawal_spec.rb
```diff
@@ -35,14 +36,19 @@ def when_i_edit_the_public_explanation
...
     body = { type: "withdrawal", explanation: converted_explanation, locale: @edition.locale }
-    stub_publishing_api_unpublish(@edition.content_id, body: body)
+    @unpublish_request = stub_publishing_api_unpublish(@edition.content_id, body: body)
 
     fill_in "public_explanation", with: @new_explanation
     click_on "Update explanation"
   end
 
   def then_i_can_see_the_updated_explanation
+    expect(@unpublish_request).to have_been_requested
```
kevindew: This doesn't seem to match the step - don't think it's necessary

spec/features/withdrawing/withdraw_spec.rb
```diff
@@ -44,12 +45,18 @@ def then_i_see_the_document_has_been_withdrawn
     withdrawal = status.details
     document_type = @edition.document_type.label.downcase
 
+    expect(@unpublish_request).to have_been_requested
```
kevindew: Again not sure we need to do this.

spec/requests/unwithdraw_spec.rb
```diff
@@ -12,7 +12,9 @@
       post unwithdraw_path(withdrawn_edition.document)
       follow_redirect!
 
-      expect(response.body).to include(I18n.t!("documents.history.entry_types.unwithdrawn"))
+      within("#app-c-metadata") do
+        expect(response.body).to include(I18n.t!("user_facing_states.published"))
+      end
```
kevindew: within doesn't work outside of a capybara session so these assertion in here will never be called.

I'd probably change this request test to just assert that it redirects to document summary and not bother following redirect or asserting or contents since the action isn't setting those.

spec/requests/unwithdraw_spec.rb
```diff
@@ -43,7 +45,9 @@
         post unwithdraw_path(withdrawn_history_mode_edition.document)
         follow_redirect!
 
-        expect(response.body).to include(I18n.t!("documents.history.entry_types.unwithdrawn"))
+        within("#app-c-metadata") do
+          expect(response.body).to include(I18n.t!("user_facing_states.published"))
+        end
```
kevindew: Similar point here - should be enough to just assert it's a successful request

spec/features/editing_content_settings/backdate_spec.rb
```diff
@@ -48,15 +49,23 @@ def and_i_enter_a_date_to_backdate_the_content_to
...
 
   def and_i_click_save
-    @request = stub_publishing_api_put_content(@edition.content_id, {})
+    @put_content_request = stub_publishing_api_put_content(@edition.content_id, {})
     click_on "Save"
   end
 
   def then_i_see_the_content_has_been_backdated
+    expect(@put_content_request).to have_been_requested
+    within("#content_settings") do
```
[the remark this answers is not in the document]
1pretz1: I've removed the selector check, originally I did this as just checking for "1 January 2019" seemed vague and could potentially be included in the body of the content so thought I'd narrow it down to content-settings!

spec/features/editing_content_settings/backdate_spec.rb
```diff
@@ -48,15 +49,23 @@ def and_i_enter_a_date_to_backdate_the_content_to
...
+      expect(page).to have_content("1 January 2019")
+      expect(page).to_not have_content(I18n.t!("documents.show.content_settings.backdate.no_backdate"))
+    end
+  end
+
+  def and_i_see_the_timeline_entry
+    click_on "Document history"
     expect(page).to have_content("1 January 2019")
```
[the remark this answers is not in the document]
1pretz1: You're correct! 👍

spec/features/editing_content_settings/backdate_spec.rb
```diff
@@ -48,15 +49,23 @@ def and_i_enter_a_date_to_backdate_the_content_to
...
-    @request = stub_publishing_api_put_content(@edition.content_id, {})
+    @put_content_request = stub_publishing_api_put_content(@edition.content_id, {})
     click_on "Save"
   end
 
   def then_i_see_the_content_has_been_backdated
+    expect(@put_content_request).to have_been_requested
+    within("#content_settings") do
+      expect(page).to have_content("1 January 2019")
+      expect(page).to_not have_content(I18n.t!("documents.show.content_settings.backdate.no_backdate"))
```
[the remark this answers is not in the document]
1pretz1: I think this is valuable as we're checking above for the presence of the text so it shows  visual change between steps & just having the date on its own seems kinda vague.

spec/features/editing_content_settings/history_mode_spec.rb
1pretz1: We certainly can!

spec/features/scheduling/scheduled_publishing_failed_spec.rb
1pretz1: Yup good point

spec/features/scheduling/scheduled_publishing_failed_spec.rb
1pretz1: Good point, changed back!

spec/features/scheduling/scheduled_publishing_spec.rb
1pretz1: Yep, it is now removed!

spec/requests/unwithdraw_spec.rb
1pretz1: Ah interesting didn't know that, I've followed your advise!

1pretz1:
Thanks for the review @kevindew, I've removed any changes I made regarding request assertions in the specs and resolved all the comments regarding this. As discussed offline Ill be doing an audit of these and try bring together some more consistency in terms of how we test requests in a separate stream of work!

spec/features/workflow/submit_for_2i_spec.rb
```diff
@@ -45,6 +49,7 @@ def and_i_see_a_link_to_the_edition
 
   def when_i_edit_the_edition
     stub_any_publishing_api_put_content
+    visit document_path(@edition.document)
```
kevindew: Should this be `click_on "Document summary"`

1pretz1: Sure

1pretz1:
Good spot on the fixup @kevindew, rebased again and should be all good to go! 😄

PR #1592, opened 2019-12-31
Replace view feature tests with view tests

kevindew:
https://trello.com/c/wgkfOJsE/1283-improve-consistency-in-content-publisher-tests

This replaces a number of feature tests where we are effectively testing views and moves them to view tests. 

This is in line with the testing strategy defined in: https://github.com/alphagov/content-publisher/pull/1590. Further details in individual commits.


app/views/documents/show/_content_settings.html.erb
```diff
@@ -22,7 +22,7 @@
     field: t("documents.show.content_settings.access_limit.title"),
     value: access_limit,
     edit: {
-      href: access_limit_path,
```
1pretz1: It must look at the current url and make some assumptions on what it thinks it should be -> 
Screenshot 2020-01-02 at 16 30 13

This is an example where rails confuddles me completely! 😆

app/views/documents/show/_content_settings.html.erb
```diff
@@ -42,7 +42,12 @@
 <% items << {
   field: t("documents.show.content_settings.gets_history_mode.title"),
   value: @edition.political? ? t("documents.show.content_settings.gets_history_mode.true_label") : t("documents.show.content_settings.gets_history_mode.false_label"),
```
1pretz1: Should this also be made into an `if` block for consistency as per the code change below?

spec/views/documents/show.html.erb_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+      assign(:edition, edition)
+      render
+
+      title = I18n.t!("documents.show.historical.title",
+                      document_type: edition.document_type.label.downcase)
+      description = I18n.t!("documents.show.historical.description",
+                            government_name: edition.government.title)
+      expect(rendered)
+        .to include(title)
+        .and include(description)
```
1pretz1: `and` is cool, didn't realise it was a thing!

spec/views/documents/show.html.erb_spec.rb
```diff
@@ -4,6 +4,37 @@
   include TopicsHelper
   before { populate_default_government_bulk_data }
 
+  describe "tags" do
+    let(:tag_field) { build(:tag_field, type: "single_tag", id: "tag_id") }
+    let(:document_type) { build(:document_type, tags: [tag_field]) }
+
+    it "shows the topics when a document has tags" do
```
1pretz1: Think this should be  "shows the tags when a document has tags"

spec/views/documents/show.html.erb_spec.rb
```diff
@@ -4,6 +4,37 @@
...
+      stub_publishing_api_has_linkables([tag], document_type: tag_field.document_type)
+      edition = build(:edition,
+                      document_type_id: document_type.id,
+                      tags: { tag_field.id => [tag["content_id"]] })
+      assign(:edition, edition)
+
+      expect(render).to have_selector("#tags", text: "Tag name")
+    end
+
+    it "shows a message when a document doesn't have topics" do
```
1pretz1: Think you need to replace 'topics' with 'tags'!

spec/views/documents/show.html.erb_spec.rb
```diff
@@ -4,6 +4,37 @@
...
+      edition = build(:edition, document_type_id: document_type.id, tags: {})
+      assign(:edition, edition)
+      expect(render).to include(I18n.t!("documents.show.tags.none"))
+    end
+
+    it "shows a message when topics can't be loaded" do
```
1pretz1: Same comment as above! 😄

app/views/documents/show/_content_settings.html.erb
kevindew: Yeah - I find it a bit annoyingly magical, particularly when it breaks in a different context 😞

spec/views/documents/show.html.erb_spec.rb
kevindew: d'oh! what poor copy and pasting on my behalf. Thanks for catching.

app/views/documents/show/_content_settings.html.erb
kevindew: That's a good shout, it's a very long line.

kevindew:
Thanks for the review @1pretz1 - I've amended accordingly.

app/views/documents/show/_content_settings.html.erb
```diff
@@ -33,16 +33,25 @@
...
     }
   } %>
 <% end %>
 
 <% items << {
   field: t("documents.show.content_settings.gets_history_mode.title"),
-  value: @edition.political? ? t("documents.show.content_settings.gets_history_mode.true_label") : t("documents.show.content_settings.gets_history_mode.false_label"),
-  edit: @edition.editable? ? { href: history_mode_path, data_attributes: { gtm: "edit-history-mode" } } : {}
+  value: if @edition.political?
+           t("documents.show.content_settings.gets_history_mode.true_label")
```
benthorner: Minor: how about `#{@edition.political?}_label`? 😮

spec/views/documents/show.html.erb_spec.rb
```diff
@@ -0,0 +1,123 @@
...
+      render
+
+      expect(rendered).to include("Title")
+      expect(rendered).to include("Summary")
+      expect(rendered).to have_selector(
+        ".app-c-metadata",
+        text: /#{I18n.t!("documents.show.metadata.created_by")}:\s*User 2/,
+      )
+      expect(rendered).to have_selector(
+        ".app-c-metadata",
```
benthorner: Minor: do we need to use a selector approach? As long as the text is specific enough, I generally prefer to just check for that, unless it's clearly ambiguous.

spec/views/documents/show.html.erb_spec.rb
benthorner: Apparently it's not cool enough for L18 😆

spec/views/documents/show.html.erb_spec.rb
kevindew: Good point - selector wasn't a preferred approach just a result of me getting more specific to get a match. Of course that was before I switched to a regex so that was just cruft left from a previous approach.

app/views/documents/show/_content_settings.html.erb
kevindew: Oh yes - nice catch 👍

spec/views/documents/show.html.erb_spec.rb
kevindew: They're now there in their cool glory 😎

kevindew:
Thanks @benthorner I've made amends accordingly

PR #1594, opened 2020-01-02
Fix backdate checker spec

ChrisBAshton:
Test was failing because 01/01/2020 is no longer a date in the future.

### DIFF
--- a/spec/lib/requirements/backdate_checker_spec.rb
+++ b/spec/lib/requirements/backdate_checker_spec.rb
@@ -9,7 +9,7 @@ RSpec.describe Requirements::BackdateChecker do
     end
 
     it "returns an issue if the date is in the future" do
-      date = Time.current.change(day: 1, month: 1, year: 2020)
+      date = 1.year.from_now
       issues = Requirements::BackdateChecker.new(date).pre_submit_issues
       expect(issues).to have_issue(:backdate_date, :in_the_future)
     end


spec/lib/requirements/backdate_checker_spec.rb
```diff
@@ -9,7 +9,7 @@
     end
 
     it "returns an issue if the date is in the future" do
-      date = Time.current.change(day: 1, month: 1, year: 2020)
+      date = Time.current.change(day: 1, month: 1, year: 2030)
```
leenagupte: If the date just needs to be in the future, could `1.year.from_now` be used?

spec/lib/requirements/backdate_checker_spec.rb
```diff
@@ -9,7 +9,7 @@
     end
 
     it "returns an issue if the date is in the future" do
-      date = Time.current.change(day: 1, month: 1, year: 2020)
+      date = Time.current.change(day: 1, month: 1, year: 1.year.from_now)
```
kevindew: This might be slightly weird as `1.year.from_now` returns an instance of `ActiveSupport::TimeWithZone` rather than an int. 

You can change this whole line to be date = `date = 1.year.from_now`

ChrisBAshton: Whoops. Yep, agreed.

PR #1595, opened 2020-01-02
Move format configuration

kevindew:

In Content Publisher we have used the formats directory in app to store configuration files. This is somewhat unconventional in a Rails app so this PR moves them to the config directory. 

More details in commits.


benthorner:
At long last!

PR #1598, opened 2020-01-03
Request tests for document workflow

kevindew:

This follows on the test strategy under discussion in https://github.com/alphagov/content-publisher/pull/1590. Towards this I've done a bunch of writing of request tests and deleting of feature tests. Rather than do one large PR I thought it was best to separate into distinct PR's that are easier to review and thought I'd start with one that covers a few.

This PR introduces a shared helper for RSpec which allows asserting whether editions in a particular state are handled accordingly with hopefully the minimum of overhead. It then adds request tests for the New Document, Preview, Publish and Review controllers and removes feature tests for these that test paths that result in erroneous responses (validation, service unavailable), the request tests also cover some outcomes that weren't covered by feature tests.


spec/requests/new_document_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+      supertype = Supertype.all.find(&:managed_elsewhere)
+      get choose_document_type_path, params: { supertype: supertype.id }
+      expect(response).to redirect_to(supertype.managed_elsewhere_url)
+    end
+
+    it "returns an unprocessable response with an issue when a super type isn't selected" do
+      get choose_document_type_path
+      expect(response).to have_http_status(:unprocessable_entity)
+      expect(response.body)
+        .to include(I18n.t!("requirements.supertype.not_selected.form_message"))
```
1pretz1: Is there a reason this expectation and https://github.com/alphagov/content-publisher/pull/1598/files#diff-a85fa69ab372a0c3eb3625fa606ceb90R56 do not use `have_selector` but the other expectations testing requirements messages do, like here - https://github.com/alphagov/content-publisher/pull/1598/files#diff-3463045769d93e0dafbeddae41166202R26?

spec/support/assert_edition_state_examples.rb
```diff
@@ -0,0 +1,33 @@
...
+  # Within the example we expect a let of an edition to be defined, which
+  # sets up an edition in an incorrect state and a let of run_request which is
+  # a lambda to call to run the request
+  #
+  # For example:
+  #   it_behaves_like "requests that assert edition state",
+  #                   "situation we're testing against",
+  #                   routes: { our_path: %i[get post delete] } do
+  #     let(:edition) { create(:edition, :certain_state) }
+  #     let(:route_params) { [edition.document] }
```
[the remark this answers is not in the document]
kevindew: Yup, I put it as an example for overriding but it doesn't seem needed. Will remove.

spec/spec_helper.rb
```diff
@@ -36,6 +36,7 @@
   config.include GovukSchemas::RSpecMatchers
   config.include AuthenticationHelper, type: ->(spec) { spec.in?(%i[feature request]) }
   config.include BulkDataHelper
+  config.include Capybara::RSpecMatchers, type: :request
```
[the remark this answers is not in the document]
kevindew: Ah yes - I had tried to avoid adding this as much as possible so I'd used include before it was added. have_content does seem a more idiomatic option.

spec/requests/new_document_spec.rb
```diff
@@ -0,0 +1,60 @@
+# frozen_string_literal: true
+
+RSpec.describe "New Document" do
+  describe "GET /documents/choose-document-type" do
+    it "shows supertype when a supertype managed by content publisher is selected" do
+      supertype = Supertype.all.reject(&:managed_elsewhere).first
+      get choose_document_type_path, params: { supertype: supertype.id }
+      expect(response).to have_http_status(:ok)
+      expect(response.body).to include(supertype.label)
```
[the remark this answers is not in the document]
kevindew: I've changed the include over.

spec/requests/new_document_spec.rb
kevindew: Yes, mostly we don't want to care about selectors. The ones that use have_selector are used to test that the message is displayed in an error context. We anticipate that the message would normally be displayed but just not in an error context.

The alternative approach to this would be to test that the flash was set for example `expect(flash["tried_to_publish"]).to be(true)` but this has the downside of not asserting behaviour.

kevindew:
Thanks for the review @1pretz1 - I've made the amends.

spec/requests/new_document_spec.rb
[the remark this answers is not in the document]
1pretz1: Thanks Kev, sorry for the inconvenience!

spec/requests/new_document_spec.rb
1pretz1: Ah okay that's good to know! 

However the two expectations I've mentioned above are both testing the flash error message contained within a `.gem-c-error-summary` element (as both are asserting requirements checks) so shouldn't they also be testing the selector like the other example?

kevindew: Ah sure I think I get you.

So in [this new document one](https://github.com/alphagov/content-publisher/pull/1598/files#diff-a85fa69ab372a0c3eb3625fa606ceb90R52-R57) we're testing that given an unprocessable request we show an issue. We're not too concerned about the context as our controller is not concerned about that.

Whereas in [this `.gem-c-error-summary` one](https://github.com/alphagov/content-publisher/pull/1598/files#diff-3463045769d93e0dafbeddae41166202R19-R28) the effect we are testing is not whether the redirected page shows the issue (as it will anyway) but that it shows it in a particular context. Thus the way to test that is to use the selector.

So testing logic is such that if we can avoid needing to use the selector then that is preferable - however that isn't the case for places where the controller is adjusting the context in which the error is presented.

1pretz1: Oh I get you, thanks for the explanation!!

PR #1602, opened 2020-01-06
Check whether data imported from Whitehall is accurate with the Publishing API

leenagupte:
Trello: https://trello.com/c/JHMoTkfA
Based on: https://github.com/alphagov/content-publisher/pull/1577
Depends on: https://github.com/alphagov/gds-api-adapters/pull/975

## What's changed
Creates a service that checks whether the data we are using to replace Whitehall data during the migration is accurate with the Publishing API. This service acts as the final buffer as to whether to complete a migration or abort.

## Why?
To give confidence when doing a bulk republishing as part of migration. 

### Matches
```
document = Document.find_by_param("e882a7b1-8605-45e7-9700-7b0fe8e068d0:en")
=> #<Document id: 671, content_id: "e882a7b1-8605-45e7-9700-7b0fe8e068d0", locale: "en", created_at: "2019-08-02 07:38:33", updated_at: "2019-12-30 14:58:44", created_by_id: 139, first_published_at: nil, imported_from: "whitehall">
irb(main):002:0> WhitehallImporter::IntegrityChecker.new(document.current_edition).valid?
=> true
irb(main):004:0> WhitehallImporter::IntegrityChecker.new(document.current_edition).problems
=> []
```

### Doesn't match
```
irb(main):008:0> document = Document.find_by_param("1dd15d42-e6b5-4fab-8492-350eca33cf11:en")
=> #<Document id: 667, content_id: "1dd15d42-e6b5-4fab-8492-350eca33cf11", locale: "en", created_at: "2019-01-10 14:44:21", updated_at: "2019-01-23 17:38:10", created_by_id: 130, first_published_at: nil, imported_from: "whitehall">
irb(main):009:0> WhitehallImporter::IntegrityChecker.new(document.live_edition).valid?
=> false
irb(main):010:0> WhitehallImporter::IntegrityChecker.new(document.live_edition).problems
=> ["image caption doesn't match"]

proposed = PreviewService::Payload.new(edition, republish: true).payload
proposed["details"]["image"]
=> {"high_resolution_url"=>nil, "url"=>nil, "alt_text"=>"Man being interviewed in front of a fire engine", "caption"=>"", "credit"=>nil}
```


lib/whitehall_importer/integrity_checker.rb
```diff
@@ -0,0 +1,98 @@
...
+      problems.empty?
+    end
+
+    def problems
+      problems = []
+      problems += content_problems
+      problems += image_problems
+      problems += organisation_problems
+
+      problems
```
erino: What about 

```ruby
[].concat(content_problems, image_problems, organisation_problems)
```

leenagupte: I didn't realise `concat` is a lot faster.

leenagupte: I could just do `content_problems + image_problems + organisation_problems`

Gemfile
```diff
@@ -24,6 +24,7 @@ gem "pg", "~> 1"
 gem "plek", "~> 3"
 gem "rinku", "~> 2"
 gem "rubyzip", "~> 2", require: "zip"
+gem "sanitize", "~> 5.1.0"
```
kevindew: Can we go with `~> 5` to be consistent with our other pins?

lib/whitehall_importer/import.rb
```diff
@@ -73,5 +74,16 @@ def first_publish_date
...
       EditionHistory.new(whitehall_document["editions"].first["revision_history"])
     end
+
+    def integrity_check(document)
+      integrity_check_edition(document.live_edition) if document.live_edition
+      integrity_check_edition(document.current_edition) if document.current_edition != document.live_edition
+    end
+
+    def integrity_check_edition(edition)
+      integrity_checker = WhitehallImporter::IntegrityChecker.new(edition)
```
kevindew: You shouldn't need this to be prefixed with "WhitehallImporter::"

lib/whitehall_importer/import.rb
```diff
@@ -73,5 +74,16 @@ def first_publish_date
     def first_edition_history
       EditionHistory.new(whitehall_document["editions"].first["revision_history"])
     end
+
+    def integrity_check(document)
+      integrity_check_edition(document.live_edition) if document.live_edition
+      integrity_check_edition(document.current_edition) if document.current_edition != document.live_edition
+    end
+
+    def integrity_check_edition(edition)
```
kevindew: These two methods may feel more consistent with the methods with the rest of this file if they start with the verb. For example `check_document_integrity` and `check_edition_integrity`

lib/whitehall_importer/integrity_checker.rb
```diff
@@ -0,0 +1,93 @@
...
+      Sanitize.clean(publishing_api_body_text).squish == Sanitize.clean(proposed_body_text).squish
+    end
+
+    def image_problems
+      proposed_image_payload = proposed_payload.dig("details", "image") || {}
+      publishing_api_image = publishing_api_content.dig("details", "image") || {}
+
+      problems = []
+
+      %w(alt_text caption).each do |attribute|
```
kevindew: You can use an each_with_object or an inject to avoid the need to create and return the problems array.

spec/factories/whitehall_export/edition_factory.rb
```diff
@@ -52,5 +52,28 @@
         ]
       end
     end
+
+    trait :published do
+      transient do
+        previous_state { "submitted" }
+      end
```
kevindew: Do we make use of this transient attribute? I have a feeling it's a c+p from scheduled where we care more about it

spec/factories/whitehall_export/edition_factory.rb
```diff
@@ -52,5 +52,28 @@
...
+          build(:revision_history_event,
+                event: "update",
+                created_at: 2.days.ago.rfc3339,
+                state: previous_state),
+          build(:revision_history_event,
+                event: "update",
+                state: "published",
+                created_at: Time.current.tomorrow.rfc3339),
+        ]
+      end
```
kevindew: I see there's other scheduled bits in here. I wonder if it'd work the same for us were we to do `build(:whitehall_export_edition, state: "published")` and avoid needing the trait?

spec/lib/whitehall_importer/import_spec.rb
```diff
@@ -107,5 +113,28 @@
...
+      allow(WhitehallImporter::IntegrityChecker)
+        .to receive(:new)
+        .and_return(instance_double(
+                      WhitehallImporter::IntegrityChecker,
+                      valid?: false,
+                      problems: ["base path doesn't match"],
+                    ))
+
+      expect { described_class.call(build(:whitehall_export_document)) }
+        .to raise_error(WhitehallImporter::AbortImportError)
```
kevindew: Should we also assert that this includes the message since we typed it out above?

spec/lib/whitehall_importer/integrity_checker_spec.rb
```diff
@@ -0,0 +1,144 @@
...
+  describe "#valid?" do
+    let(:edition) do
+      create(
+        :edition,
+        tags: {
+          primary_publishing_organisation: [SecureRandom.uuid],
+          organisations: [SecureRandom.uuid],
+        },
+      )
+    end
```
kevindew: Can we use build(:edition) rather than create so we don't need to hit the database?

spec/lib/whitehall_importer/integrity_checker_spec.rb
```diff
@@ -0,0 +1,144 @@
...
+
+    it "returns true if there aren't any problems" do
+      integrity_check = WhitehallImporter::IntegrityChecker.new(edition)
+      expect(integrity_check.valid?).to be true
+    end
+  end
+
+  describe "#problems" do
+    let(:edition) { create(:edition) }
```
kevindew: similar point on build rather than create

spec/factories/whitehall_export/edition_factory.rb
leenagupte: You caught me! I'll get rid of it.

spec/factories/whitehall_export/edition_factory.rb
leenagupte: I tried that, but it doesn't work because you also need a revision history entry for the "published" state, so I thought it was nicer to create a trait instead.

lib/whitehall_importer/import.rb
leenagupte: Oops! Will fix.

lib/whitehall_importer/integrity_checker.rb
erino: Even better

leenagupte:
@kevindew Thanks for your review. I've addressed your comments and fixed up. The only one I wasn't sure about, and have left, is the one about the number of tests, but we spoke about that IRL.

spec/factories/whitehall_export/edition_factory.rb
kevindew: No problem 👍

spec/lib/tasks/import_spec.rb
```diff
@@ -28,12 +28,14 @@
       Rake::Task["import:whitehall_document"].reenable
       allow(ResyncService).to receive(:call)
       allow(WhitehallImporter::ClearLinksetLinks).to receive(:call)
+      allow(WhitehallImporter::Import).to receive(:call).and_return(create(:document, :with_current_edition))
```
kevindew: I think we've got an issue here where this is also testing a level above the unit of code. 

I've put together a quick proposal for how we should unit test this task: https://github.com/alphagov/content-publisher/commit/a99eed212c58ac7adf7e72727934dbdffe3cfa7e

lib/whitehall_importer/integrity_checker.rb
```diff
@@ -0,0 +1,97 @@
...
+      problems << "primary_publishing_organisation doesn't match" unless primary_publishing_organisation_matches?
+      problems << "organisations don't match" unless organisations_match?
+
+      problems
+    end
+
+    def primary_publishing_organisation_matches?
+      primary_publishing_organisation =
+        publishing_api_content.dig("links", "primary_publishing_organisation") ||
+        publishing_api_links.dig("links", "primary_publishing_organisation")
```
kevindew: Do we want to capture this in a method?

we could have: 

```
def publishing_api_link(link_type)
  publishing_api_content.dig("links", link_type) ||
    publishing_api_links.dig("links", link_type)
end
```

spec/lib/whitehall_importer/integrity_checker_spec.rb
```diff
@@ -0,0 +1,168 @@
...
+        content_id: edition.content_id,
+        links: {
+          primary_publishing_organisation: edition.tags["primary_publishing_organisation"].to_a,
+          organisations: edition.tags["organisations"].to_a + edition.tags["primary_publishing_organisation"].to_a,
+        },
+      )
+
+      integrity_check = WhitehallImporter::IntegrityChecker.new(edition)
+      expect(integrity_check.valid?).to be true
+    end
```
kevindew: 👍

lib/whitehall_importer/integrity_checker.rb
leenagupte: Ha! That's what I had originally and took out! I can add it back in.

spec/lib/tasks/import_spec.rb
leenagupte: Oh! You want to remove the "creates a document" test? Well that certainly makes the rake task easier to test. I assume it's because the functionality and whether a document is created should be tested by the `WhitehallImporter` tests?

leenagupte:
@kevindew I've addressed the rest of your comments and fixed up. Cheers!

spec/lib/tasks/import_spec.rb
kevindew: Yeah we don't actually create a document in the rake task so it seems an odd thing to assert on, I guess we either used to do that or we were testing this more like an integration test.

PR #1605, opened 2020-01-06
Fix publishing-api deprecation warnings

leenagupte:
Fixes warnings for:

```
#publishing_api_isnt_available is deprecated on GdsApi::TestHelpers::PublishingApi and will be removed in a future version. Use #stub_publishing_api_isnt_available instead
#stub_default_publishing_api_put_intent is deprecated on GdsApi::TestHelpers::PublishingApi and will be removed in a future version. Use #stub_any_publishing_api_put_intent instead
.
.
.
```


kevindew:
⭐️

PR #1609, opened 2020-01-07
Change submitted for review banner

emmabeynon:
For https://trello.com/c/1zOX0xmQ/1285-submit-for-2i-review-banner

This changes the submit for review banner to use the inset prompt component and ensures it persists while a document is in a `submitted_for_review` state.

**Before**
submitted_for_review_before

**After**
submitted_for_review_after

### DIFF
--- a/app/views/documents/show/_summary_tab.html.erb
+++ b/app/views/documents/show/_summary_tab.html.erb
@@ -1,5 +1,11 @@
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
+    <% if @edition.failed_to_publish? %>
+      <%= render "documents/show/failed_to_publish_notice" %>
+    <% end %>
+
+    <%= render "documents/show/requirements" %>
+
     <% if @edition.submitted_for_review? %>
       <%= render "documents/show/submitted_for_review" %>
     <% end %>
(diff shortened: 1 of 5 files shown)


app/views/documents/show/_summary_tab.html.erb
```diff
@@ -1,6 +1,6 @@
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
-    <% if flash[:submitted_for_review] %>
+    <% if @edition.submitted_for_review? %>
```
kevindew: I think we’ll need to ponder the ordering of this now that it is persistent.

There’s various different prompts that can appear in this section and I’d expect that this one will be a lower priority than alerts such as history mode or publishing requirements. May be worth checking with Simon based on the worst case scenario.

emmabeynon: It doesn't seem that you can submit for 2i if there are outstanding publishing requirements (see https://github.com/alphagov/content-publisher/blob/master/app/interactors/review/submit_for2i_interactor.rb#L29) so I don't think that should be an issue.  I'll check with Simon about history mode.

emmabeynon: @kevindew Simon's done a more thorough review of the ordering of messages on the summary tab, which I've updated in another commit.

kevindew: Nice one 👍

PR #1610, opened 2020-01-07
1287 delete draft banner

gclssvglx:
For https://trello.com/c/bE8Uz9e6/1287-delete-draft-banner

This changes the way that the delete draft action confirmation is handled.  

Currently, an `alert` is used to request confirmation of deletion, this change introduces a new page `confirm_delete_draft` for that purpose.  

The current use of an `alert` is considered an incorrect use of the pattern. So this change brings the action in-line with other actions, such as scheduling, removing content, change history mode tag, etc.

**Before...**

before

**After...** 

after


gclssvglx:
Thanks for the comments Peter - makes sense.  Have now re-ordered the commits.

PR #1611, opened 2020-01-07
1286 undo withdraw banner

gclssvglx:
For https://trello.com/c/9M016Ztw/1286-undo-withdraw-banner

This changes the way that the undo withdraw action confirmation is handled.

Currently, an alert is used to request confirmation of undo withdraw, this change introduces a new page confirm for that purpose.

The current use of an alert is considered an incorrect use of the pattern. So this change brings the action in-line with other actions, such as scheduling, removing content, change history mode tag, etc.

**Before...** 

before

** After...**

after


app/views/unwithdraw/confirm.html.erb
```diff
@@ -0,0 +1,34 @@
+<% content_for :title, t("unwithdraw.confirm.title", title: @edition.title_or_fallback) %>
+<% content_for :back_link, render_back_link(href: document_path(@edition.document)) %>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-full">
+    <div class="govuk-grid-row">
+      <div class="govuk-grid-column-two-thirds">
+        <p class="govuk-body">
+          <%= render_govspeak(t("unwithdraw.confirm.description_govspeak")) %>
+        </p>
```
kevindew: This will be producing invalid HTML as the govspeak will be generating paragraph elements and they're not allowed to be nested within a paragraph element.

You should embed within a `div` rather than p for this.

kevindew: Just noticed that we've got this on the previous PR too btw.

spec/requests/unwithdraw_spec.rb
```diff
@@ -65,9 +65,8 @@
       login_as(managing_editor)
 
       get unwithdraw_path(withdrawn_edition.document)
-      follow_redirect!
 
-      expect(response.body).to include(I18n.t!("documents.show.unwithdraw.title"))
+      expect(response.body).to include(withdrawn_edition.title_or_fallback)
```
kevindew: Switching to "unwithdraw.confirm.title" would be a closer like for like swap here.

spec/requests/unwithdraw_spec.rb
```diff
@@ -94,9 +93,8 @@
         login_as(user)
 
         get unwithdraw_path(withdrawn_history_mode_edition.document)
-        follow_redirect!
 
-        expect(response.body).to include(I18n.t!("documents.show.unwithdraw.title"))
+        expect(response.body).to include(withdrawn_history_mode_edition.title_or_fallback)
```
kevindew: Same comment re unwithdraw.confirm.title

config/locales/en/unwithdraw/unwithdraw.yml
```diff
@@ -2,3 +2,10 @@ en:
   unwithdraw:
     no_managing_editor_permission:
       title: Only a managing editor can undo withdraw
+    confirm:
+      title: Undo the withdrawal of '%{title}'
+      description_govspeak: |
+        Are you sure you want to undo the withdrawal?
+
+        This will remove the withdrawn banner and republish
+        the content immediately. It will not send an email alert to users.
```
kevindew: For consistency this should be in `config/locales/en/unwithdraw/confirm.yml` where I understand our convention is we have locale files ordered via /controller/action.yml

spec/requests/unwithdraw_spec.rb
gclssvglx: Thanks - yes, I thought about that but unlike the delete draft we already had a dedicated unwithdraw.yml.  So, rather than create an additional file - and as it seemed to fit the need - I opted to use that.  I can change if you'd prefer.

kevindew: Yeah I think we should try maintain the consistent approach. Is the existing file breaking this convention?

app/views/unwithdraw/confirm.html.erb
gclssvglx: Thanks for this - my lack of understanding of how govspeak works.  Will change.

spec/requests/unwithdraw_spec.rb
gclssvglx: Ok, got it - will change.

spec/requests/unwithdraw_spec.rb
gclssvglx: Will change

spec/requests/unwithdraw_spec.rb
kevindew: Oops I think we chatted about this comment here: https://github.com/alphagov/content-publisher/pull/1611/files/78c5e74f4b8d48208c5bf117dad255d505e96ea0#diff-a94cb3ddb985768921b39dc10c93465c rather than this code change.

kevindew:
I checked with Simon by the way and he's happy about the key question having reduced prominence.

gclssvglx:
Cool - thanks Kevin

app/views/unwithdraw/confirm.html.erb
gclssvglx: Fixed

spec/requests/unwithdraw_spec.rb
gclssvglx: Fixed

spec/requests/unwithdraw_spec.rb
gclssvglx: Fixed

config/locales/en/unwithdraw/unwithdraw.yml
gclssvglx: Fixed

PR #1615, opened 2020-01-08
Enqueue document import

JonathanHallam:
This allows us to queue a series of jobs to import multiple documents from Whitehall based on an organisation slug and document type.

Previously importing was being done by running a rake task (`whitehall_document`) that imported a single document.  This takes that logic and executes the same process inside a Sidekiq job.

A `WhitehallMigration` is created when the process begins, and is also marked as completed (by means of a `end_time`) when the import has finished.

### DIFF
--- a/app/jobs/whitehall_document_import_job.rb
+++ b/app/jobs/whitehall_document_import_job.rb
@@ -3,5 +3,6 @@
 class WhitehallDocumentImportJob < ApplicationJob
   def perform(document_import)
     WhitehallImporter.import_and_sync(document_import)
+    document_import.whitehall_migration&.check_migration_finished
   end
 end
--- a/app/jobs/whitehall_document_import_job.rb
+++ b/app/jobs/whitehall_document_import_job.rb
@@ -1,0 +1,7 @@
+# frozen_string_literal: true
+
+class WhitehallDocumentImportJob < ApplicationJob
+  def perform(document_import)
+    WhitehallImporter.import_and_sync(document_import)
+  end
+end
--- a/lib/tasks/import.rake
+++ b/lib/tasks/import.rake
@@ -17,8 +17,11 @@ namespace :import do
 
   desc "Import a single document from Whitehall Publisher using Whitehall's internal document ID e.g. import:whitehall_document[123]"
   task :whitehall_document, [:document_id] => :environment do |_, args|
-    whitehall_export = GdsApi.whitehall_export.document_export(args.document_id)
-    whitehall_import = WhitehallImporter.import_and_sync(whitehall_export.to_hash)
+    whitehall_import = WhitehallMigration::DocumentImport.create!(
+      whitehall_document_id: args.document_id,
+      state: "pending",
+    )
+    WhitehallImporter.import_and_sync(whitehall_import)
 
     unless whitehall_import.completed?
       puts whitehall_import.state.humanize
--- a/lib/whitehall_importer.rb
+++ b/lib/whitehall_importer.rb
@@ -1,9 +1,11 @@
 # frozen_string_literal: true
 
+require "gds_api/whitehall_export"
+
 module WhitehallImporter
   def self.create_migration(organisation_content_id, document_type)
-    ActiveRecord::Base.transaction do
-      whitehall_migration = WhitehallMigration.create!(organisation_content_id: organisation_content_id,
+    whitehall_migration = ActiveRecord::Base.transaction do
+      record = WhitehallMigration.create!(organisation_content_id: organisation_content_id,
                                  document_type: document_type,
                                  start_time: Time.current)
 
@@ -13,19 +15,24 @@ module WhitehallImporter
         page["documents"].each do |document|
           WhitehallMigration::DocumentImport.create!(
             whitehall_document_id: document["document_id"],
-            whitehall_migration_id: whitehall_migration.id,
+            whitehall_migration_id: record.id,
             state: "pending",
           )
         end
       end
-
-      whitehall_migration
+      record
     end
+    whitehall_migration.document_imports.find_each do |document_import|
+      WhitehallDocumentImportJob.perform_later(document_import)
+    end
+    whitehall_migration
   end
 
-  def self.import_and_sync(whitehall_document)
-    whitehall_import = WhitehallMigration::DocumentImport.create!(
-      whitehall_document_id: whitehall_document["id"],
+  def self.import_and_sync(whitehall_import)
+    raise "Cannot import with a state of #{whitehall_import.state}" unless whitehall_import.pending?
+
+    whitehall_document = GdsApi.whitehall_export.document_export(whitehall_import.whitehall_document_id).to_h
+    whitehall_import.update!(
       payload: whitehall_document,
       content_id: whitehall_document["content_id"],
       state: "importing",
--- a/spec/factories/whitehall_migration_factory.rb
+++ b/spec/factories/whitehall_migration_factory.rb
@@ -1,0 +1,12 @@
+# frozen_string_literal: true
+
+FactoryBot.define do
+  factory :whitehall_migration do
+    sequence(:id)
+    organisation_content_id { "content_id" }
+    document_type { "NewsArticle" }
+    start_time { Time.current }
+    created_at { Time.current.rfc3339 }
+    updated_at { Time.current.rfc3339 }
+  end
+end
--- a/spec/jobs/whitehall_document_import_job_spec.rb
+++ b/spec/jobs/whitehall_document_import_job_spec.rb
@@ -11,10 +11,16 @@ RSpec.describe WhitehallDocumentImportJob do
 
   before do
     allow(WhitehallImporter).to receive(:import_and_sync)
+    allow(whitehall_migration).to receive(:check_migration_finished)
   end
 
   it "calls on the import and sync method" do
     expect(WhitehallImporter).to receive(:import_and_sync).with(whitehall_migration_document_import)
+    WhitehallDocumentImportJob.perform_now(whitehall_migration_document_import)
+  end
+
+  it "calls on the mark migration completed method" do
+    expect(whitehall_migration_document_import.whitehall_migration).to receive(:check_migration_finished)
     WhitehallDocumentImportJob.perform_now(whitehall_migration_document_import)
   end
 end
--- a/spec/jobs/whitehall_document_import_job_spec.rb
+++ b/spec/jobs/whitehall_document_import_job_spec.rb
@@ -1,0 +1,20 @@
+# frozen_string_literal: true
+
+RSpec.describe WhitehallDocumentImportJob do
+  include ActiveJob::TestHelper
+
+  let(:whitehall_migration) { create(:whitehall_migration) }
+
+  let(:whitehall_migration_document_import) do
+    create(:whitehall_migration_document_import, whitehall_migration_id: whitehall_migration["id"], state: "pending")
+  end
+
+  before do
+    allow(WhitehallImporter).to receive(:import_and_sync)
+  end
+
+  it "calls on the import and sync method" do
+    expect(WhitehallImporter).to receive(:import_and_sync).with(whitehall_migration_document_import)
+    WhitehallDocumentImportJob.perform_now(whitehall_migration_document_import)
+  end
+end
--- a/spec/lib/tasks/import_spec.rb
+++ b/spec/lib/tasks/import_spec.rb
@@ -24,25 +24,34 @@ RSpec.describe "Import tasks" do
     let(:whitehall_export) { build(:whitehall_export_document) }
 
     before do
-      allow($stdout).to receive(:puts)
       Rake::Task["import:whitehall_document"].reenable
-      stub_request(:get, "#{whitehall_host}/government/admin/export/document/123")
-        .to_return(status: 200, body: whitehall_export.to_json)
     end
 
     it "imports the export and syncs with publishing-api" do
-      import = build(:whitehall_migration_document_import, state: :completed)
+      import = build(:whitehall_migration_document_import, state: "completed")
 
-      expect(WhitehallImporter).to receive(:import_and_sync)
-                               .with(whitehall_export)
-                               .and_return(import)
+      allow(WhitehallMigration::DocumentImport).to receive(:create!).and_return(import)
+      expect(WhitehallImporter).to receive(:import_and_sync).with(import)
+
       Rake::Task["import:whitehall_document"].invoke("123")
     end
 
+    it "creates a pending whitehall migration document import" do
+      allow(WhitehallImporter).to receive(:import_and_sync)
+      allow_any_instance_of(WhitehallMigration::DocumentImport).to receive(:completed?).and_return(true)
+
+      expect { Rake::Task["import:whitehall_document"].invoke("123") }
+        .to change { WhitehallMigration::DocumentImport.pending.exists?(whitehall_document_id: 123) }
+        .to(true)
+    end
+
     it "aborts if the import fails" do
+      allow($stdout).to receive(:puts)
+
       import = build(:whitehall_migration_document_import,
                      state: :import_failed,
                      error_log: "Error importing")
+      allow(WhitehallMigration::DocumentImport).to receive(:create!).and_return(import)
       expect(WhitehallImporter).to receive(:import_and_sync).and_return(import)
 
       expect($stdout).to receive(:puts).with("Import failed")
--- a/spec/lib/whitehall_importer_spec.rb
+++ b/spec/lib/whitehall_importer_spec.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 RSpec.describe WhitehallImporter do
+  include ActiveJob::TestHelper
   describe ".create_migration" do
     let(:whitehall_host) { Plek.new.external_url_for("whitehall-admin") }
     let(:whitehall_export_page_1) { build(:whitehall_export_index, documents: build_list(:whitehall_export_index_document, 100)) }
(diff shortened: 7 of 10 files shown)


lib/whitehall_importer.rb
```diff
@@ -23,9 +25,26 @@ def self.create_migration(organisation_content_id, document_type)
...
+    documents = WhitehallMigration::DocumentImport.where(whitehall_migration_id: whitehall_migration.id)
+    documents.each do |document|
+      WhitehallDocumentImportJob.perform_later(document)
+    end
+  end
+
+  def self.mark_migration_completed(whitehall_migration_document_import)
+    running_states = %w(pending importing imported syncing)
+    whitehall_migration = WhitehallMigration.find(whitehall_migration_document_import.whitehall_migration_id)
+    if WhitehallMigration::DocumentImport.where(whitehall_migration_id: whitehall_migration.id, state: running_states).empty?
```
kevindew: These 3 lines seem a bit funky and I think they could be wrote more declaratively with ActiveRecord associations.

We could put a scope on DocumentImport to determine which ones are finished with something like:  `scope :in_progress, -> { pending.or(importing).or(imported).or(syncing) }` and then on the whitehall_migration model ask `whitehall_migration.documents.in_progress.any?`

This will mean setting up the association in ActiveRecord but we should do that anyway since these are associations.

lib/whitehall_importer.rb
```diff
@@ -23,9 +25,26 @@ def self.create_migration(organisation_content_id, document_type)
...
+      whitehall_migration.update!(
+        end_time: Time.current,
+      )
```
kevindew: Minor, this could be a single line.

lib/whitehall_importer.rb
```diff
@@ -23,9 +25,26 @@ def self.create_migration(organisation_content_id, document_type)
...
+    end
+  end
+
+  def self.import_and_sync(whitehall_import)
```
kevindew: Something to consider is that if this method is being called from ActiveJob then we should assume that we don't know what state the whitehall_import model is in (Jobs should be idempotent and resilient to retrying). 

We hit a problem here that this method immediately changes the state of the `whitehall_import` model which hinders the idempotency of the job. 

Might be that all we want to do is assert that the whitehall_import is pending when this method is called and worry further about retryability in the follow up story.

lib/whitehall_importer.rb
```diff
@@ -23,9 +25,26 @@ def self.create_migration(organisation_content_id, document_type)
...
+    whitehall_document = GdsApi.whitehall_export.document_export(whitehall_import.whitehall_document_id).to_h
```
kevindew: I can see why this has moved to here but one thing that becomes a bit unclear is who/what is responsible for dealing with whether this fails. This was a bit simpler when this was a rake task as we'd just have that task fail. Maybe this is something to be concerned about with retrying as I'm a little unsure of a clear answer - just feels a riskier introduction.

spec/jobs/whitehall_document_import_job_spec.rb
```diff
@@ -0,0 +1,28 @@
+# frozen_string_literal: true
+
+RSpec.describe WhitehallDocumentImportJob do
+  include ActiveJob::TestHelper
+
+  let(:whitehall_migration) {
+    build(:whitehall_migration)
+  }
```
kevindew: Minor: our convention dictates do and end for multi line blocks. This should probably be a single line block.

spec/jobs/whitehall_document_import_job_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+
+  let(:whitehall_migration_document_import) {
+    build(:whitehall_migration_document_import, whitehall_migration_id: whitehall_migration["id"])
+  }
+
+  before do
+    allow(WhitehallImporter).to receive(:import_and_sync).and_return("return_value")
+    allow(WhitehallImporter).to receive(:mark_migration_completed).and_return("return_value")
```
kevindew: Do we use/need these "return_value" declarations?

spec/lib/whitehall_importer_spec.rb
```diff
@@ -27,26 +28,74 @@
...
+  end
+
+  describe ".mark_migration_completed" do
+    context "with only jobs that have finished running" do
+      before do
+        @whitehall_migration = create(:whitehall_migration)
+        @whitehall_migration_document_import_completed = create(:whitehall_migration_document_import, whitehall_migration_id: @whitehall_migration["id"], state: "completed")
+        @whitehall_migration_document_import_aborted = create(:whitehall_migration_document_import, whitehall_migration_id: @whitehall_migration["id"], state: "import_aborted")
+        @whitehall_migration_document_import_failed = create(:whitehall_migration_document_import, whitehall_migration_id: @whitehall_migration["id"], state: "import_failed")
+        @whitehall_migration_document_import_sync_failed = create(:whitehall_migration_document_import, whitehall_migration_id: @whitehall_migration["id"], state: "sync_failed")
```
kevindew: Convention would dictate using let arguments rather than creating these as ivars, there's various other ones of these scattered around this file.

lib/whitehall_importer.rb
```diff
@@ -23,9 +25,26 @@ def self.create_migration(organisation_content_id, document_type)
     end
   end
 
-  def self.import_and_sync(whitehall_document)
-    whitehall_import = WhitehallMigration::DocumentImport.create!(
-      whitehall_document_id: whitehall_document["id"],
+  def self.queue_migration_jobs(whitehall_migration)
+    documents = WhitehallMigration::DocumentImport.where(whitehall_migration_id: whitehall_migration.id)
+    documents.each do |document|
```
kevindew: Looks like there should be an ActiveRecord association set up for this so that we can do `whitehall_migration.documents.each`

app/jobs/whitehall_document_import_job.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+class WhitehallDocumentImportJob < ApplicationJob
+  def perform(whitehall_migration_document_import)
+    WhitehallImporter.import_and_sync(whitehall_migration_document_import)
+  end
+
+  after_perform do |job|
+    WhitehallImporter.mark_migration_completed(job.arguments.first)
+  end
```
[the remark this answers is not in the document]
brucebolt: We agree nothing is gained (in fact, the `after_perform` won't get executed if the job fails) and have moved this into `perform` block.

app/jobs/whitehall_document_import_job.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+class WhitehallDocumentImportJob < ApplicationJob
+  def perform(whitehall_migration_document_import)
+    WhitehallImporter.import_and_sync(whitehall_migration_document_import)
+  end
+
+  after_perform do |job|
+    WhitehallImporter.mark_migration_completed(job.arguments.first)
```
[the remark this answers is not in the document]
brucebolt: Renamed this to `check_migration_finished` and moved to the model.

lib/tasks/import.rake
```diff
@@ -10,15 +10,19 @@ namespace :import do
     )
 
     whitehall_migration = WhitehallImporter.create_migration(organisation_content_id, args.document_type)
+    WhitehallImporter.queue_migration_jobs(whitehall_migration)
```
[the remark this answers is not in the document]
brucebolt: Moved this out of the rake task and into the `create_migration` method.  Tests will be updated to reflect this.

lib/whitehall_importer.rb
brucebolt: This is definitely much cleaner and has reduced the length/complexity of this code.

lib/whitehall_importer.rb
brucebolt: Moved this onto a single as part of the change to scope `in_progress`

spec/jobs/whitehall_document_import_job_spec.rb
brucebolt: Removed these since they aren't needed.

lib/whitehall_importer.rb
brucebolt: We've created an association for this and updated the relevant part of the import code.

lib/whitehall_importer.rb
JonathanHallam: We've left a comment in https://trello.com/c/xXNlisKk/1273-spike-how-and-when-should-we-retry-jobs to make sure that this is followed up when that spike gets done :)

spec/lib/whitehall_importer_spec.rb
JonathanHallam: We've removed the assignments for the objects that we're not actually using and added a `let!` for those that need an assignment. This tidies it up a bit :)

spec/lib/whitehall_importer_spec.rb
```diff
@@ -25,37 +26,63 @@
...
+    it "calls the queue_migration_jobs method" do
+      allow(WhitehallImporter).to receive(:queue_migration_jobs)
+      whitehall_migration = WhitehallImporter.create_migration("123", "NewsArticle")
+      expect(WhitehallImporter).to have_received(:queue_migration_jobs).with(whitehall_migration)
+    end
+  end
+
+  describe ".queue_migration_jobs" do
+    before do
+      @whitehall_migration = create(:whitehall_migration)
```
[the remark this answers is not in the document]
JonathanHallam: yep, have split this out and put the following line on a few to tidy it up some

spec/lib/whitehall_importer_spec.rb
```diff
@@ -25,37 +26,63 @@
...
+
+    it "queues a job for each listed document" do
+      WhitehallImporter.queue_migration_jobs(@whitehall_migration)
+      expect(WhitehallDocumentImportJob).to have_been_enqueued.exactly(5).times
+    end
   end
 
   describe ".import_and_sync" do
+    let(:whitehall_export_document) { build(:whitehall_export_document) }
+    let(:whitehall_migration_document_import) { build(:whitehall_migration_document_import, whitehall_document_id: "123", state: "pending") }
```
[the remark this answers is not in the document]
JonathanHallam: agreed, looks nicer now :)

lib/whitehall_importer.rb
```diff
@@ -13,19 +15,29 @@ def self.create_migration(organisation_content_id, document_type)
...
       end
+      record
+    end
+    queue_migration_jobs(whitehall_migration)
+    whitehall_migration
+  end
 
-      whitehall_migration
+  def self.queue_migration_jobs(whitehall_migration)
+    documents = WhitehallMigration::DocumentImport.where(whitehall_migration_id: whitehall_migration.id)
```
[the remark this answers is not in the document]
brucebolt: I've dropped this method and used the association, which has simplified this code considerably

spec/jobs/whitehall_document_import_job_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+
+  let(:whitehall_migration) { create(:whitehall_migration) }
+
+  let(:whitehall_migration_document_import) do
+    create(:whitehall_migration_document_import, whitehall_migration_id: whitehall_migration["id"])
+  end
+
+  before do
+    allow(WhitehallImporter).to receive(:import_and_sync)
+    allow(whitehall_migration).to receive(:check_migration_finished)
```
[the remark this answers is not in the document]
brucebolt: We mocked these since the `import_and_sync` method makes HTTP requests, which would need to be stubbed.  It didn't seem like the right place to test these requests, since they are being tested elsewhere.

spec/lib/tasks/import_spec.rb
```diff
@@ -11,6 +11,7 @@
         "/government/organisations/cabinet-office" => "96ae61d6-c2a1-48cb-8e67-da9d105ae381",
       )
       allow(WhitehallImporter).to receive(:create_migration).and_return(whitehall_migration_document_import)
+      allow(WhitehallImporter).to receive(:queue_migration_jobs).and_return(nil)
```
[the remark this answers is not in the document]
brucebolt: Yes, it isn't needed, so I've removed it

spec/lib/whitehall_importer_spec.rb
```diff
@@ -25,37 +26,74 @@
...
     end
+
+    it "calls the queue_migration_jobs method" do
+      allow(WhitehallImporter).to receive(:queue_migration_jobs)
+      whitehall_migration = WhitehallImporter.create_migration("123", "NewsArticle")
+      expect(WhitehallImporter).to have_received(:queue_migration_jobs).with(whitehall_migration)
+    end
+  end
+
+  describe ".queue_migration_jobs" do
```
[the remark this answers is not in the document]
brucebolt: I've removed the `queue_migration_jobs` method and now testing the queuing on the `create_migration` method

spec/lib/tasks/import_spec.rb
```diff
@@ -24,22 +25,35 @@
     let(:whitehall_export) { build(:whitehall_export_document) }
 
     before do
-      allow($stdout).to receive(:puts)
       Rake::Task["import:whitehall_document"].reenable
       stub_request(:get, "#{whitehall_host}/government/admin/export/document/123")
         .to_return(status: 200, body: whitehall_export.to_json)
```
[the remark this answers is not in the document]
brucebolt: Nope, not needed anymore

spec/lib/tasks/import_spec.rb
```diff
@@ -24,22 +25,35 @@
...
+      allow(WhitehallImporter::Import).to receive(:call).and_return(build(:document, :with_current_edition))
```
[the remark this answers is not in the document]
brucebolt: Turns out we didn't need this anyway, so have removed it

spec/models/whitehall_migration_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+
+    context "with some incomplete jobs" do
+      let!(:whitehall_migration) { create(:whitehall_migration) }
+      before do
+        create(:whitehall_migration_document_import, whitehall_migration_id: whitehall_migration["id"], state: "completed")
+        create(:whitehall_migration_document_import, whitehall_migration_id: whitehall_migration["id"], state: "pending")
+      end
+
+      it "does not update each of the end times" do
+        freeze_time do
```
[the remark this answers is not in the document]
brucebolt: Looks like it isn't doing anything, so have removed it

spec/factories/whitehall_migration_factory.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+FactoryBot.define do
+  factory :whitehall_migration do
+    sequence(:id)
+    organisation_content_id { "content_id" }
+    document_type { "NewsArticle" }
+    start_time { Time.current }
+    created_at { Time.current.rfc3339 }
```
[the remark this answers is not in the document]
brucebolt: Maybe something we could fix with a migration to rename the column (or just remove it)

spec/factories/whitehall_migration_document_import_factory.rb
```diff
@@ -7,5 +7,6 @@
     whitehall_document_id { payload["id"] }
     content_id { payload["content_id"] }
     document { association :document, imported_from: "whitehall" }
+    whitehall_migration_id { nil }
```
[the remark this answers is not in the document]
kevindew: @brucebolt did we work out if we needed this - not sure if this comment got orphaned somewhere.

spec/factories/whitehall_migration_factory.rb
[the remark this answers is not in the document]
kevindew: Yeah I think we can sort it out relatively simply - shall we flag it as further AC of this card to do after this PR is merged?

spec/lib/tasks/import_spec.rb
```diff
@@ -24,22 +24,34 @@
...
+
+      allow(WhitehallMigration::DocumentImport).to receive(:create!).and_return(import)
+      allow(WhitehallImporter).to receive(:import_and_sync).and_return(import)
 
-      expect(WhitehallImporter).to receive(:import_and_sync)
-                               .with(whitehall_export)
-                               .and_return(import)
       Rake::Task["import:whitehall_document"].invoke("123")
+
+      expect(WhitehallImporter).to have_received(:import_and_sync)
```
kevindew: If we change the allow above to be an expect we won't need to set things up for this twice.

spec/lib/tasks/import_spec.rb
```diff
@@ -24,22 +24,34 @@
...
-      stub_request(:get, "#{whitehall_host}/government/admin/export/document/123")
-        .to_return(status: 200, body: whitehall_export.to_json)
     end
 
     it "imports the export and syncs with publishing-api" do
-      import = build(:whitehall_migration_document_import, state: :completed)
+      import = build(:whitehall_migration_document_import, state: "completed")
+
+      allow(WhitehallMigration::DocumentImport).to receive(:create!).and_return(import)
+      allow(WhitehallImporter).to receive(:import_and_sync).and_return(import)
```
kevindew: We shouldn't care that this returns anything anymore.

spec/lib/tasks/import_spec.rb
```diff
@@ -24,22 +24,34 @@
...
       Rake::Task["import:whitehall_document"].invoke("123")
+
+      expect(WhitehallImporter).to have_received(:import_and_sync)
+                               .with(import)
+    end
+
+    it "creates a pending whitehall migration document import" do
+      import = build(:whitehall_migration_document_import, state: "completed")
+
+      allow(WhitehallImporter).to receive(:import_and_sync).and_return(import)
```
kevindew: same point here that we won't need the and_return

spec/factories/whitehall_migration_document_import_factory.rb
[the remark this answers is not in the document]
brucebolt: Looks like it was inadvertently introduced during a rebase.  Have removed it now.

spec/factories/whitehall_migration_factory.rb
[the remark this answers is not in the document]
brucebolt: Yes, that's a good idea

spec/lib/tasks/import_spec.rb
brucebolt: I've removed the return value since it's not needed

spec/lib/tasks/import_spec.rb
brucebolt: Removed this too

brucebolt:
We've made the proposed changes.

The rake task tests aren't the cleanest (mostly because we're testing each piece of functionality individually, there's a lot of mocking and stubbing).  I guess we could move each part of the rake task into separate methods, then have a unit test for each method (and test the rake task calls the method)?

PR #1625, opened 2020-01-09
Access limit request tests

kevindew:

This follows on the test strategy under discussion in #1590. Towards this I've done a bunch of writing of request tests and deleting of feature tests. Rather than do one large PR I thought it was best to separate into distinct PR's that are easier to review and thought I'd start with one that covers a few.

This PR introduces request tests for the access limit endpoints and removes a number of feature tests that test dead-end workflows (such as publishing api down and requirement issues) - it also consolidates the feature test into a single feature that tests that you can add and remove an access limit.


app/controllers/access_limit_controller.rb
```diff
@@ -3,25 +3,26 @@
...
 
   def edit
     @edition = Edition.find_current(document: params[:document])
+    assert_edition_state(@edition, &:editable?)
   end
 
   def update
     result = AccessLimit::UpdateInteractor.call(params: params, user: current_user)
-    issues, edition = result.to_h.values_at(:issues, :edition)
+    @edition, issues = result.to_h.values_at(:edition, :issues)
```
1pretz1: Why has the `edition` here been converted into an instance variable instead of passing to the partial via the assigns method?

spec/requests/access_limit_spec.rb
```diff
@@ -0,0 +1,89 @@
...
+    )
+  end
+
+  it_behaves_like "requests that assert edition state",
+                  "access limiting a non editable edition",
+                  routes: { access_limit_path: %i[get patch] } do
+    let(:edition) { create(:edition, :published) }
+  end
+
+  describe "GET /documents/:document/acces-limit" do
```
1pretz1: ` describe "GET /documents/:document/acces-limit" do` -> `describe "GET /documents/:document/access-limit"`

spec/requests/access_limit_spec.rb
```diff
@@ -0,0 +1,89 @@
...
+      patch access_limit_path(edition.document),
+            params: { limit_type: "primary_organisation" }
+
+      expect(response).to have_http_status(:unprocessable_entity)
+      expect(response.body).to have_content(
+        I18n.t!("requirements.access_limit.user_has_no_org.form_message"),
+      )
+    end
+
+    it "returns service unavailable with there are issues and Publishing API is down" do
```
1pretz1: `"returns service unavailable with there are issues and Publishing API is down"` -> `"returns service unavailable when there are issues and Publishing API is down"`

spec/features/editing_content_settings/remove_access_limit_spec.rb
```diff
@@ -1,67 +0,0 @@
-# frozen_string_literal: true
-
-RSpec.feature "Remove access limit" do
```
1pretz1: Nice! 🔥

app/controllers/access_limit_controller.rb
kevindew: This is to fix a bug that the API down rescue would error because it expects an edition ivar.

kevindew:
Thanks @1pretz1 - and sorry for your rebase woes!

app/controllers/access_limit_controller.rb
1pretz1: Ah okay 👍

PR #1626, opened 2020-01-09
Contact embed request tests

kevindew:

This PR introduces request tests for the contact embed endpoints, removes a number of feature test that test dead-end behaviour and consolidates the feature test to encompass js and non-js into a single file.


spec/features/editing_content/insert_contact_embed_spec.rb
```diff
@@ -1,18 +1,27 @@
...
     when_i_go_to_edit_the_edition
     and_i_click_to_insert_a_contact
-    and_i_select_a_contact
+    and_i_select_a_contact_from_the_autcomplete
     then_i_see_the_contact_preview
     when_i_click_insert_contact
     then_i_see_the_snippet_is_inserted
   end
 
+  scenario "without javascript", js: false do
```
1pretz1: Do you need to explicitly set javascript to false here for there to be no javascript?

kevindew: No - I just wanted to be explicit, rather than an implicit. Do you think it's better without?

1pretz1: Hmm, what came to my mind when I saw it was 'does the js flag from the scenario prior leak into the second scenario so you need to explicitly set it in the the second'. But now I can see you're just trying to be explicit, I think it's fine!

spec/requests/contact_embed_spec.rb
```diff
@@ -0,0 +1,72 @@
...
+    before do
+      stub_publishing_api_get_editions([contact], Contacts::EDITION_PARAMS)
+      stub_publishing_api_has_linkables([organisation],
+                                        document_type: "organisation")
+    end
+
+    it "returns just the embed code in a model context" do
+      post contact_embed_path(edition.document),
+           params: { contact_id: contact["content_id"] },
+           headers: { "Content-Publisher-Rendering-Context" => "modal" }
```
benthorner: Minor: if we're going to do this elsewhere, could we use a helper like `rendering_content("modal")`?

spec/requests/contact_embed_spec.rb
```diff
@@ -0,0 +1,72 @@
...
+
+RSpec.describe "Contact Embed" do
+  let(:edition) do
+    body_field = build(:field, id: "body", type: "govspeak")
+    document_type = build(:document_type, contents: [body_field])
+    create(:edition, document_type_id: document_type.id)
+  end
+
+  describe "GET /documents/:document/contact-embed" do
+    it "renders a service unavailable response when the Publishing API is down" do
```
benthorner: I might be missing something, but I think there's a slight risk here that we're no longer checking the error behaviour through the modal: although the responses will be correct, I'm not sure if we check the modal interprets it correctly, since it's not something that will "just work".

benthorner: I think the same may be true with requirements issues ([below](https://github.com/alphagov/content-publisher/pull/1626/commits/08fbaa896852f386064b2c1446aa5f207c6faefc#diff-eef72a8e83aead427b576a42b7b5b8b8R63)).

spec/features/editing_content/insert_contact_embed_spec.rb
```diff
@@ -1,18 +1,27 @@
...
+    given_there_is_an_edition
+    when_i_go_to_edit_the_edition
+    and_i_click_to_insert_a_contact
+    when_i_select_a_contact_without_javascript
```
benthorner: Minor: should this be an `and_` step?

kevindew:
@benthorner Ah yes darn, thanks for catching - looks like this flutters into the realm of the wall of pain card: https://trello.com/c/GVp3CoMH/1030-our-modal-scripts-dont-have-tests. I hadn't realised that at the time of writing. 

I've done a bit of experimenting and will raise a PR when I get some time to finish it. I'll discard the review on this for now.

spec/requests/contact_embed_spec.rb
kevindew: I looked across all the request tests upcoming and it's only used in this and one other file so I don't think this is a big concern currently. I think we'd want to see it used a reasonable amount more to justify a helper. 

This is shaped somewhat by it not being the clearest helper to write. I guess we'd want it to return a hash and be passed into headers. For it to be global we'd probably want to name it with headers in the name, so it'd be used like: `headers: rendering_context_header("modal")` which didn't feel a huge gain for any ambiguities it creates.

Certainly happy to re-visit if usage proliferates and causes pain.

spec/requests/contact_embed_spec.rb
kevindew: This should now be resolved by the modal-tests PR: https://github.com/alphagov/content-publisher/pull/1641

spec/features/editing_content/insert_contact_embed_spec.rb
kevindew: Thanks! sorted.

spec/requests/contact_embed_spec.rb
```diff
@@ -0,0 +1,72 @@
+# frozen_string_literal: true
+
+RSpec.describe "Contact Embed" do
+  let(:edition) do
+    body_field = build(:field, id: "body", type: "govspeak")
+    document_type = build(:document_type, contents: [body_field])
+    create(:edition, document_type_id: document_type.id)
+  end
+
+  describe "GET /documents/:document/contact-embed" do
```
benthorner: Are we missing a test here that verified the behaviour of the `new` action? I'm thinking about the modal rendering context in particular. I notice it's still covered via the feature test, so I don't think it's a major concern. More my curiosity about the apparent testing gap in this particular file.

kevindew: Sure, I see what you mean. I'd considered the layout to be an aspect of config really and have only used it as a consideration when it impacts the actual view or has logic around it, rather than the less impactful layout specifier.

It's not actually something covered by feature tests, whether we render in a modal context or not the tests will still pass as it will just render the full HTML in the modal and still find the key elements. I tested this by removing the line about layout and found the full test suite succeeded. So request tests would probably be the right layer to actually test it.

Do you think we'd benefit from testing it or that it's a glaring inconsistency?

We can probably put tests for it but I'm not sure of the best mechanism to assert on it as it's essentially the absence of key HTML elements which feels the more risky negative type of assertion than something positive (unless I'm missing something we can hook onto)

I guess if we are to do it we'd be doing it in a number of places so it'd be good to have a particular expectation for it.

benthorner: I was thinking the same thing. With the feature tests, I meant that the setting of the rendering context doesn't cause an issue in either case. It could be that the modal context triggers a buggy behaviour in the associated view.

The main oddity for me was the lack of a success case test for the GET action, which made the test seem incomplete. Are we aiming for completeness in terms of what the controller is doing, since we don't have any controller unit tests?

With the layout context specifically, we could consider using the shared example approach. However, it doesn't actually inhibit the feature, like you say, so perhaps it's not worth testing until it actually causes a problem.

kevindew: Oh I'm sorry - I was on a different wavelength.

Yes I think anytime we test an endpoint for negative scenarios there should be a pairing a positive scenario, even if it's pretty trivial just, as you say, for consistency in the test. I missed this one here - my bad 😞. I've added it now and also caught that it wasn't testing Publishing API down on the post request.

benthorner: Thanks - that's much better!

PR #1630, opened 2020-01-09
Store integrity check failures

leenagupte:
Trello: https://trello.com/c/A9GjvYyp

## What's changed and why?
The integrity check added in PR #1602 has highlighted some discrepancies between how some of the fields are handled. 

For example, if the image caption is blank, it's recorded in publishing-api as `nil`, however, the import process tries to replace this value with an empty string. Also, some fields are being sent to publishing-api with an extra space or newline at the end of them.

All of these discrepancies will cause the integrity check and therefore the import to fail. 

Debugging integrity check failures is difficult. As the integrity check runs within a transaction, when the import fails and the transaction rolls back, all of the documents, editions and revisions the import created are obliterated, which means that the proposed payload that we wanted to send to publishing-api is also lost. Trying to recreate this payload in a console is quite difficult because of the need to also create all of the editions and revisions etc.

This change keeps a copy of this payload in the database, to make it easier for the investigator to find out what's failed and why. We're already storing a copy of the export taken from Whitehall in the `whitehall_migration_document_imports` table and the proposed payload will be stored along side it. 
We don't need to store what publishing-api expects because it's easy enough to see that by calling the `get_content`, `get_live_content` and `get_links` api methods.

### DIFF
--- a/db/migrate/20200109094012_add_integrity_check_problems_to_whitehall_document_migration.rb
+++ b/db/migrate/20200109094012_add_integrity_check_problems_to_whitehall_document_migration.rb
@@ -1,0 +1,10 @@
+# frozen_string_literal: true
+
+class AddIntegrityCheckProblemsToWhitehallDocumentMigration < ActiveRecord::Migration[6.0]
+  def change
+    change_table :whitehall_migration_document_imports, bulk: true do |t|
+      t.string :integrity_check_problems, array: true, default: [], null: false
+      t.json :integrity_check_proposed_payload, default: "{}", null: false
+    end
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -358,6 +358,8 @@ ActiveRecord::Schema.define(version: 2020_01_14_140323) do
     t.datetime "updated_at", null: false
     t.bigint "document_id"
     t.bigint "whitehall_migration_id"
+    t.string "integrity_check_problems", default: [], null: false, array: true
+    t.json "integrity_check_proposed_payload", default: "{}", null: false
   end
 
   create_table "whitehall_migrations", force: :cascade do |t|
--- a/lib/whitehall_importer/import.rb
+++ b/lib/whitehall_importer/import.rb
@@ -84,7 +84,9 @@ module WhitehallImporter
     def check_edition_integrity(edition)
       integrity_checker = IntegrityChecker.new(edition)
 
-      raise AbortImportError, integrity_checker.problems.join(", ") unless integrity_checker.valid?
+      unless integrity_checker.valid?
+        raise WhitehallImporter::IntegrityCheckError.new(integrity_checker)
+      end
     end
   end
 end
--- a/lib/whitehall_importer/integrity_check_error.rb
+++ b/lib/whitehall_importer/integrity_check_error.rb
@@ -1,0 +1,15 @@
+# frozen_string_literal: true
+
+module WhitehallImporter
+  class IntegrityCheckError < AbortImportError
+    attr_reader :problems, :payload
+
+    def initialize(integrity_check)
+      @problems = integrity_check.problems
+      @payload = integrity_check.proposed_payload
+
+      publishing_status = integrity_check.edition.live? ? "live" : "draft"
+      super("#{publishing_status.titleize} integrity check failed")
+    end
+  end
+end
--- a/lib/whitehall_importer/integrity_checker.rb
+++ b/lib/whitehall_importer/integrity_checker.rb
@@ -16,6 +16,10 @@ module WhitehallImporter
       content_problems + image_problems + organisation_problems
     end
 
+    def proposed_payload
+      @proposed_payload ||= PreviewService::Payload.new(edition, republish: edition.live?).payload
+    end
+
   private
 
     def content_problems
@@ -94,10 +98,6 @@ module WhitehallImporter
 
     def organisations_match?
       proposed_payload.dig("links", "organisations")&.sort == publishing_api_link("organisations")&.sort
-    end
-
-    def proposed_payload
-      @proposed_payload ||= PreviewService::Payload.new(edition, republish: edition.live?).payload
     end
 
     def publishing_api_content
--- a/lib/whitehall_importer/integrity_checker.rb
+++ b/lib/whitehall_importer/integrity_checker.rb
@@ -27,7 +27,11 @@ module WhitehallImporter
          document_type
          schema_name).each do |attribute|
         if publishing_api_content[attribute] != proposed_payload[attribute]
-          problems << "#{attribute} doesn't match"
+          problems << problem_description(
+            "#{attribute} doesn't match",
+            publishing_api_content[attribute],
+            proposed_payload[attribute],
+          )
         end
       end
 
@@ -49,17 +53,39 @@ module WhitehallImporter
 
       %w(alt_text caption).each_with_object([]) do |attribute, problems|
         if publishing_api_image[attribute] != proposed_image_payload[attribute]
-          problems << "image #{attribute} doesn't match"
+          problems << problem_description(
+            "image #{attribute} doesn't match",
+            publishing_api_image[attribute],
+            proposed_image_payload[attribute],
+          )
         end
       end
     end
 
     def organisation_problems
       problems = []
-      problems << "primary_publishing_organisation doesn't match" unless primary_publishing_organisation_matches?
-      problems << "organisations don't match" unless organisations_match?
+
+      unless primary_publishing_organisation_matches?
+        problems << problem_description(
+          "primary_publishing_organisation doesn't match",
+          publishing_api_link("primary_publishing_organisation"),
+          proposed_payload.dig("links", "primary_publishing_organisation"),
+        )
+      end
+
+      unless organisations_match?
+        problems << problem_description(
+          "organisations don't match",
+          publishing_api_link("organisations"),
+          proposed_payload.dig("links", "organisations"),
+        )
+      end
 
       problems
+    end
+
+    def problem_description(message, expected, actual)
+      "#{message}, expected: #{expected.inspect}, actual: #{actual.inspect}"
     end
 
     def primary_publishing_organisation_matches?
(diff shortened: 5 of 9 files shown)


db/schema.rb
```diff
@@ -358,6 +358,8 @@
     t.datetime "updated_at", null: false
     t.bigint "document_id"
     t.bigint "whitehall_migration_id"
+    t.text "integrity_check_problems"
```
kevindew: Can we do this as a postgres array of strings? https://guides.rubyonrails.org/active_record_postgresql.html#array

lib/whitehall_importer/integrity_checker.rb
```diff
@@ -27,7 +31,9 @@ def content_problems
          document_type
          schema_name).each do |attribute|
         if publishing_api_content[attribute] != proposed_payload[attribute]
-          problems << "#{attribute} doesn't match"
+          problems << "#{attribute} doesn't match, " \
+            "expected: #{publishing_api_content[attribute].inspect}, " \
+            "actual: #{proposed_payload[attribute].inspect}"
```
kevindew: Might want to put this into a method or class where we pass in message, expected and actual and that can convert this to a string.

lib/whitehall_importer/import.rb
```diff
@@ -84,7 +84,7 @@ def check_document_integrity(document)
     def check_edition_integrity(edition)
       integrity_checker = IntegrityChecker.new(edition)
 
-      raise AbortImportError, integrity_checker.problems.join(", ") unless integrity_checker.valid?
+      raise WhitehallImporter::IntegrityCheckError.new(edition.state, integrity_checker.problems, integrity_checker.proposed_payload) unless integrity_checker.valid?
```
kevindew: state doesn't feel the right thing to be passing in here as it can vary into scheduled, removed etc. Aren't we more interested in whether it's live or draft?

lib/whitehall_importer/integrity_check_error.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+module WhitehallImporter
+  class IntegrityCheckError < AbortImportError
+    attr_reader :problems, :payload
+
+    def initialize(state, problems, payload)
+      @problems = problems
+      @payload = payload
+      super("#{state} Integrity check failed")
```
kevindew: This might read slightly weird if it's like "draft Integrity check failed" may want to adjust capitalisation.

lib/whitehall_importer/integrity_checker.rb
leenagupte: Yeah. I had the same thought, but hadn't gotten around to it yet.

db/schema.rb
leenagupte: I was following the pattern used on the users table, where a text field is used for `permissions` and then it's serialised in the model, but I'm happy to change it.

lib/whitehall_importer/integrity_check_error.rb
leenagupte: Yeah, that was going to be my recommendation too if I didn't have time to finish the harder stuff today..

lib/whitehall_importer/import.rb
leenagupte: Yeah, it was another thing on my To do list.

leenagupte:
@kevindew Thanks for your initial comments. This PR is ready for a full review now.

db/migrate/20200109094012_add_integrity_check_problems_to_whitehall_document_migration.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+class AddIntegrityCheckProblemsToWhitehallDocumentMigration < ActiveRecord::Migration[6.0]
+  def change
+    change_table :whitehall_migration_document_imports, bulk: true do |t|
+      t.column :integrity_check_problems, :string, array: true
+      t.column :integrity_check_proposed_payload, :json, default: "{}"
```
kevindew: Should this be null: false if we're setting it as a default empty object? Similar thoughts on the problems column should that have a default of empty and null false

Minor but it seems unusual that these have t.column rather than t.json so if we edit these lines it'd be nice to make them more consistent with other migrations

lib/whitehall_importer/import.rb
```diff
@@ -84,7 +84,10 @@ def check_document_integrity(document)
     def check_edition_integrity(edition)
       integrity_checker = IntegrityChecker.new(edition)
 
-      raise AbortImportError, integrity_checker.problems.join(", ") unless integrity_checker.valid?
+      unless integrity_checker.valid?
+        state = edition.live? ? "live" : "draft"
+        raise WhitehallImporter::IntegrityCheckError.new(state, integrity_checker.problems, integrity_checker.proposed_payload)
```
kevindew: I think we can probably pass the integrity checker result into the `WhitehallImporter::IntegrityCheckError`. 

So something like: `raise WhitehallImporter::IntegrityCheckError.new, integrity_checker`

As it looks like we've got logic in the exception and before the exception.

lib/whitehall_importer/integrity_check_error.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+module WhitehallImporter
+  class IntegrityCheckError < AbortImportError
+    attr_reader :problems, :payload
+
+    def initialize(state, problems, payload)
```
kevindew: As per previous comment we can probably handle setting various variables in here based on an integrity_check itself rather than individual attributes

lib/whitehall_importer/integrity_check_error.rb
```diff
@@ -0,0 +1,13 @@
...
+      @problems = problems
+      @payload = payload
+      super("#{state.titleize} Integrity check failed")
```
kevindew: Minor, I feel like integrity doesn't need a capital letter.

spec/lib/whitehall_importer/integrity_checker_spec.rb
```diff
@@ -162,7 +162,7 @@
       )
 
       integrity_check = WhitehallImporter::IntegrityChecker.new(edition)
-      expect(integrity_check.problems).to include("organisations don't match")
+      expect(integrity_check.problems).to include(match("organisations don't match"))
     end
```
kevindew: No too sure of the answer, as we have lot of repetition across the tests - but it feels like at least once we should be testing the description message is exactly what we expect. Not sure if now is maybe the time to add a shared helper, what do you think?

spec/lib/whitehall_importer_spec.rb
```diff
@@ -119,6 +119,29 @@
...
+      end
+
+      it "marks the import as aborted and logs the error and the payload" do
+        WhitehallImporter.import(whitehall_migration_document_import)
+
+        expect(whitehall_migration_document_import).to be_import_aborted
+        expect(whitehall_migration_document_import.integrity_check_problems).to eq(problems)
+        expect(whitehall_migration_document_import.integrity_check_proposed_payload).to eq(payload)
+      end
+    end
```
kevindew: 👍

spec/lib/whitehall_importer/integrity_checker_spec.rb
emmabeynon: I've attempted to refactor this and reduce some repetition in a new commit https://github.com/alphagov/content-publisher/pull/1630/commits/484ba19c4b17adc7f7b9f5dc8450d477b9e76d2d.  Let me know what you think.

db/migrate/20200109094012_add_integrity_check_problems_to_whitehall_document_migration.rb
emmabeynon: I've changed these to `t.string` and `t.json` and set `null: false`

lib/whitehall_importer/import.rb
emmabeynon: Now passing the state and `IntegrityChecker` instance.  It felt a bit weird to access the edition's state through the IntegrityChecker so keeping that as is.

lib/whitehall_importer/integrity_check_error.rb
emmabeynon: Now passing the `IntegrityChecker` instance instead of separate problems and payload.

lib/whitehall_importer/integrity_check_error.rb
emmabeynon: Changed 👍

emmabeynon:
Thanks @kevindew you're comments should be addressed now.

lib/whitehall_importer/import.rb
kevindew: I don't think it's weird to do that since we have edition as a public method of IntegrityChecker - I think it's a bit weirder having the bits of logic on the calling side and in the initializer of the exception. We should probably also not use the word state as it means a different thing to what we normally mean by edition state.

spec/support/integrity_checker_helper.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+
+module IntegrityCheckerHelper
+  def problem_message(attribute, expected, actual, plural = false)
+    negative_do = plural ? "don't" : "doesn't"
+    "#{attribute} #{negative_do} match, expected: #{expected.inspect}, actual: #{actual.inspect}"
+  end
```
kevindew: This looks good - though I would move this function into the test rather than create a specific helper  file for it (this'll keep it nearer to what it tests).

I also think we have only one instance of a plural (organisations) so I'd suggest just testing that manually rather than having this helper cope with that.

I think the thing we risk a bit is that we're writing the same logic in our test as in our application code and never test that. Since the logic is pretty trivial it doesn't seem a big concern though.

lib/whitehall_importer/import.rb
emmabeynon: I've changed it to `publishing_status` - is that any better?

spec/support/integrity_checker_helper.rb
emmabeynon: I've moved it into the test file.

PR #1632, opened 2020-01-09
Add defaults for 'topics' feature on formats

benthorner:
https://trello.com/c/NLAjn71N/1101-spike-and-test-approach-for-format-field-definitions

Previously we had to manually specify that topics were supported on
every format, which is unnecesary since they always are. This makes
having topics the default for a format, but inverts this for (most)
tests, in order to avoid having extra API stubs.

It seems reasonable to keep the 'topics' flag, since it represents a
'unit feature' of a format, even though it's technically mandatory.

### DIFF
--- a/app/models/document_type.rb
+++ b/app/models/document_type.rb
@@ -21,6 +21,7 @@ class DocumentType
         hash["tags"] = hash["tags"].to_a.map(&TagField.method(:new))
         hash["publishing_metadata"] = PublishingMetadata.new(hash["publishing_metadata"].to_h)
         hash["guidance"] = hash["guidance"].to_a.map(&Guidance.method(:new))
+        hash["topics"] = true # this feature is only disabled in tests
         new(hash)
       end
     end
--- a/spec/factories/document_type_factory.rb
+++ b/spec/factories/document_type_factory.rb
@@ -14,6 +14,7 @@ FactoryBot.define do
     contents { [] }
     tags { [] }
     guidance { [] }
+    topics { false }
 
     publishing_metadata do
       DocumentType::PublishingMetadata.new(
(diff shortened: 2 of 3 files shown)


spec/factories/document_type_factory.rb
```diff
@@ -14,6 +14,7 @@
     contents { [] }
     tags { [] }
     guidance { [] }
+    topics { false }
```
leenagupte: Shouldn't this also default to `true`?

benthorner: It needs to be false. I mentioned in the commit that we'll need loads of API stubs otherwise, because all the tests will start running with topic API calls enabled.

app/models/document_type.rb
```diff
@@ -21,6 +21,7 @@ def self.all
         hash["tags"] = hash["tags"].to_a.map(&TagField.method(:new))
         hash["publishing_metadata"] = PublishingMetadata.new(hash["publishing_metadata"].to_h)
         hash["guidance"] = hash["guidance"].to_a.map(&Guidance.method(:new))
+        hash["topics"] ||= true
```
kevindew: I'd have thought this ends up being set to true when ever it's false. 

Consider:

```
irb(main):004:0> hash = { topics: false }
=> {:topics=>false}
irb(main):005:0> hash[:topics] ||= true
=> true
irb(main):006:0> hash[:topics]
=> true
```
I may be missing something as I assume you'd have test failures?

Doing `hash[:topics] = hash.fetch(:topics, true)` would sort it though

benthorner: That confused me for a moment, too. It's because the tests all use the `document_type` factory, which initializes the model directly. But you're right the `||=` is redundant. I've updated to be more explicit - if it's still confusing I could add a comment to explain.

kevindew: Sure, yeah I think a comment would be useful as it's intentions will be somewhat cryptic.

benthorner: 👍- updated.

PR #1642, opened 2020-01-14
Fix hints not showing for document types

benthorner:


benthorner:
@alex-ju I was wondering about that. I only checked whether or not they do the no-visited-state behaviour, but then reasoned they don't need to, since they're not action links.

kevindew:
Nice catch - not a problem for this PR but is this something a regression we can cover somehow? Things that come to mind are either having InitializeWithHash blow up on unexpected attributes (which would cover the naming difference) and probably less effective might be a view test (as I guess we might have had a passing one of those if the input data was set as `hint` rather than `hint_govspeak`).

ChrisBAshton:
I had this same thought @kevindew - it should be tightened up in #1659 when we move to using `I18n.t!` for retrieving the hint. (If the translation doesn't exist, it will raise an exception).

PR #1643, opened 2020-01-14
Update history entry styles

alex-ju:
Update styling for history timeline:
 - highlight entries with comments
 - add a separator between editions to visually group them

<table>
<tr><td valign="top">

document-history-old

</td><td valign="top">

document-history-new

</td></tr>
</table>

[Trello card](https://trello.com/c/luEFJwLl)


app/views/documents/history.html.erb
```diff
@@ -8,21 +8,25 @@
...
+
+      <% if displayed_edition_number != entry.edition.number %>
+        <% if index > 0 %>
+          <hr class="govuk-section-break govuk-section-break--visible govuk-section-break--xl">
         <% end %>
+        <h3 class="govuk-heading-m"><%= t "documents.history.edition_header", number: entry.edition.number.ordinalize %></h3>
+        <% displayed_edition_number = entry.edition.number %>
+      <% end %>
 
+      <div class="app-timeline-entry <%= "app-timeline-entry--highlighted" if entry.details %>">
```
kevindew: I feel like it's a bit presumptuous for us to rely on the presence of a details object as the indication that we want to show a highlight. For instance that convention is just about to be broken in https://github.com/alphagov/content-publisher/pull/1644

I think it would be safer to have an array of highlighted_entry_types and then we can ask `<%= "app-timeline-entry--highlighted" if highlighted_entry_types.include?(entry.entry_type) %>">`

app/assets/stylesheets/objects/_history.scss
```diff
@@ -1,5 +1,15 @@
...
+  margin-bottom: govuk-spacing(6);
+
+  .govuk-heading-s {
+    margin-bottom: govuk-spacing(1);
+  }
+}
+
+.app-timeline-entry--highlighted {
+  background: govuk-colour("light-grey");
+  padding: govuk-spacing(3);
```
kevindew: 👍

spec/features/history/internal_note_spec.rb
```diff
@@ -25,8 +25,10 @@ def and_i_fill_in_and_submit_a_internal_note
   end
 
   def then_i_should_see_the_internal_note
-    expect(page).to have_content(I18n.t!("documents.history.entry_types.internal_note"))
-                .and have_content(@edition.timeline_entries.last.details.body)
+    within(".app-timeline-entry.app-timeline-entry--highlighted") do
+      expect(page).to have_content(I18n.t!("documents.history.entry_types.internal_note"))
+                  .and have_content(@edition.timeline_entries.last.details.body)
+    end
```
kevindew: I don't think we should use the feature tests to assert the highlighting. Our convention is to use the within option of feature tests just to distinguish particular content from others. I think using the within in this way is somewhat cryptic that it is actually a test. For example if it fails a developer may think "oh the within is just on the wrong selector".

I think if we actually want to test this logic we'd be better placed doing this by testing the particular view. In that case we can then that given this particular input we expect this particular selector. Bruce is working in a similar area in https://github.com/alphagov/content-publisher/pull/1644 by the way.

app/views/documents/history.html.erb
alex-ju: I might have misunderstood the card description here:

> Notes or history events where a user entered additional content, for example: Internal notes, Withdrawn, Backdated, removed. (this is grey by a different background as shown in the prototype)

I read 'entered additional content' as regardless of the entry type we only highlight if there is content. Happy to swap to an type-based approach if that's what we want.

spec/features/history/internal_note_spec.rb
alex-ju: I'm still trying to get the grasp of what's the best way to test the presence of a class on a specific element with capybara then where to place this test functional/view, or both. Will catch-up with Bruce.

app/views/documents/history.html.erb
kevindew: Yeah I think we should as we can't guarantee a details object being present means that we'll show the user extra information.

spec/features/history/internal_note_spec.rb
kevindew: Does this doc help at all: https://github.com/alphagov/content-publisher/blob/master/docs/testing-strategy.md ? (I really hope it does 🙏 )

alex-ju: That helped, thank you. I rebased this on #1644 since there are some changes in the way views are being structured coming up. #1644 introduces views tests for timeline entries and whitehall imported entries, but there's still no test for document history overall – which is where this assertion would need to be. I'm happy to create it, but I will need some support in creating a mock document with timeline entries for this.

app/views/documents/history.html.erb
alex-ju: I doubled check this with Simon and the design intention is to highlight the entries with content, not based on type. If I understand well checking for `entry.details` could be flaky, is there another way to check if a timeline entry has user-generated comments on it?

alex-ju: Since #1644 got merged I rebased this to master and updates the highlighting logic as discussed on the offline review. It also covers the view test discussed in the comment below.

kevindew:
Thanks Alex - that capture looks pretty nice and makes the class conditional pretty simple.

I did find it entered a slightly strange scenario where we end up with that only applying to one of the 3 conditionals we have in timeline entry. I did some work trying to fix this in https://github.com/alphagov/content-publisher/commit/1df6b871fae04d0c98e384bf07573ca40ef5f1e7 which I think is an improvement, what do you think?

I'm not totally happy with the work there though because of duplication between _whitehall_entry and _content_publisher_entry. This seems to suggest that we'd benefit from a component (as you suggested the other day). Maybe we want one for a timeline_entry so that we can call it like:

```
<%= render "govuk_publishing_components/components/timeline_entry", {
  title: t("documents.history.entry_types.whitehall_migration.#{entry.details.entry_type}",
  created_at: entry.created_at,
  created_by: entry.created_by
} do %>
  <%# extra content? or should this be injected as a variable #%>
<% end %>
```

The only oddity we seem to have if we are to introduce a component is that we have the Whitehall notice timeline entry https://github.com/alphagov/content-publisher/blob/e61052d1c8615c11ef7898725b7a6df79d786b5e/app/views/documents/_timeline_entry.html.erb#L2-L5 that sits inside the app-timeline-entry class but wouldn't have anything else in common with the component.

kevindew:
@alex-ju I've fixed the problems now and rebased. Is this good to be reviewed and merged in?

alex-ju:
> @alex-ju I've fixed the problems now and rebased. Is this good to be reviewed and merged in?

Thanks! Yes, it is ready for a second review.

app/views/documents/history/_content_publisher_entry.html.erb
```diff
@@ -38,38 +38,20 @@
...
-      <% end %>
-    </div>
-  <% else %>
-    <h4 class="govuk-heading-s">
-      <%= t "documents.history.entry_types.#{entry.entry_type}" %>
-    </h4>
+  <div class="app-timeline-entry__dateline">
+    <%= format_date_at_time(entry.created_at) %>
+    <% if entry.created_by %>
+    <%= t "documents.history.by"%> <%= entry.created_by.name %>
```
ChrisBAshton: Could you indent this?

app/views/documents/history/_whitehall_entry.html.erb
```diff
@@ -0,0 +1,19 @@
+<div class="app-timeline-entry">
+  <% if entry.details.imported_from_whitehall? %>
+    <%= render "govuk_publishing_components/components/notice", {
+      title: "#{t "documents.history.entry_types.whitehall_migration.#{entry.details.entry_type}"}",
```
ChrisBAshton: This is a little fiddly to read. Can I suggest...?

```suggestion
      title: "#{t("documents.history.entry_types.whitehall_migration.#{entry.details.entry_type}")}",
```

app/views/documents/history/_whitehall_entry.html.erb
```diff
@@ -0,0 +1,19 @@
...
+      description_text: format_date(entry.created_at)
+    } %>
+  <% else %>
+    <h4 class="govuk-heading-s">
+      <%= t "documents.history.entry_types.whitehall_migration.#{entry.details.entry_type}" %>
```
ChrisBAshton: ...and here?

app/views/documents/history/_whitehall_entry.html.erb
```diff
@@ -0,0 +1,19 @@
...
+    </h4>
+
+    <div class="app-timeline-entry__dateline">
+      <%= format_date_at_time(entry.created_at) %>
+      <% if entry.created_by %>
+      <%= t "documents.history.by" %> <%= entry.created_by.name %>
```
ChrisBAshton: And indent here?

app/views/documents/history/_content_publisher_entry.html.erb
kevindew: Sure, done.

app/views/documents/history/_whitehall_entry.html.erb
kevindew: Done - although I did it a slightly different way.

app/views/documents/history/_whitehall_entry.html.erb
kevindew: I didn't use the braces here, as I understood your suggestion, as this would make this `t` call inconsistent with our regular approach of using no braces.

app/views/documents/history/_whitehall_entry.html.erb
kevindew: Done

PR #1644, opened 2020-01-14
Allow Whitehall timeline entries to be stored separately from Content Publisher timeline entries

brucebolt:
Creates a way of storing imported timeline history from Whitehall that keeps it separate from history generated by Content Publisher.

* A new model has been created (`WhitehallMigration::TimelineEntry`) for storing timeline entries from Whitehall; this contains an `entry_type` (a key to map this to a translation) and timestamps (body and any other data can be added in later stories)
* A new type of timeline entry named `whitehall_migration` which maps to a `WhitehallMigration::TimelineEntry` has been added
* `imported_from_whitehall` timeline entries have been moved to `WhitehallMigration::TimelineEntry` (and import code updated)
* Moved timeline history entries into a separate partial, which has made it easier to differentiate the code for Content Publisher and Whitehall history
* The `WhitehallMigration::TimelineEntry` entries are shown in the document history tab
* The timeline labels for Whitehall (https://docs.google.com/spreadsheets/d/1I7TbA9glyFyhyWq5wQJAGykkXFwOJjyd--kLCycXbSY/edit#gid=0) have been added to translations

Trello card: https://trello.com/c/ITUjv2Pl


config/locales/en/documents/history.yml
```diff
@@ -38,6 +38,21 @@ en:
         access_limit_updated: "Access limit updated"
         access_limit_removed: "Access limit removed"
         political_status_changed: "History mode criteria updated"
-        imported_from_whitehall: "Imported from Whitehall"
+        whitehall_migration:
```
brucebolt: Note to self: sort these alphabetically

db/migrate/20200114125900_update_imported_from_whitehall.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+class UpdateImportedFromWhitehall < ActiveRecord::Migration[6.0]
+  def change
+    TimelineEntry.where(entry_type: :imported_from_whitehall).each do |timeline_entry|
```
brucebolt: Note to self:
```suggestion
    TimelineEntry.imported_from_whitehall.each do |timeline_entry|
```

app/models/whitehall_migration/timeline_entry.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+
+class WhitehallMigration::TimelineEntry < ApplicationRecord
```
kevindew: I think this class would make more semantic sense to live under the TimelineEntry namespace as despite being a product of a WhitehallMigration it has very little actual association with it and is likely to exist forever. Whereas at some point we'd hope we could kill off all the WhitehallMigration classes.

I'd think something like TimelineEntry::WhitehallImportedEntry or TimelineEntry::WhitehallEntry would be a good option.

app/models/whitehall_migration/timeline_entry.rb
```diff
@@ -0,0 +1,20 @@
...
+    first_created: "first_created",
+    imported_from_whitehall: "imported_from_whitehall",
+    internal_note: "internal_note",
+    new_edition: "new_edition",
+    published: "published",
+    rejected: "rejected",
+    removed: "removed",
+    scheduled: "scheduled",
+    submitted: "submitted",
+    withdrawn: "withdrawn",
```
kevindew: These look good - this list is shorter than I envisioned.

db/migrate/20200114095428_create_whitehall_migration_timeline_entries.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+class CreateWhitehallMigrationTimelineEntries < ActiveRecord::Migration[6.0]
+  def change
+    create_table :whitehall_migration_timeline_entries do |t|
+      t.string "entry_type"
+      t.text "body"
```
kevindew: I don't think we should add this body column in straight away as I'm a bit worried this will solve our problem for notes but then leave us wondering what to do for fact check which will have richer data (I think I suggested in the card that we could have a JSON column). So I'd suggest maybe the limit of this PR is to get the types in and then think about how the different extra data will be represented.

db/migrate/20200114095428_create_whitehall_migration_timeline_entries.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+class CreateWhitehallMigrationTimelineEntries < ActiveRecord::Migration[6.0]
+  def change
+    create_table :whitehall_migration_timeline_entries do |t|
+      t.string "entry_type"
```
kevindew: should this be null false?

app/models/whitehall_migration/timeline_entry.rb
```diff
@@ -0,0 +1,20 @@
...
+  }
```
kevindew: We should flag this class as read_only after creation since we wouldn't expect anyone to have reason to edit it under the normal application flow: 

https://github.com/alphagov/content-publisher/blob/master/app/models/revision.rb#L67-L69

db/migrate/20200114095428_create_whitehall_migration_timeline_entries.rb
```diff
@@ -0,0 +1,11 @@
...
+      t.text "body"
+      t.timestamps
```
kevindew: I think as we'd not expect this record to ever be updated that we'd just need a created_at here, see https://github.com/alphagov/content-publisher/blob/master/db/migrate/20190626155821_create_access_limits.rb#L6 for past approach

db/migrate/20200114125900_update_imported_from_whitehall.rb
```diff
@@ -0,0 +1,16 @@
...
+      migration_timeline_entry = WhitehallMigration::TimelineEntry.create!(
+        entry_type: :imported_from_whitehall,
+      )
+      timeline_entry.update!(
+        entry_type: :whitehall_migration,
+        details_type: "WhitehallMigration::TimelineEntry",
+        details_id: migration_timeline_entry.id,
+      )
+    end
+  end
```
kevindew: Nice work doing this, however I don't think we're going to actually need this migration since we've never actually migrated something in production. 

I imagine given data syncs we're probably not going to need this for integration so it'd probably only benefit developers and given there'll be only a couple at most we can probably just tell them what to do.

db/migrate/20200114125900_update_imported_from_whitehall.rb
kevindew: Somewhat moot given my last comment but you'd not want to do that as that change wouldn't work once https://github.com/alphagov/content-publisher/pull/1644/files#diff-73d3559ae89c6d594ff5fb797014be76L61 is applied.

app/views/documents/history.html.erb
```diff
@@ -33,7 +38,7 @@
             <% end %>
           </div>
 
-          <% if entry.internal_note? && entry.details %>
+          <% if entry.internal_note? && entry.details || entry.whitehall_migration? && entry.details.body %>
             <%= simple_format(escape_and_link(entry.details.body)) %>
           <% end %>
```
kevindew: Since this file is starting to get a bit hairy on logic I wonder if we want to have a partial that we pull in for rendering whitehall migration entries? - that could keep them somewhat distinct which will come with some pros of encapsulation and some cons for re-use (as we've done with internal note)

spec/views/documents/history.html.erb_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+             details_type: "WhitehallMigration::TimelineEntry",
+             details_id: whitehall_migration_timeline_entry.id)
+      assign(:edition, edition)
+      assign(:timeline_entries, edition.timeline_entries.page(1).per(50))
+      expect(render)
+        .to include(I18n.t("documents.history.entry_types.whitehall_migration.new_edition"))
+        .and include(whitehall_migration_timeline_entry.body)
+    end
+  end
+end
```
kevindew: I think if you had a separate view this might be easier to test as you'd have to do less DB writes to set-up creation and could focus on an individual entry.

app/models/whitehall_migration/timeline_entry.rb
brucebolt: Good point about `WhitehallMigration` eventually going away, so I've decided on `TimelineEntry::WhitehallImportedEntry` (having gone through many iterations of different names).

db/migrate/20200114095428_create_whitehall_migration_timeline_entries.rb
brucebolt: I've removed the `body` column.  We can introduce whatever we decide on in the later stories (fact check and editorial remarks).

db/migrate/20200114095428_create_whitehall_migration_timeline_entries.rb
brucebolt: Yes, it should.  Oversight not including it.

db/migrate/20200114095428_create_whitehall_migration_timeline_entries.rb
brucebolt: Done

db/migrate/20200114125900_update_imported_from_whitehall.rb
brucebolt: Very good point!  I was stuck in the world of my local environment where there are hundreds of `WhitehallMigration::DocumentImport` records.

I've removed this migration completely.

app/views/documents/history.html.erb
brucebolt: I've moved the standard timeline entry and Whitehall timeline entries into separate partials.  The code looks much cleaner and has become easier to test.

app/models/whitehall_migration/timeline_entry.rb
brucebolt: Done

app/views/documents/history.html.erb
brucebolt: I've moved the standard timeline entry and Whitehall timeline entries into separate partials. The code looks much cleaner and has become easier to test.

(cc @alex-ju)

brucebolt:
Thank you @kevindew for taking an early look at this.  I have addressed your comments and tidied the commits, so I think it's ready for full review now.

app/views/documents/history.html.erb
alex-ju: Thanks for the heads up, Bruce. I'll probably wait with #1643 to get this one merged anyway – as I didn't manage to work on it at all today.

app/views/documents/history.html.erb
```diff
@@ -15,61 +15,15 @@
           <% displayed_edition_number = entry.edition.number %>
         <% end %>
 
-        <% if entry.imported_from_whitehall? %>
+        <% if entry.whitehall_migration? && entry.details.imported_from_whitehall? %>
           <%= render "govuk_publishing_components/components/notice", {
-            title: "#{t "documents.history.entry_types.#{entry.entry_type}"}",
+            title: "#{t "documents.history.entry_types.whitehall_migration.#{entry.details.entry_type}"}",
             description_text: format_date(entry.created_at)
           } %>
```
kevindew: I think it'd make sense that this entry was in the whitehall_migration partial, that way all whitehall imported entries are in the same file.

app/views/documents/_whitehall_imported_entry.html.erb
```diff
@@ -0,0 +1,10 @@
+<h4 class="govuk-heading-s govuk-!-margin-bottom-1">
+  <%= t "documents.history.entry_types.whitehall_migration.#{entry.details.entry_type}" %>
+</h4>
+
+<div class="app-timeline-entry__dateline">
+  <%= format_date_at_time(entry.created_at) %>
+  <% if entry.created_by %>
+    by <%= name_or_fallback(entry.created_by) %>
```
kevindew: Do we need a fallback if we're checking if the user exists?

kevindew: I think convention would dictate that we'd want the by in a translation file - sorry I know that's a bit of a pain for 2 letters.

app/views/documents/history.html.erb
```diff
@@ -15,61 +15,15 @@
...
-                <p>
-                  Alternative path
-                  <%= link_to(Plek.new.website_root + removal.alternative_path,
-                              nil,
-                              class: "govuk-link") %>
-                </p>
-              <% end %>
-            <% end %>
-          <% end %>
+          <%= render partial: "documents/timeline_entry", locals: {entry: entry} %>
```
kevindew: I think we hit a bit of a naming conundrum here where this partial name would imply that this can render all timeline entries and not just non-Whitehall ones. I think we should try resolve this before merging so we don't risk it festering as an area of communication.

Approaches we could use are:

1. Keep non-Whitehall timeline entries in this view file and only use a partial for them
2. Render everything in the _timeline_entry partial and inside that have a conditional that checks for whitehall_migration
3. Something I've not thought of

of those I find 2 most appealing and would work with the approach you've taken for tests.

lib/whitehall_importer.rb
```diff
@@ -82,10 +82,14 @@ def self.sync(whitehall_import)
   end
 
   def self.create_timeline_entry(edition)
-    TimelineEntry.create_for_revision(
+    whitehall_migration_timeline_entry = TimelineEntry::WhitehallImportedEntry.create!(
```
kevindew: minor: we could probably could this something more succinct like `details` given how it's used rather than the long variable name.

lib/whitehall_importer.rb
```diff
@@ -82,10 +82,14 @@ def self.sync(whitehall_import)
   end
 
   def self.create_timeline_entry(edition)
```
kevindew: Sorry I know you didn't create this, just wondering if you could scratch an itch of mine here.

I think this method is in the wrong place here - It doesn't really seem a concern of the overall WhitehallImporter module and we'd want to abort an import if it failed (as unlikely as that may be) so I think it belongs in the WhitehallImporter::Import class within the transaction.

Any chance you could sort that as part of this 🙏 ?

app/views/documents/_whitehall_imported_entry.html.erb
Rosa-Fox: There should always be a creator given [this](https://github.com/alphagov/content-publisher/blob/master/app/controllers/documents_controller.rb#L28). I guess it just won't display anything if there was no value for `entry.created_by`, seems ok to me but I suppose we could display some text instead?

Rosa-Fox: Will change the `by` to be localised

app/views/documents/history.html.erb
Rosa-Fox: I went with option 2...

lib/whitehall_importer.rb
Rosa-Fox: Have moved!

Rosa-Fox:
@kevindew I addressed your comments in individual commits. We can crush them down after. Let me know if you require any further changes. There is a small rebase that I will do now.

app/views/documents/_timeline_entry.html.erb
```diff
@@ -0,0 +1,61 @@
+<% if entry.whitehall_migration? %>
+  <h4 class="govuk-heading-s govuk-!-margin-bottom-1">
+    <%= t "documents.history.entry_types.whitehall_migration.#{entry.details.entry_type}" %>
+  </h4>
+
+  <div class="app-timeline-entry__dateline">
+    <%= format_date_at_time(entry.created_at) %>
+    <% if entry.created_by %>
+    <%= t "documents.history.by" %> <%= name_or_fallback(entry.created_by)%>
```
kevindew: We can probably just do `entry.created_by.name` rather than needing the name_or_fallback helper as we know it's not going to be a fallback.

app/views/documents/history.html.erb
```diff
@@ -15,61 +15,13 @@
           <% displayed_edition_number = entry.edition.number %>
         <% end %>
 
-        <% if entry.imported_from_whitehall? %>
+        <% if entry.whitehall_migration? && entry.details.imported_from_whitehall? %>
           <%= render "govuk_publishing_components/components/notice", {
-            title: "#{t "documents.history.entry_types.#{entry.entry_type}"}",
+            title: "#{t "documents.history.entry_types.whitehall_migration.#{entry.details.entry_type}"}",
             description_text: format_date(entry.created_at)
           } %>
```
kevindew: We should move this entry into the _timeline_entry so that all timeline entries are rendered by the same partial.

app/views/documents/_timeline_entry.html.erb
```diff
@@ -0,0 +1,67 @@
+<% if entry.whitehall_migration? && entry.details.imported_from_whitehall? %>
+  <%= render "govuk_publishing_components/components/notice", {
+    title: "#{t "documents.history.entry_types.whitehall_migration.#{entry.details.entry_type}"}",
+    description_text: format_date(entry.created_at)
+  } %>
+<% end %>
+<% if entry.whitehall_migration? %>
```
kevindew: should this be a `elsif` ?

app/views/documents/history.html.erb
```diff
@@ -14,63 +14,7 @@
...
-                  Alternative path
-                  <%= link_to(Plek.new.website_root + removal.alternative_path,
-                              nil,
-                              class: "govuk-link") %>
-                </p>
-              <% end %>
-            <% end %>
-          <% end %>
-        <% end %>
+        <%= render partial: "documents/timeline_entry", locals: {entry: entry} %>
```
kevindew: Minor, convention would be to have spaces within the braces. For example: `{ entry: entry }`

spec/views/documents/_timeline_entry.html.erb_spec.rb
```diff
@@ -0,0 +1,42 @@
+# frozen_string_literal: true
+
+RSpec.describe "documents/_timeline_entry.html.erb" do
+  describe "timeline entry" do
```
kevindew: I don't think this section needs this describe as all things are timeline entries. I'd skip having a describe here and just have its within the main describe.

spec/views/documents/_timeline_entry.html.erb_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+    it "shows a timeline entry without an author" do
+      timeline_entry = create(:timeline_entry,
+                              created_by: nil)
+      render partial: "documents/timeline_entry", locals: { entry: timeline_entry }
+      expect(rendered).not_to have_content(I18n.t!("documents.history.by"))
+      expect(rendered).to have_content(I18n.t!("documents.history.entry_types.created"))
+    end
+
+    it "shows a Whitehall timeline entry with an author" do
```
kevindew: Looks like a copy and paste issue here as this isn't a Whitehall timeline entry.

Rosa-Fox:
Thanks @kevindew. Have fixed :)

PR #1645, opened 2020-01-14
Remove the start_time column from Whitehall Migrations

JonathanHallam:
It was surplus to requirements and also doesn't fit in with rails convention. We'll be using the `created_at` column from here on out. This also changes any references to the `start_time` column.

Related - https://github.com/alphagov/content-publisher/pull/1615

### DIFF
--- a/lib/whitehall_importer.rb
+++ b/lib/whitehall_importer.rb
@@ -6,8 +6,7 @@ module WhitehallImporter
   def self.create_migration(organisation_content_id, document_type)
     whitehall_migration = ActiveRecord::Base.transaction do
       record = WhitehallMigration.create!(organisation_content_id: organisation_content_id,
-                                 document_type: document_type,
-                                 start_time: Time.current)
+                                          document_type: document_type)
 
       whitehall_export = GdsApi.whitehall_export.document_list(organisation_content_id, document_type)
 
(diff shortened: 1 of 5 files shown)


lib/whitehall_importer.rb
```diff
@@ -6,8 +6,7 @@ module WhitehallImporter
   def self.create_migration(organisation_content_id, document_type)
     whitehall_migration = ActiveRecord::Base.transaction do
       record = WhitehallMigration.create!(organisation_content_id: organisation_content_id,
-                                 document_type: document_type,
-                                 start_time: Time.current)
+                                 document_type: document_type)
```
1pretz1: I appreciate that you haven't formatted these lines but I find its a little jarring to read, could the indentation be changed to something like: ?
```
WhitehallMigration.create!(organisation_content_id: organisation_content_id,
                           document_type: document_type)
```

JonathanHallam: yeah sure thing. I've adjusted it

JonathanHallam:
@kevindew Yeah I'll get end time done now :)

brucebolt:
If we removed end_time, how will we know when an organisation migration has been completed?

kevindew:
> If we remove end_time, how will we know when an organisation migration has been completed?

We shouldn't remove it but we should rename it to be consistent with ActiveRecord time naming. I'd suggest going with `completed_at`

PR #1647, opened 2020-01-14
Rename the Whitehall Migrations end_time column

JonathanHallam:
This makes it more in line with rails customs. It also renames the references to the column.

### DIFF
--- a/app/models/whitehall_migration.rb
+++ b/app/models/whitehall_migration.rb
@@ -4,6 +4,6 @@ class WhitehallMigration < ApplicationRecord
   has_many :document_imports
 
   def check_migration_finished
-    update!(end_time: Time.current) if document_imports.in_progress.empty?
+    update!(finished_at: Time.current) if document_imports.in_progress.empty?
   end
 end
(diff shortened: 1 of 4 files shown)


app/models/whitehall_migration.rb
```diff
@@ -4,6 +4,6 @@ class WhitehallMigration < ApplicationRecord
   has_many :document_imports
 
   def check_migration_finished
```
kevindew: Sorry minor nitpick - but it'd be consistent if we used completed in this method name as well instead of finished so same word is used for the same thing.

JonathanHallam: So this has reminded me of something, we may not want to merge this PR at all. The reason that we named this method `check_migration_finished` rather than `check_migration_completed` was because `completed` is an actual state. This method allows a few other states that imply that the migration has finished running and isn't just checking to see if something is `completed`.

Thinking we may want to change this pr to name the column something like `finished_at` to avoid confusion with the `completed` state.

kevindew: Ah yes - good thinking. If I'm right we don't actually have a state field on a migration so it wouldn't conflict with terminology there but if finished_at avoids any meaning problems then I think that's a better shout.

PR #1651, opened 2020-01-15
Add page info to pagination links

1pretz1:
The page information was missed on the first pass of implementing the pagination
feature, I believe because the prototype was out of date. Now the
pagination links display information around what page is currently being
displayed.

This comes after a product review from @Tobiogunsina.

## Before
Screenshot 2020-01-15 at 11 54 04
Screenshot 2020-01-15 at 11 54 10

## After 
Screenshot 2020-01-15 at 11 42 13
Screenshot 2020-01-15 at 13 08 28

Trello:
https://trello.com/c/TsRTLbiW/1276-move-document-history-from-tab-to-secondary-navigation-and-add-pagination

### DIFF
--- a/app/views/documents/history.html.erb
+++ b/app/views/documents/history.html.erb
@@ -75,11 +75,21 @@
     <% end %>
 
     <%
-    pages = {}
-    pages[:previous_page] = { url: document_history_path(page: @timeline_entries.prev_page),
-                              title: "Previous page" } if @timeline_entries.prev_page
-    pages[:next_page] = { url: document_history_path(page: @timeline_entries.next_page),
-                          title: "Next page" } if @timeline_entries.next_page
+      pages = {}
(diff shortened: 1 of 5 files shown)


kevindew:
Have only quickly scanned this PR so haven't seen the code. But the screenshot in after looks wrong to have both page 2 as the previous and next page.

1pretz1:
That's how the index page currently displays the page information @kevindew, shall I change them both to display the actual previous and next page, so for example 1 (previous page) and 3 (next page) for page two ?

kevindew:
> That's how the index page currently displays the page information @kevindew, shall I change them both to display the actual previous and next page, so for example 1 (previous page) and 3 (next page) for page two ?

Yup that might well be the bug @JonathanHallam found today - so worth checking if he's also fixing it.

1pretz1:
Updated @kevindew

PR #1653, opened 2020-01-15
Rename services to match a verb/noun pattern

benthorner:
https://github.com/alphagov/content-publisher/issues/1256

https://trello.com/c/f1QaCH7x/231-our-services-directory-is-a-mess

This tries to bring some closure to our work on the code in the `app/services` 
directory. In many cases, I think the new names are more precise than 
the old, shorter ones. It's meant to be a set of open suggestions, though, 
so feel free to suggest alternatives.

From the commit:

    This renames a few services to follow a pattern where the single action
    a service performs is described in terms of a verb, like 'create' or
    'publish', operating on a noun, like 'filename' or 'scheduled
    publishing'. In the case of 'edition' vs 'draft', I've gone with
    'draft', which is less ambiguous than an arbitrary edition.

### DIFF
--- a/app/controllers/documents_controller.rb
+++ b/app/controllers/documents_controller.rb
@@ -69,7 +69,7 @@ class DocumentsController < ApplicationController
   def generate_path
     edition = Edition.find_current(document: params[:document])
     assert_edition_state(edition, &:editable?)
-    base_path = PathGeneratorService.call(edition.document, params[:title])
+    base_path = GenerateBasePathService.call(edition.document, params[:title])
     render plain: base_path
   end
 
--- a/app/interactors/access_limit/update_interactor.rb
+++ b/app/interactors/access_limit/update_interactor.rb
@@ -31,7 +31,7 @@ private
 
     if LIMIT_TYPES.exclude?(limit_type)
       context.fail! if edition.access_limit.nil?
-      EditEditionService.call(edition, user, access_limit: nil)
+      EditDraftEditionService.call(edition, user, access_limit: nil)
       return
     end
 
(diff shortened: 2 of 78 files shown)


benthorner:
Immediately after writing this, I'm wondering if 'draft edition' would have been a better compromise!

kevindew:
This looks good Ben - thanks for doing this. I'd agree on your idea about Draft Edition. I feel like just having draft contrasts with the more precise nature of suggestions like ResyncDocumentService, and we know that draft is terminology also used in the context of assets.

This looks a nice improvement to the purpose of some of the classes, for example it's so much easier to know the role of "CreateFileAttachmentBlobService" compared to "FileAttachmentBlobService".

benthorner:
@kevindew thanks for taking a look! I've made the suggested change, but compromised on `FailsafeDraftPreviewService`, which is more speakable than `FailsafePreviewDraftEditionService`.

Also renamed `PublishAssetService` -> `PublishAsset`**s**`Service`, since that's what it does.

gclssvglx:
For me, the only one I'd question is: UniqueFilenameService to GenerateFilenameService.  It feels like we would loose a key bit of information - the uniqueness of the filename.  I know it's wordy, but can I suggest GenerateUniqueFilenameService?

benthorner:
@gclssvglx good question. I think you're right - updated. I've also renamed `GeneratePathService` -> `Generate`**Base**`PathService, as that was also ambiguous.

gclssvglx:
Thanks for doing this Ben - as a newbie, the renaming is definitely making these clearer in intent and easier to understand :-)

PR #1658, opened 2020-01-16
Use I18n for supertypes

JonathanHallam:
Trello - https://trello.com/c/hAJkXW1T/1325-use-i18n-to-manage-copy-for-each-supertype

This moves the copy from a yaml file to be internationalised. This is more in line with how i18n is meant to be used and will make configuration easier once we start iterating the supertype model.

### DIFF
--- a/app/views/new_document/choose_document_type.html.erb
+++ b/app/views/new_document/choose_document_type.html.erb
@@ -1,11 +1,11 @@
-<% content_for :browser_title, @supertype.label %>
+<% content_for :browser_title, t("supertypes.#{@supertype.id}.label") %>
 <% content_for :back_link, render_back_link(href: new_document_path) %>
 
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
     <%= form_tag create_document_path, data: { gtm: "confirm-document-type" } do %>
       <%= render "govuk_publishing_components/components/radio", {
-        heading: @supertype.label,
+        heading: t("supertypes.#{@supertype.id}.label"),
         is_page_heading: true,
         name: "document_type",
         error_items: @issues&.items_for(:document_type),
--- a/spec/models/supertype_spec.rb
+++ b/spec/models/supertype_spec.rb
@@ -6,8 +6,8 @@ RSpec.describe Supertype do
       describe "Supertype #{supertype.id}" do
         it "has the required attributes for #{supertype.id}" do
           expect(supertype.id).to_not be_blank
-          expect(supertype.label).to_not be_blank
-          expect(supertype.description).to_not be_blank
+          expect(I18n.t("supertypes.#{supertype.id}.label")).to_not be_blank
+          expect(I18n.t("supertypes.#{supertype.id}.description")).to_not be_blank
         end
       end
     end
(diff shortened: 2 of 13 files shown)


app/views/new_document/choose_document_type.html.erb
```diff
@@ -1,11 +1,11 @@
-<% content_for :browser_title, @supertype.label %>
+<% content_for :browser_title, I18n.t("supertypes.#{@supertype.id}.label") %>
```
benthorner: We can actually just do `t`, instead of `I18n.t` in the views.

spec/models/supertype_spec.rb
```diff
@@ -6,8 +6,6 @@
       describe "Supertype #{supertype.id}" do
         it "has the required attributes for #{supertype.id}" do
           expect(supertype.id).to_not be_blank
-          expect(supertype.label).to_not be_blank
```
benthorner: I think we're losing something here in that we expect the associated locale keys to exist. Perhaps we can test for that instead?

spec/requests/new_document_spec.rb
```diff
@@ -4,9 +4,10 @@
   describe "GET /documents/choose-document-type" do
     it "shows supertype when a supertype managed by content publisher is selected" do
       supertype = Supertype.all.reject(&:managed_elsewhere).first
-      get choose_document_type_path, params: { supertype: supertype.id }
+      supertype_id = supertype.id
```
benthorner: Minor: the character difference between `supertype_id` and `supertype.id` suggests this variable isn't necessary.

spec/requests/new_document_spec.rb
```diff
@@ -4,9 +4,10 @@
...
+      get choose_document_type_path, params: { supertype: supertype_id }
       expect(response).to have_http_status(:ok)
-      expect(response.body).to have_content(supertype.label)
+      expect(response.body).to have_content(I18n.t("supertypes.#{supertype_id}.label"))
```
benthorner: We should use `I18n.t!` in tests, as the body is quite likely to have an empty string, which is the `t` fallback.

app/views/new_document/choose_document_type.html.erb
JonathanHallam: nice one, thanks :)

spec/requests/new_document_spec.rb
JonathanHallam: good spot, not completely sure why I added this

spec/requests/new_document_spec.rb
JonathanHallam: thanks :)

spec/models/supertype_spec.rb
JonathanHallam: Have added this in :)

spec/models/supertype_spec.rb
```diff
@@ -6,8 +6,8 @@
...
-          expect(supertype.description).to_not be_blank
+          expect(I18n.t("supertypes.#{supertype.id}.label")).to_not be_blank
+          expect(I18n.t("supertypes.#{supertype.id}.description")).to_not be_blank
```
ChrisBAshton: Just a warning that `t` will silently fail with "translation missing: supertypes.foo.description", so even if the translation doesn't exist, it will never be blank 😬 might be worth changing all of these to `t!`.

PR #1659, opened 2020-01-16
Use I18n to manage copy for each document type

ChrisBAshton:
This PR rips out copy from `config/document_types.yml` and into a new `config/locales/en/document_types.yml`, so that we can internationalise Content Publisher in the future. `config/document_types.yml` is now solely about the configuration of each document type, which would apply to any language.

The [Trello card](https://trello.com/c/AuzTdQVt/1100-use-i18n-to-manage-copy-for-each-document-type) makes it quite clear that we want to avoid duplicating text/config between similar document types, and proposes a `document_types.default` namespace. The only duplication I've come across so far is in the tagging of the `news_story` and `press_release` document types, whose configurations remain under `config/document_types.yml`; I have used standard YAML inheritance to avoid duplicating the values there. It should be simple enough to use this approach inside the I18n files when the time comes; for now, creating a 'default' doctype copy feels like a premature optimisation.

### DIFF
--- a/app/helpers/document_type_helper.rb
+++ b/app/helpers/document_type_helper.rb
@@ -1,0 +1,20 @@
+# frozen_string_literal: true
+
+module DocumentTypeHelper
+  def t_doctype_exists?(i18n_key)
+    I18n.exists?(i18n_key) || I18n.exists?(doctype_default_key(i18n_key))
+  end
+
+  def t_doctype(i18n_key)
+    match = I18n.t!(I18n.exists?(i18n_key) ? i18n_key : doctype_default_key(i18n_key))
+    match.is_a?(String) ? match : match.stringify_keys
+  end
+
+private
+
+  def doctype_default_key(i18n_key)
+    parts = i18n_key.split(".")
+    parts[1] = "default"
+    parts.join(".")
+  end
+end
--- a/app/helpers/document_type_helper.rb
+++ b/app/helpers/document_type_helper.rb
@@ -1,15 +1,21 @@
 # frozen_string_literal: true
 
 module DocumentTypeHelper
-  def t_doctype_exists?(i18n_key)
+  def t_doctype_field?(edition, partial_i18n_key)
+    i18n_key = prefix_doctype_field(edition, partial_i18n_key)
     I18n.exists?(i18n_key) || I18n.exists?(doctype_default_key(i18n_key))
   end
 
-  def t_doctype(i18n_key)
+  def t_doctype_field(edition, partial_i18n_key)
+    i18n_key = prefix_doctype_field(edition, partial_i18n_key)
     I18n.t!(I18n.exists?(i18n_key) ? i18n_key : doctype_default_key(i18n_key))
   end
 
 private
+
+  def prefix_doctype_field(edition, partial_i18n_key)
+    "document_types.#{edition.document_type.id}.fields.#{partial_i18n_key}"
+  end
 
   def doctype_default_key(i18n_key)
     parts = i18n_key.split(".")
--- a/app/helpers/document_type_helper.rb
+++ b/app/helpers/document_type_helper.rb
@@ -6,8 +6,7 @@ module DocumentTypeHelper
   end
 
   def t_doctype(i18n_key)
-    match = I18n.t!(I18n.exists?(i18n_key) ? i18n_key : doctype_default_key(i18n_key))
-    match.is_a?(String) ? match : match.stringify_keys
+    I18n.t!(I18n.exists?(i18n_key) ? i18n_key : doctype_default_key(i18n_key))
   end
 
 private
--- a/app/models/document_type.rb
+++ b/app/models/document_type.rb
@@ -4,7 +4,7 @@ class DocumentType
   include InitializeWithHash
 
   attr_reader :contents, :id, :label, :managed_elsewhere, :publishing_metadata,
-              :path_prefix, :tags, :guidance_govspeak, :description, :images,
+              :path_prefix, :tags, :guidance_govspeak, :images,
               :topics, :check_path_conflict
 
   def self.find(id)
--- a/app/models/document_type.rb
+++ b/app/models/document_type.rb
@@ -4,7 +4,7 @@ class DocumentType
   include InitializeWithHash
 
   attr_reader :contents, :id, :label, :managed_elsewhere, :publishing_metadata,
-              :path_prefix, :tags, :guidance_govspeak, :description, :hint_govspeak, :images,
+              :path_prefix, :tags, :guidance_govspeak, :description, :images,
               :topics, :check_path_conflict
 
   def self.find(id)
--- a/app/models/document_type.rb
+++ b/app/models/document_type.rb
@@ -18,7 +18,13 @@ class DocumentType
 
       hashes.map do |hash|
         hash["contents"] = hash["contents"].to_a.map(&Field.method(:new))
-        hash["tags"] = hash["tags"].to_a.map(&TagField.method(:new))
+        hash["tags"] = (hash["tags"] || []).map do |tag_config|
+          tag_i18n_data = "document_types.#{hash['id']}.fields.#{tag_config['id']}"
+          if I18n.exists?(tag_i18n_data)
+            TagField.new(tag_config.merge(I18n.t!(tag_i18n_data).stringify_keys))
+          end
+        end
+        hash["tags"] = hash["tags"].compact
         hash["publishing_metadata"] = PublishingMetadata.new(hash["publishing_metadata"].to_h)
         hash["topics"] = true # this feature is only disabled in tests
         new(hash)
--- a/app/models/document_type.rb
+++ b/app/models/document_type.rb
@@ -1,6 +1,8 @@
 # frozen_string_literal: true
 
 class DocumentType
+  extend DocumentTypeHelper # for static methods
+  include DocumentTypeHelper
   include InitializeWithHash
 
   attr_reader :contents, :id, :managed_elsewhere, :publishing_metadata,
@@ -19,8 +21,8 @@ class DocumentType
         hash["contents"] = hash["contents"].to_a.map(&Field.method(:new))
         hash["tags"] = (hash["tags"] || []).map do |tag_config|
           tag_i18n_data = "document_types.#{hash['id']}.fields.#{tag_config['id']}"
-          if I18n.exists?(tag_i18n_data)
-            TagField.new(tag_config.merge(I18n.t!(tag_i18n_data).stringify_keys))
+          if t_doctype_exists?(tag_i18n_data)
+            TagField.new(tag_config.merge(t_doctype(tag_i18n_data)))
           end
         end
         hash["tags"] = hash["tags"].compact
@@ -36,7 +38,7 @@ class DocumentType
   end
 
   def label
-    I18n.t!("document_types.#{id}.label")
+    t_doctype("document_types.#{id}.label")
   end
 
   class TagField
--- a/app/models/document_type.rb
+++ b/app/models/document_type.rb
@@ -3,7 +3,7 @@
 class DocumentType
   include InitializeWithHash
 
-  attr_reader :contents, :id, :label, :managed_elsewhere, :publishing_metadata,
+  attr_reader :contents, :id, :managed_elsewhere, :publishing_metadata,
               :path_prefix, :tags, :guidance_govspeak, :images,
               :topics, :check_path_conflict
 
@@ -28,6 +28,10 @@ class DocumentType
 
   def managed_elsewhere_url
     Plek.new.external_url_for(managed_elsewhere.fetch("hostname")) + managed_elsewhere.fetch("path")
+  end
+
+  def label
+    I18n.t!("document_types.#{id}.label")
   end
 
   class TagField
--- a/app/models/document_type.rb
+++ b/app/models/document_type.rb
@@ -1,8 +1,6 @@
 # frozen_string_literal: true
 
 class DocumentType
-  extend DocumentTypeHelper # for static methods
-  include DocumentTypeHelper
   include InitializeWithHash
 
   attr_reader :contents, :id, :managed_elsewhere, :publishing_metadata, :label,
@@ -19,13 +17,7 @@ class DocumentType
 
       hashes.map do |hash|
         hash["contents"] = hash["contents"].to_a.map(&Field.method(:new))
-        hash["tags"] = (hash["tags"] || []).map do |tag_config|
-          tag_i18n_data = "document_types.#{hash['id']}.fields.#{tag_config['id']}"
-          if t_doctype_exists?(tag_i18n_data)
-            TagField.new(tag_config.merge(t_doctype(tag_i18n_data)))
-          end
-        end
-        hash["tags"] = hash["tags"].compact
+        hash["tags"] = hash["tags"].to_a.map(&TagField.method(:new))
         hash["publishing_metadata"] = PublishingMetadata.new(hash["publishing_metadata"].to_h)
         hash["topics"] = true # this feature is only disabled in tests
         new(hash)
--- a/app/models/document_type.rb
+++ b/app/models/document_type.rb
@@ -4,8 +4,7 @@ class DocumentType
   include InitializeWithHash
 
   attr_reader :contents, :id, :managed_elsewhere, :publishing_metadata,
-              :path_prefix, :tags, :guidance_govspeak, :images,
-              :topics, :check_path_conflict
+              :path_prefix, :tags, :images, :topics, :check_path_conflict
 
   def self.find(id)
     item = all.find { |document_type| document_type.id == id }
--- a/app/models/document_type.rb
+++ b/app/models/document_type.rb
@@ -5,7 +5,7 @@ class DocumentType
   include DocumentTypeHelper
   include InitializeWithHash
 
-  attr_reader :contents, :id, :managed_elsewhere, :publishing_metadata,
+  attr_reader :contents, :id, :managed_elsewhere, :publishing_metadata, :label,
               :path_prefix, :tags, :images, :topics, :check_path_conflict
 
   def self.find(id)
@@ -35,10 +35,6 @@ class DocumentType
 
   def managed_elsewhere_url
     Plek.new.external_url_for(managed_elsewhere.fetch("hostname")) + managed_elsewhere.fetch("path")
-  end
-
-  def label
-    t_doctype("document_types.#{id}.label")
   end
 
   class TagField
--- a/app/models/document_type.rb
+++ b/app/models/document_type.rb
@@ -20,7 +20,6 @@ class DocumentType
         hash["contents"] = hash["contents"].to_a.map(&Field.method(:new))
         hash["tags"] = hash["tags"].to_a.map(&TagField.method(:new))
         hash["publishing_metadata"] = PublishingMetadata.new(hash["publishing_metadata"].to_h)
-        hash["guidance"] = hash["guidance"].to_a.map(&Guidance.method(:new))
         hash["topics"] = true # this feature is only disabled in tests
         new(hash)
       end
@@ -31,18 +30,9 @@ class DocumentType
     Plek.new.external_url_for(managed_elsewhere.fetch("hostname")) + managed_elsewhere.fetch("path")
   end
 
-  def guidance_for(id)
-    @guidance.find { |guidance| guidance.id == id }
-  end
-
   class TagField
     include InitializeWithHash
     attr_reader :id, :label, :type, :document_type, :hint
-  end
-
-  class Guidance
-    include InitializeWithHash
-    attr_reader :id, :title, :body_govspeak
   end
 
   class PublishingMetadata
--- a/app/views/documents/fields/_govspeak_input.html.erb
+++ b/app/views/documents/fields/_govspeak_input.html.erb
@@ -1,5 +1,5 @@
 <% document_contents_guidance = capture do %>
-  <%= edition.document_type.guidance_for(field.id) ? render_govspeak(edition.document_type.guidance_for(field.id).body_govspeak) : nil %>
+  <%= I18n.exists?("document_types.#{edition.document_type.id}.fields.#{field.id}.guidance") ? render_govspeak(t("document_types.#{edition.document_type.id}.fields.#{field.id}.guidance.body_govspeak")) : nil %>
   <h3 class="govuk-heading-s"><%= t("content.edit.fields.govspeak.title") %></h3>
   <p class="govuk-body">
     <%= link_to t("content.edit.fields.govspeak.guidance"), "https://www.gov.uk/guidance/how-to-publish-on-gov-uk/markdown", class: "govuk-link", target: "_blank" %>
@@ -12,9 +12,9 @@
     id: field.id,
     html_for: "#{field.id}-field"
   }
-  if edition.document_type.guidance_for(field.id)
+  if I18n.exists?("document_types.#{edition.document_type.id}.fields.#{field.id}.guidance")
     field_contextual_guidance.merge!({
-      title: edition.document_type.guidance_for(field.id).title,
+      title: t("document_types.#{edition.document_type.id}.fields.#{field.id}.guidance.title"),
       content: document_contents_guidance,
       guidance_id: "#{field.id}-guidance"
     })
--- a/app/views/documents/fields/_govspeak_input.html.erb
+++ b/app/views/documents/fields/_govspeak_input.html.erb
@@ -1,5 +1,5 @@
 <% document_contents_guidance = capture do %>
-  <%= I18n.exists?("document_types.#{edition.document_type.id}.fields.#{field.id}.guidance") ? render_govspeak(t("document_types.#{edition.document_type.id}.fields.#{field.id}.guidance.body_govspeak")) : nil %>
+  <%= t_doctype_exists?("document_types.#{edition.document_type.id}.fields.#{field.id}.guidance") ? render_govspeak(t_doctype("document_types.#{edition.document_type.id}.fields.#{field.id}.guidance.body_govspeak")) : nil %>
   <h3 class="govuk-heading-s"><%= t("content.edit.fields.govspeak.title") %></h3>
   <p class="govuk-body">
     <%= link_to t("content.edit.fields.govspeak.guidance"), "https://www.gov.uk/guidance/how-to-publish-on-gov-uk/markdown", class: "govuk-link", target: "_blank" %>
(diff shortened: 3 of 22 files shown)


app/models/document_type.rb
```diff
@@ -35,6 +36,14 @@ def guidance_for(id)
     @guidance.find { |guidance| guidance.id == id }
   end
 
+  def hint_govspeak
+    begin
+      govspeak_to_html(I18n.t!("document_types.#{id}.guidance.body_govspeak"))
```
benthorner: Could we just use `t` (vs. `t!`)? I think that would remove the need for this change, although we'll probably need a variable in the view.

app/models/document_type.rb
```diff
@@ -32,8 +31,13 @@ def managed_elsewhere_url
     Plek.new.external_url_for(managed_elsewhere.fetch("hostname")) + managed_elsewhere.fetch("path")
   end
 
-  def guidance_for(id)
-    @guidance.find { |guidance| guidance.id == id }
+  def guidance_exists?(field_id)
```
benthorner: This seems like a good idea, to avoid repeating parts of the key everywhere. I think it would be better-placed in a helper, since it's primarily a view concern, and the point of this card is remove such concerns from the model.

config/locales/en/tags/edit.yml
```diff
@@ -5,3 +5,19 @@ en:
       description: Add tags which describe what the content is about. Content will appear in lists on GOV.UK based on each tag.
       api_down: This content can’t be edited right now. We’re having trouble getting the data we need for you to make changes on this page.
       organisation_warning: Access to this document is restricted. Adding or removing organisation tags will change who can access it.
+    types:
```
benthorner: This was a bit of a surprise, and not really consistent with the pattern for these files, since these strings are also used in the document summary. I think we should keep these locales close to the document type definitions for now, so they can participate in the same 'defaulting' mechanism we're aiming for in general. We may want to do more work with the locales, but this change is in excess of what we need to do for the card, and I'm not sure if it will cause problems for the upcoming work.

config/locales/en/document_types.yml
```diff
@@ -0,0 +1,112 @@
+en:
+  default:
```
benthorner: We should switch these, so the file name corresponds to the namespace within. This should be safe, since we're not likely to have a `default` document type 🤞

config/locales/en/tags/edit.yml
ChrisBAshton: I wasn't too sure where to put these, so would appreciate your thoughts on it. When I last showed you the code, it lived in `config/locales/en/document_types.yml`, which we agreed wasn't a great idea as locale files typically only have one top-level key, not two.

I don't think it should continue to live in `config/document_types.yml` as that means we'd have hardcoded English outside of the I18n namespace. My understanding is that the primary aim of the card is to move all associated English into I18n.

So it looks like we're moving towards having these in a new file: `config/locales/en/tags.yml`. Does that sound right?

kevindew:
Nice work on this Chris 👍 

One thing I think we should consider in regards to any naming is that this system is intended to be a specific case of trickier locale information and thus should be distinguished from the other system, `I18n`, that'll continue to exist. So I think methods like `t` and `t!` are to similar to those provided by I18n (consider the reading experience of a file that uses this approach and that) and that perhaps I18nHelper is a bit too broad a name (depending on whatever future scope). 

> @benthorner  
> It may also be helpful to reconsider a (somewhat subtle) suggestion on the card: having an I18n helper (like t_doctype) that's specific to document types should make the overall approach simpler.

That seems an improvement in distinguishing. It does however introduce the portmanteau of doctype into the app which means it's not referencing any existing terminology.

> One of the crucial aims of this work is to get view/English concerns out of the model. I think we still need to do some work to achieve this.

I wouldn't mind catching up with you on this point - mainly as it wasn't something covered on the card so we've not chatted about it - I see there being risks whether this information is accessed entirely through the model or entirely through view helpers and it'd be good to have discussed them before having committed to a direction.

ChrisBAshton:
Thanks for the comments, both. `I18nHelper` (and its `t`/`t!` conventions) were deliberate naming choices as I see the inheritance behaviour being useful outside of just doctypes. I treated it as an extension of the library, a bit like `fs-extra` vs the native `fs` in Node, which exposes all of the underlying methods and some additional useful methods.

If we genuinely only see the inheritance functionality being applied to doctypes, then a `t_doctype` helper would indeed be the way forward, and comes with the advantage of model/view separation. It's hard to reach a conclusion when the only thing we can currently inherit are the field labels for 'title', 'summary' and 'body' - I still think I'm missing some context on exactly why this is desired. Perhaps that might be better tackled as a second PR, leaving this first PR to be solely about internationalising the document_types without making further optimisations?

app/models/document_type.rb
ChrisBAshton: Unfortunately not, as `t` silently fails and returns "translation missing: document_types.news_story.guidance.body_govspeak". The safest way to force a `nil` return is to throw an exception via `t!` and catch it.

app/models/document_type.rb
ChrisBAshton: I'll move this to `I18nHelper` (or whatever that helper becomes) 👍

app/models/document_type.rb
ChrisBAshton: Note to self after call with Ben: this should be accomplishable with something like `t.has_key?("...")` rather than relying on exceptions. (Will certainly make the code nicer!)

config/locales/en/tags/edit.yml
ChrisBAshton: Note to self after call with Ben: tags & fields can be treated as the same thing, as per the gist.

config/locales/en/document_types.yml
ChrisBAshton: Note to self after call with Ben: it doesn't matter that the 'managed elsewhere' document types (`speech`, etc) would inherit the default properties. In fact, they should be left in `config/document_types.yml` rather than moved into I18n as they will be deleted later.

ChrisBAshton: Actually, not moving the the "managed elsewhere" document types is more work than expected, as we then have two ways of retrieving `label`/`description` 🤔 so code will end up something like this:

```
def description
  return t!("document_types.#{id}.description") if I18n.exists?("document_types.#{id}.description")

  hash["description"]
end
```

Is there any harm in transferring these psuedo-document-types across? I think it will make them easier to delete, as opposed to more difficult.

benthorner: Good point! 

On your comment: if it's easier to extract the label and description now, then let's go for it, as I agree doing some kind of switch is unnecessarily complicated.

@kevindew also pointed out that we should think carefully about whether any of these fields are 'inherent properties' of the object (like maybe label), rather than auxiliary things (definitely guidance). On that note: if extracting a field like `label` has difficult consequences (maybe in the tests), then perhaps we should continue to treat it as a property.

app/helpers/document_type_helper.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+
+module DocumentTypeHelper
+  def t_doctype_key_exists?(i18n_key)
+    I18n.exists?(i18n_key) || I18n.exists?(default_key(i18n_key))
+  end
+
+  def t_doctype(i18n_key)
+    match = I18n.t!(I18n.exists?(i18n_key) ? i18n_key : default_key(i18n_key))
+    match.is_a?(String) ? match : match.stringify_keys
```
benthorner: Could you help we understand what we need this for?

app/models/document_type.rb
```diff
@@ -17,7 +17,9 @@ def self.all
       hashes = YAML.load_file(Rails.root.join("config/document_types.yml"))
 
       hashes.map do |hash|
-        hash["contents"] = hash["contents"].to_a.map(&Field.method(:new))
+        hash["contents"] = hash["contents"].to_a.map do |content|
```
benthorner: I think we should look at other ways of doing this, as it seems to be making the model more complicated. One way would be to add an equivalent method within the `Field` object. But since `label` is used so infrequently, can't we just do the translation in the view? Looking at [one of the tests](https://github.com/alphagov/content-publisher/blob/master/spec/features/editing_tags/edit_tags_spec.rb), it doesn't look like we're relying on it - if we are, we could think about using an ID instead.

app/views/new_document/choose_document_type.html.erb
```diff
@@ -13,7 +13,7 @@
           {
             value: document_type.id,
             text: document_type.label,
-            hint_text: document_type.description,
+            hint_text: t("document_types.#{document_type.id}.description"),
```
benthorner: It's a shame this caused such an issue with the tests. While the factory solution is pretty neat, could we think about making the description optional? I agree that's not exactly a refactoring, but we [have a precedent for this sort of thing](https://github.com/alphagov/content-publisher/blob/bd10dd6c488102b53ba2ce1fa85fa923d16bdd46/app/models/document_type.rb#L24), where it's beneficial for the tests. The upcoming work on testing formats should ensure that any "real" format do provide a description, so I don't see a great risk of tweaking the behaviour here to make things easier.

spec/factories/document_type_factory.rb
```diff
@@ -28,7 +27,7 @@
     after(:build) do |document_type|
       DocumentType.all << document_type
       Supertype.all.first.document_types << document_type
-      en = { document_types: { document_type.id.to_sym => { description: "..." } } }
+      en = { document_types: { document_type.id.to_sym => { label: document_type.id, description: "..." } } }
```
benthorner: Could we use the defaulting mechanism to specify a "fallback" label? I hesitate here because the label is definitely required, but it's tempting because it reduces the coupling between the models and the locales. The central question here is whether one should be able to add a new document type without any locale information; while the upcoming work on tests should discourage that, functionally it seems reasonable. What do you think?

app/models/document_type.rb
```diff
@@ -20,7 +20,13 @@ def self.all
         hash["contents"] = hash["contents"].to_a.map do |content|
           Field.new(content.merge("label" => I18n.t!("document_types.#{hash['id']}.fields.#{content['id']}.label")))
         end
-        hash["tags"] = hash["tags"].to_a.map(&TagField.method(:new))
+        hash["tags"] = (hash["tags"] || []).map do |tag_config|
```
benthorner: Similarly here. Finding other ways to achieve the lookup should mean we can avoid the need for static helper methods in the later commit.

app/views/documents/edit.html.erb
```diff
@@ -20,10 +20,10 @@
       id: "title",
       html_for: "title-field"
     }
-    if I18n.exists?("document_types.#{@edition.document_type.id}.fields.title.guidance")
+    if t_doctype_key_exists?("document_types.#{@edition.document_type.id}.fields.title.guidance")
       title_contextual_guidance.merge!({
-        title: t("document_types.#{@edition.document_type.id}.fields.title.guidance.title"),
-        content: t("document_types.#{@edition.document_type.id}.fields.title.guidance.body_govspeak"),
+        title: t_doctype("document_types.#{@edition.document_type.id}.fields.title.guidance.title"),
```
benthorner: I think we should get the helper to do more work here: `t_doctype("document_types` is repetitive. Since the helper is specific to doctypes, it makes sense to move away from strict I18n keys, and at least lose the first element of the key.

benthorner: We should avoid the interpolation here, since we repeat it a lot. We could think about having multiple arguments to the helper, and maybe even having a separate one for fields.

app/views/documents/fields/_govspeak_input.html.erb
```diff
@@ -1,5 +1,5 @@
 <% document_contents_guidance = capture do %>
-  <%= I18n.exists?("document_types.#{edition.document_type.id}.fields.#{field.id}.guidance") ? render_govspeak(t("document_types.#{edition.document_type.id}.fields.#{field.id}.guidance.body_govspeak")) : nil %>
+  <%= t_doctype_key_exists?("document_types.#{edition.document_type.id}.fields.#{field.id}.guidance") ? render_govspeak(t_doctype("document_types.#{edition.document_type.id}.fields.#{field.id}.guidance.body_govspeak")) : nil %>
```
benthorner: Since guidance is optional, perhaps we could use a dedicated `t_doctype_guidance` helper that's allowed to return nil - this would DRY-up the ternary, as well as most of the elements of the locale key.

app/helpers/document_type_helper.rb
ChrisBAshton: It's used here: `TagField.new(tag_config.merge(t_doctype(tag_i18n_data)))`

We're trying to merge tag metadata:

```
    - id: primary_publishing_organisation
      type: single_tag
      document_type: organisation
```

...with i18n info:

```
        primary_publishing_organisation:
          label: Primary organisation
          hint: The organisation responsible for publishing and maintaining this content.
```

Rather than manually retrieving the i18n value for both `label` and `hint`, this code pulls the i18n value for `primary_publishing_organisation` as a whole, so that we can merge _all_ the i18n values into the tag metadata.

But by default I18n will return hashes as symbol/value pairs instead of string/value pairs, which is different to the hashes we retrieve from parsing the YAML in our `self.all` method. So I use `stringify_keys` on the hash.

spec/factories/document_type_factory.rb
ChrisBAshton: If we did provide a fallback label, it would still be a hardcoded string (unless YAML can do something clever I don't know about!). This would break a number of tests where a document type has to be chosen, as over the course of the test suite various document types build up, so our test would be trying to pick from this:

```html
<select>
  <option value="the-fallback-label">the-fallback-label</option>
  <option value="the-fallback-label">the-fallback-label</option>
  <option value="the-fallback-label">the-fallback-label</option>
  <option value="the-fallback-label">the-fallback-label</option>
  ...
</select>
```

...and then RSpec complains about an ambiguous match.

app/models/document_type.rb
ChrisBAshton: I've now removed the static helper method - didn't like it!

app/views/documents/edit.html.erb
ChrisBAshton: Agreed 👍

ChrisBAshton: I considered this, but couldn't decide on a way forward that was much better than the original. We need to pass the document type ID, the field name and then the name of the guidance we want to retrieve, so would end up with something like this:

```diff
- t_doctype("document_types.#{@edition.document_type.id}.fields.title.guidance.title"),
+ t_doctype_field_guidance(@edition.document_type.id, "title", "title"),
```

This felt like one of those cases where some ugly long strings were cleaner than a poor abstraction, but I'll happily go in this direction if you think otherwise.

Another thing I considered was some form of on-the-fly object-orientation, e.g.:

```diff
- t_doctype("document_types.#{@edition.document_type.id}.fields.title.guidance.title"),
+ DocumentTypeHelper.new(@edition.document_type).guidance_for("title").title,
```

This reads cleaner but doesn't map well with the 'helper module' approach we've gone for - it feels more like a presenter. We don't yet have any presenters in Content Publisher and I didn't want to establish a precedent.

app/views/documents/fields/_govspeak_input.html.erb
ChrisBAshton: Great idea 👍 I wasn't sure what to call it but this sounds nice.

app/helpers/document_type_helper.rb
benthorner: I think I noticed this after I commented 😄. However, since the main goal is still to get this stuff out of the models, could we look at getting the views to do this instead?

app/views/new_document/choose_document_type.html.erb
```diff
@@ -13,12 +13,12 @@
           {
             value: document_type.id,
             text: document_type.label,
-            hint_text: document_type.description,
+            hint_text: t_doctype("document_types.#{document_type.id}.description"),
```
ChrisBAshton: Use I18n for these - we're moving them later.

app/views/new_document/choose_document_type.html.erb
```diff
@@ -13,12 +13,12 @@
...
             data_attributes: {
               gtm: "choose-document-type",
               "gtm-value": document_type.label,
             },
-            conditional: document_type.hint_govspeak ? render_govspeak(document_type.hint_govspeak) : nil,
+            conditional: t_doctype_key_exists?("document_types.#{document_type.id}.guidance.body_govspeak") ? render_govspeak(t_doctype("document_types.#{document_type.id}.guidance.body_govspeak")) : nil,
```
ChrisBAshton: Use I18n for these - we're moving them later.

spec/factories/document_type_factory.rb
ChrisBAshton: Clearing the dynamically created document types is something we should look at in a future PR.

app/models/document_type.rb
```diff
@@ -31,20 +39,15 @@ def managed_elsewhere_url
     Plek.new.external_url_for(managed_elsewhere.fetch("hostname")) + managed_elsewhere.fetch("path")
   end
 
-  def guidance_for(id)
-    @guidance.find { |guidance| guidance.id == id }
+  def label
```
ChrisBAshton: Move this back to document_types.yml in separate commit & explain why.

app/helpers/document_type_helper.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+
+module DocumentTypeHelper
+  def t_doctype_key_exists?(i18n_key)
+    I18n.exists?(i18n_key) || I18n.exists?(default_key(i18n_key))
+  end
+
+  def t_doctype(i18n_key)
+    match = I18n.t!(I18n.exists?(i18n_key) ? i18n_key : default_key(i18n_key))
```
ChrisBAshton: This can just be `t!`

app/views/documents/edit.html.erb
ChrisBAshton: Talked IRL, decided on this as a way forward:

```
t_doctype(@edition, "fields.title.guidance.title"),
```

This will only appear in `app/views/documents/fields/_govspeak_input.html.erb`, as we will no longer need to use the helper in `app/views/documents/edit.html.erb` (we'll use I18n directly).

app/views/documents/fields/_govspeak_input.html.erb
ChrisBAshton: This idea is superseded by using raw I18n instead (we don't anticipate setting any defaults).

spec/factories/document_type_factory.rb
ChrisBAshton: We've decided to keep 'label' in config/document_types.yml - there will be a commit that encapsulates this.

app/views/new_document/choose_document_type.html.erb
ChrisBAshton: Have now made it optional (locally -will push soon).

app/helpers/document_type_helper.rb
ChrisBAshton: `tag_field.label` is called in _tags.html, _single_tag_input.html and _multi_tag_input.html. If I use `t` in the views, I get errors like this (for both label and hint):

`translation missing: en.document_types.8b5139bd.fields.tag_id.label`

...which I can solve by adding an arbitrary default label (doesn't feel like the right approach!):

```ruby
t("document_types.#{@edition.document_type.id}.fields.#{tag_field.id}.label", default: "Tag label missing")
```

We could come to the same conclusion as we came to for the document type labels - i.e. to keep each tag `label`/`hint` in config/document_types.yml. Or we can keep this model code which combines the metadata and the I18n keys.

benthorner: Ah, that's annoying. I think this is happening because the tests are creating random IDs, which won't be possible for much longer (at least, not for content fields). We could reasonably change them to use a known ID, since the total set of fields isn't meant to be dynamic. Does that help?

ChrisBAshton: I can look into it tomorrow morning 🤔 also need to fix my build after all the refactoring. Hope to wrap this up soon!

ChrisBAshton: I have now done this 👍 thanks for the suggestion.

app/views/documents/edit.html.erb
ChrisBAshton: Now done 👍

app/models/document_type.rb
```diff
@@ -3,9 +3,8 @@
 class DocumentType
   include InitializeWithHash
 
-  attr_reader :contents, :id, :label, :managed_elsewhere, :publishing_metadata,
-              :path_prefix, :tags, :guidance_govspeak, :description, :hint_govspeak, :images,
-              :topics, :check_path_conflict
+  attr_reader :contents, :id, :managed_elsewhere, :publishing_metadata, :label,
```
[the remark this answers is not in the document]
ChrisBAshton: Agreed it would be great to extract this. If we move it back into I18n, the problem is the dynamically generated document types lead to a `translation missing: document_types.abc123.description` type error. We'd have to make a choice:

1) convert `document_type.description` to `t("document_types.#{document_type.id}.description", default: "")` - downside being introducing a `default` that is only ever used by our tests; something we considered and rejected for `label`.
2) inject I18n key into our tests every time we generate a document type; something we considered and rejected for `label`.

I couldn't think of a way of doing it that we hadn't already dismissed in the case of `label`, so for consistency moved it back out 😞

spec/helpers/document_type_helper_spec.rb
```diff
@@ -0,0 +1,61 @@
...
+          fields: {
+            title: {
+              label: "My title",
+              description: "...",
+            },
+          },
+        },
+      },
+    }
+    I18n.backend.store_translations(:en, en)
```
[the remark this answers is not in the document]
ChrisBAshton: Ooh good point 👍

app/models/document_type.rb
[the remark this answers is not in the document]
benthorner: Ah, so I think I wasn't clear enough about what I meant by 'optional'. I meant doing [something like this](https://github.com/alphagov/content-publisher/pull/1659/files#diff-7ee41565bf7db9c2b08e740bd04bfe64R23) - guess that's option 3 😉.

The factory illustrates that it's optional, in that we've been creating document types with no description - the ternary would be equivalent to that.

Does that make more sense?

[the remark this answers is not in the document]
ChrisBAshton: Hmm that _is_ an option 3 - but is still using application code to cater for a situation we'll only encounter in our tests. For me it's equivalent to option 1.

So I guess we need to decide if it's better to have some redundant 'optional' logic in the view and a cleaner separation of 'description' into I18n, or a cleaner view but some icky config 🤔 I'm starting to think option 1 is the best option - what do you think?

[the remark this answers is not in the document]
benthorner: So you're right that the app code is catering for the tests. That's not a bad thing in itself - we do it all the time with class interfaces - as long as it doesn't surprise/confuse people. If we went with option 1, my only concern is that it's inconsistent with all other translations (i.e. something new), which would be more surprising than the ternary pattern we have elsewhere. In itself, seeing code to cope with an optional description wouldn't surprise me.

The problem with `label` is that it's not optional: we have to put something there, and all of the alternatives were surprising. Conversely, it's plausible that a document type may not have a description. Do agree we should avoid option 2!

I think the case is pretty strong, but I also think we could leave it as-is for now and look at it in a separate PR, so it doesn't block this one going in.

spec/helpers/document_type_helper_spec.rb
```diff
@@ -0,0 +1,61 @@
+# frozen_string_literal: true
+
+RSpec.describe DocumentTypeHelper, type: :helper do
```
benthorner: Minor comment from @kevindew - the `type` should be inferred automatically.

spec/features/editing_tags/edit_tags_spec.rb
```diff
@@ -17,8 +19,8 @@
   end
 
   def given_there_is_an_edition
-    multi_tag_field = build(:tag_field, type: "multi_tag", id: "multi_tag_id")
-    single_tag_field = build(:tag_field, type: "single_tag", id: "single_tag_id")
+    multi_tag_field = build(:tag_field, type: "multi_tag", id: multi_tag_id)
```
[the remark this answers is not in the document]
ChrisBAshton: I don't think traits would actually help make edit_tags_spec.rb any more readable, as this test suite looks at the HTML of the page and asserts values/IDs. If the ID we're asserting against is actually hidden inside the factory, we'd have two places to update the ID if it ever changes.

However, adding the traits does allow me to simplify some of the other tag tests. I've encapsulated it as a separate refactor commit.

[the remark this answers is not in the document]
benthorner: Yes, simplifying the tests was what I was thinking of here. I don't think it will make it any more or less readable. While the tests still need to use the IDs, the benefit of this change should be that they don't need to know precisely what they are, since we can just ask the field each time.

app/helpers/document_type_helper.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+module DocumentTypeHelper
+  def t_doctype_exists?(edition, partial_i18n_key)
+    i18n_key = prefix_document_type(edition, partial_i18n_key)
+    I18n.exists?(i18n_key) || I18n.exists?(default_key(i18n_key))
+  end
+
+  def t_doctype_field(edition, partial_i18n_key)
```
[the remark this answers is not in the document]
ChrisBAshton: I think this would become a confusing abstraction, as we sometimes need to access a piece of guidance (`title.guidance.title`) and sometimes a label (`#{tag_field.id}.label`). A hybrid approach of I18n and prefixing seems a good way forward for now.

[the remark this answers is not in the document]
benthorner: Hmm...I think we'll see a stronger need for this son, since `#{field.id}` will become the norm, as part of the upcoming work on "flexible fields". On the other hand, I get that writing `, "title")` is somewhat different to writing `, "guidance.title")`, in that the interpretation of the argument is different. I don't think it's a blocker for this PR, but perhaps something to revisit as we stop the hard-coding for title and summary.

spec/helpers/edition_url_helper_spec.rb
```diff
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-RSpec.describe EditionUrlHelper, type: :helper do
+RSpec.describe EditionUrlHelper do
```
benthorner: Oops! Looks like this got missed in the past 😬

PR #1660, opened 2020-01-17
Fix govspeak render in document type description

alex-ju:
Follow up on https://github.com/alphagov/content-publisher/pull/1642.

Document type description may contain links which means the associated content needs to be rendered as govspeak to be styled properly.

### Before
Screenshot 2020-01-17 at 08 57 59

### After
Screenshot 2020-01-17 at 08 55 51

### DIFF
--- a/app/views/new_document/choose_document_type.html.erb
+++ b/app/views/new_document/choose_document_type.html.erb
@@ -18,7 +18,7 @@
               gtm: "choose-document-type",
               "gtm-value": document_type.label,
             },
-            conditional: document_type.hint_govspeak ? tag.div(govspeak_to_html(document_type.hint_govspeak), class: "govuk-body") : nil,
+            conditional: document_type.hint_govspeak ? render_govspeak(document_type.hint_govspeak) : nil,
             bold: true,
           }
         }
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -1,19 +1,13 @@
 # frozen_string_literal: true
 
 module ApplicationHelper
-  def govspeak_to_html(govspeak)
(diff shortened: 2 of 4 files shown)


app/views/new_document/choose_document_type.html.erb
```diff
@@ -18,7 +18,7 @@
               gtm: "choose-document-type",
               "gtm-value": document_type.label,
             },
-            conditional: document_type.hint_govspeak ? tag.div(govspeak_to_html(document_type.hint_govspeak), class: "govuk-body") : nil,
+            conditional: document_type.hint_govspeak ? render_govspeak(document_type.hint_govspeak) : nil,
```
ChrisBAshton: There are a few other cases where we call `govspeak_to_html` in the views, we should probably replace those calls with `render_govspeak` too?

That would then suggest we could remove the `govspeak_to_html` method altogether but I think it would still be needed by the tests 🤔

alex-ju: Agree. I think we're currently using `govspeak_to_html` in `render_govspeak` – have to double-check. I guess the question is shall we keep it and make it private or move the `govspeak_to_html` logic straight into `render_govspeak`?

ChrisBAshton: Yes, ideally we could move the `govspeak_to_html` logic into `render_govspeak` and delete the method. Not sure if swapping `govspeak_to_html` for `render_govspeak` inside the tests will break things (the additional dependency / runtime complexities associated with the `render` call). Worth a try!

alex-ju: Did that refactor in a separate commit and tested visually to make sure it renders correctly.

PR #1662, opened 2020-01-17
Allow autocomplete component to be shown only when JavaScript is enabled

alex-ju:
The autocomplete component (being an enhanced input), as any other form element rely on `govuk-form-group` class to generate consistent spacing across elements. This class assumes the form elements are at a sibling level (and not nested in unnecessary div wrappers) to style the last element without spacing. This updates the way we apply the `app-js-only` class to this instance of the component by applying the class to the component based on a `jsonly` flag.

### Before
Screenshot 2020-01-17 at 11 00 51

### After
Screenshot 2020-01-17 at 11 00 37

[Trello card](https://trello.com/c/rO5ssI8S)

### DIFF
--- a/app/views/components/_autocomplete.html.erb
+++ b/app/views/components/_autocomplete.html.erb
@@ -10,10 +10,12 @@
   select ||= {}
   input ||= {}
   search ||= false
+  jsonly ||= false
 
   root_classes = %w(app-c-autocomplete govuk-form-group)
   root_classes << "app-c-autocomplete--#{width}" if width
   root_classes << "app-c-autocomplete--search" if search
+  root_classes << "app-js-only" if jsonly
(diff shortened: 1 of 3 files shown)


benthorner:
> This updates the way we apply the app-js-only class to this instance of the component 

I'm a little confused by the description. Am I right in thinking the aim of this PR is: to support only showing the autocomplete component when JS is enabled? (And although we did this before, it wasn't a complete solution, which caused the weird spacing issue?)

alex-ju:
> I'm a little confused by the description. Am I right in thinking the aim of this PR is: to support only showing the autocomplete component when JS is enabled? (And although we did this before, it wasn't a complete solution, which caused the weird spacing issue?)

That is right. I've updated the PR title in a try to make it more clear. The current solution is applied from outside the component using an additional container which causes spacing issues.

PR #1665, opened 2020-01-20
Multiple lead orgs

1pretz1:
We now want to import editions which have multiple lead organisations
but use the 'primary' (lowest numerical lead_ordering) to assign to the
edition.

A check is already done [here](https://github.com/alphagov/whitehall/blob/master/app/controllers/admin/export/document_controller.rb#L45) to decipher if the lead organisation passed in matches the primary lead org of the latest edition, this makes sure we are only importing documents which the most current version is owned by the organisation queried in whitehall.

Trello:
https://trello.com/c/7t6Nv8YT/1321-import-editions-with-multiple-lead-organisations

### DIFF
--- a/lib/whitehall_importer/create_revision.rb
+++ b/lib/whitehall_importer/create_revision.rb
@@ -95,10 +95,6 @@ module WhitehallImporter
         raise AbortImportError, "Lead organisation missing"
       end
 
-      if primary_publishing_organisations.count > 1
-        raise AbortImportError, "Cannot have more than one lead organisation"
-      end
-
       primary_publishing_organisation = primary_publishing_organisations.min { |o| o["lead_ordering"] }
 
       [primary_publishing_organisation["content_id"]]
--- a/lib/whitehall_importer/create_revision.rb
+++ b/lib/whitehall_importer/create_revision.rb
@@ -54,8 +54,8 @@ module WhitehallImporter
         ),
         tags_revision: TagsRevision.new(
           tags: {
-            "primary_publishing_organisation" => primary_publishing_organisation(whitehall_edition["organisations"]),
-            "organisations" => supporting_organisations(whitehall_edition["organisations"]),
+            "primary_publishing_organisation" => primary_publishing_organisation,
+            "organisations" => supporting_organisations,
             "role_appointments" => tags(whitehall_edition["role_appointments"]),
             "topical_events" => tags(whitehall_edition["topical_events"]),
             "world_locations" => tags(whitehall_edition["world_locations"]),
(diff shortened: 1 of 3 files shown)


spec/factories/whitehall_migration_factory.rb
```diff
@@ -3,7 +3,7 @@
 FactoryBot.define do
   factory :whitehall_migration do
     sequence(:id)
-    organisation_content_id { "content_id" }
+    organisation_content_id { SecureRandom.uuid }
```
brucebolt: 👍

spec/lib/whitehall_importer/create_revision_spec.rb
```diff
@@ -181,6 +181,20 @@
...
+
+      revision = described_class.call(document_import, whitehall_edition)
+
+      expect(revision.primary_publishing_organisation_id)
+        .to eq(first_lead_organisation["content_id"])
+    end
+
     it "sets supporting organisations" do
       lead_organisation = build(:whitehall_export_organisation, :lead)
       supporting_organisation = build(:whitehall_export_organisation)
```
brucebolt: I don't think we're testing the scenario of there being multiple lead organisations.  The second lead organisation has lead=true, so I believe https://github.com/alphagov/content-publisher/pull/1665/commits/a6e5dd9b6274c6cac65c82023be7f253e6b72bbd#diff-eb91f6f587e00528b32cc8dc205dcd0bL109 would mean this doesn't get added as a supporting organisation either.  Maybe this test could be updated (or another added) to reflect this scenario?

1pretz1: You're absolutely right Bruce, I've created a context for multiple lead orgs and two unit tests off of your suggestion to better describe what each test is actually testing instead of pushing them into one.

1pretz1:
Good catch @brucebolt, you're right!! I've now updated the code to reflect this, serves me right for just brainlessly following the A/C's on the card without thinking. 😄

lib/whitehall_importer/create_revision.rb
```diff
@@ -82,30 +82,34 @@ def translation
...
-      supporting_organisations = organisations.reject do |organisation|
+    def primary_publishing_organisations
+      whitehall_edition["organisations"].select do |organisation|
         organisation["lead"]
       end
+    end
 
-      supporting_organisations.map { |organisation| organisation["content_id"] }
+    def first_lead_organisation
+      @first_lead_organisation ||= primary_publishing_organisations.min { |o| o["lead_ordering"] }
```
brucebolt: I'm wondering if it might be an idea to sort the return value from `primary_publishing_organisations`, then this method probably wouldn't be needed (since you could just use `.first` in `primary_publishing_organisation`)?

1pretz1: Sure, you do lose the memoization aspect but as we're only looping through hash's on small arrays it doesn't really make a difference!

1pretz1:
I've made the changes you requested @brucebolt! :)

PR #1671, opened 2020-01-21
Image request tests

kevindew:

This PR adds request tests for image and lead image endpoints. It then removes feature tests that test the dead-end aspect of the feature


kevindew:
Thanks for the review @benthorner 

As discussed yesterday the reason for the 404 testing is that we introduced logic that meant that not only does the image have to exist but it also has to be associated with the edition model. So the actual test is not that the record exists but that it is on the right item. So this mostly tests that a user can't somehow cause a strange mess by editing an image with an id that doesn't match the document and then the system somehow conflating those two things together.

I've put together a shared example for this to reduce the duplication and overhead of this though and applied it to file attachments.

PR #1672, opened 2020-01-21
Add request tests for tags and topics

kevindew:

This PR adds request tests for tags and topic endpoints and removes the dead-end feature tests that are now covered by the request tests.


spec/requests/tags_spec.rb
```diff
@@ -0,0 +1,94 @@
...
+      stub_publishing_api_has_linkables(
+        [{ "content_id" => current_user.organisation_content_id,
+           "internal_name" => "Organisation" }],
+        document_type: tag_field.document_type,
+      )
+
+      get tags_path(edition.document)
+      expect(response).to have_http_status(:ok)
+      expect(response.body)
+        .to have_content(I18n.t!("tags.edit.organisation_warning"))
```
benthorner: Thinking back to our discussion on the different types of test: would this be better suited (and perhaps more easily implemented) in a view test? It seems to be more focussed on testing a property of the view, than the nature of the request/response.

spec/requests/tags_spec.rb
```diff
@@ -0,0 +1,94 @@
...
+    end
+  end
+
+  describe "PATCH /documents/:document/tags" do
+    before { stub_any_publishing_api_put_content }
+
+    it "redirects to document path on successful request" do
+      tag_field = build(:tag_field, type: "multi_tag", id: "field")
```
benthorner: Minor: do we need any tag fields for this to work?

spec/requests/tags_spec.rb
kevindew: Yes, you're right - I've moved this into a view test.

spec/requests/tags_spec.rb
kevindew: No, it doesn't I've removed this. Thanks

PR #1673, opened 2020-01-21
Add request tests for scheduling

kevindew:

This PR adds request tests for schedule and schedule proposal endpoints. It also fixes an issue where we risked a 500 error based on incorrect params provided to both endpoints.


spec/requests/schedule_proposal_spec.rb
```diff
@@ -0,0 +1,87 @@
...
+
+    it "returns successfully" do
+      get schedule_proposal_path(edition.document)
+
+      expect(response).to have_http_status(:ok)
+    end
+  end
+
+  describe "POST /documents/:document/schedule-proposal" do
+    before { stub_any_publishing_api_put_content }
```
benthorner: Minor: I don't think we need this.

spec/requests/schedule_proposal_spec.rb
```diff
@@ -0,0 +1,87 @@
...
+        expect(response).to have_http_status(:unprocessable_entity)
+        expect(response.body).to have_content(
+          I18n.t!("requirements.schedule_action.not_selected.form_message"),
+        )
+      end
+    end
+  end
+
+  describe "DELETE /documents/:document/schedule-proposal" do
+    before { stub_any_publishing_api_put_content }
```
benthorner: Minor: I don't think we need this.

spec/requests/schedule_spec.rb
```diff
@@ -0,0 +1,148 @@
...
+  it_behaves_like "requests that assert edition state",
+                  "managing the scheduling of a non-scheduled edition",
+                  routes: { edit_schedule_path: %i[get patch],
+                            schedule_path: %i[delete],
+                            scheduled_path: %i[get] } do
+    let(:edition) { create(:edition, :published) }
+  end
+
+  describe "GET /documents/:document/schedule/new" do
+    it "redirects to document summary with errors when the content isn't publishable" do
```
benthorner: This seemed a bit overkill, but then I noticed that it's [missing from the interactor spec](https://github.com/alphagov/content-publisher/blob/master/spec/interactors/schedule/update_interactor_spec.rb#L71). Could we plug the hole there, and just have one spec to focus on "issues" here? It's pretty minor, really - the testing strategy was tricky to figure out, since the corresponding feature test has three cases, and here we only have two.

kevindew: I have a bit of confusion here as you're linking to the update interactor but this is the new endpoint. I also missed the success case test, which I've since added.

But anyway on the new endpoint there are three responses the controller can return and there are now three tests which cover each.

kevindew:
Thanks @benthorner - You were right those stubs were unnecessary. 

ivar is shorthand for instance variable. I've amended the commit message since it was intended to confuse and when I googled the first result was someone confused on stack overflow (https://stackoverflow.com/questions/33371300/what-is-ivar-in-rails-controller).

spec/requests/schedule_spec.rb
benthorner: Oops - you're right. I *was* looking at the wrong thing!

PR #1675, opened 2020-01-21
Consistent withdraw request tests

kevindew:

This does some refactoring of the existing withdraw and unwithdraw request tests to make them consistent with the conventions of request tests.

They also catch a couple of issues. It allows to return a 503 on an API error rather than redirect for a withdraw that hits a Publishing API error (this required resolving a hash keys issue on the layout) and fixes that the assert edition status on the Withdraw::CreateInteractor wasn't working.

### DIFF
--- a/app/interactors/withdraw/create_interactor.rb
+++ b/app/interactors/withdraw/create_interactor.rb
@@ -24,7 +24,7 @@ private
   def find_and_lock_edition
     context.edition = Edition.lock.find_current(document: params[:document])
 
-    assert_edition_state(@edition, assertion: "is published or withdrawn") do
+    assert_edition_state(edition, assertion: "is published or withdrawn") do
       edition.published? || edition.published_but_needs_2i? || edition.withdrawn?
     end
   end
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -65,14 +65,15 @@
       <% end %>
 
       <% if flash["alert_with_description"] %>
+        <% alert = flash["alert_with_description"].stringify_keys %>
         <%= render "govuk_publishing_components/components/error_alert", {
-          message: flash["alert_with_description"].fetch("title"),
+          message: alert.fetch("title"),
           data_attributes: {
             gtm: "alert-with-description",
-            "gtm-value" => flash["alert_with_description"]["title"],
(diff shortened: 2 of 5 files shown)


app/interactors/withdraw/create_interactor.rb
```diff
@@ -24,7 +24,7 @@ def call
   def find_and_lock_edition
     context.edition = Edition.lock.find_current(document: params[:document])
 
-    assert_edition_state(@edition, assertion: "is published or withdrawn") do
+    assert_edition_state(edition, assertion: "is published or withdrawn") do
```
1pretz1: Good catch!

spec/requests/unwithdraw_spec.rb
```diff
@@ -11,104 +11,144 @@
   end
 
   describe "POST /documents/:document/unwithdraw" do
-    let(:withdrawn_edition) { create(:edition, :withdrawn) }
+    let(:edition) { create(:edition, :withdrawn) }
```
1pretz1: V minor but in the commit message you mention feature tests when I assume you mean request tests.

kevindew: Thanks - I've fixed that.

kevindew:
Thanks @1pretz1 - that's a good point about terminology like "regular user" I've done some tweaking to remove that phrasing.

app/views/layouts/application.html.erb
```diff
@@ -65,14 +65,15 @@
...
 
       <% if flash["alert_with_description"] %>
+        <% alert = flash["alert_with_description"].stringify_keys %>
         <%= render "govuk_publishing_components/components/error_alert", {
-          message: flash["alert_with_description"].fetch("title"),
+          message: alert.fetch("title"),
           data_attributes: {
             gtm: "alert-with-description",
-            "gtm-value" => flash["alert_with_description"]["title"],
+            "gtm-value" => alert["title"],
```
ChrisBAshton: Why do we have two ways of retrieving the `"title"`?

kevindew: That'll be because this fetch is superfluous after the one on line 70. It does make for a jarring read though so I've amended it to be consistent.

PR #1681, opened 2020-01-22
Controller Reshuffle

JonathanHallam:
Trello - https://trello.com/c/ZTRjP9eT/1099-break-up-the-actions-in-the-documentscontroller

The Documents controller has been largely unchanged since we introduced the concept of editions. This moves functionality from the documents controller into the editions controller and a newly created content controller to better fit with our data models. It also adjusts the copy and tests accordingly.

### DIFF
--- a/app/controllers/content_controller.rb
+++ b/app/controllers/content_controller.rb
@@ -1,0 +1,39 @@
+# frozen_string_literal: true
+
+class ContentController < ApplicationController
+  def edit
+    @edition = Edition.find_current(document: params[:document])
+    assert_edition_state(@edition, &:editable?)
+    @revision = @edition.revision
+  end
+
+  def update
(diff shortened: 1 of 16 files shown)


app/controllers/content_controller.rb
```diff
@@ -0,0 +1,39 @@
...
+
+class ContentController < ApplicationController
+  def edit
+    @edition = Edition.find_current(document: params[:document])
+    assert_edition_state(@edition, &:editable?)
+    @revision = @edition.revision
+  end
+
+  def update
+    result = Editions::UpdateInteractor.call(params: params, user: current_user)
```
benthorner: Should this be under the `Content` namespace?

JonathanHallam: yep, should have been, just fixing it now :)

PR #1684, opened 2020-01-22
Add body to timeline_entry_whitehall_imported_entries

Rosa-Fox:
- Required for upcoming implementation of: https://trello.com/c/o8PhjdUc/1312-map-whitehall-fact-check-requests-fact-checking-tab-in-whitehall-to-content-publisher-timeline-entries-on-import and https://trello.com/c/tNsWXZZj/1314-map-whitehall-editorial-remarks-notes-tab-in-whitehall-to-content-publisher-timeline-entries-on-import

- Will allow us to store the notes body content as well as the fact check data

Co-authored-by: Graham Lewis <graham.lewis@digital.cabinet-office.gov.uk>

### DIFF
--- a/db/migrate/20200122110833_add_body_to_timeline_entry_whitehall_imported_entries.rb
+++ b/db/migrate/20200122110833_add_body_to_timeline_entry_whitehall_imported_entries.rb
@@ -1,0 +1,7 @@
+# frozen_string_literal: true
+
+class AddBodyToTimelineEntryWhitehallImportedEntries < ActiveRecord::Migration[6.0]
+  def change
+    add_column :timeline_entry_whitehall_imported_entries, :body, :json, default: "{}", null: false
+  end
+end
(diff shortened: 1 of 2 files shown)


db/migrate/20200122110833_add_entry_body_to_timeline_entry_whitehall_imported_entries.rb
```diff
@@ -0,0 +1,5 @@
+class AddEntryBodyToTimelineEntryWhitehallImportedEntries < ActiveRecord::Migration[6.0]
+  def change
+    add_column :timeline_entry_whitehall_imported_entries, :entry_body, :json
```
kevindew: I'm not sure you need to prefix this name with entry since the class is already an entry.

Do you want this to be null too? It can be quite common to have JSON fields as non-null and default to an empty hash. Null might well serve a purpose here so you may want it, but just flagging as it can be confusing if you have to check regularly that item isn't null for usage.

gclssvglx: Thanks for the comments.  Fixed as requested and made JSON field not null with default.

PR #1688, opened 2020-01-23
Whitehall notes and fact check - import

gclssvglx:
Whitehall stores document history in three places. This PR addresses the import of two of these into Content Publisher:

-  Editorial remarks (Notes tab in Whitehall) 
- Fact check requests (Fact checking tab in Whitehall)

Trello: https://trello.com/c/o8PhjdUc/1312-map-whitehall-editorial-remarks-notes-tab-in-whitehall-and-fact-check-requests-fact-checking-tab-in-whitehall-to-content-publish

### DIFF
--- a/lib/whitehall_importer/create_edition.rb
+++ b/lib/whitehall_importer/create_edition.rb
@@ -34,6 +34,7 @@ module WhitehallImporter
                 end
 
       create_revision_history(edition)
+      create_notes(edition)
 
       edition.tap { |e| access_limit(e) }
     end
@@ -184,18 +185,41 @@ module WhitehallImporter
         entry_type = history.imported_entry_type(event, edition_number)
         next if entry_type.nil?
 
-        details = TimelineEntry::WhitehallImportedEntry.create!(
-          entry_type: entry_type,
-        )
-        TimelineEntry.create!(
-          entry_type: :whitehall_migration,
-          created_at: event["created_at"],
-          created_by_id: user_ids[event["whodunnit"]],
-          edition: edition,
-          document: edition.document,
-          details: details,
-        )
+        details = create_whitehall_imported_entry(entry_type)
+        create_timeline_entry(details,
+                              edition,
+                              event["created_at"],
+                              user_ids[event["whodunnit"]])
       end
+    end
+
+    def create_notes(edition)
+      whitehall_edition["editorial_remarks"].each do |event|
+        contents = {
+          body: event["body"],
+        }
+        details = create_whitehall_imported_entry("internal_note", contents)
+        create_timeline_entry(details,
+                              edition, event["created_at"], event["author_id"])
+      end
+    end
+
+    def create_whitehall_imported_entry(entry_type, contents = {})
+      TimelineEntry::WhitehallImportedEntry.create!(
+        entry_type: entry_type,
+        contents: contents,
+      )
+    end
+
+    def create_timeline_entry(details, edition, created_at, created_by = nil)
+      TimelineEntry.create!(
+        entry_type: :whitehall_migration,
+        created_by_id: created_by,
+        created_at: created_at,
+        edition: edition,
+        document: edition.document,
+        details: details,
+      )
     end
   end
 end
--- a/lib/whitehall_importer/create_edition.rb
+++ b/lib/whitehall_importer/create_edition.rb
@@ -35,6 +35,7 @@ module WhitehallImporter
 
       create_revision_history(edition)
       create_notes(edition)
+      create_fact_checks(edition)
 
       edition.tap { |e| access_limit(e) }
     end
@@ -202,6 +203,32 @@ module WhitehallImporter
         create_timeline_entry(details,
                               edition, event["created_at"], event["author_id"])
       end
+    end
+
+    def create_fact_checks(edition)
+      whitehall_edition["fact_check_requests"].each do |event|
+        create_fact_check_request(edition, event)
+        create_fact_check_response(edition, event) if event["comments"].present?
+      end
+    end
+
+    def create_fact_check_request(edition, event)
+      contents = {
+        email_address: event["email_address"],
+        instructions: event["instructions"],
+      }
+      details = create_whitehall_imported_entry("fact_check_request", contents)
+      create_timeline_entry(details,
+                            edition, event["created_at"], event["requestor_id"])
+    end
+
+    def create_fact_check_response(edition, event)
+      contents = {
+        email_address: event["email_address"],
+        comments: event["comments"],
+      }
+      details = create_whitehall_imported_entry("fact_check_response", contents)
+      create_timeline_entry(details, edition, event["updated_at"])
     end
 
     def create_whitehall_imported_entry(entry_type, contents = {})
--- a/spec/factories/whitehall_export/editorial_remark_event.rb
+++ b/spec/factories/whitehall_export/editorial_remark_event.rb
@@ -1,0 +1,14 @@
+# frozen_string_literal: true
+
+FactoryBot.define do
+  factory :whitehall_export_editorial_remark_event, class: Hash do
+    skip_create
+
+    sequence(:id)
+    body { "Note about the revision" }
+    sequence(:author_id)
+    created_at { Time.current.rfc3339 }
+
+    initialize_with { attributes.stringify_keys }
+  end
+end
--- a/spec/factories/whitehall_export/fact_check_event_factory.rb
+++ b/spec/factories/whitehall_export/fact_check_event_factory.rb
@@ -1,0 +1,19 @@
+# frozen_string_literal: true
+
+FactoryBot.define do
+  factory :whitehall_export_fact_check_event, class: Hash do
+    skip_create
+
+    sequence(:id)
+    sequence(:edition_id)
+    key { "redacted-1" }
+    email_address { "someone@somewhere.com" }
+    instructions { "Do something" }
+    comments { "hello" }
+    sequence(:requestor_id)
+    created_at { Time.current.rfc3339 }
+    updated_at { Time.current.rfc3339 }
+
+    initialize_with { attributes.stringify_keys }
+  end
+end
--- a/spec/factories/whitehall_export/revision_history_event.rb
+++ b/spec/factories/whitehall_export/revision_history_event.rb
@@ -4,9 +4,10 @@ FactoryBot.define do
   factory :whitehall_export_revision_history_event, class: Hash do
     skip_create
 
+    sequence(:id)
     event { "create" }
     state { "draft" }
-    whodunnit { 1 }
+    sequence(:whodunnit)
     created_at { Time.current.rfc3339 }
 
     initialize_with { attributes.stringify_keys }
--- a/spec/factories/whitehall_export/revision_history_event.rb
+++ b/spec/factories/whitehall_export/revision_history_event.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 FactoryBot.define do
-  factory :revision_history_event, class: Hash do
+  factory :whitehall_export_revision_history_event, class: Hash do
     skip_create
 
     event { "create" }
--- a/spec/lib/whitehall_importer/create_edition_spec.rb
+++ b/spec/lib/whitehall_importer/create_edition_spec.rb
@@ -62,7 +62,8 @@ RSpec.describe WhitehallImporter::CreateEdition do
 
     it "does not mark the edition as synced with Publishing API" do
       whitehall_edition = build(:whitehall_export_edition)
-      edition = described_class.call(document_import: document_import, whitehall_edition: whitehall_edition)
+      edition = described_class.call(document_import: document_import,
+                                     whitehall_edition: whitehall_edition)
 
       expect(edition.revision_synced).to be false
     end
@@ -88,59 +89,83 @@ RSpec.describe WhitehallImporter::CreateEdition do
       expect(edition.editors.count).to eq(2)
     end
 
-    it "sets the timeline history for revision history" do
-      freeze_time do
-        first_user = create(:user)
-        second_user = create(:user)
+    context "has editorial remarks" do
+      let(:user) { create(:user) }
+      let(:whitehall_user_id) { rand(100) }
+      let(:user_ids) { { whitehall_user_id => user.id } }
+      let(:create_event) do
+        build(:whitehall_export_revision_history_event,
+              event: "create",
+              whodunnit: whitehall_user_id,
+              created_at: 1.week.ago.noon)
+      end
 
-        user_ids = {
-          1 => first_user.id,
-          2 => second_user.id,
-        }
-
-        whitehall_edition = build(
-          :whitehall_export_edition,
-          revision_history: [
-            build(:whitehall_export_revision_history_event,
-                  whodunnit: 1,
-                  event: "create",
-                  state: "draft",
-                  created_at: 2.days.ago),
-            build(:whitehall_export_revision_history_event,
-                  whodunnit: 2,
-                  event: "update",
-                  state: "published",
-                  created_at: 1.day.ago),
-            build(:whitehall_export_revision_history_event,
-                  whodunnit: 1,
-                  event: "update",
-                  state: "superseded",
-                  created_at: 2.days.ago),
-          ],
-        )
-
+      it "imports a create revision history event" do
+        whitehall_edition = build(:whitehall_export_edition,
+                                  revision_history: [create_event])
         edition = described_class.call(document_import: document_import,
                                        whitehall_edition: whitehall_edition,
                                        user_ids: user_ids)
+        timeline_entry = edition.timeline_entries.first
+        expect(timeline_entry.attributes)
+          .to match a_hash_including("entry_type" => "whitehall_migration",
+                                     "created_by_id" => user.id,
+                                     "created_at" => 1.week.ago.noon)
+        expect(timeline_entry.details.attributes)
+          .to match a_hash_including("entry_type" => "first_created",
+                                     "contents" => {})
+      end
 
-        expect(edition.timeline_entries.first).to be_whitehall_migration
-        expect(edition.timeline_entries.first.details).to be_first_created
-        expect(edition.timeline_entries.first.created_at).to eq(2.days.ago)
-        expect(edition.timeline_entries.first.created_by_id)
-          .to eq(first_user.id)
+      it "imports a published revision history event" do
+        publish_event = build(:whitehall_export_revision_history_event,
+                              whodunnit: whitehall_user_id,
+                              event: "update",
+                              state: "published",
+                              created_at: 1.day.ago.noon)
+        whitehall_edition = build(:whitehall_export_edition,
+                                  revision_history: [create_event,
+                                                     publish_event])
+        edition = described_class.call(document_import: document_import,
+                                       whitehall_edition: whitehall_edition,
+                                       user_ids: user_ids)
+        timeline_entry = edition.timeline_entries.order(:created_at).last
+        expect(timeline_entry.attributes)
+          .to match a_hash_including("entry_type" => "whitehall_migration",
+                                     "created_by_id" => user.id,
+                                     "created_at" => 1.day.ago.noon)
+        expect(timeline_entry.details.attributes)
+          .to match a_hash_including("entry_type" => "published",
+                                     "contents" => {})
+      end
 
-        expect(edition.timeline_entries.last).to be_whitehall_migration
-        expect(edition.timeline_entries.last.details).to be_published
-        expect(edition.timeline_entries.last.created_at).to eq(1.day.ago)
-        expect(edition.timeline_entries.last.created_by_id)
-          .to eq(second_user.id)
+      it "imports an editorial remark event" do
+        event = build(:whitehall_export_editorial_remark_event,
+                      author_id: user.id,
+                      body: "Another note",
+                      created_at: 1.day.ago.noon)
+        whitehall_edition = build(:whitehall_export_edition,
+                                  revision_history: [create_event],
+                                  editorial_remarks: [event])
+        edition = described_class.call(document_import: document_import,
+                                       whitehall_edition: whitehall_edition,
+                                       user_ids: user_ids)
+        timeline_entry = edition.timeline_entries.order(:created_at).last
+        expect(timeline_entry.attributes)
+          .to match a_hash_including("entry_type" => "whitehall_migration",
+                                     "created_by_id" => user.id,
+                                     "created_at" => 1.day.ago.noon)
+        expect(timeline_entry.details.attributes)
+          .to match a_hash_including("entry_type" => "internal_note",
+                                     "contents" => { "body" => "Another note" })
       end
     end
 
     context "when importing an access limited edition" do
       it "creates an access limit" do
-        whitehall_edition = build(:whitehall_export_edition, access_limited: true)
-        edition = described_class.call(document_import: document_import, whitehall_edition: whitehall_edition)
+        whitehall_edition = build(:whitehall_export_edition,
+                                  access_limited: true)
+        edition = described_class.call(document_import: document_import,
+                                       whitehall_edition: whitehall_edition)
 
         expect(edition.access_limit).to be_present
         expect(edition.access_limit).to be_tagged_organisations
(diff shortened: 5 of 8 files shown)


lib/whitehall_importer/create_edition.rb
```diff
@@ -194,5 +196,66 @@ def create_revision_history(edition)
...
+
+    def create_notes(edition)
+      whitehall_edition["editorial_remarks"].each do |event|
+        details = TimelineEntry::WhitehallImportedEntry.create!(
+          entry_type: "internal_note",
+          body: {
+            "body": event["body"],
+          }.to_json,
+        )
+        TimelineEntry.create!(
```
emmabeynon: Since we've got a few instances of creating a TimelineEntry, perhaps it would be worth DRYing it up by extracting it out into a `whitehall_migration_timeline_entry` (or similar) method to which we pass the edition, `created_at`, `created_by` and `details`.

lib/whitehall_importer/create_edition.rb
```diff
@@ -194,5 +196,66 @@ def create_revision_history(edition)
...
+        details: details,
+      )
+    end
+
+    def create_fact_check_response(edition, event)
+      details = TimelineEntry::WhitehallImportedEntry.create!(
+        entry_type: "fact_check_response",
+        body: {
+          "email_address": event["email_address"],
+          "comments":      event["comments"],
```
emmabeynon: Looks like we have some excess whitespace here.

spec/lib/whitehall_importer/create_edition_spec.rb
```diff
@@ -99,8 +99,14 @@
...
-            build(:revision_history_event, whodunnit: 2, event: "update", state: "published", created_at: 1.day.ago),
+            build(:revision_history_event, whodunnit: 1, event: "create", state: "draft", created_at: 5.days.ago),
+            build(:revision_history_event, whodunnit: 2, event: "update", state: "published", created_at: 4.days.ago),
+          ],
+          editorial_remarks: [
+            build(:note_event, author_id: 1, body: "Another note", state: "published", created_at: 2.days.ago),
+          ],
+          fact_check_requests: [
+            build(:fact_check_event, requestor_id: 2, email_address: "someone@somewhere.com", comments: "Hello World", state: "published", created_at: 3.days.ago, updated_at: 1.day.ago),
           ],
```
emmabeynon: Instead of passing static values here, I think we could assign each of these events to a variable, pass them to the edition and refer to them in the expect statements.  I don't think we care too much about what the values are, we just care that they're assigned to the correct attributes in the timeline entry. For example:
```
editiorial_remark = build(:note_event)
whitehall_edition = build(
... #some code
editorial_remarks: [editorial remark],
# rest of the code
```
`expect(edition.timeline_entries.third.created_at).to eq(editorial_remark.created_at)`

lib/whitehall_importer/create_edition.rb
```diff
@@ -194,5 +196,66 @@ def create_revision_history(edition)
...
+        create_fact_check_response(edition, event) if event["comments"].present?
+      end
+    end
+
+    def create_fact_check_request(edition, event)
+      details = TimelineEntry::WhitehallImportedEntry.create!(
+        entry_type: "fact_check_request",
+        body: {
+          "email_address": event["email_address"],
+          "instructions":  event["instructions"],
```
emmabeynon: We've got some excess whitespace here.

spec/lib/whitehall_importer/create_edition_spec.rb
```diff
@@ -111,13 +117,27 @@
...
+        expect(edition.timeline_entries.third.created_by_id).to eq(first_user.id)
+
+        expect(edition.timeline_entries.fourth).to be_whitehall_migration
+        expect(edition.timeline_entries.fourth.details).to be_fact_check_request
+        expect(edition.timeline_entries.fourth.created_at).to eq(3.days.ago)
+        expect(edition.timeline_entries.fourth.created_by_id).to eq(second_user.id)
+
+        expect(edition.timeline_entries.fifth).to be_whitehall_migration
+        expect(edition.timeline_entries.fifth.details).to be_fact_check_response
+        expect(edition.timeline_entries.fifth.created_at).to eq(1.day.ago)
```
kevindew: This looks a bit like we're testing lots of cases all in a single test whereas I think we probably want to have context or describe blocks to break these things up, as it seems far more than we can cover with `it "sets the timeline history"`


for example:

`context "when the edition has editorial remarks"` or `describe "editorial remarks"`

It also seems a bit of a smell to have quite as many expectations as we have going on and I wonder if we're missing some assertions (such as the text of internal notes and fact checks).

You might be able to switch some out like:

```
timeline_entry = edition.timeline_entries.third
expect(timeline_entry.attributes).to match a_hash_including(type: :whitehall_migration, created_at: 2.days.ago, created_by_id: first_user.id)
expect(timeline_entry.details.attributes).to match a_hash_including(type: :internal_note, body: a_hash_including({ "body" => "note" })
```

lib/whitehall_importer/create_edition.rb
```diff
@@ -197,5 +199,66 @@ def create_revision_history(edition)
         )
       end
     end
+
+    def create_notes(edition)
+      whitehall_edition["editorial_remarks"].each do |event|
+        details = TimelineEntry::WhitehallImportedEntry.create!(
+          entry_type: "internal_note",
+          body: {
+            "body": event["body"],
```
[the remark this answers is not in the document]
gclssvglx: Will rename "body" to "contents" so, `"contents": event["body"],`

lib/whitehall_importer/create_edition.rb
gclssvglx: @emmabeynon Thanks for the comments.  Will pull this and the `details = TimelineEntry::WhitehallImportedEntry...` into methods and DRY up.

lib/whitehall_importer/create_edition.rb
gclssvglx: Sorry, lining up is a personal pref, will adjust my brain ;-)

lib/whitehall_importer/create_edition.rb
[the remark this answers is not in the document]
kevindew: Would probably more consistent with existing naming to rename the column. Then the column would be consistent with a different JSON column we have and we could continue using body as per note.

`TimelineEntry.create!(entry_type: "internal_note", contents: { body: event["body"] })`

lib/whitehall_importer/create_edition.rb
```diff
@@ -33,7 +33,9 @@ def call
                   create_edition(status: status, current: current, edition_number: edition_number)
                 end
 
-      create_revision_history(edition)
+      create_revision_history(edition) if whitehall_edition["revision_history"]
+      create_notes(edition) if whitehall_edition["editorial_remarks"]
+      create_fact_checks(edition) if whitehall_edition["fact_check_requests"]
```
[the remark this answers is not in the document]
gclssvglx: I believe we can remove the conditionals as you suggest, as these elements will always either be an empty array or an array with 1+ element - see: https://github.com/alphagov/whitehall/blob/917085dca58fb58dbf65c7e226ad445cc1f27bdd/app/presenters/document_export_presenter.rb#L84, https://github.com/alphagov/whitehall/blob/917085dca58fb58dbf65c7e226ad445cc1f27bdd/app/presenters/document_export_presenter.rb#L92 and https://github.com/alphagov/whitehall/blob/917085dca58fb58dbf65c7e226ad445cc1f27bdd/app/presenters/document_export_presenter.rb#L98

spec/lib/whitehall_importer/create_edition_spec.rb
gclssvglx: Have updated as requested.

spec/lib/whitehall_importer/create_edition_spec.rb
gclssvglx: Have updated as requested.

spec/lib/whitehall_importer/create_edition_spec.rb
```diff
@@ -86,38 +86,148 @@
...
-    it "sets the timeline history for revision history" do
-      freeze_time do
-        first_user = create(:user)
-        second_user = create(:user)
+    context "has editorial remarks" do
+      let(:first_user) { create(:user) }
+      let(:second_user) { create(:user) }
 
-        user_ids = {
+      let(:user_ids) {
```
emmabeynon: Typically if the block spans multiple lines we use `do...end` instead of curly braces.

gclssvglx: No worries - all changed.  Thanks.

lib/whitehall_importer/create_edition.rb
```diff
@@ -184,18 +186,62 @@ def create_revision_history(edition)
...
+        details = create_timeline_import_entry("internal_note", body)
+        create_timeline_entry(details, edition, event["created_at"], user_ids[event["author_id"]])
+      end
+    end
+
+    def create_fact_checks(edition)
+      whitehall_edition["fact_check_requests"].each do |event|
+        create_fact_check_request(edition, event)
+        create_fact_check_response(edition, event) if event["comments"].present?
+      end
```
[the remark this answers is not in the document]
gclssvglx: Yeah, nice (I think) - odd as `created_at` is set for request and `updated_at` for response.

lib/whitehall_importer/create_edition.rb
```diff
@@ -184,18 +186,62 @@ def create_revision_history(edition)
...
+        "comments": event["comments"],
+      }
+      details = create_timeline_import_entry("fact_check_response", body)
+      create_timeline_entry(details, edition, event["updated_at"])
+    end
+
+    def create_timeline_import_entry(entry_type, contents = {})
+      TimelineEntry::WhitehallImportedEntry.create!(
+        entry_type: entry_type,
+        contents: contents.to_json,
```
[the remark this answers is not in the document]
gclssvglx: Without `to_json` I get failing tests.  I haven't looked at the error in too much detail, but I'd like to get a sense of how big a deal removing it is before spending time on it.

lib/whitehall_importer/create_edition.rb
```diff
@@ -184,18 +186,62 @@ def create_revision_history(edition)
...
+        details = create_timeline_import_entry(entry_type)
+        create_timeline_entry(details, edition, event["created_at"], user_ids[event["whodunnit"]])
       end
     end
+
+    def create_notes(edition)
+      whitehall_edition["editorial_remarks"].each do |event|
+        body = {
+          "body": event["body"],
+        }
```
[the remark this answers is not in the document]
gclssvglx: Updated - thanks Kev

lib/whitehall_importer/create_edition.rb
```diff
@@ -184,18 +186,62 @@ def create_revision_history(edition)
...
-        )
+        details = create_timeline_import_entry(entry_type)
+        create_timeline_entry(details, edition, event["created_at"], user_ids[event["whodunnit"]])
       end
     end
+
+    def create_notes(edition)
+      whitehall_edition["editorial_remarks"].each do |event|
+        body = {
+          "body": event["body"],
```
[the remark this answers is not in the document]
gclssvglx: Updated - thanks Kev

lib/whitehall_importer/create_edition.rb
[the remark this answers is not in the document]
kevindew: What error do you get? you should be able to pass a hash in, so your error might be a sign of a bigger problem. Makes me wonder if somehow the column is a string 🤔

kevindew:
Argh sorry Graham Github seems to have lost the big comment I made on the last review :( just going to try find it.

lib/whitehall_importer/create_edition.rb
```diff
@@ -184,18 +186,62 @@ def create_revision_history(edition)
...
-        )
+        details = create_timeline_import_entry(entry_type)
+        create_timeline_entry(details, edition, event["created_at"], user_ids[event["whodunnit"]])
       end
     end
+
+    def create_notes(edition)
+      whitehall_edition["editorial_remarks"].each do |event|
+        contents = {
+          contents: event["body"],
```
kevindew: shouldn't this be a body key here?

spec/lib/whitehall_importer/create_edition_spec.rb
```diff
@@ -86,38 +86,148 @@
...
+        expect(timeline_entry.attributes).to match a_hash_including(
+          "entry_type" => "whitehall_migration",
+          "created_at" => draft_revision["created_at"],
+          "created_by_id" => first_user.id,
+        )
+        expect(timeline_entry.details.attributes).to match a_hash_including(
+          "entry_type" => "first_created",
+          "contents" => "{}",
+        )
+      end
```
kevindew: Argh I wrote a long comment here before but it seems lost in an abyss somewhere.

I found this test, and the other ones, somewhat confusing and inconsistent with the other tests in this file because the actual work for testing the code is done in the `let(:edition)` block. Whereas other tests in this file (and mostly across the app) try to focus on executing the code under test in the test itself. 

I think these tests can be improved by doing more in the test itself. I think this would have benefits in making the tests isolated (at the moment you need to read a lot of code to understand why say an editiorial remark mark looks up the third timeline entry of an edition) and reduce the amount of lets to go through which can be a bit of a smell (see: https://thoughtbot.com/blog/lets-not)

As an example:

```
it "imports a create revision history as a first_created event" do
  create_revision_event = build(:revision_history_event,
                                whodunnit: 1,
                                event: "create",
                                state: "draft",
                                created_at: 5.days.ago.noon)
  whitehall_edition = build(:whitehall_export_edition, revision_history: [create_revision])
  edition = described_class.call(document_import: document_import,
                                 whitehall_edition: whitehall_edition,
                                 user_ids: user_ids)
  timeline_entry = edition.timeline_entries.first
  expect(timeline_entry.attributes)
    .to match a_hash_including("entry_type" => "whitehall_migration",
                               "created_at" => 5.days.ago.noon,
                               "created_by_id" => first_user.id)
  expect(timeline_entry.details.attributes)
    .to match a_hash_including("entry_type" => "first_created",
                               "contents" => "{}")
end
```

There's a good chance this may increase lines of code slightly but I think it'll be beneficial to the overall code.

spec/lib/whitehall_importer/create_edition_spec.rb
```diff
@@ -86,38 +86,148 @@
...
+        timeline_entry = edition.timeline_entries.fourth
+        expect(timeline_entry.attributes).to match a_hash_including(
+          "entry_type" => "whitehall_migration",
+          "created_at" => fact_check["created_at"],
+          "created_by_id" => second_user.id,
+        )
+        expect(timeline_entry.details.attributes).to match a_hash_including(
+          "entry_type" => "fact_check_request",
+          "contents" => "{\"email_address\":\"someone@somewhere.com\",\"instructions\":\"Do something\"}",
+        )
```
kevindew: I think something has gone wrong for this to be a string. I'd expect this to be a hash.

lib/whitehall_importer/create_edition.rb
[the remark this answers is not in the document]
gclssvglx: Yes, looking at the error (and there's many of 'em) - it does look like the JSON column is being represented as a String at various points.  I'll look into it...

```
  1) WhitehallImporter::CreateEdition#call has editorial remarks has a draft revision
     Failure/Error:
       expect(timeline_entry.details.attributes).to match a_hash_including(
         "entry_type" => "first_created",
         "contents" => "{}",
       )
     
       expected {"contents"=>{}, "created_at"=>2020-01-29 15:02:27.442141000 +0000, "entry_type"=>"first_created", "id"=>1329} to match (a hash including {"entry_type" => "first_created", "contents" => "{}"})
       Diff:
       @@ -1,2 +1,5 @@
       -(a hash including {"entry_type" => "first_created", "contents" => "{}"})
       +"contents" => {},
       +"created_at" => 2020-01-29 15:02:27.442141000 +0000,
       +"entry_type" => "first_created",
       +"id" => 1329,
       
     # ./spec/lib/whitehall_importer/create_edition_spec.rb:176:in 'block (4 levels) in <top (required)>'
```

lib/whitehall_importer/create_edition.rb
gclssvglx: As in `body: event["body"]`?  I think I'm getting a tad confused, but the only place we really need `body` is in the `event`, so I've just updated everything else (to be consistent) to `contents`.

spec/lib/whitehall_importer/create_edition_spec.rb
gclssvglx: Hopefully, this is now resolved!  Thanks - nice spot.

spec/lib/whitehall_importer/create_edition_spec.rb
gclssvglx: Thanks for the suggestion and keeping me on track!  Tests updated.

gclssvglx:
@emmabeynon @kevindew Would you mind re-reviewing this for me please?

spec/lib/whitehall_importer/create_edition_spec.rb
```diff
@@ -86,38 +86,109 @@
...
+                      comments: "Hello World",
+                      created_at: 2.days.ago.noon,
+                      updated_at: 1.day.ago.noon)
+        whitehall_edition = build(:whitehall_export_edition,
+                                  fact_check_requests: [event])
+        edition = described_class.call(document_import: document_import,
+                                       whitehall_edition: whitehall_edition,
+                                       user_ids: user_ids)
+        # The first entry is the creation event - the response created_by_id is nil
+        timeline_entry = edition.timeline_entries.order(:created_at).second
```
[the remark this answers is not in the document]
gclssvglx: Actually, there are only 2 entries in this case: the creation event, and the fact check event, so there should be no risk of a third entry. The fact check event is a single event that covers both the fact check request and response. Granted this is a tad counter-intuitive - hence my comment.  We could use `.last` here - and happy to do that if you prefer - but `.second` and `.last` are the same entry.  As we've already tested the creation event elsewhere, we're not interested in that event here, so we can ignore the first event.

spec/factories/whitehall_export/fact_check_event_factory.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+FactoryBot.define do
+  factory :fact_check_event, class: Hash do
+    skip_create
+
+    id { 1 }
+    edition_id { 1 }
```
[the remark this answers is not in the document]
gclssvglx: I added this after our discussion on keeping the API the same, my thinking was "yes, we know it, but in order to keep the API consistent we should have an edition_id in the factory".  If my thinking is incorrect I'm happy to remove (or make a sequence).  Please let me know your preference.

spec/factories/whitehall_export/editorial_remark_event.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+FactoryBot.define do
+  factory :editorial_remark_event, class: Hash do
+    skip_create
+
+    id { 1 }
```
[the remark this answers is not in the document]
gclssvglx: Updated as requested.

spec/factories/whitehall_export/fact_check_event_factory.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+FactoryBot.define do
+  factory :fact_check_event, class: Hash do
+    skip_create
+
+    id { 1 }
```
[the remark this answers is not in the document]
gclssvglx: Updated as requested.

spec/factories/whitehall_export/revision_history_event.rb
```diff
@@ -4,6 +4,7 @@
   factory :revision_history_event, class: Hash do
     skip_create
 
+    id { 1 }
```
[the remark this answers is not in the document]
gclssvglx: Updated as requested.

spec/factories/whitehall_export/fact_check_event_factory.rb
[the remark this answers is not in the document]
gclssvglx: Updated as requested.

gclssvglx:
@kevindew...

I've addressed the long lines as requested.  I guess there is a question there about cleaning up other parts of the overall file that you've not changed (when do you/should you do this?).  I've opted to only update and clean the methods where changes have been made as part of this piece of work. Maybe we should add the Rubocop fixed-width text rule to force the issue and not rely on personal preference? 

You asked me to separate into distinct commits.  I thought about a logical split to do this and couldn't really see one that made sense (to me), or told the story better than simply test vs application code changes (so, 2 commits).  Have left as one commit for now - pending a better suggestion.

kevindew:
> @kevindew...
> 
> I've addressed the long lines as requested. I guess there is a question there about cleaning up other parts of the overall file that you've not changed (when do you/should you do this?). I've opted to only update and clean the methods where changes have been made as part of this piece of work. Maybe we should add the Rubocop fixed-width text rule to force the issue and not rely on personal preference?

Yes that sounds reasonable. It's unfortunate that we don't use a rubocop rule for this, there used to be one back in the day for GDS but it was switched off as it was a bit too restrictive (80 chars). I don't think it'd be a fun ride getting consensus on what an absolute limit should be, however the working definition is we should aim for 80 characters see: https://github.com/alphagov/styleguides/blob/7a9045bf653e6f12d614217b30d70b7798ece5bd/ruby.md#general (this styleguide has since been removed but since nothing has superseded this advice it should still be followed). I'd suggest setting your text editor to have a 80 char margin and try follow it where appropriate for consistent GDS Ruby code. 

> You asked me to separate into distinct commits. I thought about a logical split to do this and couldn't really see one that made sense (to me), or told the story better than simply test vs application code changes (so, 2 commits). Have left as one commit for now - pending a better suggestion.

Right, I'd go for the commits to be like:

1) Rename revision_history_event factory
2) Change whodunnit to be a sequence
3) Add imported internal notes to the codebase
4) Add imported fact checks to the codebase.

While you're changing the code could you change the [`create_timeline_import_entry` method](https://github.com/alphagov/content-publisher/pull/1688/files#diff-81283073dacc62b4b7c3997787cc5288R231) - it caught me a bit by a surprise in as I wasn't sure what a timeline_import_entry was. I'd suggest going for `create_whitehall_imported_entry` since it creates a TimelineEntry::WhitehallImportedEntry model.

Thanks Graham this looks just about done.

gclssvglx:
@emmabeynon @kevindew Would you mind re-reviewing this for me.  Thanks.

gclssvglx:
> LGTM ⭐️

Thanks Emma - and thanks for all your help and comments on this 😃

lib/whitehall_importer/create_edition.rb
```diff
@@ -184,18 +186,67 @@ def create_revision_history(edition)
...
+        create_fact_check_request(edition, event)
+        create_fact_check_response(edition, event) if event["comments"].present?
+      end
+    end
+
+    def create_fact_check_request(edition, event)
+      contents = {
+          email_address: event["email_address"],
+          instructions: event["instructions"],
+        }
```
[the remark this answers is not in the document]
gclssvglx: Damn - thanks Kevin!  Fixed.

gclssvglx:
> Nice one Graham.
> 
> Not a blocker for this one but I'd suggest adding more information on the commit messages on why we are making the changes we are.

@kevindew Thanks for all your help and comments on this Kevin.  Sorry, it's been a bit of a pain!  I've updated the commit messages and fixed the indentation issue.  Would you like to re-review or are you happy for me to merge?

lib/whitehall_importer/create_edition.rb
```diff
@@ -184,18 +186,67 @@ def create_revision_history(edition)
...
+      TimelineEntry::WhitehallImportedEntry.create!(
+        entry_type: entry_type,
+        contents: contents,
+      )
+    end
+
+    def create_timeline_entry(details, edition, created_at, created_by = nil)
+      TimelineEntry.create!(
+        entry_type: :whitehall_migration,
+        created_by_id: created_by,
```
brucebolt: I'm a bit late to the party here, but this will try to insert the user's Whitehall user ID, rather than their Content Publisher ID (which can be looked up in `user_ids`, i.e. `user_ids[created_by]`).

gclssvglx: Thanks for this Bruce.  Resolved via PR: https://github.com/alphagov/content-publisher/pull/1738.

PR #1689, opened 2020-01-23
Ignore superseded revision history state changes from Whitehall

brucebolt:
Whitehall has a change of state to 'superseded' which happens when an existing published edition is replaced with a new published edition. The existing one is given a new state of superseded, however this is not to be shown to users.

We should therefore skip any history entries that have a state change to superseded, as detailed in https://docs.google.com/spreadsheets/d/1I7TbA9glyFyhyWq5wQJAGykkXFwOJjyd--kLCycXbSY/edit#gid=0.

This fix allows us to import 53 of NDA's documents which are currently failing.

Trello: https://trello.com/c/3AV7mZTG

### DIFF
--- a/lib/whitehall_importer/edition_history.rb
+++ b/lib/whitehall_importer/edition_history.rb
@@ -91,6 +91,7 @@ module WhitehallImporter
       when "rejected" then "rejected"
       when "scheduled" then "scheduled"
       when "submitted" then "submitted"
+      when "superseded" then nil
       when "withdrawn" then "withdrawn"
       else
         raise(AbortImportError, "Edition history has an unsupported state #{event['state']}")
--- a/lib/whitehall_importer/create_edition.rb
+++ b/lib/whitehall_importer/create_edition.rb
@@ -181,8 +181,11 @@ module WhitehallImporter
 
     def create_revision_history(edition)
       whitehall_edition["revision_history"].each do |event|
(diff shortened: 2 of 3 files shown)


lib/whitehall_importer/edition_history.rb
```diff
@@ -91,6 +91,7 @@ def imported_entry_type(event, edition_number)
       when "rejected" then "rejected"
       when "scheduled" then "scheduled"
       when "submitted" then "submitted"
+      when "superseded" then "superseded"
```
kevindew: maybe we want to return nil here and bail out on ones that are nil rather than being specific for superseded?

brucebolt: The only other case would be 'deleted'.  But I've been unable to find any examples of this to see how Whitehall presents the history.  It does not affect the NDA.

kevindew: Yeah I think if it's deleted it's no longer included in Whitehall's default scope so we don't export them. Scenario we'd hit them is if someone un-deleted something. Not sure if that's ever happened though.

brucebolt: I haven't come across any cases like that.  I'll change this to return nil, then it'll be simpler to update for any future instances of this.

PR #1690, opened 2020-01-23
Iterate our route config after a controller reshuffle

JonathanHallam:
https://github.com/alphagov/content-publisher/pull/1681 introduced some inconsistency with our routes. This fixes that.

The above pull request made it so that we had the same paths going to different controllers based on the request method. This makes paths consistent with controllers and updates the tests accordingly.

### DIFF
--- a/app/views/content/edit.html.erb
+++ b/app/views/content/edit.html.erb
@@ -6,14 +6,17 @@
   <% content_for :title, t("content.edit.title", title: @edition.title_or_fallback) %>
 <% end %>
 
-<%= form_for @edition.document,
-  html: { class: "app-c-contextual-guidance__form" },
-  data: {
-    module: "edit-document-form",
-    "url-preview-path": generate_path_path,
-    "warn-before-unload": "true",
-    gtm: "edit-document-submit"
-  } do |f| %>
+<%= form_tag(
+      content_path(@edition.document),
+      method: :patch,
+      html: { class: "app-c-contextual-guidance__form" },
+      data: {
+        module: "edit-document-form",
+        "url-preview-path": generate_path_path(@edition.document),
+        "warn-before-unload": "true",
+        gtm: "edit-document-submit"
+      }
+    ) do |f| %>
 
   <%
     title_contextual_guidance = {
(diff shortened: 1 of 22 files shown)


app/views/content/edit.html.erb
```diff
@@ -6,14 +6,17 @@
...
-<%= form_for @edition.document,
-  html: { class: "app-c-contextual-guidance__form" },
-  data: {
-    module: "edit-document-form",
-    "url-preview-path": generate_path_path,
-    "warn-before-unload": "true",
-    gtm: "edit-document-submit"
-  } do |f| %>
+<%= form_tag(
+      content_path,
```
kevindew: it'd be good to do `content_path(@edition.document)`. Using Rails' magic context can end up being inconsistent with other places.

app/views/content/edit.html.erb
```diff
@@ -6,14 +6,17 @@
...
+      method: :patch,
+      html: { class: "app-c-contextual-guidance__form" },
+      data: {
+        module: "edit-document-form",
+        "url-preview-path": generate_path_path,
```
kevindew: I know you didn't introduce but would be good to pass in the document here too.

spec/requests/documents_spec.rb
```diff
@@ -3,9 +3,9 @@
 RSpec.describe "Documents" do
```
kevindew: We should move some of these around to match the controllers. So we should have a "Content" request test for those checks and delete_draft ones should be moved to Editions request tests. We'll also have to update the paths in the describe methods.

app/views/content/edit.html.erb
JonathanHallam: nice one, sorted

app/views/content/edit.html.erb
JonathanHallam: will do 👍

spec/requests/documents_spec.rb
JonathanHallam: All sorted after in person conversation with you :)

PR #1691, opened 2020-01-23
Ensure document creation is rolled back when an import exception occurs

emmabeynon:
For https://trello.com/c/g0GfY8PF/1364-fix-import-script

Our problem is that we mutate the document_import object during the
import, which conflicts with our 'all or nothing' strategy for
transactions. What happens is that we create a document and various
other models in the database and then, when a transaction fails, roll
them back. However as we set the document attribute on a document_import
the rolled back model still exists in memory and when we save the
document_import it tries to save the document model meaning we at best
get an exception and at worst get a half created document.

An approach to resolve this is to reload the document_import object at
the point of the exception occurring.

A different, more complex context based approach, is considered in
https://github.com/alphagov/content-publisher/compare/atomic-imports?expand=1

Couple of other tweaks this work did:

- Move setting a document as imported to be in the transaction. That way, if
  this step raises an exception we haven't persisted a document but failed to
set the status.
- We no longer need `WhitehallImporter::Import#call` to return a document to
  `WhitehallImporter`, since no longer handle assigning the document to the
document import there. This change caused some tests to fail as they were
relying on receiving the document from `WhitehallImporter::Import#call` to
check various values on the document. These have been updated.

### DIFF
--- a/spec/lib/whitehall_importer/import_spec.rb
+++ b/spec/lib/whitehall_importer/import_spec.rb
@@ -18,15 +18,16 @@ RSpec.describe WhitehallImporter::Import do
     it "aborts if a document already exists" do
       content_id = create(:document).content_id
       import_data = build(:whitehall_export_document, content_id: content_id)
-      document_import = build(:whitehall_migration_document_import, payload: import_data)
+      document_import = create(:whitehall_migration_document_import, payload: import_data)
       expect { described_class.call(document_import) }
         .to raise_error(WhitehallImporter::AbortImportError)
     end
 
     it "sets the document as being imported from Whitehall" do
-      document = described_class.call(build(:whitehall_migration_document_import))
+      document_import = build(:whitehall_migration_document_import, document: nil)
+      described_class.call(document_import)
 
-      expect(document).to be_imported_from_whitehall
+      expect(document_import.document).to be_imported_from_whitehall
     end
 
     it "sets the timeline entry as Imported from Whitehall" do
@@ -37,10 +38,11 @@ RSpec.describe WhitehallImporter::Import do
     end
 
     it "associates the created document with the import record" do
-      import_record = build(:whitehall_migration_document_import)
-      document = described_class.call(import_record)
+      import_record = build(:whitehall_migration_document_import, document: nil)
+      whitehall_document_content_id = import_record.payload["content_id"]
+      described_class.call(import_record)
 
-      expect(import_record.document).to eq(document)
+      expect(import_record.document.content_id).to eq(whitehall_document_content_id)
     end
 
     it "creates users who have never logged into Content Publisher" do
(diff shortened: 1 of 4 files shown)


kevindew:
Nice job - only thing I'd add is it'd be nice to capture this in a test so it's captured for regressions.

You could probably test it by raising an exception when update! gets called on the document object and then asserting that document_import.changed? is false

emmabeynon:
@kevindew I've added the test in.  I found I needed to assert that the error was raised in order to effectively rescue it when the test runs; do you know if there's another way to handle it?  I couldn't seem to find an alternative approach.

spec/lib/whitehall_importer_spec.rb
```diff
@@ -95,12 +95,12 @@
...
       end
 
-      it "marks the import as imported" do
-        WhitehallImporter.import(whitehall_migration_document_import)
-        expect(whitehall_migration_document_import).to be_imported
+      it "does not raise an error" do
+        expect { WhitehallImporter.import(whitehall_migration_document_import) }
+          .not_to raise_error
       end
     end
```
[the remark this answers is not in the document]
emmabeynon: I've removed this now 👍

emmabeynon:
@kevindew Thanks for that - I've amended the test as you've suggested.

spec/lib/whitehall_importer/import_spec.rb
```diff
@@ -163,8 +165,20 @@
...
+
+    it "does not update the document import if the transaction fails" do
+      document_import = create(:whitehall_migration_document_import)
+      updated_at = document_import.updated_at
+      document_import.updated_at = 2.days.from_now
+      allow(document_import).to receive(:update!).and_raise("forced error")
+
+      expect { described_class.call(document_import) }.to raise_error("forced error")
+      expect(document_import.changed?).to be false
+      expect(document_import.updated_at.to_i).to eq(updated_at.to_i)
```
kevindew: I was a bit puzzled by the to_i, I tried it on my machine and this passes with `expect(document_import.updated_at).to eq(updated_at)` so I don't think that's needed.

emmabeynon: I added `to_i` as it wasn't passing for me locally.  I got:
```
expected: 2020-01-23 15:26:25.365678900 +0000
            got: 2020-01-23 15:26:25.365678000 +0000
```

I found a few solutions that suggested using `to_i` to overcome it e.g. https://stackoverflow.com/questions/20403063/trouble-comparing-time-with-rspec

kevindew: Ah I see - sorry I didn't clock that as a precision check. Maybe I got lucky when I ran the test 🤔 time precision is a pain.

We've got around this in other parts of the suite by using times that are fixed to a second. I'd suggest switching `2.days.from_now` to `Date.tomorrow.noon` - which is consistent with what we do in the factories: https://github.com/alphagov/content-publisher/commit/6edd08fcacd10a16d18124840799f276da7e493e so then you'll not need to convert to integer.

emmabeynon: Seems like just passing a specific date to the factory made it pass without the need for `to_i` 👍

emmabeynon: Seems like just passing a specific date to the factory made it pass without the need for to_i 👍

PR #1692, opened 2020-01-23
Use keyword args for all services

benthorner:
https://github.com/alphagov/content-publisher/issues/1256
https://trello.com/c/f1QaCH7x/231-our-services-directory-is-a-mess

Previously we allowed a mixture of plain and keyword arguments for classes
in 'app/services'. This changes all services to use keyword arguments only,
noting the increase in verbosity in certain places is balanced by the
increased meaning, since we now say what each of the arguments is (for). A
good example of this is:

    job.arguments.first vs. edition_id: job.arguments.first

### DIFF
--- a/app/services/delete_draft_edition_service.rb
+++ b/app/services/delete_draft_edition_service.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 class DeleteDraftEditionService < ApplicationService
-  def initialize(document, user)
+  def initialize(document:, user:)
     @document = document
     @user = user
   end
@@ -58,7 +58,7 @@ private
       end
     end
 
-    AssignEditionStatusService.call(edition, user, :discarded)
+    AssignEditionStatusService.call(edition: edition, user: user, state: :discarded)
     edition.update!(current: false)
   end
 
--- a/app/services/edit_draft_edition_service.rb
+++ b/app/services/edit_draft_edition_service.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 class EditDraftEditionService < ApplicationService
-  def initialize(edition, user, **attributes)
+  def initialize(edition:, user:, **attributes)
     @edition = edition
     @user = user
     @attributes = attributes
--- a/app/controllers/documents_controller.rb
+++ b/app/controllers/documents_controller.rb
@@ -30,7 +30,8 @@ class DocumentsController < ApplicationController
   def generate_path
     edition = Edition.find_current(document: params[:document])
     assert_edition_state(edition, &:editable?)
-    base_path = GenerateBasePathService.call(edition.document, params[:title])
+    base_path = GenerateBasePathService.call(document: edition.document,
+                                             proposed_title: params[:title])
     render plain: base_path
   end
 
--- a/app/interactors/access_limit/update_interactor.rb
+++ b/app/interactors/access_limit/update_interactor.rb
@@ -31,7 +31,7 @@ private
 
     if LIMIT_TYPES.exclude?(limit_type)
       context.fail! if edition.access_limit.nil?
-      EditDraftEditionService.call(edition, user, access_limit: nil)
+      EditDraftEditionService.call(edition: edition, user: user, access_limit: nil)
       return
     end
 
@@ -42,7 +42,9 @@ private
                                    limit_type: limit_type,
                                    revision_at_creation: edition.revision)
 
-    EditDraftEditionService.call(edition, user, access_limit: access_limit)
+    EditDraftEditionService.call(edition: edition,
+                                 user: user,
+                                 access_limit: access_limit)
   end
 
   def check_for_issues
@@ -72,6 +74,6 @@ private
   end
 
   def update_preview
-    FailsafeDraftPreviewService.call(edition)
+    FailsafeDraftPreviewService.call(edition: edition)
   end
 end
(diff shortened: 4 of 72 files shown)


app/services/edit_draft_edition_service.rb
```diff
@@ -2,9 +2,9 @@
 
 class EditDraftEditionService < ApplicationService
   def initialize(edition:, user:, **attributes)
-    @edition = edition
-    @user = user
-    @attributes = attributes
+    super(edition: edition,
+          user: user,
+          attributes: attributes)
```
ChrisBAshton: You hint in the commit message that this was necessary, but could you explain a bit more in the commit about why it's necessary?

I guess it's because we want any additional arbitrary values to come under an `attributes` key in this one case. I wonder if it's worth changing the `EditDraftEditionService` to look for attributes under the top level rather than under `attributes`, in which case this special behaviour wouldn't be necessary?

app/services/delete_draft_edition_service.rb
```diff
@@ -1,10 +1,6 @@
 # frozen_string_literal: true
 
 class DeleteDraftEditionService < ApplicationService
-  def initialize(document:, user:)
```
ChrisBAshton: Typo in commit message: `prooves`

kevindew:
Just to record our discussion from yesterday. I'm somewhat ambivalent to this change. I can understand the desire to standardise and for consistency that provides. However I'm not sure there is a need to standardise on the argument layout for these classes.

I think of this in a similar way as any other Ruby method (and these are just other examples of them) where the author has the decision of where and when to use keyword arguments. There's always a trade off between verbosity and code that is easy to understand. 

In this work you've definitely highlighted some cases that become more readable with keyword arguments though I don't know that it's something we need to standardise on and that some of the arguments would indeed be fine without keyword args.

app/services/delete_draft_edition_service.rb
benthorner: Fixxed.

benthorner:
@kevindew @ChrisBAshton I (amusingly) share your ambivalence with respect to some of the services and/or the overall change. While we're not trying to force services to take a particular argument (like `edition`), since services can be used for operations on anything, I would find it helpful to have less choice/verbosity in the interface design for services. On the other hand, this could also be re-framed as improvements to specific services.

In order to get to a decision, can you suggest which services you'd prefer *not* to use keyword args? I'm content with all of them being this way if it creates consistency. However, if we can agree on a reduced scope for this PR, then we could just go with that. If we can't agree on which services should *not* use keyword args, then it seems simpler to just enforce them, so we don't have to repeat the decision for future (iterations of) services.

benthorner:
Having had a further discussion with @kevindew about this, we agreed to drop the overall goal of trying to provide a common interface for services, as it's not clearly beneficial e.g. DRYing-up the initializers could still lead to confusion about how instance variables are set. 

We could still benefit from the clarity of some of the interface changes proposed here, where the use of a service is unclear because the arguments are not named variables e.g. `true`, `false`, `nil`, `<some_complex_expression>`. Using this as a guide, how about the following interfaces:

```
# some of these have defaults for the keyword args
AssignEditionStatusService.call(edition, user:, status:, status_details:)
DeleteDraftDocumentService.call(document, user)
EditDraftEditionService.call(edition, user, **attributes)
FailsafeDraftPreviewService.call(edition)
GenerateBasePathService.call(document, proposed_title:)
GenerateUniqueFilenameService.call(existing_filenames:, suggested_title:)
PreviewAssetService.call(edition, asset)
PreviewDraftEditionService.call(edition, republish:)
PublishAssetsService.call(edition, live_edition:)
PublishDraftEditionService.call(edition, user, with_review:)
RemoveDocumentService.call(edition, removal)
RescueScheduledPublishingService.call(edition_id:)
ResyncDocumentService.call(document)
SchedulePublishService.call(edition, user, scheduling)
WithdrawDocumentService.call(edition, user, public_explanation:)
```

You can search this PR to see examples of confusing `Service.call`s. The only one I'm not sure about is `WithdrawDocumentService`: the usage with `params[:public_explanation]` is pretty clear and `nil` is not permitted, so it could be a normal arg; the fact that it's not a well-defined class pushes me to give it a clear naming, since it could feasibly be passed via a more complex expression.

kevindew:
Thanks Ben - I like the direction this is going 👍 

I think you're right that WithdrawDocumentService benefits from the explanation argument.

There's a few where I'm not sure we named the argument correctly in the class that look a little more exposed with this change:

**`GenerateBasePathService.call(document, proposed_title:)`**

I feel like the proposed part of this is a bit strange as it sounds like we're making the proposal to this class - whereas this service doesn't care what status the title has. It doesn't even really care that it's a title just whatever we want to generate the slug from, though generalising too much isn't fun for naming 🙂

Also we've got the max items args, not sure if you're intending to remove that or just not mentioned here

**`GenerateUniqueFilenameService.call(existing_filenames:, suggested_title:)`**

This seems to have changed from a name attribute to a title whereas I think title is the wrong word since we're dealing with a filename. 

I wonder if we got the order wrong and it should be `GenerateUniqueFilenameService.call(filename:, existing_filenames:)`

**`PublishAssetsService.call(edition, live_edition:)`**

I think the naming on this has become less clear since we also use this service to sync a live edition assets in the resync service. As we may end up with a call like this `PublishAssetsService.call(live_edition, live_edition: nil)` which feels pretty confusing. I think we suffer in general with re-use in the resync service where some of the services we call don't feel intuitive to the situation but the class does what we need it to do so there's definitely wider things at play but a short term aspect could be to change live_edition to supersede_edition.

Thanks again, and sorry if I've blown the scope open.

benthorner:
Thanks @kevindew - you're right, there were a few errors in my comment above. I agree with all your suggestions, which I've re-implemented as a PR to supersede this one: https://github.com/alphagov/content-publisher/pull/1731.

PR #1694, opened 2020-01-24
Test index filtering Publishing API down in view

kevindew:

The feature test that is being replaced is used to test a particular
conditional in the view. Our test strategy suggests that these should be
covered by view tests.


app/views/documents/index/_filters.html.erb
```diff
@@ -1,4 +1,4 @@
-<%= form_tag({}, method: :get, data: { gtm: "filter-documents" }) do %>
+<%= form_tag(documents_path, method: :get, data: { gtm: "filter-documents" }) do %>
```
ChrisBAshton: Can you explain in the commit message why this change is necessary? If we're only moving/refactoring the test then we shouldn't really need to touch application code.

kevindew:
@ChrisBAshton Thanks - just updated the commit.

PR #1697, opened 2020-01-24
Create new revision for images where metadata has changed

brucebolt:
In some Whitehall documents, the image metadata (e.g. alt_text and caption) has changed between editions.  This affects 8 news articles from NDA.

An example is https://whitehall-admin.integration.publishing.service.gov.uk/government/admin/export/document/404026 where the suffix "OBE" has been removed from the name in the caption and alt_text values.

We were only storing this data from the first edition, which led to the integrity check failing in cases where these values had been updated in a later edition (in the example above, we were proposing to publish the name with OBE, that had been removed in the second edition)

This change creates a new revision for an existing image where either the alt_text or caption has changed between editions.  The correct values for alt_text and caption will therefore be assigned to the relevant historic editions.

Trello card: https://trello.com/c/3AV7mZTG

### DIFF
--- a/lib/whitehall_importer/create_revision.rb
+++ b/lib/whitehall_importer/create_revision.rb
@@ -28,7 +28,7 @@ module WhitehallImporter
       raise AbortImportError, "Edition has an unsupported document type" unless SUPPORTED_DOCUMENT_TYPES.include?(whitehall_edition[document_type_key])
 
       file_attachment_revisions = find_or_create_file_attachment_revisions(whitehall_edition["attachments"])
-      image_revisions = find_or_create_image_revisions(whitehall_edition["images"])
+      image_revisions = image_revisions(whitehall_edition["images"])
       Revision.create!(
         document: document,
         number: document.next_revision_number,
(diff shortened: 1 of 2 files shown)


lib/whitehall_importer/create_revision.rb
```diff
@@ -119,7 +119,12 @@ def tags(associations)
...
-        revision = if already_imported
+        revision = if already_imported &&
+            (already_imported.image_revision.alt_text != image["alt_text"] ||
+             already_imported.image_revision.caption != image["caption"])
+                     WhitehallImporter::CreateImageRevision
+                      .call(document_import, image, memo.map(&:filename), already_imported.image_revision.image)
+                   elsif already_imported
                      already_imported.image_revision
                    else
                      WhitehallImporter::CreateImageRevision
```
kevindew: I wonder if we should move all of this into a method since it's quite a lot of logic for a loop - or maybe the alt check may feel like the need for a new method.

lib/whitehall_importer/create_revision.rb
```diff
@@ -119,7 +119,12 @@ def tags(associations)
     def find_or_create_image_revisions(images)
       images.reduce([]) do |memo, image|
         already_imported = WhitehallMigration::AssetImport.find_by(original_asset_url: image["url"])
-        revision = if already_imported
+        revision = if already_imported &&
+            (already_imported.image_revision.alt_text != image["alt_text"] ||
+             already_imported.image_revision.caption != image["caption"])
+                     WhitehallImporter::CreateImageRevision
+                      .call(document_import, image, memo.map(&:filename), already_imported.image_revision.image)
+                   elsif already_imported
```
kevindew: I wonder if we can do some of this without needing to call the CreateImageRevision class as we should be able to re-use a bunch of data and skip a bunch of steps the class takes.

For example if we have the already_imported.image_revision we should be able to do:

```
new_image_revision = Image::Revision.create!(
  image: already_imported.image_revision.image
  metadata_revision: Image::MetadataRevision.new(
    caption: image["caption"],
    alt_text: image["alt_text"],
  ),
  blob_revision: already_imported.image_revision.blob_revision,
)
```

This will mean that it re-uses the same asset files

brucebolt: Good point, it means I haven't needed to touch `CreateImageRevision` at all now.

lib/whitehall_importer/create_revision.rb
brucebolt: Moved all this logic into a new method and just calling the method within the loop.  Looks much cleaner now.

brucebolt:
Thanks for the review @kevindew.  I've made the proposed changes which have made the code cleaner and reduced the number of classes I've had to modify.

brucebolt:
No, that won't be an issue here.  We're always using the value from the Whitehall export, which is an empty string when not present.  The only issue comes in the integrity check, which is resolved in #1704.

PR #1701, opened 2020-01-27
Replace generic (body) fields with specific ones

benthorner:

This introduces a new "BodyField" class to represent the 'body' field and
its associated attributes. We want to start co-locating the various
behaviours (requirements, payload, etc.) of the 'body' field with its data
("Tell-Don't-Ask"), and this class provides a starting point.


spec/models/document_type/body_field_spec.rb
```diff
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
```
ChrisBAshton: This could be fixed up to the commit where body_field_spec is introduced.

app/models/document_type/body_field.rb
```diff
@@ -16,4 +16,20 @@ def payload(edition)
...
+    issues = Requirements::CheckerIssues.new
+    return issues if GovspeakDocument.new(revision.contents[id], edition).valid?
+
+    issues << Requirements::Issue.new(id, :invalid_govspeak)
+    issues
+  end
+
+  def pre_publish_issues(_edition, revision)
+    issues = Requirements::CheckerIssues.new
+    return issues if revision.contents[id].present?
```
ChrisBAshton: I think this migrated code has lost some of its clarity - the previous version is more scannable and obvious that it returns a `:blank` issue, and is also more easily extended to add additional issues because it doesn't return early. I wonder if this would be better written as:

```ruby
issues = Requirements::CheckerIssues.new
issues << Requirements::Issue.new(id, :blank) if revision.contents[id].blank?
issues
```

I'm guessing you've written it as-is for consistency with the `pre_preview_issues` method above, which would be harder to rewrite that way due to line length. 🤔

spec/models/document_type/body_field_spec.rb
benthorner: Good spot - fixed.

app/models/document_type/body_field.rb
benthorner: Sure. I agree that the nesting in the previous version helps indicate the structure better. I tend to avoid long inline conditionals, but I'm aware that's subjective. For consistency and extensibility, I've changed it to the nested `if` form. Does that help?

ChrisBAshton: Looks better 👍 thanks. Now have a failing build though!

benthorner:
Closing in favour of https://github.com/alphagov/content-publisher/pull/1705

PR #1705, opened 2020-01-28
Introduce Field classes for contents

benthorner:

This introduces a new "BodyField" class to represent the 'body' field and
its associated attributes. Using a 'Field' class enables us to co-locate the 
various behaviours (requirements, payload, etc.) of the 'body' field with its 
data ("Tell-Don't-Ask"). With this new class as a foundation, we then 
introduce other 'Field' classes to remove the hard-coding of the title, 
base path and summary fields, where these are treated as 'contents'.

Some pain points to address in follow-up PRs:

   - [ ] Reduce verbosity around requirements issues in tests
   - [ ] Reduce verbosity of document types (IDs) in test
   - [ ] Reduce verbosity of Pub API payload (use symbol keys)
   - [ ] Explore only passing a revision for requirements checks


app/interactors/content/update_interactor.rb
```diff
@@ -52,15 +57,13 @@ def update_preview
...
+  end
 
-    params.require(:revision)
-      .permit(:update_type, :change_note, :title, :summary, contents: contents_params)
-      .tap do |p|
-        p[:title] = p[:title]&.strip
-        p[:summary] = p[:summary]&.strip
-        p[:base_path] = GenerateBasePathService.call(edition.document, p[:title])
-      end
+  def update_params
```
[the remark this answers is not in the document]
benthorner: I wasn't concerned about it, and now I see that it's different to the original interpretation, since there are two such methods. On the other hand, I also look how it distinguishes `change_note`, which is tangential to the purpose of the page. But `content_params` sounds better overall. Updated.

spec/factories/document_type_factory.rb
```diff
@@ -11,11 +11,15 @@
...
     label { SecureRandom.alphanumeric(8) }
-    contents { [] }
     tags { [] }
     guidance { [] }
     topics { false }
 
+    contents do
+      [DocumentType::TitleAndBasePathField.new,
+       DocumentType::SummaryField.new]
+    end
```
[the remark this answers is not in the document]
benthorner: I'm not sure what you mean here; could you help me understand? Since the purpose of the factory is to create an instance of the object, it seemed naturally to use its constructor directly.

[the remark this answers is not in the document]
kevindew: Yes sorry. It is fine to initialise the classes in the factory. 

I was referring to how we're having to pass in classes in the tests: https://github.com/alphagov/content-publisher/blob/720412706763e62fe936f1fdc738100beb946a14/spec/lib/whitehall_importer/integrity_checker_spec.rb#L64-L69

and thought it'd be nicer if the call to the factory could be more like

```
build(:document_type, content_fields: %w[title_and_base_path summary body])
```

where the factory could convert those strings in the same way YAML does to get the classes?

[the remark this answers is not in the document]
benthorner: Ah, I see. That would be a little easier. I have a point on the card to try and address this with a `newsesque` trait, since we generally have (or don't care about) all three fields together. If we go with the trait option, we could add support for field IDs in future too. Perhaps we can discuss this further on the subsequent PR I'm intending to raise for this?

app/models/document_type/body_field.rb
```diff
@@ -5,7 +5,37 @@ def id
...
+      },
+    }
+  end
+
+  def updater_params(_edition, params)
+    params
+      .require(:revision)
+      .permit(contents: [:body])
+      .to_h
+      .deep_symbolize_keys
```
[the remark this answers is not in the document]
benthorner: Yeah, I thought it might look odd. I found working with `params` directly to be quite frustrating. For example `#to_h` actually returns the string key version, and I was worried our downstream code might be expecting one or the other. I think this means you'd need a slightly different `expect` statement - having tried a few!

I've had a go at rewriting the method to try and make it less surprising, but also more deterministic in the type of data it returns. Could you take another look?

benthorner:
@kevindew thanks!

To summarise your points:  as we iterate, we may want to move some things around (e.g. `Field` classes) and extend functionality (e.g. labels on the content summary) in future.

With the views, I was wondering that the more specific location of them (by action) may actually be beneficial: although it would be 'ideal' to re-use them if the need arises, in another sense they are simply meant to be part of the behaviour of the field - if that behaviour needs to change depending on where it's shown/input, then it still seems reasonable to have a separate view.

I think that covers all your comments. Would you like to take another look?

app/interactors/content/update_interactor.rb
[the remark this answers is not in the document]
kevindew: Ah seems we've focused on different words in the compound. I wasn't fazed by the update word but by the params word based on it implying it's an ActionController::Parameters object.

app/models/document_type/body_field.rb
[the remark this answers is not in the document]
kevindew: `to_h` on an ActionController::Parameters returns an instance of HashWithIndifferentAccess so you can work with either variant. 

The way you've changed it looks an improvement though 👍

spec/factories/document_type_factory.rb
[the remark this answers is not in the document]
kevindew: Yeah, no problem for dealing with on a later PR.

I think the closer we can have this configuration to what's in the YAML the better by the way, be even better if the building of the fields could be done by the same code that does the building of the YAML config.

app/interactors/content/update_interactor.rb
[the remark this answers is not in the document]
benthorner: Oh dear ‼️ . I can't say I'm concerned about the use of `_params`, as they are parameters (technically, 'arguments'). Other words like `_attributes` and `_data` spring to mind, but it seems overkill to expend more effort on an internal method.

app/models/document_type/body_field.rb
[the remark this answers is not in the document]
benthorner: Huh, that's weird - I may not have used the combination of `#to_h` and `match` when I was trying to get it to work. Oh well - I still like the determinism.

app/models/document_type/body_field.rb
```diff
@@ -0,0 +1,42 @@
...
+      issues << Requirements::Issue.new(id, :blank)
+    end
+
+    issues
+  end
+
+  def updater_params(_edition, params)
+    body = params.require(:revision).require(:contents)[:body]
+    { contents: { body: body } }
+  end
```
kevindew: This method is placed in quite a different ordering to the other field classes. It seems rather minor as we don't often talk or enforce any method placements, but this just felt a bit jarring in context.

I am wondering if it makes sense to do the `require(:revision)` (and deal with the bad request exception) in the interactor and then do dig on the resulting params. 

for example: `{ contents: { body: params.dig(:contents, :body) }`

Or if that doesn't appeal maybe we don't even need to worry about using require here since we're plucking out one value rather than using it in a permit context for mass assignment. I think it'd be fine to just do `params.dig(:revision, :contents, :body)` and we're already at risk of this returning nil here.

spec/factories/document_type_factory.rb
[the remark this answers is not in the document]
benthorner: I think we used to do that sort of thing for classes like `DocumentType`, but I remember it made them harder to work with in tests, probably because the generation of the transient hash wasn't very intuitive, given the layer of indirection it represents.

Found the commit that describes the pain: https://github.com/alphagov/content-publisher/pull/622/commits/24b4fd21805c3edd94ab4b64a675c7cde29736ee

I like the random "foo" at the end 😆

app/models/document_type/body_field.rb
benthorner: That's OK - I did notice it, and then thought "surely no one will mind" 😄. Updated.

The `require` poses an interesting question: how much ownership should fields have of their params? So far, I've been thinking about the params for each field independently of each other and the interactor; it's up to the field (input) to specify its params, and the interactor shouldn't (or at least, doesn't need to) have any knowledge of them or impose any restriction on them. This is just what seemed intuitive to me, and I'm open to change.

I've left the code as-is for now, and I'm interested to know if you think we should make any alternative decisions as part of this PR? <- fine with me, just not sure

kevindew: Yeah, I thought you might be thinking in that capacity. I think we hit a somewhat weird situation where revision is this overall grouping for the form but these field classes have individual concerns. So it seems a bit fishy to me that these field classes are caring about revision as that feels a wrapper for the higher level above the abstraction.

After a bit of pondering I don't think we really need the revision wrapper, it doesn't seem to cause us harm but then I think it also implies shared data that we're making more independent.

Simplest thing forward would probably be to cut down the long chained require statement into `params.dig(:revision, :contents, :body)`

app/models/document_type/title_and_base_path_field.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class DocumentType::TitleAndBasePathField
+  def id
+    "title_and_base_path"
```
ChrisBAshton: Have you considered something like `self.class.name.demodulize.underscore` so that we don't have to repeat it for each field type? Though we'd need a parent field class that both of these inherit from.

config/locales/en/document_types.yml
```diff
@@ -80,33 +80,33 @@ en:
 
     fatality_notice:
       description: Initial fatality notices and subsequent obituaries of forces and MOD personnel
-
```
ChrisBAshton: These space fixes could be moved to their own commit.

config/locales/en/document_types.yml
```diff
@@ -2,7 +2,7 @@ en:
   document_types:
     default:
       fields:
-        title:
+        title_and_base_path:
```
ChrisBAshton: I think this renaming could be moved to a separate commit, where the argument for renaming can be made 👍

spec/models/document_type/title_and_base_path_field_spec.rb
```diff
@@ -0,0 +1,14 @@
+RSpec.describe DocumentType::TitleAndBasePathField do
+  describe "#payload" do
+    it "returns a hash with title and routing attributes" do
+      edition = build(:edition, title: "Some title", base_path: "/foo/bar/baz")
```
ChrisBAshton: For consistency, is it worth making the base path derive from the title? i.e. `/some-title`?

spec/models/document_type/summary_field_spec.rb
```diff
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
```
ChrisBAshton: This could be fixed up to the commit where the spec file is introduced.

spec/models/document_type/title_and_base_path_field_spec.rb
```diff
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
```
ChrisBAshton: And here.

app/models/document_type/title_and_base_path_field.rb
benthorner: Indeed I did, although I didn't think it would be that easy! While it's tempting to change it here, since it would enforce consistency, I'd prefer to do that in a subsequent PR, along with a few other refactorings. I'll make a note...

config/locales/en/document_types.yml
benthorner: I think it's reasonable to have this change here, although I appreciate we should try to limit the scope of a commit if possible. There's an effort/reward ratio to changing this, and I'm not sure it's worth it in this case. 

As someone who likes to read through the sequence of commits, I'd prefer to see a slightly larger diff than read through an extra commit that may or may not do more than just fix whitespacing (never trust a commit title 😉 ).

spec/models/document_type/summary_field_spec.rb
benthorner: Sure - fixed.

spec/models/document_type/title_and_base_path_field_spec.rb
benthorner: And here :-).

config/locales/en/document_types.yml
benthorner: Oops! I realise I didn't offer any explanation in the commit. I think it's important to have the change in this commit, since it provides a little motivation for it - I've updated the commit message.

spec/models/document_type/title_and_base_path_field_spec.rb
benthorner: I'm struggling a little to see where you're coming from here. Maybe I'm used to test data being pretty random! In this case, there's an argument for keeping it the same as the test it was extracted from. And consistency may be a false friend here: the 'consistent' base path would actually be `/prefix/some-title`.

ChrisBAshton: Only suggesting as the title usually drives the base path, so `Some title` would become `some-title`. I don't think it affects the readability of the test particularly - it's not a hill I'd die on! `/prefix/some-title` does seem more accurate than `/foo/bar/baz` though, and gives someone glancing through the codebase a better idea that one affects the other.

PR #1708, opened 2020-01-28
Fix 'translation missing' errors for supertype selection

ChrisBAshton:
Hint text is only defined for some supertypes, not others.
An attempt was made to only display hint text if it is not
'nil', however it is never 'nil' because the fallback value
is a string 'translation missing: en.supertypes.not-sure.hint'

This commit does an explicit check using 'exists?' and
removes a redundant 'hint' var that is only used in the
ternary.

Fixes this bug:

Screenshot 2020-01-28 at 16 38 15

### DIFF
--- a/app/views/new_document/choose_supertype.html.erb
+++ b/app/views/new_document/choose_supertype.html.erb
@@ -14,11 +14,11 @@
         error_items: @issues&.items_for(:supertype),
         items: Supertype.all.map do |supertype|
           id = supertype.id
-          label = I18n.t("supertypes.#{id}.label")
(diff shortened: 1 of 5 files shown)


app/views/new_document/choose_supertype.html.erb
```diff
@@ -15,7 +15,6 @@
         items: Supertype.all.map do |supertype|
           id = supertype.id
           label = I18n.t("supertypes.#{id}.label")
```
kevindew: Can this also be switched to just `t` so that it raises on a missing translation?

PR #1710, opened 2020-01-29
Iterate tests: use schema to validate config

ChrisBAshton:
Uses a schema for comprehensive testing of our document type configs, and replaces feature tests with request/view tests (see commits).

Trello card: https://trello.com/c/ihcpYeWf/1320-iterate-how-we-test-formats


spec/factories/document_type_factory.rb
```diff
@@ -1,5 +1,18 @@
 # frozen_string_literal: true
 
+if defined? RSpec.configure # conditional needed when running lint
```
benthorner: Could we take a shorter approach of just deleting the class variable, so `DocumentType.all` resets by itself? I think it would be reasonable to add a small amount of logic in the `DocumentType` class to support this, if that helps.

Although the cleanup code is related to the action of the factory, it's unusual to see RSpec hooks here, which is apparent since we have to do an `if defined?`. Can we move this to `spec_helper.rb`, or its own helper?

spec/views/new_document/choose_supertype.html.erb_spec.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+RSpec.describe "new_document/choose_supertype.html.erb" do
```
benthorner: Good call 👍

spec/factories/document_type_factory.rb
ChrisBAshton: I liked that all of the messy code is in one place rather than leaking out, but agree that hooking into RSpec functions inside a factory is unconventional. Have tried a few approaches for 'deleting the class variable', haven't had much success so far 😕 

I added this to spec_helper.rb:

```ruby
  config.before :all do
    # This is required to clean up after multiple `build(:document_type)` calls
    [DocumentType.all, Supertype.all.first.document_types].map(&:clear)
  end
```

...and added another line to top of the `self.all` method in document_type.rb:

```
return @all if defined?(@all) && @all.count.positive?

@all = begin
  # the rest of the method is unchanged
```

But would get a failing test along the lines of:

```
  1) News article format 
     Failure/Error: choose DocumentType.find("news_story").label
     
     Capybara::ElementNotFound:
       Unable to find radio button "News story" that is not disabled
     # ./spec/features/formats/news_article_spec.rb:23:in `when_i_choose_this_document_type'
     # ./spec/features/formats/news_article_spec.rb:12:in `block (2 levels) in <top (required)>'
```

So I reverted to the more verbose method of deleting the 'random' document types. Tried cutting this down to reduce the number of loops but for some reason would always get this error, so it wasn't working:

```
  1) DocumentType.all should create a DocumentType for each one in the YAML
     Failure/Error: expect(DocumentType.all.count).to eq(document_types.count)
     
       expected: 13
            got: 21
     
       (compared using ==)
     # ./spec/models/document_type_spec.rb:31:in `block (3 levels) in <top (required)>'
```

Now that I've moved this code out into spec_helper.rb and removed the `if RSpec.defined?` conditional, I'm inclined to leave this as-is. We can refactor it (or remove its necessity altogether) in a future PR.

benthorner: What about `def self.clear; @all = nil; end` in `DocumentType`? Since we're treating the class as a mutable collection in the tests already, it seems reasonable to support cleanup better. I think it's worth trying for something a little leaner in this PR, as right now we have both time and knowledge with which to think about the problem. If the previous suggestion isn't any good, perhaps we should have a chat and think about what the fundamental difficulty is here - these models are just a little bit quirky!

ChrisBAshton: Have had another go 👍

spec/spec_helper.rb
```diff
@@ -42,6 +42,11 @@
     Rails.application.load_seed
   end
 
+  config.before :all do
+    # This is required to clean up after multiple `build(:document_type)` calls
```
benthorner: Minor: I think this is implicit, noting the other `clear` methods in here!

app/models/document_type.rb
```diff
@@ -12,7 +12,9 @@ def self.find(id)
   end
 
   def self.all
-    @all ||= begin
+    return @all if defined?(@all) && @all.present?
```
benthorner: Could you help me understand why this line is necessary? I was thinking the `||=` should still do the trick.

ChrisBAshton: You're quite right 👍 I originally set `@all = []` in the `self.clear` method, which made this line necessary. Then I followed your suggestion of setting it to `nil` and ended up building on it instead of removing it. Have now removed!

spec/spec_helper.rb
ChrisBAshton: Removed!

PR #1714, opened 2020-01-30
Allow import from Whitehall to be filtered by document subtype

brucebolt:
We may wish to import documents of a single subtype (e.g. only news stories and press releases for the news article type) from Whitehall, since Content Publisher does not yet support the other subtypes of news article.

An optional subtype parameter has been added to the document index endpoint in Whitehall (https://github.com/alphagov/whitehall/pull/5303), so these are the corresponding changes in Content Publisher.

Trello card: https://trello.com/c/dmS7BYoG

### DIFF
--- a/db/migrate/20200130090021_add_whitehall_migration_subtypes.rb
+++ b/db/migrate/20200130090021_add_whitehall_migration_subtypes.rb
@@ -1,0 +1,7 @@
+# frozen_string_literal: true
+
+class AddWhitehallMigrationSubtypes < ActiveRecord::Migration[6.0]
+  def change
+    add_column :whitehall_migrations, :document_subtypes, :text, array: true
+  end
+end
--- a/lib/tasks/import.rake
+++ b/lib/tasks/import.rake
@@ -3,13 +3,15 @@
 require "gds_api/whitehall_export"
 
 namespace :import do
-  desc "Import all documents matching an organisation and document type from Whitehall Publisher, e.g. import:whitehall_migration[\"cabinet-office\",\"NewsArticle\"]"
-  task :whitehall_migration, %i[organisation_slug document_type] => :environment do |_, args|
+  desc "Import all documents matching an organisation, document type and optional list of document subtypes from Whitehall Publisher, e.g. import:whitehall_migration[\"cabinet-office\",\"NewsArticle\",\"NewsStory,PressRelease\"]"
+  task :whitehall_migration, %i[organisation_slug document_type document_subtypes] => :environment do |_, args|
     organisation_content_id = GdsApi.publishing_api.lookup_content_id(
       base_path: "/government/organisations/#{args.organisation_slug}",
     )
 
-    whitehall_migration = WhitehallImporter.create_migration(organisation_content_id, args.document_type)
+    document_subtypes = args.document_subtypes ? args.document_subtypes.split(",") : []
+
+    whitehall_migration = WhitehallImporter.create_migration(organisation_content_id, args.document_type, document_subtypes)
 
     documents_to_import = WhitehallMigration::DocumentImport.where(whitehall_migration_id: whitehall_migration.id).count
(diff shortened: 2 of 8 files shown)


spec/lib/gds_api/whitehall_export_spec.rb
```diff
@@ -8,16 +8,16 @@
     let(:whitehall_export_page_2) { build(:whitehall_export_index, documents: build_list(:whitehall_export_index_document, 10)) }
 
     before do
-      stub_request(:get, "#{whitehall_host}/government/admin/export/document?lead_organisation=123&type=NewsArticle&page_count=100&page_number=1")
+      stub_request(:get, "#{whitehall_host}/government/admin/export/document?lead_organisation=123&type=NewsArticle&subtypes[]=NewsStory&subtypes[]=PressRelease&page_count=100&page_number=1")
```
emmabeynon: Since these API requests are quite lengthy, we could extract it out to a helper method.  We have an example of this already [here](https://github.com/alphagov/content-publisher/blob/master/spec/support/topics_helper.rb#L6).  It might be premature refactoring to create a new helper file at this stage, but we could still have the method in this file.  We'd also have the added bonus of being able to do something like this when we're asserting that the request has been made:
```
api_request = stub_whitehall_export_api_document
expect(whitehall_document_list.next).to have_requested(api_request)
```

lib/tasks/import.rake
```diff
@@ -4,12 +4,12 @@ require "gds_api/whitehall_export"
 
 namespace :import do
   desc "Import all documents matching an organisation and document type from Whitehall Publisher, e.g. import:whitehall_migration[\"cabinet-office\",\"NewsArticle\"]"
```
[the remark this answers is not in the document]

db/migrate/20200130090021_add_whitehall_migration_subtypes.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class AddWhitehallMigrationSubtypes < ActiveRecord::Migration[6.0]
+  def change
+    add_column :whitehall_migrations, :document_subtypes, :text
```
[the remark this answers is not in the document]
brucebolt: Yes, I think that's a good idea, have made this change.

spec/lib/gds_api/whitehall_export_spec.rb
brucebolt: I've moved all the stubbing into a separate method and am now asserting against the stub, rather than having a URL hardcoded twice.

spec/lib/gds_api/whitehall_export_spec.rb
```diff
@@ -36,4 +35,18 @@
       expect(whitehall_adapter.document_export("123")).to have_requested(:get, "#{whitehall_host}/government/admin/export/document/123")
     end
   end
+
+  def stub_publishing_api_has_index_page(lead_organisation, document_type, document_subtypes, page_number, items_on_page)
```
emmabeynon: Should this be called something like `stub_whitehall_api_has_document_index`?

brucebolt: Yes, that seems like a better name.  Have updated.

spec/lib/gds_api/whitehall_export_spec.rb
```diff
@@ -2,22 +2,21 @@
 
 RSpec.describe GdsApi::WhitehallExport do
   describe "document_list" do
-    let(:whitehall_adapter) { GdsApi::WhitehallExport.new(Plek.find("whitehall-admin")) }
     let(:whitehall_host) { Plek.new.external_url_for("whitehall-admin") }
```
emmabeynon: I think we can get rid of this now that we're calling `#stub_whitehall_api_has_document_index`.

brucebolt: Yes, this isn't needed, so I've removed it

PR #1717, opened 2020-01-31
Ignore missing images on import when organisation default is used

brucebolt:
News articles with no user uploaded image were being rejected when importing from Whitehall.  This was because the integrity check was failing.

In the case of no image being specified for a news article, we fall back to the organisation default, or failing that to a placeholder.

Whitehall submits the default organisation or placeholder image to Publishing API as part of the payload, with an alt_text of "placeholder": https://github.com/alphagov/whitehall/blob/9396ec5a69750bd41c106fa7517d8c3476988d31/app/presenters/lead_image_presenter_helper.rb#L18-L24

Content Publisher does not do this.  Instead no images are sent to Publishing API and the image is determined by the frontend application: https://github.com/alphagov/government-frontend/blob/2ef6483d9fb96baf5af0294a07cbddb9c0e4d420/app/presenters/content_item/news_image.rb#L3-L12

Therefore we can ignore failures on these images (identified by the presence of "placeholder" alt_text and no images on the document) since government-frontend will handle the image choice.

### DIFF
--- a/lib/whitehall_importer/integrity_checker.rb
+++ b/lib/whitehall_importer/integrity_checker.rb
@@ -57,6 +57,7 @@ module WhitehallImporter
 
       %w(alt_text caption).each_with_object([]) do |attribute, problems|
         if publishing_api_image[attribute] != proposed_image_payload[attribute]
+          next if default_image?(proposed_image_payload, publishing_api_image, attribute)
           next if attribute == "caption" && publishing_api_image[attribute].nil? && proposed_image_payload[attribute].empty?
 
           problems << problem_description(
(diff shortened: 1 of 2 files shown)


lib/whitehall_importer/integrity_checker.rb
```diff
@@ -57,6 +57,7 @@ def image_problems
 
       %w(alt_text caption).each_with_object([]) do |attribute, problems|
         if publishing_api_image[attribute] != proposed_image_payload[attribute]
+          next if attribute == "alt_text" && proposed_image_payload.empty? && publishing_api_image[attribute] == "placeholder"
```
emmabeynon: I wonder if it might be helpful to extract this out into its own method that describe the logic?  It might not be immediately obvious why we're making this check otherwise.  Something like `#placeholder_image?` maybe?

brucebolt: Good idea.  I've moved this into a method, and also added a commit to put the other conditional statement into a named method.

brucebolt:
Thanks for the review Emma.  I've used your suggested approach to the testing which has massively simplified the changes being made here, and also used this opportunity to move a previously implicit test (image with caption as an empty string mismatching with Publishing API's nil value) to it's own explicit test scenario.

PR #1718, opened 2020-01-31
Tweaks to flexible fields

benthorner:

This PR contains a few improvements to the work on flexible fields that
would have made the original PR too big. Please see the commits for more
details.

### DIFF
--- a/app/interactors/content/update_interactor.rb
+++ b/app/interactors/content/update_interactor.rb
@@ -58,7 +58,9 @@ private
   end
 
   def change_note_params
-    { update_type: params[:update_type], change_note: params[:change_note] }.compact
+    return {} if edition.first?
+
+    { update_type: params[:update_type], change_note: params[:change_note] }
   end
 
   def content_params
--- a/app/interactors/content/update_interactor.rb
+++ b/app/interactors/content/update_interactor.rb
@@ -58,7 +58,7 @@ private
   end
 
   def change_note_params
-    params.require(:revision).permit(:update_type, :change_note)
+    { update_type: params[:update_type], change_note: params[:change_note] }.compact
   end
 
   def content_params
--- a/app/services/preview_draft_edition_service/payload.rb
+++ b/app/services/preview_draft_edition_service/payload.rb
@@ -41,7 +41,7 @@ class PreviewDraftEditionService::Payload
       payload[:bulk_publishing] = true
     end
 
-    payload.deep_stringify_keys
+    payload
   end
 
 private
@@ -67,6 +67,7 @@ private
       .merge(roles_and_people(role_appointments))
       .merge(organisations: links.uniq)
       .merge(government: [edition.government&.content_id].compact)
+      .symbolize_keys
   end
 
   def image
--- a/app/services/preview_draft_edition_service/payload.rb
+++ b/app/services/preview_draft_edition_service/payload.rb
@@ -14,35 +14,34 @@ class PreviewDraftEditionService::Payload
 
   def payload
     payload = {
-      "locale" => edition.locale,
-      "schema_name" => publishing_metadata.schema_name,
-      "document_type" => document_type.id,
-      "publishing_app" => PUBLISHING_APP,
-      "rendering_app" => publishing_metadata.rendering_app,
-      "update_type" => edition.update_type,
-      "details" => details,
-      "links" => links,
-      "access_limited" => access_limited,
-      "auth_bypass_ids" => auth_bypass_ids,
+      locale: edition.locale,
+      schema_name: publishing_metadata.schema_name,
+      document_type: document_type.id,
+      publishing_app: PUBLISHING_APP,
+      rendering_app: publishing_metadata.rendering_app,
+      update_type: edition.update_type,
+      details: details,
+      links: links,
+      access_limited: access_limited,
+      auth_bypass_ids: auth_bypass_ids,
     }
-    payload["change_note"] = edition.change_note if edition.major?
+    payload[:change_note] = edition.change_note if edition.major?
 
     document_type.contents.each do |field|
-      attributes = field.payload(edition)
-      payload.deep_merge!(attributes.deep_stringify_keys)
+      payload.deep_merge!(field.payload(edition))
     end
 
     if edition.backdated_to.present?
-      payload["first_published_at"] = edition.backdated_to
-      payload["public_updated_at"] = edition.backdated_to if edition.first?
+      payload[:first_published_at] = edition.backdated_to
+      payload[:public_updated_at] = edition.backdated_to if edition.first?
     end
 
     if republish
-      payload["update_type"] = "republish"
-      payload["bulk_publishing"] = true
+      payload[:update_type] = "republish"
+      payload[:bulk_publishing] = true
     end
 
-    payload
+    payload.deep_stringify_keys
   end
 
 private
@@ -50,7 +49,7 @@ private
   def access_limited
     return {} unless edition.access_limit
 
-    { "organisations" => edition.access_limit_organisation_ids }
+    { organisations: edition.access_limit_organisation_ids }
   end
 
   def auth_bypass_ids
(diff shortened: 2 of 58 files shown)


app/interactors/content/update_interactor.rb
```diff
@@ -58,7 +58,7 @@ def update_preview
   end
 
   def change_note_params
-    params.require(:revision).permit(:update_type, :change_note)
+    params.permit(:update_type, :change_note)
```
kevindew: This will raise a warning about the other parameters. You'd see something like "Unpermitted parameters: :title, :summary, :contents". May want to change this to something like `{ update_type: params[:update_type], change_note: params[:change_note] }`

Just a thought too - it doesn't look we do anything to stop these params going in when we're on the first edition so we might want this to have a `return {} if edition.first?`

benthorner:
We might want to be careful deploying f117e0cbb63d593d333935b9525b334024fa1038, since it will create a temporary incompatibility for users who already have the content edit screen loaded.

app/services/preview_draft_edition_service/payload.rb
```diff
@@ -14,43 +14,42 @@ def initialize(edition, republish: false)
...
 
     if republish
-      payload["update_type"] = "republish"
-      payload["bulk_publishing"] = true
+      payload[:update_type] = "republish"
+      payload[:bulk_publishing] = true
     end
 
-    payload
+    payload.deep_stringify_keys
```
kevindew: do you know if we actually need this? would have thought it gets converted to JSON at some point so key types aren't a concern.

I guess it might break the WhitehallImporter code that also uses this class (I think the class should also be moved but that's something for outside this scope)

spec/factories/edition_factory.rb
```diff
@@ -15,7 +15,8 @@
     transient do
       content_id { SecureRandom.uuid }
       locale { I18n.available_locales.sample }
-      document_type_id { build(:document_type, path_prefix: "/prefix").id }
+      document_type_id { document_type.id }
+      document_type { build(:document_type, path_prefix: "/prefix") }
```
kevindew: This is nice 👍

app/interactors/content/update_interactor.rb
benthorner: Oh yes, sorry - you did mention. `#slice` and `#to_h` are also protected (with an exception, it turns out). I've also pushed an extra commit to add that new check in 👍.

app/services/preview_draft_edition_service/payload.rb
benthorner: We don't, and your guess was correct. I had another go and it turned out to be simpler than when I tried the first time - I've added another commit.

benthorner:
@kevindew I think this is ready for another look now 🔹

spec/factories/edition_factory.rb
JonathanHallam: +1

app/interactors/content/update_interactor.rb
```diff
@@ -58,7 +58,9 @@ def update_preview
   end
 
   def change_note_params
-    params.require(:revision).permit(:update_type, :change_note)
+    return {} if edition.first?
+
+    { update_type: params[:update_type], change_note: params[:change_note] }.compact
```
kevindew: Do we need the compact here? Since we don't do this for fields I wonder if it's an inconsistency we need or not.

lib/whitehall_importer/integrity_checker.rb
```diff
@@ -95,28 +95,40 @@ def problem_description(message, expected, actual)
...
+        publishing_api_links.dig(:links, link_type)
     end
 
     def publishing_api_links
-      @publishing_api_links ||= GdsApi.publishing_api.get_links(edition.content_id).to_h
+      @publishing_api_links ||= GdsApi
+        .publishing_api
+        .get_links(edition.content_id)
+        .to_h
+        .deep_symbolize_keys
```
kevindew: Minor: I wonder if this might benefit from having the method call on the same line and then inline. For example:

```
      @publishing_api_links ||= GdsApi.publishing_api
                                      .get_links(edition.content_id)
                                      .to_h
                                      .deep_symbolize_keys
```

spec/interactors/content/update_interactor_spec.rb
```diff
@@ -3,57 +3,63 @@
...
 
     it "updates the preview" do
       expect(FailsafeDraftPreviewService).to receive(:call).with(edition)
-      Content::UpdateInteractor.call(params: build_params, user: user)
+      Content::UpdateInteractor.call(params: params, user: user)
     end
 
     it "raises an error when the edition isn't editable" do
-      params = build_params(document: create(:edition, :published).document)
+      params.merge!(document: (create :edition, :published).document.to_param)
```
kevindew: This use of parenthesis seems somewhat unconventional of wrapping the full method call in parenthesis rather than the arguments.

spec/interactors/content/update_interactor_spec.rb
```diff
@@ -3,57 +3,63 @@
...
-      params = build_params(title: "New title", summary: "New summary")
-
       expect { Content::UpdateInteractor.call(params: params, user: user) }
         .to change { edition.reload.title }.to("New title")
         .and change { edition.reload.summary }.to("New summary")
+        .and change { edition.reload.change_note }.to("New note")
+    end
+
+    it "ignores change notes for first editions" do
+      params.merge!(document: (create :edition).document.to_param)
```
kevindew: And here..

spec/interactors/content/update_interactor_spec.rb
```diff
@@ -3,57 +3,63 @@
...
-                     summary: SecureRandom.alphanumeric(10))
-      ActionController::Parameters.new(
-        document: document.to_param,
-        revision: { title: title, summary: summary },
-      )
+    let(:params) do
+      ActionController::Parameters.new(document: edition.document.to_param,
+                                       title: "New title",
+                                       summary: "New summary",
+                                       change_note: "New note")
```
kevindew: Should this also have update type?

Minor but I'm not sure if it's a bit of a readability detriment having this function switch to a let and having to do mutating of the variable.

app/interactors/content/update_interactor.rb
benthorner: Yes, this was a surprise. The problem is that the presence of the key will modify the revision, as the value being `nil` is significant in general e.g. for lead images. So we need to remove redundant keys - the tests helpfully fail otherwise.

spec/interactors/content/update_interactor_spec.rb
benthorner: Oh yeah - I see what you mean. Updated 👍

lib/whitehall_importer/integrity_checker.rb
benthorner: Sure - updated 👍

spec/interactors/content/update_interactor_spec.rb
benthorner: I did, briefly. Eventually I decided to take it out, since RSpec makes it hard to do a `to_not change [...].and_not change [...]` for the `change_note` and the `update_type`. I thought just checking the change note would suffice.

With the param building, I agree it's a minor point. I made the change due to having to modify the parameters and the definition (two things) of the former `build_params` method, and the slight surprise of having a helper method in a unit spec. Perhaps I'm being oversensitive to the presence of a method. Are you content to leave it as-is?

kevindew: Cool no worries about the update_type

Yeah I don't mind leaving this in. Though I very much think we should accept and welcome methods in unit tests though, after all they're the same situation as importing a helper and are often more understandable than the use of lets.

app/interactors/content/update_interactor.rb
kevindew: Ah bummer - nils and empty strings are so annoying. I wonder if we should just have loads of our string fields be not null with defaults of empty string to avoid these challenges.

benthorner: This sounds interesting. Could you explain more what you mean here?

kevindew: Sure, no worries.

So in the scenario where we have a table like: 

```
create_table "our_table", force: :cascade do |t|
  t.string "our_field"
end
```

We have the situation where our_field can have a value of an empty string or null which means we have to worry about two different empty states. Whereas in some ways life might be easier if we have a default of empty string:

```
create_table "our_table", force: :cascade do |t|
  t.string "our_field", null: false, default: ''
end
```

Then there is only one empty state. 

We've hit a couple of issues like this with migration where Publishing API has an empty string whereas we have a nil value in our DB.

Loosely ties into an article I saw yesterday actually: https://mitchum.blog/null-values-in-sql-queries/

benthorner: Ahhh, I see. I thought you might be referring to some HTML form behaviour that would transform the nulls into empty strings. Turns out we can expect the parameters to be non-null in practice, so I've modified this commit to avoid `compact` 😸.

PR #1719, opened 2020-01-31
Remove checks for 'put_content' in features

benthorner:

Previously we added these steps to many features as a way to check the
tangential outcomes of a user action, including the document state and API
requests to Publishing API and Asset Manager, which arise from the
PreviewDraftEditionService. Checking API requests in features was pragmatic
when we had limited tests at lower levels, and the idea was that each
feature would cover its specific part of the Publishing API payload. Since
then, we've introduced the
'failsafe preview' feature, such that its far less important or
'optional' that a request is made, since a fallback exists. The checker for
draft user state was added with a similar mindset, to ensure that the
edition state is (re)set appropriately on edit, which is now covered by
'editable?' assertions.

This removes the checks for API requests and draft user state, and renames
or removes the associated step method to leave one that checks for a
timeline entry only.

### DIFF
--- a/spec/features/editing_images/upload_image_spec.rb
+++ b/spec/features/editing_images/upload_image_spec.rb
@@ -8,7 +8,7 @@ RSpec.feature "Upload an image", js: true do
     and_i_crop_the_image
     and_i_fill_in_the_metadata
     then_i_see_the_new_image
-    and_the_preview_creation_succeeded
+    and_i_see_the_timeline_entry
   end
 
   scenario "inline image" do
@@ -18,7 +18,7 @@ RSpec.feature "Upload an image", js: true do
     and_i_crop_the_image
     and_i_fill_in_the_metadata
     then_i_see_the_snippet_is_inserted
-    and_the_preview_creation_succeeded
+    and_i_see_the_timeline_entry
   end
 
   def given_there_is_an_edition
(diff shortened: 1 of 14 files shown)


benthorner:
@1pretz1 you might be interested in this :-)

spec/features/editing_images/upload_image_spec.rb
```diff
@@ -49,12 +49,11 @@ def and_i_upload_a_new_image
   def and_i_crop_the_image
     stub_publishing_api_put_content(@edition.content_id, {})
     click_on "Continue"
-    reset_executed_requests!
```
JonathanHallam: Why has this been removed / why was it there in the first place?

1pretz1:
I am interested indeed, nice work @benthorner!

spec/features/editing_images/upload_image_spec.rb
benthorner: So this was because the upload process involves multiple preview generation requests. We were trying to check the last one of them, so we `reset_executed_requests!`, which was easier than trying to specify which request number we wanted to check. Does that help?

JonathanHallam: yep, thanks :)

PR #1722, opened 2020-02-03
Stripped 'news_article_spec.rb' down to its bare essentials

ChrisBAshton:
- Added 'fill_in "revision[contents][body]"' to make this a valid doc
- Checks are made against Document.last, rather than Document.first,
  as this makes more sense
- Avoid stubbing role appointment links, as these aren't required for
  a minimum viable document
- Choose only a primary publishing organisation tag, as the others
  aren't required for a minimum viable document
- Move 'reset_executed_requests!' call to later in the test, so that
  it is more obvious why it is needed
- Assert that the 'Publish' link exists. This isn't strictly necessary
  but is a peace-of-mind quick check that our UI is showing that the
  document is publishable when it is publishable.

...and then simplified further:

- We only care that a user can create a document from scratch and take
  the minimal steps necessary to make the document valid against its
  schema. This validation alone should give us confidence that it can
  preview (and publish) correctly.
- Previewing doesn't require a body (publishing does), so I've removed
  the 'fill_in body' code from the previous commit. It does not need
  any tagging either, and therefore no publishing API stubbing in the
  'before' block.
- This also meant I could remove the 'reset_executed_requests!' hack.

Done as part of https://trello.com/c/ihcpYeWf/1320-iterate-how-we-test-formats

### DIFF
--- a/spec/features/formats/news_article_spec.rb
+++ b/spec/features/formats/news_article_spec.rb
@@ -3,16 +3,10 @@
 RSpec.describe "News article format" do
   include TopicsHelper
 
-  before do
-    stub_any_publishing_api_put_content
-    stub_any_publishing_api_no_links
-  end
-
   scenario do
     when_i_choose_this_document_type
     and_i_fill_in_the_form_fields
-    and_i_add_some_tags
-    then_i_can_publish_the_document
+    then_the_document_should_be_previewable
   end
 
   def when_i_choose_this_document_type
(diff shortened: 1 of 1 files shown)


spec/features/formats/news_article_spec.rb
```diff
@@ -25,54 +19,16 @@ def when_i_choose_this_document_type
   end
 
   def and_i_fill_in_the_form_fields
+    base_path = Edition.last.document_type.path_prefix + "/a-great-title"
+    stub_publishing_api_has_lookups(base_path => Document.last.content_id)
     fill_in "revision[title]", with: "A great title"
     fill_in "revision[summary]", with: "A great summary"
```
benthorner: I think even this is unnecessary!

ChrisBAshton: It is necessary, as otherwise the `PUT /content/` request is never made (presumably because the base path is not associated with this document, so a new document is created rather than making an edit to an existing one).

benthorner: Ah, I meant the `summary`. Should be unnecessary - trying in the real app, for example.

ChrisBAshton: Ah yes! Removed 👍

PR #1723, opened 2020-02-03
Whitehall notes and fact check - display

gclssvglx:
Whitehall stores document history in three places. This PR addresses the display of two of these into Content Publisher:

- Editorial remarks (Notes tab in Whitehall)

- Fact check requests (Fact checking tab in Whitehall)

For fact checks, there are two things we need to maintain: a fact check request (i.e. a request sent to a department or individual) and a fact check response (i.e. the response coming back from the recipient(s) of the original request).

In all cases, Whitehall entries with content are highlighted to the user, entries without content are not.

**Before...**

before

**After...**

after

Trello: https://trello.com/c/tNsWXZZj/1314-user-interface-display-for-imported-whitehall-editorial-remarks-notes-tab-in-whitehall-and-fact-check-requests-fact-checking-tab

### DIFF
--- a/app/views/documents/history/_whitehall_entry.html.erb
+++ b/app/views/documents/history/_whitehall_entry.html.erb
@@ -1,4 +1,10 @@
-<div class="app-timeline-entry">
+<% whitehall_timeline_entry_content = capture do %>
+  <% if entry.details.internal_note? && entry.details.contents["body"].present? %>
+    <%= simple_format(escape_and_link(entry.details.contents["body"])) %>
+  <% end %>
+<% end %>
+
+<div class="app-timeline-entry <%= "app-timeline-entry--highlighted" if whitehall_timeline_entry_content %>">
   <% if entry.details.imported_from_whitehall? %>
     <%= render "govuk_publishing_components/components/notice", {
       title: t("documents.history.entry_types.whitehall_migration.#{entry.details.entry_type}"),
@@ -15,5 +21,11 @@
         <%= t "documents.history.by" %> <%= entry.created_by.name %>
       <% end %>
     </div>
+
+    <% if whitehall_timeline_entry_content %>
+      <div class="app-timeline-entry__content">
+        <%= whitehall_timeline_entry_content %>
+      </div>
+    <% end %>
   <% end %>
 </div>
--- a/app/views/documents/history/_whitehall_entry.html.erb
+++ b/app/views/documents/history/_whitehall_entry.html.erb
@@ -2,6 +2,16 @@
   <% if entry.details.internal_note? && entry.details.contents["body"].present? %>
     <%= simple_format(escape_and_link(entry.details.contents["body"])) %>
   <% end %>
+
+  <% if entry.details.fact_check_request? && entry.details.contents["instructions"].present? %>
+    <%= simple_format(escape_and_link(entry.details.contents["instructions"])) %>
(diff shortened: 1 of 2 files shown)


app/views/documents/history/_whitehall_entry.html.erb
```diff
@@ -15,5 +15,19 @@
...
+      <%= simple_format(escape_and_link(entry.details.contents["body"])) %>
+    <% end %>
+
+    <% if entry.details.fact_check_request? %>
+      <%= simple_format(escape_and_link(entry.details.contents["instructions"])) %>
+      <%= simple_format(escape_and_link(entry.details.contents["email_address"])) %>
+    <% end %>
+
+    <% if entry.details.fact_check_response? %>
+    <%= simple_format(escape_and_link(entry.details.contents["comments"])) %>
```
emmabeynon: Looks like this is missing an indent

spec/views/documents/history/_whitehall_entry.html.erb_spec.rb
```diff
@@ -20,4 +20,55 @@
     expect(rendered).to have_content(I18n.t!("documents.history.by") + " John Smith")
     expect(rendered).to have_content(I18n.t!("documents.history.entry_types.whitehall_migration.new_edition"))
   end
+
+  it "shows an internal note imported timeline entry with an author" do
```
emmabeynon: Since we have tests for a timeline entry with an author, and there's no special behaviour for displaying the author when it comes to internal notes, fact check requests or fact check responses, I don't think we need to test for these again.  We do however have logic for what we show for each of these timeline entry types, so it would be good to test for these.  For example, we should have a test that the body of the internal note is displayed.

app/views/documents/history/_whitehall_entry.html.erb
```diff
@@ -1,4 +1,4 @@
-<div class="app-timeline-entry">
+<div class="app-timeline-entry app-timeline-entry--highlighted">
```
kevindew: It looks like this will highlight every entry from Whitehall whereas we only want to highlight a few of them (internal note, fact checks). 

It's worth looking at the approach we used in https://github.com/alphagov/content-publisher/blob/b3b61fe7514ff7362192e34c0403edc1237bd134/app/views/documents/history/_content_publisher_entry.html.erb#L40 to determine whether something gets a highlight - we should probably have this covered with a unit test too.

We have a story to make them into a shared component for later to deal with the duplicated logic: https://trello.com/c/iadUckIy/1429-convert-the-timeline-entry-partials-into-a-component

gclssvglx: @kevindew  I read [this](https://github.com/alphagov/content-publisher/pull/1643#discussion_r367391070) and probably misunderstood Simon's intent.  I've updated as per your suggestion and will write a test as part of @emmabeynon's [review](https://github.com/alphagov/content-publisher/pull/1723#discussion_r374180783).

Personally, I prefer the approach you put forward [here](https://github.com/alphagov/content-publisher/pull/1643#discussion_r366489776) - it feels much more flexible.

spec/views/documents/history/_whitehall_entry.html.erb_spec.rb
gclssvglx: @emmabeynon Thanks for your comments and suggestions on this.  I've updated the tests as appropriate and added new tests for the highlighting.

PR #1724, opened 2020-02-03
Retry Whitehall Document Import Job

emmabeynon:
For https://trello.com/c/Y3hKP4W9/1366-handle-retrying-whitehall-import-job

This PR implements retrying the Whitehall Document Import Job when a `GdsApi::BaseError` is raised.  When other errors are raised, or the retry limit is exhausted, we update the Whitehall Document Import with the appropriate state and a dump of the error that was raised.  This work has involved moving some existing functionality around which hopefully defines some class responsibilities more clearly.  

This PR does not cover calling the Whitehall Export API to lock and unlock a document, or mark a document as migrated.  This will be covered in a subsequent PR.


lib/whitehall_importer/import.rb
```diff
@@ -19,6 +19,10 @@ def call
       end
 
       ActiveRecord::Base.transaction do
+        document_import.update!(
+          payload: whitehall_document,
```
brucebolt: I think we put this outside the transaction originally so the payload and content_id will still get written to the database if the rest of the import fails.  It looks like that won't be the case anymore.

app/jobs/whitehall_document_import_job.rb
```diff
@@ -18,7 +18,12 @@ def perform(document_import)
...
   def self.handle_error(document_import, error)
-    state = document_import.imported? ? "sync_failed" : "import_failed"
-    document_import.update!(state: state, error_log: error.inspect)
+    case error
+    when WhitehallImporter::AbortImportError
+      document_import.update!(state: "import_aborted", error_log: error.inspect)
+    else
+      state = document_import.imported? ? "sync_failed" : "import_failed"
+      document_import.update!(state: state, error_log: error.inspect)
+    end
```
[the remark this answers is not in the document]
brucebolt: Ah, looking at a later commit, this probably won't be possible.

lib/whitehall_importer/create_migration.rb
```diff
@@ -0,0 +1,42 @@
...
+          page["documents"].each do |document|
+            WhitehallMigration::DocumentImport.create!(
+              whitehall_document_id: document["document_id"],
+              whitehall_migration_id: record.id,
+              state: "pending",
+            )
+          end
+        end
+        record
+      end
```
kevindew: For brevity we might want to break this method up now we have the luxury of this being in a dedicated class.

As a suggestion would be:

```
    def call
      whitehall_migration = ActiveRecord::Base.transaction do
        record = create_migration
        document_list.each { |page| create_document_imports(record, page) }
        record
      end
      
      whitehall_migration.document_imports.find_each do |document_import|
        WhitehallDocumentImportJob.perform_later(document_import)
      end
      
      whitehall_migration
    end
```

lib/whitehall_importer/create_migration.rb
```diff
@@ -0,0 +1,42 @@
...
+
+        whitehall_export = GdsApi.whitehall_export.document_list(organisation_content_id, document_type)
+
+        whitehall_export.each do |page|
+          page["documents"].each do |document|
+            WhitehallMigration::DocumentImport.create!(
+              whitehall_document_id: document["document_id"],
+              whitehall_migration_id: record.id,
+              state: "pending",
+            )
```
kevindew: Sorry I know you didn't introduce this.

It seems a bit unusual that this creates this association without adding it to the record.
For idiomatic Rails I'd have thought we wanted something like:

```
record.document_import << WhitehallMigration::DocumentImport.create!

# though I think we can do it shorter with: 

record.create_document_import!(whitehall_document_id: document["document_id"], state: "pending")
```

lib/whitehall_importer/create_migration.rb
```diff
@@ -0,0 +1,42 @@
...
+    attr_reader :organisation_content_id, :document_type
+
+    def self.call(*args)
+      new(*args).call
+    end
+
+    def initialize(organisation_content_id, document_type)
+      @organisation_content_id = organisation_content_id
+      @document_type = document_type
+    end
```
kevindew: I'm not sure quite on the status but we'll probably have to rebase this against Bruce's change that adds the sub types.

lib/whitehall_importer/import.rb
```diff
@@ -32,15 +40,17 @@ def call
...
-      document_import.reload
-      raise
     end
 
   private
 
+    def export_whitehall_document
+      GdsApi.whitehall_export.document_export(document_import.whitehall_document_id)
+                             .to_h
+    end
```
kevindew: Rather than have this method response set the instance variable I wonder if we should have it as a lazily loaded variable. For example:

```
    def whitehall_document
      @whitehall_document ||= GdsApi.whitehall_export
                                    .document_export(document_import.whitehall_document_id)
                                    .to_h
    end
```

lib/whitehall_importer/sync.rb
```diff
@@ -0,0 +1,27 @@
+# frozen_string_literal: true
+
+module WhitehallImporter
+  class Sync
+    attr_reader :whitehall_import
+
+    def self.call(whitehall_import)
+      new(whitehall_import).call
+    end
```
kevindew: If you go for:

```
    def self.call(*args)
      new(*args).call
    end
```

Then we won't have to worry about keeping the attributes of this method in sync with initialize.

app/jobs/whitehall_document_import_job.rb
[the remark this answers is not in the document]
emmabeynon: Yeah I did have a similar thought, but it didn't quite work out unfortunately.

lib/whitehall_importer/create_migration.rb
emmabeynon: Good idea 👍

lib/whitehall_importer/create_migration.rb
emmabeynon: Indeed 😬

lib/whitehall_importer/import.rb
emmabeynon: Ah good to know - I'll move it back out 👍

lib/whitehall_importer/import.rb
emmabeynon: Done

lib/whitehall_importer/create_migration.rb
emmabeynon: `record.document_imports.create!` worked for me 👍

lib/whitehall_importer/sync.rb
emmabeynon: Done

emmabeynon:
@brucebolt Thanks for spotting that issue with the single document rake - it's a bit of a concern that the tests didn't pick that up!  I think I've found a solution that will make the tests actually fail in future, and also made changes so that `#import_and_sync` is no longer called.  https://trello.com/c/ffU41pHl/1391-adapt-import-single-document-rake-task-to-create-a-migration will provide a more elegant solution soon.

@kevindew Thanks for the review - your comments should be addressed now 🌟

spec/lib/whitehall_importer/sync_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+      expect(whitehall_migration_document_import).to be_completed
+    end
+
+    it "raises if the WhitehallMigration::DocumentImport doesn't have a state of imported" do
+      whitehall_migration_document_import = build(:whitehall_migration_document_import)
+      expect { described_class.call(whitehall_migration_document_import) }
+        .to raise_error(RuntimeError, "Cannot sync with a state of pending")
+    end
+
+    context "when the sync fails" do
```
kevindew: I don't think we need this context anymore because we don't do anything in this class for dealing with this exception (I'm assuming this is just something remaining from a C+P of the old tests)

emmabeynon: Yes...I'm relieved there's only one! 😅

lib/whitehall_importer/create_migration.rb
emmabeynon: This has been rebased now 👍

emmabeynon:
@kevindew No in the case of the rake task tests the rake tasks just didn't seem to be invoked, and the tests just passed. Reenabling the tasks before each test seemed to do the trick.  I do understand the concerns about so much mocking going on, though don't see a way around having to stub the API requests.

kevindew:
> @kevindew No in the case of the rake task tests the rake tasks just didn't seem to be invoked, and the tests just passed. Reenabling the tasks before each test seemed to do the trick. I do understand the concerns about so much mocking going on, though don't see a way around having to stub the API requests.

Ah cool no worries. We'll just keep an eye out then. If we hit situations where this breaks yet our tests still pass we'll know we have a problem.

PR #1727, opened 2020-02-04
Strip exif data from imported images

1pretz1:
We need to strip exif data from images we import due to whitehall not
acknowledging these attributes, for example - rotation.

This causes problems in the browser if an image has rotation exif data
as in certain circumstances the browser will display the image rotated
(image open in new tab) and sometimes not.

The problem also persists when we import a rotated image to content
publisher as we respect the orientation before we strip the data which
ends up potentially rotating the image in a way the user didn't expect
which then fails validation, E.g width = height.

To get round this issue we want to strip the image first to ensure
content-publisher sees the image in the same way whitehall does.

Trello:
https://trello.com/c/b3vqfsA5/1395-deal-with-rotated-images-in-import


lib/whitehall_importer/create_image_revision.rb
```diff
@@ -56,7 +56,11 @@ def create_blob_revision(temp_image)
     def normalise_image(file)
       upload_checker = Requirements::ImageUploadChecker.new(file)
       abort_on_issue(upload_checker.issues)
-      normaliser = ImageNormaliser.new(file)
+
+      raw_image = MiniMagick::Image.open(file.path)
+      raw_image.combine_options(&:strip)
```
emmabeynon: Does `raw_image.strip` have the same effect? https://www.rubyguides.com/2018/12/minimagick-gem

spec/lib/whitehall_importer/create_image_revision_spec.rb
```diff
@@ -86,6 +86,19 @@
       end
     end
 
+    context "Image has exif data" do
+      let(:whitehall_image) do
+        build(:whitehall_export_image, fixture_file: "960x640-rotated.jpg")
+      end
+
+      it "should strip the exif data from the image" do
+        revision = described_class.call(document_import, whitehall_image, ["valid-image.jpg"])
```
emmabeynon: I wonder if it's worth checking for the present of exif data before processing the image in order to be certain it has been stripped at the end?

lib/whitehall_importer/create_image_revision.rb
1pretz1: Ah ha, you're absolutely correct! I've moved it all onto one line as it boiled down to fairly simple code 💥

spec/lib/whitehall_importer/create_image_revision_spec.rb
1pretz1: Hmm I know what you mean but you can't really as the image doesn't exist before the point of execution. I mean you could test the image fixture but you shouldn't need to because you should trust a fixture is what it says it is.

I pondered whether a test is even needed as you're pretty much just testing methods provided by minimagick but decided we should probably have one. :)

1pretz1:
I've addressed your comments @emmabeynon and tested on integration so should be ready to go! 🚀

spec/lib/whitehall_importer/create_image_revision_spec.rb
emmabeynon: Sure, is it worth renaming the image to specify that it has exif data?  I suppose my concern is if the fixture somehow gets replaced with a image that doesn't have it (if that's even possible)...I don't know much about exif to be honest 🤷‍♀

lib/whitehall_importer/create_image_revision.rb
```diff
@@ -56,7 +56,9 @@ def create_blob_revision(temp_image)
     def normalise_image(file)
       upload_checker = Requirements::ImageUploadChecker.new(file)
       abort_on_issue(upload_checker.issues)
-      normaliser = ImageNormaliser.new(file)
+
+      stripped_image = MiniMagick::Image.open(file.path).strip
```
kevindew: Does this create a new file or modify the existing file? just wondering with regard to the need of the file and/or if it modifies the fixture.

spec/lib/whitehall_importer/create_image_revision_spec.rb
```diff
@@ -86,6 +86,19 @@
...
+
+        image = MiniMagick::Image.open(revision.image.image_revisions.last.blob_revision.blob)
```
kevindew: I'd have thought you could do this with just `revision.blob`

spec/lib/whitehall_importer/create_image_revision_spec.rb
1pretz1: I mean I can rename to include exif but the 'rotate' implies that it has exif orientation data and it also follows the naming convention of another fixture that presumably also contains exif orientation data - '640x960-rotated.jpg' so I think the naming is okay unless you still think it's not clear?

lib/whitehall_importer/create_image_revision.rb
1pretz1: It creates a new temp file so we have to pass in the resulting stripped image rather than relying on the original as the paths diverge.

spec/lib/whitehall_importer/create_image_revision_spec.rb
1pretz1: You can indeed, that looks much cleaner!

lib/whitehall_importer/create_image_revision.rb
kevindew: Great - glad it doesn't modify the source.

PR #1730, opened 2020-02-05
Increase test log level to 'warn'

benthorner:
This is a shameless tweak to improve performance when running tests
in Docker, which is sensitive to heavy disk IO. I'm pretty sure we
don't have a use case for info-level logs in tests. I'm also open to
making this an environment variable, which might be a more reusable
approach across some of the other apps. I'm seeing about a 30-40s
improvement in test runtime on my machine.

Before

    govuk-docker-run bundle exec rspec  1.59s user 0.27s system 0% cpu 3:44.85 total

After

    govuk-docker-run bundle exec rspec  1.55s user 0.27s system 0% cpu 3:02.77 total

### DIFF
--- a/config/environments/test.rb
+++ b/config/environments/test.rb
@@ -33,6 +33,9 @@ Rails.application.configure do
   # Store uploaded files on the local file system in a temporary directory.
   config.active_storage.service = :local
 
+  # Use a higher log level to avoid unnecessary clutter and disk IO.
+  config.log_level = :warn
+
   config.action_mailer.perform_caching = false
 
   # Tell Action Mailer not to deliver emails to the real world.


kevindew:
⭐️ for providing benchmarks.

That's a great improvement 👍 

I don't have any qualms with doing this as I think it's incredibly low the amount of times I benefit from info level logs in tests. I don't think we need an env var either really as I imagine in the rare scenario you'd want to change this you could just edit the file - I imagine mostly people reach for a debugger before considering that.

On my local machine I see a much smaller drop in time but still a drop:

Before: Finished in 1 minute 59 seconds (files took 6.12 seconds to load)
After: Finished in 1 minute 52.36 seconds (files took 5.73 seconds to load)

Have you experimented with making the log directory delegated on docker to see if that helps performance at all? Also, although not a practical long term thing, have you tried mounting the logs and/or tmp on a tempfs to see how big an impact that has?

kevindew:
> On my machine I did a cold run on Ben's branch, finishing in 3m47s. Then switched to master, where it completed in 3m01s 😕
> 
> But spoke to Ben and we figured the second run must have benefited from some caching. When I re-ran the tests in Ben's branch, it came down to 2m26.5s 🎉

Yup that time difference looks likely to be the time it takes to build the assets 🤢

benthorner:
Thanks @kevindew and @ChrisBAshton for trying this out. Here's some more benchmarks for my machine.

After (again)

    govuk-docker-run bundle exec rspec  1.56s user 0.27s system 1% cpu 2:54.57 total

Before (again)

    govuk-docker-run bundle exec rspec  1.65s user 0.38s system 0% cpu 3:52.63 total

After (+tmpfs stuff)

    govuk-docker-run bundle exec rspec  1.63s user 0.28s system 1% cpu 2:58.19 total

@kevindew I'd been toying with tmpfs stuff (it isn't clear from the docs that it works with Macs). For logs, it would be a shame to make them completely unavailable. FYI you need to explicitly set the `uid` of the mount point to `1000` in order for it to work properly i.e.

```
tmpfs:
  - /govuk/content-publisher/tmp/storage:uid=1000
  - /govuk/content-publisher/log:uid=1000
```

It looks like it doesn't provide any particular improvements over the log approach. And the Redis issue is new to me 🤔.

PR #1732, opened 2020-02-05
Remove feature spec for deleting later editions

benthorner:
Previously this spec was doing a lot of work to check the edge case
behaviour that we switch back to a live edition when deleting a new draft.
This is covered by the unit spec for the associated service.


alex-ju:
May need a rebase, but good to be merged otherwise.

PR #1733, opened 2020-02-05
Remove 'revision' param for edition checkers

benthorner:

Previously we passed the 'edition' and the 'revision' between checkers for
requirements issues, often defaulting the 'revision' to the current one for
the edition. Having both objects available lead to confusion about which to
use to access data, and was originally because the revision had updated
data and was not (yet) assigned to the edition.

### DIFF
--- a/app/interactors/content/update_interactor.rb
+++ b/app/interactors/content/update_interactor.rb
@@ -38,7 +38,7 @@ private
     issues = Requirements::CheckerIssues.new
 
     edition.document_type.contents.each do |field|
-      issues += field.pre_update_issues(edition, revision)
+      issues += field.pre_update_issues(edition)
     end
 
     context.fail!(issues: issues) if issues.any?
--- a/app/interactors/content/update_interactor.rb
+++ b/app/interactors/content/update_interactor.rb
@@ -32,6 +32,7 @@ private
     updater.assign(change_note_params)
     context.fail! unless updater.changed?
     context.revision = updater.next_revision
+    EditDraftEditionService.call(edition, user, revision: revision)
   end
 
   def check_for_issues
@@ -45,7 +46,6 @@ private
   end
 
   def update_edition
-    EditDraftEditionService.call(edition, user, revision: revision)
     edition.save!
   end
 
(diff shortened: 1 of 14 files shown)


benthorner:
Notes to self: having done a little tinkering with this, I found that it's quite hard to (optionally) use raw params in the checkers; the point where I stopped was with `the title_and_base_path`:

```
def pre_update_issues(edition, params)
   <custom_checks>
   issues + pre_preview_issues(edition, <contrived_params>)
end
```

The `ImageRevisionChecker` is another example. Changing the checkers to use contrived params seems like a step backwards, so I have two suggestions that mean we could continue with the object/model approach:

```
# make the field checks agnostic to a revision
def pre_update_issues(edition_or_revision)
...
```

```
# solve the cleanliness problem in the interactor
temp_edition = edition.dup.assign_attributes(revision)
...
field.pre_update_issues(temp_edition)
```

The `Images::UpdateInteractor` is interesting: we put the dirty image revision into the context (for replay in the view), which means we no longer have the clean one; as we've discussed, this isn't actually a problem, and just illustrates how subtle this is!

I'm tempted to go for the second approach, which we can also test for in the interactor. A more extensive use of params in checkers would be better as a pattern for avoiding this pitfall in future, so we could still explore doing something like the following:

```
def pre_update_issues(edition, params)
   <custom_checks>
   issues + title_issues(edition, params[:title])
end

def pre_preview_issues(edition)
  title_issues(edition, edition.title)
end
```

So maybe not so hard after all 😸 .

kevindew:
> A more extensive use of params in checkers would be better as a pattern for avoiding this pitfall in future, so we could still explore doing something like the following:
> 
> ```
> def pre_update_issues(edition, params)
>    <custom_checks>
>    issues + title_issues(edition, params[:title])
> end
> 
> def pre_preview_issues(edition)
>   title_issues(edition, edition.title)
> end
> ```
> 
> So maybe not so hard after all 😸 .

This last approach is my preferred one 👍 👍

benthorner:
Superseded by https://github.com/alphagov/content-publisher/pull/1753

PR #1744, opened 2020-02-07
Ignore .DS_Store

ChrisBAshton:
I considered this instruction:
 git config --global core.excludesfile '~/.gitignore_global'

But we ignore .DS_Store in lots of alphagov repos, and can't rely on
everybody to set up their git config in the same way. Ignoring it
explicitly is a common enough convention that it's worth doing here.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,6 +19,7 @@
 
 /public/assets
 .byebug_history
+.DS_Store
 
 /coverage
 /public/packs


kevindew:
Although a minor concern I don’t think we should do this as it creates inconsistencies and breaks Rails conventions. One it contradicts the comment about using gitignore_global for operating system files and two it is the only item in here that is not a product of the application or its dependencies.

Personally I’m for following the Rails convention on this of only adding files produced by the application or dependencies as this is often a slippery slope of adding extra stuff unrelated to the repo. Take the specialist publisher gitignore for example that has .idea ignored, for someone to understand why that is ignored you have to know about IntelliJ text editors producing that directory. I’d actually think a better step would be removing these from other repos as I think the sooner developers are acquainted with gitignore_global the better.

issyl0:
I 100% agree with @kevindew here. People ignoring non-app files in app-level `.gitignore`s is one of my pet peeves. Let's remove them all.

PR #1746, opened 2020-02-07
Update Whitehall document when importing

emmabeynon:
For https://trello.com/c/Y3hKP4W9/1366-handle-retrying-whitehall-import-job

This calls the Whitehall export API at various stages of the import and sync job so that Whitehall knows about a document's status.

- We call the lock endpoint before a document is exported from Whitehall
- We call the unlock endpoint if a document import fails, not the sync
- We call the migrated endpoint if the document imports and syncs successfully

### DIFF
--- a/app/jobs/whitehall_document_import_job.rb
+++ b/app/jobs/whitehall_document_import_job.rb
@@ -19,6 +19,10 @@ class WhitehallDocumentImportJob < ApplicationJob
   end
 
   def self.handle_error(document_import, error)
+    if document_import.pending?
+      GdsApi.whitehall_export.unlock_document(document_import.whitehall_document_id)
+    end
+
     case error
     when WhitehallImporter::IntegrityCheckError
       document_import.update!(
--- a/app/jobs/whitehall_document_import_job.rb
+++ b/app/jobs/whitehall_document_import_job.rb
@@ -20,7 +20,11 @@ class WhitehallDocumentImportJob < ApplicationJob
 
   def self.handle_error(document_import, error)
     if document_import.pending?
-      GdsApi.whitehall_export.unlock_document(document_import.whitehall_document_id)
+      begin
+        GdsApi.whitehall_export.unlock_document(document_import.whitehall_document_id)
+      rescue StandardError => e
+        logger.warn("Failed to unlock Whitehall document: #{e.inspect}")
+      end
     end
 
     case error
--- a/spec/jobs/whitehall_document_import_job_spec.rb
+++ b/spec/jobs/whitehall_document_import_job_spec.rb
@@ -184,4 +184,9 @@ RSpec.describe WhitehallDocumentImportJob do
   def stub_whitehall_api_unlock_document(document_id)
     stub_request(:post, "#{whitehall_host}/government/admin/export/document/#{document_id}/unlock")
   end
+
+  def stub_whitehall_api_unlock_document_error(document_id, error)
+    stub_request(:post, "#{whitehall_host}/government/admin/export/document/#{document_id}/unlock")
+      .to_raise(error)
+  end
 end
--- a/spec/jobs/whitehall_document_import_job_spec.rb
+++ b/spec/jobs/whitehall_document_import_job_spec.rb
@@ -26,7 +26,7 @@ RSpec.describe WhitehallDocumentImportJob do
     allow(WhitehallImporter::Sync).to receive(:call)
                                   .and_return(completed_document_import)
     allow(whitehall_migration).to receive(:check_migration_finished)
-    stub_whitehall_api_unlock_document(
+    stub_whitehall_unlock_document(
       whitehall_migration_document_import.whitehall_document_id,
     )
   end
@@ -56,7 +56,7 @@ RSpec.describe WhitehallDocumentImportJob do
     end
 
     it "unlocks the document in Whitehall" do
-      request = stub_whitehall_api_unlock_document(
+      request = stub_whitehall_unlock_document(
         whitehall_migration_document_import.whitehall_document_id,
       )
 
@@ -79,7 +79,7 @@ RSpec.describe WhitehallDocumentImportJob do
     end
 
     it "does not unlock the document in Whitehall" do
-      request = stub_whitehall_api_unlock_document(
+      request = stub_whitehall_unlock_document(
         imported_document_import.whitehall_document_id,
       )
 
@@ -121,7 +121,7 @@ RSpec.describe WhitehallDocumentImportJob do
     let(:error) { StandardError.new }
     before do
       allow(WhitehallImporter::Import).to receive(:call).and_raise(error)
-      stub_whitehall_api_unlock_document(
+      stub_whitehall_unlock_document(
         whitehall_migration_document_import.whitehall_document_id,
       )
     end
@@ -185,7 +185,7 @@ RSpec.describe WhitehallDocumentImportJob do
     let(:error) { StandardError.new }
     before do
       allow(WhitehallImporter::Import).to receive(:call).and_raise("import error")
-      stub_whitehall_api_unlock_document_error(
+      stub_whitehall_unlock_document_error(
         whitehall_migration_document_import.whitehall_document_id, error
       )
     end
@@ -201,11 +201,11 @@ RSpec.describe WhitehallDocumentImportJob do
     end
   end
 
-  def stub_whitehall_api_unlock_document(document_id)
+  def stub_whitehall_unlock_document(document_id)
     stub_request(:post, "#{whitehall_host}/government/admin/export/document/#{document_id}/unlock")
   end
 
-  def stub_whitehall_api_unlock_document_error(document_id, error)
+  def stub_whitehall_unlock_document_error(document_id, error)
     stub_request(:post, "#{whitehall_host}/government/admin/export/document/#{document_id}/unlock")
       .to_raise(error)
   end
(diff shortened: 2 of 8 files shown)


spec/jobs/whitehall_document_import_job_spec.rb
```diff
@@ -46,61 +49,90 @@
...
+  context "when a GdsApi::BaseError exception is raised" do
+    let(:error) { GdsApi::BaseError.new }
+    before do
+      allow(WhitehallImporter::Import).to receive(:call).and_raise(error)
+    end
 
+    it "retries the job" do
+      WhitehallDocumentImportJob.perform_now(whitehall_migration_document_import)
+
+      expect(WhitehallDocumentImportJob).to have_been_enqueued
```
1pretz1: Does this actually test the job has been retried rather than just queued originally?

app/jobs/whitehall_document_import_job.rb
```diff
@@ -15,10 +15,19 @@ class WhitehallDocumentImportJob < ApplicationJob
   def perform(document_import)
     document_import = WhitehallImporter::Import.call(document_import)
     document_import = WhitehallImporter::Sync.call(document_import)
+
+    if document_import.completed?
+      GdsApi.whitehall_export.document_migrated(document_import.whitehall_document_id)
```
kevindew: I feel like we should be doing this within the sync method since it's part of the GOV.UK sync, and would then also benefit from retries.

I wonder if we have to ponder the ordering of this too - I'm thinking it might be better that we tell Whitehall we're migrated as the first task in sync, so we don't risk having duplicate items in the search index.

app/jobs/whitehall_document_import_job.rb
```diff
@@ -15,10 +15,19 @@ class WhitehallDocumentImportJob < ApplicationJob
...
+    end
+
     document_import.whitehall_migration&.check_migration_finished
   end
 
   def self.handle_error(document_import, error)
+    if document_import.pending?
+      GdsApi.whitehall_export.unlock_document(document_import.whitehall_document_id)
```
kevindew: We probably have to consider what happens if this fails as if we hit an exception here none of our other exception handling code will run.

app/jobs/whitehall_document_import_job.rb
```diff
@@ -15,10 +15,19 @@ class WhitehallDocumentImportJob < ApplicationJob
   def perform(document_import)
     document_import = WhitehallImporter::Import.call(document_import)
     document_import = WhitehallImporter::Sync.call(document_import)
+
+    if document_import.completed?
+      GdsApi.whitehall_export.document_migrated(document_import.whitehall_document_id)
+    end
+
     document_import.whitehall_migration&.check_migration_finished
```
kevindew: Did we plan to include this as part of error handling too or am I thinking of a different story or PR?

spec/lib/whitehall_importer/import_spec.rb
```diff
@@ -17,48 +17,47 @@
       allow(WhitehallImporter::IntegrityChecker)
         .to receive(:new)
         .and_return(instance_double(WhitehallImporter::IntegrityChecker, valid?: true))
+      stub_whitehall_api_document_export(
```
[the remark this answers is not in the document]
emmabeynon: I decided to use API to make it clear that we're interacting with the API part of Whitehall (even it's not technically called Whitehall API).  I can change it if you think it's important to drop it.

spec/jobs/whitehall_document_import_job_spec.rb
emmabeynon: I believe so.  If you look at the errors on which we don't retry, we test that the job has not been enqueued e.g. https://github.com/alphagov/content-publisher/blob/master/spec/jobs/whitehall_document_import_job_spec.rb#L148

app/jobs/whitehall_document_import_job.rb
emmabeynon: Would it perhaps make sense only to tell Whitehall that the document is migrated once the Publishing API publish call has been made?  Otherwise if we tell Whitehall it's migrated prior to that, we could have the situation where the document isn't in any index at all.

app/jobs/whitehall_document_import_job.rb
emmabeynon: Different PR 👍

app/jobs/whitehall_document_import_job.rb
kevindew: Yup, that's a risk too. I'd imagine we look less broken just missing the item than having it duplicated. 

Though thinking about it the risk of duplicating is mostly only likely from the subsequent tasks running (clear linkset links and migrate assets) so the sweet spot might be to run this before migrate assets as that's the item with the least consequences for failing.

spec/lib/whitehall_importer/import_spec.rb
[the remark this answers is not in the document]
1pretz1: It's not overly important no, it was just an observation I made when reading through the tests and then looking at the names of the stubs we currently have in gds-api-adapters. Will leave the decision up to thee! 😄

spec/jobs/whitehall_document_import_job_spec.rb
1pretz1: Ah ha cool, just checking! 👍

spec/lib/whitehall_importer/import_spec.rb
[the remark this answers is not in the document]
emmabeynon: Done 👍

app/jobs/whitehall_document_import_job.rb
emmabeynon: Done 👍

app/jobs/whitehall_document_import_job.rb
emmabeynon: Good catch - I've written an approach to this here https://github.com/alphagov/content-publisher/pull/1746/commits/18a26d567f0b4dfc0fa2b5430e300edc3c195c55

spec/jobs/whitehall_document_import_job_spec.rb
```diff
@@ -46,61 +49,90 @@
...
+  end
 
-    it "updates the document import state to 'import_failed' if the import failed" do
-      perform_enqueued_jobs do
-        WhitehallDocumentImportJob.perform_now(whitehall_migration_document_import)
-      end
+  context "when the sync fails" do
+    let(:error_message) { "sync failed error" }
+    before do
+      allow(WhitehallImporter::Import).to receive(:call)
```
kevindew: Do we need to do this? I notice we have a global mock.

spec/lib/gds_api/whitehall_export_spec.rb
```diff
@@ -1,14 +1,21 @@
 # frozen_string_literal: true
 
 RSpec.describe GdsApi::WhitehallExport do
-  describe "document_list" do
-    let(:whitehall_adapter) { GdsApi::WhitehallExport.new(Plek.find("whitehall-admin")) }
+  let(:whitehall_adapter) { GdsApi::WhitehallExport.new(Plek.find("whitehall-admin")) }
+  let(:whitehall_host) { Plek.new.external_url_for("whitehall-admin") }
+  let(:document_id) { "123" }
 
+  describe "document_list" do
```
kevindew: Sorry I know you didn't introduce this and you're probably already aware of the convention. However would be good to prefix this method name with a # as per http://www.betterspecs.org/#describe

app/jobs/whitehall_document_import_job.rb
```diff
@@ -19,19 +19,31 @@ def perform(document_import)
   end
 
   def self.handle_error(document_import, error)
+    errors = [error.inspect]
+
+    if document_import.pending?
+      begin
+        GdsApi.whitehall_export.unlock_document(document_import.whitehall_document_id)
+      rescue GdsApi::BaseError => e
```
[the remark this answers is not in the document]
emmabeynon: Change to StandardError 👍

app/jobs/whitehall_document_import_job.rb
```diff
@@ -19,19 +19,31 @@ def perform(document_import)
...
+        errors << e.inspect
```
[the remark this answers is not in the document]
emmabeynon: I figured we needed to capture the error somewhere; logging sounds good, I've changed it to do that.

spec/jobs/whitehall_document_import_job_spec.rb
emmabeynon: Good spot - I've removed it now.

spec/lib/gds_api/whitehall_export_spec.rb
emmabeynon: Done 👍

emmabeynon:
Thanks @kevindew - your comments should be addressed now.

app/jobs/whitehall_document_import_job.rb
```diff
@@ -19,6 +19,14 @@ def perform(document_import)
   end
 
   def self.handle_error(document_import, error)
+    if document_import.pending?
+      begin
+        GdsApi.whitehall_export.unlock_document(document_import.whitehall_document_id)
+      rescue StandardError => e
+        logger.warn(e.inspect)
```
kevindew: Sorry to be a pain - may want to prefix this with an intro line `logger.warn("Failed to unlock whitehall document: #{e.inspect}")`

emmabeynon: Done 👍

PR #1748, opened 2020-02-07
Introduce government data alert

JonathanHallam:
Trello - https://trello.com/c/FaHAEajH/1318-alert-2ndline-when-we-have-no-cached-data-on-governments

This iterates out healthcheck endpoint to include a check for government data, a connection to redis and to the database. This will allow us to alert 2nd line and our slack channel in the event that government data is unavailable. 

### Relies on
https://github.com/alphagov/govuk-developer-docs/pull/2301
https://github.com/alphagov/govuk-puppet/pull/10129

### DIFF
--- a/app/controllers/healthcheck_controller.rb
+++ b/app/controllers/healthcheck_controller.rb
@@ -1,0 +1,12 @@
+class HealthcheckController < ApplicationController
+  skip_before_action :authenticate_user!
+
+  def index
+    healthcheck = GovukHealthcheck.healthcheck([
+                    GovukHealthcheck::SidekiqRedis,
+                    GovukHealthcheck::ActiveRecord,
+                    Healthcheck::GovernmentDataCheck,
+                  ])
+    render json: healthcheck
+  end
+end
--- a/lib/bulk_data/cache.rb
+++ b/lib/bulk_data/cache.rb
@@ -6,7 +6,7 @@ module BulkData
 
     class << self
       delegate :cache, to: :instance
-      delegate :clear, :middleware, to: :cache
+      delegate :clear, :exist?, :middleware, to: :cache
     end
 
     attr_reader :cache
@@ -36,10 +36,6 @@ module BulkData
     def self.delete(key)
       cache.delete(key)
       cache.delete("#{key}:created")
-    end
-
-    def self.clear
-      cache.clear
     end
 
   private
--- a/lib/bulk_data/government_repository.rb
+++ b/lib/bulk_data/government_repository.rb
@@ -44,5 +44,13 @@ module BulkData
     rescue GdsApi::BaseError
       raise RemoteDataUnavailableError
     end
+
+    def cache_age
+      Time.current - Cache.written_at(CACHE_KEY)
+    end
+
+    def cache_populated?
+      Cache.exist?(CACHE_KEY)
+    end
   end
 end
--- a/lib/healthcheck/government_data_check.rb
+++ b/lib/healthcheck/government_data_check.rb
@@ -1,0 +1,36 @@
+module Healthcheck
+  class GovernmentDataCheck
+    attr_reader :government_repo
+    def initialize
+      @government_repo = BulkData::GovernmentRepository.new
+    end
+
+    def name
+      :government_data_check
+    end
+
+    def status
+      return GovukHealthcheck::CRITICAL unless government_repo.cache_populated?
+      return GovukHealthcheck::WARNING if government_repo.cache_age > 6.hours
+
+      GovukHealthcheck::OK
+    end
+
+    def message
+      return "No government data availible" if status == GovukHealthcheck::CRITICAL
+
+      warning_details_content if status == GovukHealthcheck:: WARNING
+    end
+
+    def enabled?
+      true
+    end
+
+  private
+
+    def warning_details_content
+      data_age = government_repo.cache_age.to_i
+      "Government data not refreshed in #{data_age / 1.hour} hours."
+    end
+  end
+end
--- a/spec/lib/healthcheck/government_data_check_spec.rb
+++ b/spec/lib/healthcheck/government_data_check_spec.rb
@@ -1,0 +1,50 @@
+RSpec.describe Healthcheck::GovernmentDataCheck do
+  let(:gov_data_check) { described_class.new }
+
+  describe "#status" do
+    context "when government data is unavailable" do
+      it "returns critical" do
+        expect(gov_data_check.status).to be(:critical)
+      end
+    end
+
+    context "when government data is set to expire" do
+      it "gives a warning" do
+        populate_default_government_bulk_data
+        travel_to(7.hours.from_now) do
+          expect(gov_data_check.status).to be(:warning)
+        end
+      end
+    end
+
+    context "when everything is fine" do
+      it "returns ok" do
+        populate_default_government_bulk_data
+        expect(gov_data_check.status).to be(:ok)
+      end
+    end
+  end
+
+  describe "#message" do
+    context "when government data is unavailable" do
+      it "displays an appropriate message" do
+        expect(gov_data_check.message).to eq("No government data availible")
+      end
+    end
+
+    context "when government data is set to expire" do
+      it "displays an appropriate message" do
+        populate_default_government_bulk_data
+        travel_to(Time.zone.now + 6.5.hours)
+        expect(gov_data_check.message).to eq("Government data not refreshed in 6 hours.")
+      end
+    end
+
+    context "when everything is fine" do
+      it "displays an appropriate message" do
+        populate_default_government_bulk_data
+        expect(gov_data_check.message).to be_nil
+      end
+    end
+  end
+end
--- a/spec/requests/healthcheck_spec.rb
+++ b/spec/requests/healthcheck_spec.rb
@@ -1,0 +1,17 @@
+RSpec.describe "Healthcheck" do
+  describe "GET /healthcheck" do
+    before do
+      populate_default_government_bulk_data
+    end
+
+    it "returns a 200 HTTP status" do
+      get healthcheck_path
+      expect(response).to have_http_status(:ok)
+    end
+
+    it "includes a status in the response body" do
+      get healthcheck_path
+      expect(JSON.parse(response.body)).to have_key("status")
+    end
+  end
+end
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -101,7 +101,7 @@ Rails.application.routes.draw do
     delete "/file-attachments/:file_attachment_id" => "file_attachments#destroy"
   end
 
-  get "/healthcheck", to: proc { [200, {}, %w[OK]] }
+  get "/healthcheck", to: "healthcheck#index"
 
   get "/how-to-use-publisher" => "publisher_information#how_to_use_publisher", as: :how_to_use_publisher
   get "/beta-capabilities" => "publisher_information#beta_capabilities", as: :beta_capabilities


cbaines:
A more general comment, but with the Email Alert API, the Platform Health Team has been working on trying to get the per-machine healthcheck back to just being about the health of the app on that machine [1].

1: https://github.com/alphagov/email-alert-api/commits/master/app/controllers/healthcheck_controller.rb

At least in the AWS world, the ELB's can use healthchecks to decide which machines to route traffic to. At least currently for the Email Alert API, it's not looking at the status of the healthcheck, but that would be good to change.

Regarding the `Healthcheck::GovernmentDataCheck` aspect of the health check, is this something that can differ between machines?

kevindew:
Thanks @cbaines for the input - this would be a check that applies to all machines

Are there decisions (or better documentation) on what we aspire to have on /healthcheck and for things that aren't covered there how. I gather for Email Alert API it's moved checks out of using Ruby and instead to monitor graphite metrics, which makes me unsure of precedent to follow for Ruby based checks if they don't use /healthcheck.

cbaines:
> Are there decisions (or better documentation) on what we aspire to have on /healthcheck and for things that aren't covered there how. I gather for Email Alert API it's moved checks out of using Ruby and instead to monitor graphite metrics, which makes me unsure of precedent to follow for Ruby based checks if they don't use /healthcheck.

I don't think anyone has made any decisions or documentation. 

I guess here, I don't see much significance in where the check is happening and what language it's implemented in. For example, you could have Content Publisher run a Sidekiq job each minute that computes the age of the government repository data, and pushes that value to Graphite (through statsd). You could then have a Icinga check that looks at that value, and checks if it's over 6 hours. Even though it's Icinga doing the checking, I think it's still possible to do what I'm guessing you want to. Also, you get the added benefit of being able to look at the trend in the data through Graphite.

kevindew:
> I don't think anyone has made any decisions or documentation.

I see, hmm maybe we need PH to do some thinking about purpose of this endpoint. It looks a bit like Email Alert API has gone in an opinionated direction of its own that leaves a few questions/discussions. Maybe an RFC might be the way to go to get comments and documentation at the same time.


> I guess here, I don't see much significance in where the check is happening and what language it's implemented in. For example, you could have Content Publisher run a Sidekiq job each minute that computes the age of the government repository data, and pushes that value to Graphite (through statsd). You could then have a Icinga check that looks at that value, and checks if it's over 6 hours. Even though it's Icinga doing the checking, I think it's still possible to do what I'm guessing you want to. Also, you get the added benefit of being able to look at the trend in the data through Graphite.

Well I think the things that lost there is the ease of the check. We take something that's a quick Ruby method and replace it with logic split across two repos (puppet and here) and then need an additional asynchronous job process. That feels a bit cumbersome in comparison :-/ 

If we have two purposes for the endpoint that we're trying to break up maybe we need two endpoints? Or if we want to be regularly updating graphite for lots of apps (I assume many are in similar boat) maybe we need something in govuk_app_config that provides common functionality.

cbaines:
> > I guess here, I don't see much significance in where the check is happening and what language it's implemented in. For example, you could have Content Publisher run a Sidekiq job each minute that computes the age of the government repository data, and pushes that value to Graphite (through statsd). You could then have a Icinga check that looks at that value, and checks if it's over 6 hours. Even though it's Icinga doing the checking, I think it's still possible to do what I'm guessing you want to. Also, you get the added benefit of being able to look at the trend in the data through Graphite.
> 
> Well I think the things that lost there is the ease of the check. We take something that's a quick Ruby method and replace it with logic split across two repos (puppet and here) and then need an additional asynchronous job process. That feels a bit cumbersome in comparison :-/
> 
> If we have two purposes for the endpoint that we're trying to break up maybe we need two endpoints? Or if we want to be regularly updating graphite for lots of apps (I assume many are in similar boat) maybe we need something in govuk_app_config that provides common functionality.

I can see us ending up with some kind of `/metrics` endpoint at some point (Prometheus...), but at least for now with statsd and Graphite, just fitting the code in where relevant seems OK to me.

Anyway, I don't see anything wrong with what's being done here, the only thing to avoid is the healthcheck timing out unnecessarily, and it doesn't look like there's a possibility of that here. I just wanted to make you aware of what had been happening elsewhere on GOV.UK.

kevindew:
> I can see us ending up with some kind of `/metrics` endpoint at some point (Prometheus...), but at least for now with statsd and Graphite, just fitting the code in where relevant seems OK to me.

Okey doke 👍 
 
> Anyway, I don't see anything wrong with what's being done here, the only thing to avoid is the healthcheck timing out unnecessarily, and it doesn't look like there's a possibility of that here. I just wanted to make you aware of what had been happening elsewhere on GOV.UK.

Ah sure that's cool, I think I misinterpreted a bit that we were on a new direction. Apologies if I came across like that. Thanks for the comms on it, it's really helpful to having your teams insight too.

lib/bulk_data/cache.rb
```diff
@@ -22,6 +22,10 @@ def self.read(key)
       cache.read(key) || (raise NoEntryError)
     end
 
+    def self.populated?(key)
+      cache.exist?(key)
+    end
```
kevindew: Rather than explicitly adding a method I think it'd be preferable to delegate `exist?` to the cache instance: https://github.com/alphagov/content-publisher/blob/c14461000856ef638303eb63b1e4315d638e13ec/lib/bulk_data/cache.rb#L11

I think we also can kill of the clear method as that is already delegated.

lib/bulk_data/government_repository.rb
```diff
@@ -46,5 +46,13 @@ def populate_cache(older_than: nil)
     rescue GdsApi::BaseError
       raise RemoteDataUnavailableError
     end
+
+    def age_of_cache
+      (Time.zone.now - Cache.written_at(CACHE_KEY)).seconds
```
kevindew: Time.current is preferred to using Time.zone.now, I wonder if I can make that a linter rule as it comes up a bit 🤔 

I'm not sure we need the `.seconds` as I believe this will execute the same with it or without it.

lib/healthcheck/government_data_check.rb
```diff
@@ -0,0 +1,44 @@
...
+
+module Healthcheck
+  class GovernmentDataCheck
+    def name
+      :government_data_check
+    end
+
+    def status
+      return GovukHealthcheck::CRITICAL unless government_data_availible?
+
```
kevindew: It feels a bit unusual to have new lines between the guard clauses.

lib/healthcheck/government_data_check.rb
```diff
@@ -0,0 +1,44 @@
...
+    end
+
+    def enabled?
+      true
+    end
+
+  private
+
+    def government_data_availible?
+      BulkData::GovernmentRepository.new.cache_populated?
```
kevindew: It'd make sense to have the repository as an instance variable rather than creating a new instance each time. 

This would also make the code shorter so you could probably kill this method and instead have the guard clause be: `return GovukHealthcheck::CRITICAL unless repository.cache_populated?`

lib/healthcheck/government_data_check.rb
```diff
@@ -0,0 +1,44 @@
...
+    end
+
+    def data_stale?
+      repository = BulkData::GovernmentRepository.new
+      repository.age_of_cache > 6.hours
+    end
+
+    def warning_details_content
+      data_age = BulkData::GovernmentRepository.new.age_of_cache
+      "Government data not refreshed in #{data_age / 3600} hours #{data_age / 60} minutes"
```
kevindew: This doesn't look correct - and also seems to be missing a test - as if you had 7200 as the age you'd get an output of "Government data not refreshed in 2 hours 120 minutes". You'll probably want to pull out the modulus operator.

Also using ActiveSupport::Duration and division for conversion makes some pretty strange intermediary numbers for example:
```
irb(main):034:0> 7200.seconds / 3600
=> 2 seconds
```

So might be actually be a not very friendly class to use and better to use an integer.

spec/lib/healthcheck/government_data_check_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+    context "government data is unavailable" do
+      it "returns critical" do
+        gov_data_check = Healthcheck::GovernmentDataCheck.new
+        expect(gov_data_check.status).to be(:critical)
+      end
+    end
+
+    context "government data is set to expire" do
+      it "gives a warning" do
+        BulkData::Cache.write("government-v1", "value")
```
kevindew: The use of this id makes an unnecessary coupling between this test and the constant in GovernmentRepository. I'd suggest instead using the helper instead: https://github.com/alphagov/content-publisher/blob/c074cd3ebb6c885f1fa3d7112cd7af6bbb4d01da/spec/support/bulk_data_helper.rb#L4

lib/bulk_data/government_repository.rb
JonathanHallam: Absolutely right on the `.seconds`. I'm happy to do the linter rule. Do you think that needs doing just in content publisher or in https://github.com/alphagov/rubocop-govuk?

spec/lib/healthcheck/government_data_check_spec.rb
JonathanHallam: Nice one, forgot that you could do this, have adjusted.

lib/healthcheck/government_data_check.rb
JonathanHallam: Rookie error on this one, have adjusted the output and written a test for it. I've left the `cache_age` method returning an `ActiveSupport::Duration` and turned that into an integer on the `warning_details_content` method. It means the return clauses above look a bit nicer as I can check for `@government_repo.cache_age > 6.hours` rather than `@government_repo.cache_age > 21600`

lib/bulk_data/government_repository.rb
```diff
@@ -46,5 +46,13 @@ def populate_cache(older_than: nil)
     rescue GdsApi::BaseError
       raise RemoteDataUnavailableError
     end
+
+    def age_of_cache
```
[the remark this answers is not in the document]
JonathanHallam: done :)

lib/bulk_data/cache.rb
JonathanHallam: clear method is now gone :)

lib/healthcheck/government_data_check.rb
JonathanHallam: yep, sorry, had to leave one of them in because the linter complained though

lib/healthcheck/government_data_check.rb
JonathanHallam: Done, that really cleared it up, thanks

db/schema.rb
```diff
@@ -371,10 +371,10 @@
   create_table "whitehall_migrations", force: :cascade do |t|
     t.text "organisation_content_id"
     t.text "document_type"
-    t.text "document_subtypes", array: true
     t.datetime "finished_at"
     t.datetime "created_at", precision: 6, null: false
     t.datetime "updated_at", precision: 6, null: false
+    t.text "document_subtypes", array: true
```
kevindew: Looks like this has slipped in on an unrelated commit?

lib/healthcheck/government_data_check.rb
```diff
@@ -0,0 +1,38 @@
...
+    def initialize
+      @government_repo = BulkData::GovernmentRepository.new
+    end
+
+    def name
+      :government_data_check
+    end
+
+    def status
+      return GovukHealthcheck::CRITICAL unless @government_repo.cache_populated?
```
kevindew: Convention would be to use an attribute reader for government_repo

lib/healthcheck/government_data_check.rb
```diff
@@ -0,0 +1,38 @@
...
+
+    def enabled?
+      true
+    end
+
+  private
+
+    def warning_details_content
+      data_age = @government_repo.cache_age.to_i
+      "Government data not refreshed in #{data_age / 3600} hours #{data_age % 60} minutes"
```
kevindew: The arithmetic is still wrong here.

For example if you have 4000 seconds it is 1 hours 6 minutes. Whereas this will say "Government data not refreshed in 1 hours 40 minutes" - which also reveals plural issues.

spec/lib/healthcheck/government_data_check_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+    end
+  end
+
+  describe "#details" do
+    let(:gov_data_check) { described_class.new }
+
+    it "displays appropriate messages" do
+      government_repo = BulkData::GovernmentRepository.new
+      gov_data_check.instance_variable_set(:@government_repo, government_repo)
+      allow(government_repo).to receive(:cache_age).and_return(1.second)
```
kevindew: I think this now returns an integer

spec/lib/healthcheck/government_data_check_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+      expect(gov_data_check.details[:warning]).to eq("Government data not refreshed in 0 hours 1 minutes")
```
kevindew: Should 1 second really output 1 minute?

lib/tasks/populate_government_data.rake
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+
+namespace :app do
+  desc "Populate government data"
+  task populate_government_data: :environment do
+    PopulateBulkDataJob.perform_now
+  end
+end
```
kevindew: I can see a few issues here (namespace app, task focused on government data but the job named differently) though mostly I'm wondering why we need the task since it's one line, may as well suggest someone runs the command in the console.

spec/requests/healthcheck_spec.rb
```diff
@@ -0,0 +1,25 @@
...
+    context "when the healthchecks pass" do
+      before do
+        populate_default_government_bulk_data
+      end
+
+      it "returns a status of 'ok'" do
+        get healthcheck_path
+        expect(JSON.parse(response.body)["status"]).to eq("ok")
+      end
+    end
```
kevindew: I don't think this request needs to be concerned about what the statuses are - since there is no logic in the controller about what is returned just which tests are run. 

I think you'd be fine to test that the healthcheck does two things:

1) returns a 200 HTTP status
2) returns JSON with a status attribute (doesn't matter what the value is)

db/schema.rb
JonathanHallam: Honestly have no idea how this got in, I've not had to look at this :/ at any rate I've removed it.

lib/healthcheck/government_data_check.rb
JonathanHallam: I've done this, can I ask why this is convention though?

lib/healthcheck/government_data_check.rb
JonathanHallam: Changed the working of this after a face to face conversation. Decided that it was accurate enough to just have hours given that this won't show up unless the data has been missing for more than 6.

spec/lib/healthcheck/government_data_check_spec.rb
JonathanHallam: Sorted

spec/lib/healthcheck/government_data_check_spec.rb
JonathanHallam: Have now dropped this entirely after face to face conversation.

lib/tasks/populate_government_data.rake
JonathanHallam: Have removed

spec/requests/healthcheck_spec.rb
JonathanHallam: Nice, I've changed the tests and removed the context block now that it doesn't make sense

lib/healthcheck/government_data_check.rb
kevindew: Mostly just what GDS went with when it started, which will then be from it being a very common Ruby convention.

There's a bunch of arguments why attr_reader is the preferred approach online: https://github.com/rubocop-hq/ruby-style-guide/issues/538 (which proposed the other approach), https://ivoanjo.me/blog/2017/09/20/why-i-always-use-attr_reader-to-access-instance-variables/ but I don't think any are earth shattering reasons. For me I'm just all for consistency, which trumps my own minor preference for attr_reader.

lib/healthcheck/government_data_check.rb
```diff
@@ -0,0 +1,39 @@
...
+      @government_repo = BulkData::GovernmentRepository.new
+    end
+
+    def name
+      :government_data_check
+    end
+
+    def status
+      return GovukHealthcheck::CRITICAL unless government_repo.cache_populated?
+      return GovukHealthcheck::WARNING if government_repo.cache_age > 21600
```
kevindew: This can still be 6.hours:

```
irb(main):001:0> 3700 > 1.hour
=> true
irb(main):002:0> 3500 > 1.hour
=> false
```

lib/healthcheck/government_data_check.rb
```diff
@@ -0,0 +1,39 @@
...
+
+    def enabled?
+      true
+    end
+
+  private
+
+    def warning_details_content
+      data_age = government_repo.cache_age.to_i
+      "Government data not refreshed in #{data_age / 3600} hours."
```
kevindew: If we're changing integers we could change the 3600 to 1.hour not sure if that's much of a reading win or not?

spec/lib/healthcheck/government_data_check_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+      end
+    end
+  end
+
+  describe "#details" do
+    let(:gov_data_check) { described_class.new }
+
+    it "displays appropriate messages" do
+      government_repo = BulkData::GovernmentRepository.new
+      gov_data_check.instance_variable_set(:@government_repo, government_repo)
```
kevindew: This seems a quite unusual mocking approach. A more common one would be expect to receive new and return instance double. For example: https://github.com/alphagov/content-publisher/blob/e06613a8a3de96245ec7055e362a6dae6da5775f/spec/services/resync_document_service_spec.rb#L115-L118

However I think you do better continuing the approach you did earlier in this test and using the `travel_to` method

lib/healthcheck/government_data_check.rb
```diff
@@ -0,0 +1,39 @@
...
+      return GovukHealthcheck::WARNING if government_repo.cache_age > 21600
+
+      GovukHealthcheck::OK
+    end
+
+    def details
+      {
+        critical: "No government data availible",
+        warning: warning_details_content,
+      }
```
kevindew: Sorry I missed this before. Is it normal that these methods will return the critical and warning messages even if the healthcheck is not in a critical or warning state? It feels like that is information that'd show in icinga.

Having a look around I'm not too sure what the purpose and structure of details needs to be. It seems common to have a value, total and thresholds (https://github.com/alphagov/govuk_app_config/blob/master/lib/govuk_app_config/govuk_healthcheck/threshold_check.rb#L23-L32) though I don't know what icinga parses or not.

spec/requests/healthcheck_spec.rb
```diff
@@ -0,0 +1,19 @@
...
+
+RSpec.describe "Healthcheck" do
+  describe "GET /healthcheck" do
+    before do
+      populate_default_government_bulk_data
+    end
+
+    it "returns a 200 HTTP status" do
+      get healthcheck_path
+      expect(response.status).to eq(200)
```
kevindew: minor, we've done this elsewhere with `expect(response).to have_http_status(:ok)`

lib/healthcheck/government_data_check.rb
JonathanHallam: I could have sworn that I'd seen it done like this in email alert api but I can't find it now. It does seem to be how [DGU does it](https://github.com/alphagov/datagovuk_publish/blob/a8468e62cb94ff755215b9cf23da345c5f4c31bc/app/models/healthcheck/recurring_jobs/ckan_org_sync.rb#L25) but you're right at this doesn't look entirely normal. I'll have a crack at it.

kevindew: Sure, they just ripped loads out of Email Alert API so maybe can find them in the past, though it'd be best to have a current one so we can look at it in icinga.

Very confusing 🤔

JonathanHallam: ah found it https://github.com/alphagov/email-alert-api/commit/0c17e757179cb4b0308bf0b49bb3b19842c15bc2#diff-9fccbed8f77bf59ae535edc077ce5e21

JonathanHallam: doesn't mean that it worked but I'm glad I wasn't losing my marbles. I might test it in integration this afternoon, as it stands visiting `/healthcheck` will show both results but I don't know what it'll do in icinga so I'll be able to see what actually works.

kevindew: Ah I see, hmm that's a bit weird too as the details for that would just be numbers if I'm understanding that correctly rather than strings 🤔

lib/healthcheck/government_data_check.rb
JonathanHallam: Nice, thanks, have updated

lib/healthcheck/government_data_check.rb
JonathanHallam: Yeah i think it's worth it. Have adjusted, didn't realise you could be quite so fluid with durations and numbers.

spec/lib/healthcheck/government_data_check_spec.rb
JonathanHallam: that seems like a while ago! I've updated the tests to use `travel_to` do you think it's worth having two tests here or would one do, I'm not sure what I'm gaining by testing for 6 and 15 hours...

lib/healthcheck/government_data_check.rb
JonathanHallam: It's all weird! I'm trying to test this in integration, I've managed to make it give a warning but not actually raise anything in icinga. This is because of an outstanding puppet change so I'll be looking to deploy puppet to integration in order to test this. I just want to make sure that the messages show up how I expect in Icinga. I'm assuming that we don't really care what it looks like when a person visits `/healthcheck`?

spec/requests/healthcheck_spec.rb
JonathanHallam: Nice, need to get better with native rspec matchers. Have adjusted.

lib/healthcheck/government_data_check.rb
```diff
@@ -0,0 +1,37 @@
...
+    def status
+      return GovukHealthcheck::CRITICAL unless government_repo.cache_populated?
+      return GovukHealthcheck::WARNING if government_repo.cache_age > 6.hours
+
+      GovukHealthcheck::OK
+    end
+
+    def message
+      "No government data availible" if status == GovukHealthcheck::CRITICAL
+      warning_details_content if status == GovukHealthcheck:: WARNING
```
kevindew: This isn't going to work correctly when the status is critical as it isn't the last statement is the second conditional so this will return nil.

You'll need a return statement on the first line or to refactor this a bit. Probably best unit testing it though so this is covered.

lib/healthcheck/government_data_check.rb
```diff
@@ -0,0 +1,37 @@
+# frozen_string_literal: true
```
kevindew: You don't need these anymore since https://github.com/alphagov/content-publisher/pull/1769

spec/lib/healthcheck/government_data_check_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+    context "when everything is fine" do
+      it "returns ok" do
+        populate_default_government_bulk_data
+        gov_data_check = described_class.new
+        expect(gov_data_check.status).to be(:ok)
+      end
+    end
+  end
+
+  describe "#details" do
```
kevindew: Looks like this needs renaming for message

spec/lib/healthcheck/government_data_check_spec.rb
```diff
@@ -0,0 +1,42 @@
+# frozen_string_literal: true
+
+RSpec.describe Healthcheck::GovernmentDataCheck do
+  describe "#status" do
+    context "when government data is unavailable" do
+      it "returns critical" do
+        gov_data_check = described_class.new
```
kevindew: I'd wonder if doing this a bunch is worth the 5 characters we save?

lib/healthcheck/government_data_check.rb
JonathanHallam: This was resolved by checking in integration. Turns out that the details method doesn't seem to be shown anywhere. I've replaced it with the "message" method that will show up in Icinga. I've also made changes to the doc to make it clearer what's happening here https://github.com/alphagov/govuk_app_config/pull/135

lib/healthcheck/government_data_check.rb
JonathanHallam: Absolutely right on this, I didn't realise that's how this worked :/ I've added in unit tests to make sure this gets caught.

lib/healthcheck/government_data_check.rb
JonathanHallam: Didn't realise that'd happened, nice! Thanks 👍

spec/lib/healthcheck/government_data_check_spec.rb
JonathanHallam: Should have spotted this, sorry. Fixed.

spec/lib/healthcheck/government_data_check_spec.rb
JonathanHallam: I've changed this file about a bit and moved this to a `let` block at the top. It's cleaned the file up some especially given the new tests I introduced.

lib/healthcheck/government_data_check.rb
```diff
@@ -0,0 +1,35 @@
...
+    def status
+      return GovukHealthcheck::CRITICAL unless government_repo.cache_populated?
+      return GovukHealthcheck::WARNING if government_repo.cache_age > 6.hours
+
+      GovukHealthcheck::OK
+    end
+
+    def message
+      return "No government data availible" if status == GovukHealthcheck::CRITICAL
+      return warning_details_content if status == GovukHealthcheck:: WARNING
```
kevindew: This last return is redundant

spec/requests/healthcheck_spec.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
```
kevindew: This frozen string literal was missed.

app/controllers/healthcheck_controller.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
```
kevindew: Another FSL here too

lib/healthcheck/government_data_check.rb
JonathanHallam: Yeah, i put it in there to keep the linter happy without a new line but it does look a bit nicer with it, have adjusted

spec/requests/healthcheck_spec.rb
JonathanHallam: yeah sorry, noticed this immediately after I sent it over, fixed now

PR #1749, opened 2020-02-07
Add instructions to navigate topics

alex-ju:
Add instructions on how to navigate topics for screen reader users.

co-author: @pollygreen 

Raising this as a draft. It should be merged after the miller-columns-element gets bumped to [1.3.0](https://github.com/alphagov/miller-columns-element/pull/33).

Update: This can now be reviewed and merged as [we're now using the latest version of miller-columns element](https://github.com/alphagov/content-publisher/pull/1750)

### DIFF
--- a/app/views/topics/edit.html.erb
+++ b/app/views/topics/edit.html.erb
@@ -81,12 +81,15 @@
     } %>
 
     <div class="govuk-form-group">
-      <miller-columns id="miller-columns" class="miller-columns" for="taxonomy" selected="selected-items">
-      <ul id="taxonomy" class="govuk-list">
-        <% Topic.govuk_homepage(@edition.document_topics.index).children.each do |topic| %>
-          <% unroll(topic) %>
-        <% end %>
-      </ul>
+      <p id="topics-navigation-instructions" class="govuk-body govuk-visually-hidden">
+        <%= t("topics.edit.navigation_instructions") %>
+      </p>
+      <miller-columns id="miller-columns" class="miller-columns" for="taxonomy" selected="selected-items" aria-describedby="topics-navigation-instructions">
+        <ul id="taxonomy" class="govuk-list">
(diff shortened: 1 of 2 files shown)


app/views/topics/edit.html.erb
```diff
@@ -81,12 +81,15 @@
     } %>
 
     <div class="govuk-form-group">
-      <miller-columns id="miller-columns" class="miller-columns" for="taxonomy" selected="selected-items">
-      <ul id="taxonomy" class="govuk-list">
-        <% Topic.govuk_homepage(@edition.document_topics.index).children.each do |topic| %>
-          <% unroll(topic) %>
-        <% end %>
-      </ul>
+      <p id="topics-navigation-instructions" class="govuk-body govuk-visually-hidden">
```
ChrisBAshton: Have you considered adding `aria-hidden` to this? Users who have CSS disabled will see this text, and we may not want them to. I've had a quick search and can't really find any guidance either way, though [the spec does use aria-hidden in one of its examples](https://www.w3.org/TR/WCAG20-TECHS/ARIA1.html#ARIA1-ex4).

alex-ju: I wouldn't be bothered by the fact that screen reader users will be able to hear the instructions in a reading mode, right before landing on the element (as per its current DOM positioned), but we can indeed add `aria-hidden` indeed if we want them not to reach it (will need to re-test across assistive tech to make sure we're not breaking their experience).

It will be however visible to sighted users that have CSS disabled, indeed – but again, the instructions might actually be of help and they shouldn't do any harm. I guess this would be the tradeoff of having the content decoupled from the component.

ChrisBAshton: Oh silly me - yes adding `aria-hidden="true` would still be visible to users who have disabled CSS 🤦‍♂ don't think I'd had my coffee yet. But my point was that screen reader users may encounter the text out of context - and as you've explained, the context is fine. Please ignore my suggestion!

PR #1753, opened 2020-02-10
Use params directly in requirements checkers

benthorner:

This resolves a pain point from the work to implement
flexible fields, where we redundantly pass an edition
and a revision as part of requirements checks.

Resolving this was more involved than I expected, since
it meant changing the approach we have for checking
requirements to avoid relying on an in-memory revision.

Please see the commits for more details. There's still
some remaining work to do in order to make our approach
consistent for file attachments and images.

### DIFF
--- a/app/interactors/content/update_interactor.rb
+++ b/app/interactors/content/update_interactor.rb
@@ -38,7 +38,7 @@ private
     issues = Requirements::CheckerIssues.new
 
     edition.document_type.contents.each do |field|
-      issues += field.pre_update_issues(edition, revision)
+      issues += field.pre_update_issues(edition, content_params)
     end
 
     context.fail!(issues: issues) if issues.any?
@@ -64,7 +64,7 @@ private
   end
 
   def content_params
-    edition.document_type.contents.reduce({}) do |hash, field|
+    @content_params ||= edition.document_type.contents.reduce({}) do |hash, field|
       hash.merge!(field.updater_params(edition, params))
     end
   end
--- a/app/interactors/content/update_interactor.rb
+++ b/app/interactors/content/update_interactor.rb
@@ -10,8 +10,8 @@ class Content::UpdateInteractor < ApplicationInteractor
   def call
     Edition.transaction do
       find_and_lock_edition
-      update_revision
       check_for_issues
+      update_revision
 
       update_edition
       create_timeline_entry
@@ -37,7 +37,7 @@ private
   def check_for_issues
     issues = Requirements::CheckerIssues.new
 
-    edition.document_type.contents.each do |field|
+    fields.each do |field|
       issues += field.pre_update_issues(edition, content_params)
     end
 
@@ -64,8 +64,12 @@ private
   end
 
   def content_params
-    @content_params ||= edition.document_type.contents.reduce({}) do |hash, field|
+    @content_params ||= fields.reduce({}) do |hash, field|
       hash.merge!(field.updater_params(edition, params))
     end
+  end
+
+  def fields
+    @fields ||= edition.document_type.contents
   end
 end
--- a/app/interactors/tags/update_interactor.rb
+++ b/app/interactors/tags/update_interactor.rb
@@ -5,15 +5,14 @@ class Tags::UpdateInteractor < ApplicationInteractor
            :user,
            :edition,
            :revision,
-           :revision_updater,
            :issues,
            to: :context
 
   def call
     Edition.transaction do
       find_and_lock_edition
-      update_revision
       check_for_issues
+      update_revision
 
       update_edition
       create_timeline_entry
@@ -29,9 +28,10 @@ private
   end
 
   def update_revision
-    context.revision_updater = Versioning::RevisionUpdater.new(edition.revision, user)
-    revision_updater.assign(tags: update_params)
-    context.revision = revision_updater.next_revision
+    updater = Versioning::RevisionUpdater.new(edition.revision, user)
+    updater.assign(tags: update_params)
+    context.fail! unless updater.changed?
+    context.revision = updater.next_revision
   end
 
   def check_for_issues
@@ -40,8 +40,6 @@ private
   end
 
   def update_edition
-    context.fail! unless revision_updater.changed?
-
     EditDraftEditionService.call(edition, user, revision: revision)
     edition.save!
   end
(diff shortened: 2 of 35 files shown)


app/interactors/tags/update_interactor.rb
```diff
@@ -29,9 +28,10 @@ def find_and_lock_edition
   end
 
   def update_revision
-    context.revision_updater = Versioning::RevisionUpdater.new(edition.revision, user)
-    revision_updater.assign(tags: update_params)
-    context.revision = revision_updater.next_revision
+    updater = Versioning::RevisionUpdater.new(edition.revision, user)
+    updater.assign(tags: update_params)
+    context.fail! unless updater.changed?
```
ChrisBAshton: This is basically equivalent to `throw "Foo" if bar`; in these cases our linter usually enforces a newline before the next instruction in the block.

Should we do the same for `context.fail!` (even though the linter doesn't recognise it as an exception)? I notice it had a newline prior to this change.
Better yet, is there a way we can get the linter to recognise that these instructions are exception-raisers?

benthorner: You've spotted my compulsion to remove redundant newlines 😄. While I do agree with using linting to keep us consistent, I'd like to suggest we pass on it this time, and we can revisit if it becomes a recurring comment. Is that OK?

ChrisBAshton: Sure 👍

app/interactors/content/update_interactor.rb
```diff
@@ -10,8 +10,8 @@ class Content::UpdateInteractor < ApplicationInteractor
   def call
     Edition.transaction do
       find_and_lock_edition
-      update_revision
       check_for_issues
+      update_revision
 
       update_edition
```
kevindew: Probably for another day but I wonder if we want to combine the update edition and revision steps in interactors where they follow each other.

app/models/document_type/body_field.rb
```diff
@@ -17,21 +17,23 @@ def updater_params(_edition, params)
...
+    unless GovspeakDocument.new(params[:contents][:body], edition).valid?
       issues.create(id, :invalid_govspeak)
     end
 
     issues
   end
 
-  alias_method :pre_update_issues, :pre_preview_issues
+  def pre_preview_issues(edition)
+    pre_update_issues(edition, contents: edition.contents.symbolize_keys)
```
kevindew: The introduction of contents here feels a bit weird, not sure if it's a smell - the symbolize keys makes me feel more like it's a smell as it feels like we're contorting attributes.

I did wonder if we might instead want the params for body to be just `params[:body]` and that this might feel a little less weird if it was `pre_update_issues(edition, body: edition.contents['body"])` - I figure that might either slot in quite nicely or create a brutal challenge of different places expecting different things.

Another thing I noticed in this class is that is that we seem to mix whether we use the id value or a body symbol. I'd imagine that for nearly all cases we could use the id as whenever we're using an ActionController::Parameters we have a hash with indifferent access.

app/views/content/edit/_change_notes.html.erb
```diff
@@ -3,9 +3,9 @@
       label: {
         text: t("content.edit.change_note.title"),
       },
-      id: "revision-change-note",
+      id: "edition-change-note",
```
kevindew: I guess this was renamed due to us using edition instead? I believe it got it it's original id from it having a name of `revision[change_note]`. I think it might be more consistent with other fields if had an id of `change-note-field` since we don't reference edition in the name (and I assume we're reserving the actual "change-note" id for linking from index).

app/views/content/edit/_summary_field.html.erb
```diff
@@ -3,10 +3,10 @@
     id: "summary",
     html_for: "summary-field"
   }
-  if t_doctype_field?(@edition, "summary.guidance")
+  if t_doctype_field?(edition, "summary.guidance")
```
kevindew: Aw locals are always nicer to look at 🙂

spec/lib/requirements/backdate_checker_spec.rb
```diff
@@ -4,19 +4,19 @@
   describe "#pre_submit_issues" do
     it "returns no issues if there are none" do
       date = Time.current.change(day: 1, month: 1, year: 2019)
-      issues = Requirements::BackdateChecker.new(date).pre_submit_issues
+      issues = subject.pre_update_issues(date)
```
kevindew: Oh - when did subjects start appear in the tests? We have some precedent for `described_class` but until now I wasn't aware this was being used in the tests - seems it was introduced with fields and is now spreading?

Why did this get introduced now without any prior discussion? I remember it was an area of contention in the early days which I am imagine you are also aware of.

app/interactors/content/update_interactor.rb
benthorner: That occurred to me too, and it would not be possible to do so. Some of the `update_revision` or `update_edition` methods can be quite long, though. Having them separate has the small advantage for me of explicitly articulating the steps we take for the revision, and the steps we take for the edition.

app/models/document_type/body_field.rb
benthorner: I agree, it does seem a little contrived. I can add a little explanation about the form of the params: for content checkers, the params that are passed in are the ones generated by the `updater_params` methods, since for `title_and_base_path` we need the `base_path` and don't want to be generating it twice. This seems reasonable, since the name of the method suggests we're going to be checking the things we're trying to use for the update. Doing it this way means the params have to be of the form `contents: { body: ... }`.

The pain point I had when implementing this is that `edition.contents` is not a has with indifferent access (or frankly, just a has with symbols). Switching everything to use string keys to accommodate this seems like a step backwards, so I'd prefer to try and address the stringiness of the JSON columns, if possible. Unfortunately, there doesn't seem to be an easy way of doing this without overriding methods :-(.

app/views/content/edit/_change_notes.html.erb
benthorner: Sure, I can do another PR to change this.

spec/lib/requirements/backdate_checker_spec.rb
benthorner: Aha, so I wasn't aware it was contentious. I do remember a discussion around this and having all tests do their own setup, execution and checks in a self-contained way. Since we now have a fair few `let`s, it didn't seem like an issue anymore.

Looking at the specs, it does seem to be something I've introduced recently. We can look at reverting it (and doing more typing 😢) if you can help me understand why we should avoid it. [This article](https://thoughtbot.com/blog/lets-not) looks familiar, but is a little extreme. We seem to be in a state of some inconsistency here, and perhaps we should try and write down want we do/n't want, or think about some linting.

PR #1755, opened 2020-02-10
Fix discarded draft attribution in timeline entries

kevindew:
Trello: https://trello.com/c/yqXt0RGc/1417-remove-documents-at-dwps-request

It turns out that we have a bug in the code used to assign the status of a timeline entry. The `DeleteDraftEditionService` accepts a document as input and modifies the current_edition of that. As the edition that is used in the interactor is not the same object as `document.current_edition` the edition in the interactor becomes stale and using that to populate the timeline entry uses an out of date status.

Looking on production we have 175 discarded timeline entries and only 32 of those are attributed to the correct status:

```
irb(main):010:0> TimelineEntry.draft_discarded.map { |te| te.status.state }.each_with_object(Hash.new(0)) { |state, memo| memo[state] += 1 }
=> {"discarded"=>32, "draft"=>129, "submitted_for_review"=>14}
```
The problems seem to kick in between March 20th to 25th this year - which makes [this change (by me)](https://github.com/alphagov/content-publisher/commit/62871d94412d6c4cbb9d0dc28411128bb08d0a5e) the likely culprit. Once this is merged I plan to fix those in the DB manually.

To resolve this I've changed the `DeleteDraftEditionService` to operate on an edition, which felt more natural anyway given the naming the class.

### DIFF
--- a/spec/services/delete_draft_edition_service_spec.rb
+++ b/spec/services/delete_draft_edition_service_spec.rb
@@ -4,25 +4,25 @@ RSpec.describe DeleteDraftEditionService do
   let(:user) { create :user }
 
   describe ".call" do
-    it "raises an exception if there is not a current edition" do
-      document = create :document
+    it "raises an exception if the edition is not current" do
+      edition = create :edition, current: false
 
-      expect { DeleteDraftEditionService.call(document, user) }
-        .to raise_error "Trying to delete a document without a current edition"
+      expect { DeleteDraftEditionService.call(edition, user) }
+        .to raise_error "Only current editions can be deleted"
     end
 
     it "raises an exception if the current edition is live" do
       document = create :document, :with_live_edition
 
-      expect { DeleteDraftEditionService.call(document, user) }
-        .to raise_error "Trying to delete a live document"
+      expect { DeleteDraftEditionService.call(document.current_edition, user) }
+        .to raise_error "Trying to delete a live edition"
     end
 
     it "attempts to delete the document preview" do
       document = create :document, :with_current_edition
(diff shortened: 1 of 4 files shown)


spec/services/delete_draft_edition_service_spec.rb
```diff
@@ -1,235 +1,210 @@
...
+        expect { DeleteDraftEditionService.call(edition, user) }
+          .to raise_error(GdsApi::BaseError)
+
+        expect(edition.revision_synced?).to be(true)
+      end
     end
 
-    it "raises an error when a base path cannot be deleted" do
-      edition = create :edition
+    describe "Asset Manager communication" do
```
benthorner: Could we extract this part into a service?

spec/services/delete_draft_edition_service_spec.rb
```diff
@@ -1,235 +1,210 @@
...
-      stub_publishing_api_unreserve_path(document.current_edition.base_path)
-      discard_draft_error = {
-        error: {
-          code: 422,
-          message: "New Publishing API problem",
-        },
-      }
-      stub_any_publishing_api_discard_draft
-        .to_return(status: 422, body: discard_draft_error.to_json)
+      it "does not delete path reservations for published documents" do
```
benthorner: Could we make a `DeletePathReservationService` to extract this?

benthorner: (Especially noting the detailed edge case of having had multiple base paths.)

kevindew:
Thanks @benthorner I'll add something to the card AC about deciding what to tell about this (whether just fix it or add to wall of pain)

PR #1756, opened 2020-02-10
Fix import of embedded files in news articles

gclssvglx:
When migrating two of NDA's news articles from Whitehall to Content Publisher, we encounter integrity check failures on the body.  The cause of this is the presence of image markdown in the proposed payload which we have failed to convert to an actual image, i.e. we're sending '[Image:wood_640_x_960.jpg]' rather than '<img ....'.

The documents are:

- [ID: 431866 Content ID: 809d70b1-3e9b-4833-a28b-41366a925805](https://www.gov.uk/government/news/barrnon-and-wood-move-to-next-stage-of-innovation-competition)

- [ID: 370367 Content ID: 5a88536b-7831-4184-b5fd-c40275ba7558](https://www.gov.uk/government/news/awards-recognise-supply-chain-excellence)

It is expected that a similar situation will occur with attachments due to the similarity of their handling.
    
On inspection of the original document body of these documents and their revisions, it is noticed that in the first case a user had manually removed one new line (e.g. `\r\n!!2\r\n`)...

**431866 Revisions**

1. `\r\n\r\n!!2\r\n\r\n`

1. `\r\n!!2\r\n`

1. `\r\n!!2\r\n`

In the second case two images are present with only one new line between them (e.g. `\n\n!!2\n!!3`)...

**370367 Revisions**

1. `\r\n\r\n!!2\r\n!!3\r\n\r\n`, `\r\n\r\n!!4\r\n!!5\r\n\r\n`, `\r\n\r\n!!6\r\n!!7\r\n\r\n`, `\r\n\r\n!!8\r\n\r\n`, `\r\n\r\n!!9\r\n!!10\r\n\r\n`
             
1. `\r\n\r\n!!2\r\n!!3\r\n\r\n`, `\r\n\r\n!!4\r\n!!5\r\n\r\n`, `\r\n\r\n!!6\r\n!!7\r\n\r\n`, `\r\n\r\n!!8\r\n\r\n`, `\r\n\r\n!!9\r\n!!10\r\n\r\n`

Files are defined in GovSpeak as `!!2` for images and `!@2` for attachments.  Both require two new lines to be entered prior to the file (e.g. `\n\n!!2` or `\n\n!@2`) in order to be correctly converted into HTML correctly.
    
We need to fix the import so these files are correctly parsed, and the proposed payload for Publishing API matches the currently published document.

Trello: https://trello.com/c/x42Ye088/1406-fix-import-of-embedded-images-in-news-articles

### DIFF
--- a/lib/whitehall_importer/embed_body_references.rb
+++ b/lib/whitehall_importer/embed_body_references.rb
@@ -17,8 +17,14 @@ module WhitehallImporter
 
     def call
       body_with_embeds = embed_contacts(body, contacts)
-      body_with_embeds = embed_images(body_with_embeds, images)
-      body_with_embeds = embed_attachments(body_with_embeds, attachments)
+      body_with_embeds = embed_files(body_with_embeds,
+                                     files: images,
+                                     old_pattern: "!!",
+                                     new_pattern: "Image")
+      body_with_embeds = embed_files(body_with_embeds,
+                                     files: attachments,
+                                     old_pattern: "!@",
+                                     new_pattern: "Attachment")
       body_with_embeds = embed_inline_attachments(body_with_embeds, attachments)
       body_with_embeds
     end
@@ -33,19 +39,13 @@ module WhitehallImporter
       end
     end
 
-    def embed_images(body, images)
-      body&.gsub(/!!(\d+)/) do
-        whitehall_image_index = Regexp.last_match[1].to_i
-        image_name = images[whitehall_image_index - 1]
-        image_name.present? ? "[Image:#{image_name}]" : ""
-      end
-    end
-
-    def embed_attachments(body, attachments)
-      body&.gsub(/!@(\d+)/) do
-        whitehall_attachment_index = Regexp.last_match[1].to_i
-        attachment_name = attachments[whitehall_attachment_index - 1]
-        attachment_name.present? ? "[Attachment:#{attachment_name}]" : ""
+    def embed_files(body, files:, old_pattern:, new_pattern:)
+      body&.gsub(/(\A|\n\n|\r\n\r\n|\n|\r\n)#{old_pattern}(\d+)/) do
+        prefix = Regexp.last_match[1]
+        file_index = Regexp.last_match[2].to_i
+        file_name = files[file_index - 1]
+        prefix = prefix * 2 if ["\r\n", "\n"].include?(prefix)
+        "#{prefix}[#{new_pattern}:#{file_name}]" if file_name.present?
       end
     end
 
--- a/spec/lib/whitehall_importer/embed_body_references_spec.rb
+++ b/spec/lib/whitehall_importer/embed_body_references_spec.rb
@@ -12,31 +12,87 @@ RSpec.describe WhitehallImporter::EmbedBodyReferences do
       expect(govspeak_body).to eq("[Contact:#{content_id}]")
     end
 
-    it "converts Whitehall image syntax to Content Publisher syntax" do
-      govspeak_body = described_class.call(
-        body: "!!1 test !!2",
-        images: ["foo.png", "bar.jpg"],
-      )
+    it "converts Whitehall image bang embeds to govspeak Image embeds" do
+      body = described_class.call(body: "!!1", images: ["file.jpg"])
 
-      expect(govspeak_body).to eq("[Image:foo.png] test [Image:bar.jpg]")
+      expect(body).to eq("[Image:file.jpg]")
     end
 
-    it "removes any image markdown that doesn't resolve to an image" do
-      govspeak_body = described_class.call(
-        body: "Bar !!2 Baz",
-        images: ["foo.png"],
-      )
+    it "converts Whitehall attachment bang embeds to govspeak Attachment embeds" do
+      body = described_class.call(body: "!@1", attachments: ["file.pdf"])
 
-      expect(govspeak_body).to eq("Bar  Baz")
+      expect(body).to eq("[Attachment:file.pdf]")
     end
 
-    it "converts Whitehall attachment syntax to Content Publisher syntax" do
-      govspeak_body = described_class.call(
-        body: "!@1 test !@2",
-        attachments: ["file.pdf", "download.csv"],
-      )
+    it "ignores Whitehall image bang embeds that are neither start of the string or preceeded by new lines" do
+      body = described_class.call(body: "!!1 test !!2",
+                                  images: ["file.png", "file.jpg"])
 
-      expect(govspeak_body).to eq("[Attachment:file.pdf] test [Attachment:download.csv]")
+      expect(body).to eq("[Image:file.png] test !!2")
+    end
+
+    it "ignores Whitehall attachment bang embeds that are neither start of the string or preceeded by new lines" do
+      body = described_class.call(body: "!@1 test !@2",
+                                  attachments: ["file.pdf", "file.csv"])
+
+      expect(body).to eq("[Attachment:file.pdf] test !@2")
+    end
+
+    it "ignores Whitehall image bang embeds that do not resolve to an Image" do
+      body = described_class.call(body: "!!2 test", images: ["file.png"])
+
+      expect(body).to eq(" test")
+    end
+
+    it "ignores Whitehall attachment bang embeds that do not resolve to an Attachment" do
+      body = described_class.call(body: "!@2 test", attachments: ["file.pdf"])
+
+      expect(body).to eq(" test")
+    end
+
+    it "converts image and attachment embeds prefixed by single new lines to be prefixed by two" do
+      prefix_conversions = [
+        { prefix: "\n", conversion: "\n\n" },
+        { prefix: "\r\n", conversion: "\r\n\r\n" },
+      ]
+
+      prefix_conversions.each do |prefix_conversion|
+        prefixed = [
+          "#{prefix_conversion[:prefix]}!!1",
+          "#{prefix_conversion[:prefix]}!@1",
+        ]
+
+        converted = []
+        converted << "#{prefix_conversion[:conversion]}[Image:file.jpg]"
+        converted << "#{prefix_conversion[:conversion]}[Attachment:file.pdf]"
+
+        body = described_class.call(
+          body: prefixed.join,
+          images: ["file.jpg"],
+          attachments: ["file.pdf"],
+        )
+
+        expect(body).to eq(converted.join)
+      end
+    end
+
+    it "doesn't change the prefixes of images and attachments preceeded by two or more new lines" do
+      prefixes = [
+        "\n\n", "\r\n\r\n",
+        "\n\n\n", "\r\n\r\n\r\n",
+        "\n\n\n\n", "\r\n\r\n\r\n\r\n"
+      ]
+
+      prefixes.each do |prefix|
+        body = described_class.call(
+          body: "#{prefix}!!1#{prefix}!@1",
+          images: ["file.jpg"],
+          attachments: ["file.pdf"],
+        )
+
+        expect(body)
+          .to eq("#{prefix}[Image:file.jpg]#{prefix}[Attachment:file.pdf]")
+      end
     end
 
     it "converts Whitehall inline attachment syntax to Content Publisher syntax" do
@@ -45,16 +101,8 @@ RSpec.describe WhitehallImporter::EmbedBodyReferences do
         attachments: ["file.pdf", "download.csv"],
       )
 
-      expect(govspeak_body).to eq("[AttachmentLink:file.pdf] test [AttachmentLink:download.csv]")
-    end
-
-    it "removes any attachment markdown that doesn't resolve to an attachment" do
-      govspeak_body = described_class.call(
-        body: "Bar !@2 Baz [InlineAttachment:3] << removed, but !@1 continues to work",
-        attachments: ["foo.pdf"],
-      )
-
-      expect(govspeak_body).to eq("Bar  Baz  << removed, but [Attachment:foo.pdf] continues to work")
+      expect(govspeak_body)
+        .to eq("[AttachmentLink:file.pdf] test [AttachmentLink:download.csv]")
     end
   end
 end


lib/whitehall_importer/embed_body_references.rb
```diff
@@ -34,13 +34,19 @@ def embed_contacts(body, contacts)
     end
 
     def embed_images(body, images)
+      body = replace_one_new_line_before_an_image_with_two(body)
+
       body&.gsub(/!!(\d+)/) do
```
kevindew: Rather than needing an additional similar substitution for this could we do the look up here based on regex last match?

I expect we have a few scenarios too based on whether windows or unix new lines are used. For example:

User input: `\r\n\r\n!!1` Expected Output: `\r\n\r\n[Image: file.jpg]`
User input: `\n\n!!1` Expected Output: `\n\n[Image: file.jpg]`
User input: `\r\n!!1` Expected Output: `\r\n\r\n[Image: file.jpg]`
User input: `\n!!1` Expected Output: `\n\n[Image: file.jpg]`
User input: `!!1` Expected Output: `\n\n[Image: file.jpg]`

And you might be able to manage that with a regex of: `/(\r\n\r\n|\n\n|\r\n|\n)?(!!\d)/` and then you'll have the separator value available in Regexp.last_match[1] to decide whether to change to two new lines in the case of nil or to repeat once in the case of "\n" or "\r\n"

spec/lib/whitehall_importer/embed_body_references_spec.rb
```diff
@@ -30,6 +30,16 @@
...
     end
 
+    it "ensures that one new line before an image is replaced by two" do
+      govspeak_body = described_class.call(
+        body: "hello\r\n!!1\n!!2",
+        images: ["foo.png", "bar.jpg"],
+      )
+
+      expect(govspeak_body)
+        .to eq("hello\r\n\n[Image:foo.png]\n\n[Image:bar.jpg]")
```
kevindew: This comes out a little whacky with the mix of windows and unix line breaks in the first example where I'd expect `\r\n\r\n`

lib/whitehall_importer/embed_body_references.rb
```diff
@@ -34,13 +34,19 @@ def embed_contacts(body, contacts)
...
         whitehall_image_index = Regexp.last_match[1].to_i
         image_name = images[whitehall_image_index - 1]
         image_name.present? ? "[Image:#{image_name}]" : ""
       end
     end
 
+    def replace_one_new_line_before_an_image_with_two(body)
+      body&.gsub(/(.+)(?=(\n{1}!!\d+))/, "\\1\n")
```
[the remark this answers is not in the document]
gclssvglx: @kevindew Currently, there are no known failing documents due to 0 lines before an image.  Both failing documents referenced on the [card](https://trello.com/c/x42Ye088/1406-fix-import-of-embedded-images-in-news-articles) are due to a single (not double) new line being entered by the user.  I feel this highlights a UI or training issue.  The expectation is that the user previews the document before saving - but there is nothing in place to ensure they do.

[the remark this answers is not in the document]
gclssvglx: @kevindew `Attachments` are not flagged on the card, I'll add that on and handle them in the code.

lib/whitehall_importer/embed_body_references.rb
gclssvglx: @kevindew Thanks for the comments.  I'll look into this.

spec/lib/whitehall_importer/embed_body_references_spec.rb
gclssvglx: Yes, it is a bit funky!  I was trying to show that `\r`'s aren't touched by this code, but may exist.  However, what I didn't want to do is write duplicate tests (with and without `\r`).  Is the best (most pragmatic) solution to simply remove all `\r`'s first?

kevindew:
Following chat this morning and looking at [Whitehall's handling](https://github.com/alphagov/govspeak/blob/master/lib/govspeak.rb#L223-L228) we want to match Whitehall's behaviour which does seem to require the item to be on a new line.

so revised regex match is probably `(\A|\n\n|\r\n\r\n|\n|\r\n)` and if we get one of the last two items we want to double them (`\n` -> `\n\n`, `\r\n` -> `\r\n\r\n`)

gclssvglx:
All input on this PR is of course more than welcome.  However, there is one area in particular that I would value your thoughts on...  

[This test](https://github.com/alphagov/content-publisher/blob/288f4c22a07f7d8715b58aac14743529a4bb8818/spec/lib/whitehall_importer/embed_body_references_spec.rb#L51) concerns both `Attachment`s and `InlineAttachment`s, as such an update was required to add two new lines prior to the expected `Attachment` output.  As a consequence of that, the rendered HTML output will now look odd!

One potential solution would be to remove the `InlineAttachment` element from the test.  It could then be made to reflect [this](https://github.com/alphagov/content-publisher/blob/288f4c22a07f7d8715b58aac14743529a4bb8818/spec/lib/whitehall_importer/embed_body_references_spec.rb#L24) `Image`  test for consistency.  `InlineAttachment`s could then be handled by their own - specific - test or tests.

lib/whitehall_importer/embed_body_references.rb
```diff
@@ -33,20 +33,23 @@ def embed_contacts(body, contacts)
...
       end
     end
 
-    def embed_attachments(body, attachments)
-      body&.gsub(/!@(\d+)/) do
-        whitehall_attachment_index = Regexp.last_match[1].to_i
-        attachment_name = attachments[whitehall_attachment_index - 1]
-        attachment_name.present? ? "[Attachment:#{attachment_name}]" : ""
-      end
+    def convert_new_lines(new_lines)
```
kevindew: A situation that isn't covered here is the one where the regex matches on `\A` - as that will get changed to a new line.

Though based on the revised criteria in https://github.com/alphagov/content-publisher/pull/1756#issuecomment-584568077 we should just need to be doing the doubling and can probably negate the no new line case.

gclssvglx:
> Given an input of: `Bar !@2 Baz [InlineAttachment:3] << removed, but !@1 continues to work`
> And: `!@2` does not exist
> We'll receive a govspeak output of: `Bar \n\n Baz << removed but \n\n[Attachment:foo.pdf] continues to work`
> 
> Which will mean that the existing rendering will separate Bar and Baz onto new paragraphs?

Sorry, yes it is confusing.  What is produced is...

```
Bar  Baz  << removed, but \n\n[Attachment:foo.pdf] continues to work
```
So, a double space between `Bar` and `Baz` and two new lines before `[Attachment:...]`.  When rendered as HTML aren't we going to get something like...

```
Bar  Baz  << removed, but

{IMAGE} continues to work
```

As a test I guess the (visual) output doesn't really matter, but I thought you might prefer to separate the `Attachment` and `InlineAttachment` tests.  If you're happy - I'm happy!

kevindew:
> > Given an input of: `Bar !@2 Baz [InlineAttachment:3] << removed, but !@1 continues to work`
> > And: `!@2` does not exist
> > We'll receive a govspeak output of: `Bar \n\n Baz << removed but \n\n[Attachment:foo.pdf] continues to work`
> > Which will mean that the existing rendering will separate Bar and Baz onto new paragraphs?
> 
> Sorry, yes it is confusing. What is produced is...
> 
> ```
> Bar  Baz  << removed, but \n\n[Attachment:foo.pdf] continues to work
> ```
> 
> So, a double space between `Bar` and `Baz` and two new lines before `[Attachment:...]`. When rendered as HTML aren't we going to get something like...
> 
> ```
> Bar  Baz  << removed, but
> 
> {IMAGE} continues to work
> ```
> 
> As a test I guess the (visual) output doesn't really matter, but I thought you might prefer to separate the `Attachment` and `InlineAttachment` tests. If you're happy - I'm happy!

Sure I think you've missed a nuance from the revised regex.

With the revised one of `(\A|\n\n|\r\n\r\n|\n|\r\n)(!@[\d])` the string you provided of: `Bar !@2 Baz [InlineAttachment:3] << removed, but !@1 continues to work` would not match instead for matches you'd need: `Bar \n!@2 Baz [InlineAttachment:3] << removed, but \n!@1 continues to work` - this matches what is in [Whitehall for attachments](https://github.com/alphagov/whitehall/blob/master/app/helpers/govspeak_helper.rb#L251) and [Govspeak for images](https://github.com/alphagov/govspeak/blob/8edb84b3192bb4722fef85af10849dbc70612af9/lib/govspeak.rb#L223)

spec/lib/whitehall_importer/embed_body_references_spec.rb
```diff
@@ -12,31 +12,135 @@
...
+      it "converts no carridge returns" do
+        govspeak_body = described_class.call(
+          body: "!!1",
+          images: ["file.jpg"],
+        )
+
+        expect(govspeak_body).to eq("[Image:file.jpg]")
+      end
+
+      it "converts Whitehall image syntax to Content Publisher syntax" do
```
kevindew: Should this test be `it "ignores images that are neither start of the string or preceeded by new lines"` ?

spec/lib/whitehall_importer/embed_body_references_spec.rb
```diff
@@ -12,31 +12,135 @@
...
+        govspeak_body = described_class.call(
+          body: "!!1 test !!2",
+          images: ["file.png", "file.jpg"],
+        )
+
+        expect(govspeak_body).to eq("[Image:file.png] test !!2")
+      end
+
+      it "removes any image markdown that doesn't resolve to an image" do
```
kevindew: This description doesn't look correct since in this example we actually ignore the markdown rather than remove it.

spec/lib/whitehall_importer/embed_body_references_spec.rb
```diff
@@ -12,31 +12,135 @@
...
+        expect(govspeak_body).to eq("Bar !!2 Baz")
+      end
     end
 
-    it "converts Whitehall attachment syntax to Content Publisher syntax" do
-      govspeak_body = described_class.call(
-        body: "!@1 test !@2",
-        attachments: ["file.pdf", "download.csv"],
-      )
+    context "embedded attachments" do
```
kevindew: Seems a bit of a bummer that we repeat all the same tests again - hmm we could try a shared example but that will have quite a lot of arguments so probably may cause more harm than good.

Maybe we can combine things to try make them more concise.

For example:

```
it "converts Whitehall image bang embeds to govspeak Image embeds" do
  body = described_class.call(body: "!!1", images: ["foo.png"])
  expect(body).to eq("[Image:foo.png]")
end

it "converts Whitehall attachment bang embeds to govspeak Image embeds" do
  body = described_class.call(body: "!@1", attachments: ["foo.pdf"])
  expect(body).to eq("[Attachment:foo.pdf]")
end

it "converts image and attachment embeds prefixed by single new lines to be prefixed by two" do
  body = described_class.call(body: "\n!!1\n!@1", images: ["foo.png"], attachments: ["foo.pdf"])
  expect(body).to eq("\n\n[Image:foo.png]\n\n[Attachment:foo.pdf]")
  body = described_class.call(body: "\r\n!!1\r\n!@1", images: ["foo.png"], attachments: ["foo.pdf"])
  expect(body).to eq("\r\n\r\n[Image:foo.png]\r\n\r\n[Attachment:foo.pdf]")
end

it "doesn't change the prefixes of images and attachments preceeded by two or more new lines" do
  body = described_class.call(body: "\n\n!!1\n\n\n!@1", images: ["foo.png"], attachments: ["foo.pdf"])
  expect(body).to eq("\n\n[Image:foo.png]\n\n\n[Attachment:foo.pdf]")

  body = described_class.call(body: "\r\n\r\n\r\n!!1\r\n\r\n!@1", images: ["foo.png"], attachments: ["foo.pdf"])
  expect(body).to eq("\r\n\r\n\r\n[Image:foo.png]\r\n\r\n[Attachment:foo.pdf]")
end
```

lib/whitehall_importer/embed_body_references.rb
```diff
@@ -33,20 +33,19 @@ def embed_contacts(body, contacts)
       end
     end
 
-    def embed_images(body, images)
-      body&.gsub(/!!(\d+)/) do
-        whitehall_image_index = Regexp.last_match[1].to_i
-        image_name = images[whitehall_image_index - 1]
-        image_name.present? ? "[Image:#{image_name}]" : ""
+    def embed_files(body, files, file_type_indicator, file_type_name)
```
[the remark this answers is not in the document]
gclssvglx: Cool - makes things much more readable.  Like it!

lib/whitehall_importer/embed_body_references.rb
```diff
@@ -33,20 +33,19 @@ def embed_contacts(body, contacts)
...
+      body&.gsub(/(\A|\n\n|\r\n\r\n|\n|\r\n)#{file_type_indicator}(\d+)/) do
+        whitehall_new_lines = Regexp.last_match[1]
+        whitehall_file_index = Regexp.last_match[2].to_i
+
+        file_name = files[whitehall_file_index - 1]
+        new_lines = convert_new_lines(whitehall_new_lines)
```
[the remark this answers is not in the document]
gclssvglx: Good call 👍

spec/lib/whitehall_importer/embed_body_references_spec.rb
gclssvglx: Yup, makes sense.  There's also one further down for Attachments - we'll handle that further down in your suggestions.

spec/lib/whitehall_importer/embed_body_references_spec.rb
gclssvglx: Agree.  Thanks

spec/lib/whitehall_importer/embed_body_references_spec.rb
gclssvglx: Umm, I'm not a big fan of munging concepts like this, but I see your point!

I've had a go - let me know if I've DRYed it too far.

spec/lib/whitehall_importer/embed_body_references_spec.rb
```diff
@@ -12,49 +12,106 @@
...
 
-      expect(govspeak_body).to eq("Bar  Baz")
+      expect(body).to eq("[Attachment:file.pdf] test !@2")
     end
 
-    it "converts Whitehall attachment syntax to Content Publisher syntax" do
-      govspeak_body = described_class.call(
-        body: "!@1 test !@2",
-        attachments: ["file.pdf", "download.csv"],
+    it "ignore any image markdown that doesn't resolve to an image" do
```
kevindew: I don't think this text is correct as I believe this didn't resolve because this has no new lines before it rather than because the image doesn't resolve. 

If the image doesn't resolve I thought we ended up with an empty string?

spec/lib/whitehall_importer/embed_body_references_spec.rb
```diff
@@ -12,49 +12,106 @@
...
 
-      expect(govspeak_body).to eq("[Attachment:file.pdf] test [Attachment:download.csv]")
+      expect(body).to eq("Bar !!2 Baz")
     end
 
-    it "converts Whitehall inline attachment syntax to Content Publisher syntax" do
-      govspeak_body = described_class.call(
-        body: "[InlineAttachment:1] test [InlineAttachment:2]",
-        attachments: ["file.pdf", "download.csv"],
+    it "ignore any attachment markdown that doesn't resolve to an attachment" do
```
kevindew: Same point as above about this description not matching what is being tested.

spec/lib/whitehall_importer/embed_body_references_spec.rb
```diff
@@ -12,49 +12,106 @@
...
+    it "converts image and attachment embeds prefixed by single new lines to be prefixed by two" do
+      prefix_conversions = [
+        { prefix: "\n", conversion: "\n\n" },
+        { prefix: "\r\n", conversion: "\r\n\r\n" },
+      ]
+
+      prefix_conversions.each do |prefix_conversion|
+        prefixed = []
+        prefixed << "#{prefix_conversion[:prefix]}!!1"
+        prefixed << "#{prefix_conversion[:prefix]}!@1"
```
kevindew: Minor but could do this more directly and concisely with: 

```
        prefixed = ["#{prefix_conversion[:prefix]}!!1",
"#{prefix_conversion[:prefix]}!@1"]
```

spec/lib/whitehall_importer/embed_body_references_spec.rb
```diff
@@ -12,49 +12,106 @@
...
+      prefixes.each do |prefix|
+        body = described_class.call(
+          body: "#{prefix}!!1#{prefix}!@1",
+          images: ["file.jpg"],
+          attachments: ["file.pdf"],
+        )
+
+        expect(body)
+          .to eq("#{prefix}[Image:file.jpg]#{prefix}[Attachment:file.pdf]")
+      end
```
kevindew: Cool this works - looped expectations are frowned upon in tests but can see the value here.

gclssvglx: I think there are - rare - occasions when this does work.  I feel this is one of those occasions too.

spec/lib/whitehall_importer/embed_body_references_spec.rb
gclssvglx: Agree.

spec/lib/whitehall_importer/embed_body_references_spec.rb
gclssvglx: Good spot - thanks.  Have update the test to clarify the intent.

spec/lib/whitehall_importer/embed_body_references_spec.rb
gclssvglx: As above.

PR #1757, opened 2020-02-10
Check for issues before updating images

benthorner:

This is a continuation of https://github.com/alphagov/content-publisher/pull/1753.

### DIFF
--- a/app/views/images/edit.html.erb
+++ b/app/views/images/edit.html.erb
@@ -46,7 +46,7 @@ content_for :title, t(title_key, title: @edition.title_or_fallback)
       },
       id: "image-alt-text",
       name: "image_revision[alt_text]",
-      value: @image_revision.alt_text,
+      value: params.dig(:image_revision, :alt_text) || @image_revision.alt_text,
       error_items: @issues&.items_for(:alt_text),
       describedby: "image-alt-text-guidance"
     } %>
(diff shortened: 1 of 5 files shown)


app/views/images/edit.html.erb
```diff
@@ -86,20 +86,22 @@ content_for :title, t(title_key, title: @edition.title_or_fallback)
...
       name: "image_revision[credit]",
-      value: @image_revision.credit,
+      value: params.dig(:image_revision, :credit) || @image_revision.credit,
       error_items: @issues&.items_for(:credit),
       describedby: "image-credit-guidance"
     } %>
   <% end %>
 
   <% unless rendering_context == "modal" %>
+    <% is_lead = params[:lead_image] || (@edition.lead_image_revision == @image_revision) %>
```
kevindew: We've got a bit of a tricky problem here of when to decide when this is checked or not on a form resubmission. As if I fill in this form and untick this box but hit a different validation error the form won't remember that it's not checked as params[:lead_image] will not exist. 

What we could probably use is something on params that is an indicator the form has been submitted. This could be achieved by nesting this option inside the grouping of image_revision. So we could do `params[:image_revision] ? params.dig("image_revision", "lead_image") : (@edition.lead_image_revision == @image_revision)`

benthorner: Oh yeah, good spot - I think I managed to spot this myself somewhere else. Updated 👍

PR #1758, opened 2020-02-10
Improve i18n coverage for document type creation

ChrisBAshton:
Before:
=> Coverage report generated for I18n to /govuk/content-publisher/coverage/i18n. 476 / 502 keys (94.82%) covered.
After:
=> Coverage report generated for I18n to /govuk/content-publisher/coverage/i18n. 498 / 502 keys (99.2%) covered.

This PR comes out of work originating in Trello card: https://trello.com/c/ihcpYeWf/1320-iterate-how-we-test-formats but is dependent on #1768.


spec/views/content/edit.html.erb_spec.rb
```diff
@@ -0,0 +1,19 @@
...
+    it "should render correctly for every document type" do
+      document_types_we_manage = DocumentType.all.filter { |dt| dt.managed_elsewhere.blank? }
+      document_types_we_manage.each do |document_type|
+        edition = create(:edition, document_type: document_type)
+        assign(:edition, edition)
+        assign(:revision, edition.revision)
+        render
+
+        document_type.contents.each do |field|
+          expect(response).to render_template("content/edit/_#{field.id}_field")
```
kevindew: This doesn't feel like something we want to add as Rails has moved in the direction away from this. See: https://github.com/rails/rails/issues/18950 for more. Also seems a massive shame to add an extra dependency for ensuring coverage.

app/models/document_type.rb
```diff
@@ -36,6 +36,10 @@ def managed_elsewhere_url
     Plek.new.external_url_for(managed_elsewhere.fetch("hostname")) + managed_elsewhere.fetch("path")
   end
 
+  def parent
+    Supertype.all.find { |supertype| supertype.document_types.map(&:id).include?(id) }
```
kevindew: would `Supertype.all.find { |supertype| supertype.document_types.include?(self) }` work? If not I'd presume we might want to override equality in document_type so it does - though is this stuff all about to change with the work you and Leena are doing?

spec/views/content/edit.html.erb_spec.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+RSpec.describe "content/edit.html.erb" do
+  describe "edit document form" do
+    it "should render correctly for every document type" do
+      document_types_we_manage = DocumentType.all.filter { |dt| dt.managed_elsewhere.blank? }
+      document_types_we_manage.each do |document_type|
```
kevindew: I wonder if this can be covered implicitly by feature tests for each format ensuring we hit the edit view? I'm guessing for coverage we need each to visit each supported documents edit content page, or is it more complex than that?

spec/views/new_document/choose_document_type.html.erb_spec.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+RSpec.describe "new_document/choose_document_type.html.erb" do
+  describe "choose document type form" do
+    it "should render correctly for every document type" do
+      DocumentType.all.each do |document_type|
+        assign(:supertype, document_type.parent)
+        render
+        expect(rendered).to have_content(document_type.label)
+      end
```
kevindew: Loops are normally frowned upon tests, though I'm not sure there'll particularly nice way to do this without a loop as feature tests would need a lot of steps.


If we do need a loop it's probably better to do this with the it inside the loop. For example:
```
    DocumentType.all.each do |document_type|
      it "should render correctly for #{document_type.name}" do
        assign(:supertype, document_type.parent)
        render
        expect(rendered).to have_content(document_type.label)
      end
    end
```

Assuming there's not some loading related issue that stops you using DocumentType.all until runtime.

I assume this also is going to change very soon though?

spec/views/content/edit.html.erb_spec.rb
ChrisBAshton: Agreed it's a shame to introduce a dependency, and that testing which file is rendered feels overly tightly coupled to implementation detail.

I considered searching for specific strings in the rendered output, corresponding to each input field, but the ids don't correspond to anything in the HTML form, so it would have to be a hardcoded lookup. Alternatively I could `expect(response.scan("<textarea")).to have_exactly(3).items` - but that relies on us using only textarea for our inputs 😕

spec/views/content/edit.html.erb_spec.rb
ChrisBAshton: I think on balance feature tests are a better idea. We had one already, but removed it here: https://github.com/alphagov/content-publisher/pull/1710/commits/66e153d8a0bd39831266e91132fffa66bf7afdf7

I could look at reintroducing it, but without the repetition between press_release and news_story.

app/models/document_type.rb
ChrisBAshton: I'm excited about overriding equality - not done that before! 🎉 
Yes it's tricky to do this work in parallel with the multi-level document type work. I figured as this test iteration work is smallest, it will probably be completed & merged first; I can then rebase the other branch. But I'm equally happy parking this for now and then tacking this i18n coverage work on at the end of the other branch.

spec/views/new_document/choose_document_type.html.erb_spec.rb
ChrisBAshton: +1 for moving the `it` block inside the loop. I'll do that.

This will change, but not by very much - we'd probably iterate over a `DocumentTypeSelection` instead of a `DocumentType` and `assign(:document_type_selection, document_type_selection)`.

spec/views/content/edit.html.erb_spec.rb
ChrisBAshton: Have talked IRL with Ben and settled on a schema validation approach - it's more expressive than attempting to validate fields in pure RSpec.

spec/views/content/edit.html.erb_spec.rb
ChrisBAshton: Spoke to Ben IRL about this, and decided as "rendering an edit form" isn't a self-contained feature, then we should not really write a feature test for it (especially as we're adding many more formats soon!)

Hopefully have now covered this satisfactorily in the schema test.

spec/models/document_type_spec.rb
```diff
@@ -12,7 +12,19 @@
         validator = JSON::Validator.fully_validate(document_type_schema, document_type)
         expect(validator).to(
           be_empty,
-          "Validation for #{document_type['id']} failed: \n\t#{validator.join("\n\t")}",
+          "Document type validation for #{document_type['id']} failed: \n\t#{validator.join("\n\t")}",
+        )
+      end
+    end
+
+    it "should have locale keys that conform to the document type locale schema" do
```
benthorner: Since we're expecting to have this for a third time soon, perhaps it's time for a helper?

ChrisBAshton: Done 👍

kevindew:
Just a quick drive by comment where I've not remotely read enough to have full context. I notice we've created our own json schema matcher and I know thoughtbot have a gem for one: https://github.com/thoughtbot/json_matchers just wondering if we'd benefit from that than DIYing it. Key advantage I'd imagine is the gem might give nicer output of problems.

kevindew:
Also I noticed from the thoughtbot gem that they suggest schemas being within spec/support which might be a good idea for us too if we're only using them as part of testing and not part of app runtime.

spec/models/document_type_selection_spec.rb
```diff
@@ -4,15 +4,22 @@
   let(:document_type_selections) { YAML.load_file(Rails.root.join("config/document_type_selections.yml")) }
 
   describe "all configured document types selections are valid" do
-    let(:document_type_selection_schema) { JSON.parse(File.read("config/schemas/document_type_selection.json")) }
-
     it "conforms to the document type selection schema" do
+      expect(document_type_selections).to all(match_json_schema("document_type_selection"))
+    end
```
kevindew: Yay no loop 👍

spec/models/document_type_selection_spec.rb
```diff
@@ -4,15 +4,22 @@
...
+      end
+    end
+
+    it "has locale keys for every option, conforming to the document type selection locale schema" do
+      options = document_type_selections.map { |s| s["options"] }.flatten
+      options.each do |document_type_selection|
+        translations = I18n.t("document_type_selections.#{document_type_selection['id']}").deep_stringify_keys
+        expect(translations).to match_json_schema("document_type_selection_locale")
       end
     end
```
kevindew: It seems a bit strange for us to be testing that these locales match a schema and feels rather unconventional. I seem to remember us previously testing views which felt like a scenario that added more value, what's the rationale for switching and/or are the views problematic?

ChrisBAshton: Thanks for the review, Kevin. Agreed it feels unconventional, but so did an attempt at the view tests, since nothing was really _tested_ - we simply visited a bunch of URLs, perhaps making an assertion too as a token gesture. At the time of writing, it would also have meant maintaining a hardcoded array of URLs (now Leena has implemented the `document_type_selection_id` parameter, we could construct a list of all options dynamically).

The schema test does lend some benefits, e.g. failing if an unused property is accidentally added. It makes failure more visible than a bit of text in the console output, and it can be run in isolation as opposed to having to run the whole test suite to test locale usage. Perhaps we should look at this as the beginning of a replacement for the i18n coverage checker in the long term.

In my chat with Ben IRL, we didn't exactly settle on a schema test, but his suggestion was to write a test to grab the locale hash for each document type and check it contains certain keys. This would rewrite a solution already solved by schema tests, IMO, and would also give false positives since the act of retrieving the hash would satisfy the i18n coverage checker even if the hash contained unused keys.

spec/models/document_type_spec.rb
```diff
@@ -5,13 +5,13 @@
...
+
+    it "has locale keys that conform to the document type locale schema" do
       document_types.each do |document_type|
-        validator = JSON::Validator.fully_validate(document_type_schema, document_type)
-        expect(validator).to(
-          be_empty,
-          "Validation for #{document_type['id']} failed: \n\t#{validator.join("\n\t")}",
-        )
+        translations = I18n.t("document_types.#{document_type['id']}").deep_stringify_keys
+        expect(translations).to match_json_schema("document_types_locale")
```
benthorner: Minor: slight inconsistency between `document_type` and `document_type*s*_locale`.

spec/models/document_type_selection_spec.rb
benthorner: I can add a little to this: I explored bumping coverage for fields with view tests, and the result is indeed contrived, since there's nothing to actually test. My worry would be that someone would just delete the tests, since it's not clear what it's testing. The schema (or at least, model-level) approach makes the intent explicit.

For document types, I think the view approach is less clear, since I reckon we'd need to visit all the different branches of the selection tree in order to exercise all the locales.

The schema approach for locales did seem a little strange at first. Since they're just YAML files, it seems reasonable and conventional to use a schema-driven approach to ensure they're valid.

PR #1763, opened 2020-02-11
Follow-up tweaks for #1753

benthorner:

### DIFF
--- a/app/views/content/edit/_change_notes.html.erb
+++ b/app/views/content/edit/_change_notes.html.erb
@@ -3,7 +3,7 @@
       label: {
         text: t("content.edit.change_note.title"),
       },
-      id: "edition-change-note",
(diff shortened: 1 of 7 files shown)


app/views/content/edit/_change_notes.html.erb
```diff
@@ -3,7 +3,7 @@
       label: {
         text: t("content.edit.change_note.title"),
       },
-      id: "edition-change-note",
+      id: "change-note-textarea",
```
kevindew: We have a precedent to suffixing these with "field". This precedent was started before field had extra meaning to it but I still think it's ok and preferable to have something less specific.

benthorner: Sure, updated. I think this convention has passed me by!

PR #1764, opened 2020-02-11
Enable rubocop-rspec

benthorner:
https://trello.com/c/OQEupQZq/1529-the-way-we-write-tests-is-inconsistent

This PR enables rubocop-rspec with a minimal set of overrides to make the
change practical (e.g. instance vars in feature specs). After applying the
bulk of the changes and corrections, I've then added extra commits to reduce
our overrides further - these are open for discussion!

Depends on: https://github.com/alphagov/content-publisher/pull/1763


spec/views/new_document/guidance.html.erb_spec.rb
```diff
@@ -2,7 +2,7 @@
 
 RSpec.describe "new_document/guidance.html.erb" do
   describe "guidance for publishers" do
-    it "should render the associated Markdown content" do
+    it "renders the associated Markdown content" do
```
kevindew: That's rather clever for the autocorrect

spec/features/editing_content/edit_edition_spec.rb
```diff
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
-RSpec.feature "Edit an edition" do
-  scenario do
+RSpec.describe "Edit an edition" do
+  it do
```
kevindew: These scenario blocks seem to be the biggest casualty of this switch over hmm - I guess it was always a bit of a hack that we had empty descriptions.

We've also got the minor annoyance of convention to deal with too, for example: https://gds-way.cloudapps.digital/standards/testing-with-rspec.html#example-integration-test - although that has the oddity of `Rspec.scenario` - and then in the unit test above we've got a subject 🤦‍♂

benthorner:
@kevindew I've updated this as discussed, and added a couple of TODOs to the card, for rules we want to explore disabling in future work.

benthorner:
@leenagupte is there a time when it would not cause major merge conflicts?

PR #1766, opened 2020-02-12
Fix import of documents with PDF attachments from Whitehall to Content Publisher

brucebolt:
When doing a test migration of NDA's news stories, we found three articles with an embedded PDF attachment did not import with an error:

```
NoMethodError: undefined method `tempfile' for #<Tempfile:/tmp/open-uri20200129-1-g41dgq>
/govuk/content-publisher/app/services/create_file_attachment_blob_service.rb:37:in `number_of_pages'
```

This manual testing revealed there were a number of issues that we had not previously seen through the existing automated tests, which are resolved through this PR:

- Adds require statements for `gds_api/whitehall_export` that were missing
- Adds a `tempfile` method for the imported PDF, solving the issue giving the error above
- Uses a string for a missing asset URL when this would otherwise be nil (i.e. assets not yet synced to asset manager), which otherwise causes a `link_to` to fail
- Ignores inconsistencies in file size presentation (e.g. `131KB` in documents published by Whitehall, but `131 KB` in those published by Content Publisher)

Additionally, the body checking part of the integrity checker has been moved into a subclass, which has made testing this code under specific scenarios much simpler.

Trello card: https://trello.com/c/86Adk1bO

### DIFF
--- a/lib/whitehall_importer/create_file_attachment_revision.rb
+++ b/lib/whitehall_importer/create_file_attachment_revision.rb
@@ -100,5 +100,9 @@ module WhitehallImporter
     def content_type
       nil
     end
+
+    def tempfile
+      self
+    end
   end
 end
--- a/lib/whitehall_importer/integrity_checker.rb
+++ b/lib/whitehall_importer/integrity_checker.rb
@@ -90,11 +90,13 @@ module WhitehallImporter
     end
 
     def primary_publishing_organisation_matches?
-      proposed_payload.dig(:links, :primary_publishing_organisation) == publishing_api_link(:primary_publishing_organisation)
+      proposed_payload.dig(:links, :primary_publishing_organisation).presence ==
+        publishing_api_link(:primary_publishing_organisation).presence
     end
 
     def organisations_match?
-      proposed_payload.dig(:links, :organisations)&.sort == publishing_api_link(:organisations)&.sort
+      proposed_payload.dig(:links, :organisations).to_a.sort ==
+        publishing_api_link(:organisations).to_a.sort
     end
 
     def publishing_api_content
--- a/lib/whitehall_importer/integrity_checker.rb
+++ b/lib/whitehall_importer/integrity_checker.rb
@@ -37,16 +37,12 @@ module WhitehallImporter
         end
       end
 
-      problems << "body text doesn't match" unless body_text_matches?
+      problems << "body text doesn't match" unless BodyTextCheck.new(
+        proposed_payload.dig(:details, :body),
+        publishing_api_content.dig(:details, :body),
+      ).sufficiently_similar?
 
       problems
-    end
-
-    def body_text_matches?
-      proposed_body_text = proposed_payload.dig(:details, :body)
-      publishing_api_body_text = publishing_api_content.dig(:details, :body)
-
-      Sanitize.clean(publishing_api_body_text).squish == Sanitize.clean(proposed_body_text).squish
     end
 
     def image_problems
--- a/spec/lib/whitehall_importer/integrity_checker/body_text_check_spec.rb
+++ b/spec/lib/whitehall_importer/integrity_checker/body_text_check_spec.rb
@@ -15,6 +15,32 @@ RSpec.describe WhitehallImporter::IntegrityChecker::BodyTextCheck do
       expect(integrity_check.sufficiently_similar?).to be true
     end
 
+    it "returns true even if there is a mismatch in inline atttachment URL filesize" do
+      proposed_body = %(
+        <p>
+          <span class="gem-c-attachment-link">
+            <a class="govuk-link" href="filename.pdf" target="_blank">Test File</a>
+            (<span class="gem-c-attachment-link__attribute"><abbr title="Portable Document Format" class="gem-c-attachment-link__abbr">PDF</abbr></span>,
+              <span class="gem-c-attachment-link__attribute">391 KB</span>, <span class="gem-c-attachment-link__attribute">9 pages</span>)
+          </span>
+        </p>
+      )
+
+      publishing_api_body = %(
+        <p>
+          <span class="attachment-inline">
+            <a href="/filename.pdf">Test File</a>
+            (<span class="type">PDF</span>,
+              <span class="file-size">391KB</span>,
+              <span class="page-length">9 pages</span>)
+          </span>
+        </p>
+      )
+
+      integrity_check = described_class.new(proposed_body, publishing_api_body)
+      expect(integrity_check.sufficiently_similar?).to be true
+    end
+
     it "returns false when the body text doesn't match" do
       integrity_check = described_class.new("Some text", "Some different text")
       expect(integrity_check.sufficiently_similar?).to be false
--- a/spec/lib/whitehall_importer/integrity_checker/body_text_check_spec.rb
+++ b/spec/lib/whitehall_importer/integrity_checker/body_text_check_spec.rb
@@ -1,0 +1,13 @@
+RSpec.describe WhitehallImporter::IntegrityChecker::BodyTextCheck do
+  describe "#sufficiently_similar?" do
+    it "retuns true if the proposed payload matches" do
+      integrity_check = described_class.new("Some text", "Some text")
+      expect(integrity_check.sufficiently_similar?).to be true
+    end
+
+    it "returns false when the body text doesn't match" do
+      integrity_check = described_class.new("Some text", "Some different text")
+      expect(integrity_check.sufficiently_similar?).to be false
+    end
+  end
+end
--- a/spec/lib/whitehall_importer/integrity_checker/body_text_check_spec.rb
+++ b/spec/lib/whitehall_importer/integrity_checker/body_text_check_spec.rb
@@ -5,6 +5,16 @@ RSpec.describe WhitehallImporter::IntegrityChecker::BodyTextCheck do
       expect(integrity_check.sufficiently_similar?).to be true
     end
 
+    it "returns true when the body text whitespace does not match" do
+      integrity_check = described_class.new("Some text", "Some     text")
+      expect(integrity_check.sufficiently_similar?).to be true
+    end
+
+    it "returns true when the HTML does not match" do
+      integrity_check = described_class.new("<b>Some text</b>", "Some text")
+      expect(integrity_check.sufficiently_similar?).to be true
+    end
+
     it "returns false when the body text doesn't match" do
       integrity_check = described_class.new("Some text", "Some different text")
       expect(integrity_check.sufficiently_similar?).to be false
--- a/spec/lib/whitehall_importer/integrity_checker_spec.rb
+++ b/spec/lib/whitehall_importer/integrity_checker_spec.rb
@@ -20,21 +20,14 @@ RSpec.describe WhitehallImporter::IntegrityChecker do
     end
 
     let(:publishing_api_item) do
-      {
-        content_id: edition.content_id,
-        base_path: edition.base_path,
-        title: edition.title,
-        description: edition.summary,
-        document_type: edition.document_type.id,
-        schema_name: edition.document_type.publishing_metadata.schema_name,
-        details: {
-          body: GovspeakDocument.new(edition.contents["body"], edition).payload_html,
-        },
-        links: {
-          primary_publishing_organisation: edition.tags["primary_publishing_organisation"].to_a,
-          organisations: edition.tags["organisations"].to_a + edition.tags["primary_publishing_organisation"].to_a,
-        },
-      }
+      default_publishing_api_item(edition,
+                                  details: {
+                                    body: GovspeakDocument.new(edition.contents["body"], edition).payload_html,
+                                  },
+                                  links: {
+                                    primary_publishing_organisation: edition.tags["primary_publishing_organisation"].to_a,
+                                    organisations: edition.tags["organisations"].to_a + edition.tags["primary_publishing_organisation"].to_a,
+                                  })
     end
 
     it "returns true if there aren't any problems for edition without image or attachment" do
@@ -69,17 +62,7 @@ RSpec.describe WhitehallImporter::IntegrityChecker do
     end
 
     it "compares against organisations in linkset links if there no edition links" do
-      stub_publishing_api_has_item(
-        content_id: edition.content_id,
-        base_path: edition.base_path,
-        title: edition.title,
-        description: edition.summary,
-        document_type: edition.document_type.id,
-        schema_name: edition.document_type.publishing_metadata.schema_name,
-        details: {
-          body: GovspeakDocument.new(edition.contents["body"], edition).payload_html,
-        },
-      )
+      stub_publishing_api_has_item(default_publishing_api_item(edition))
 
       stub_publishing_api_has_links(
         content_id: edition.content_id,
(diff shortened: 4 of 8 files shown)


lib/whitehall_importer/integrity_checker.rb
```diff
@@ -48,9 +48,19 @@ def body_text_matches?
       proposed_body_text = proposed_payload.dig(:details, :body)
       publishing_api_body_text = publishing_api_content.dig(:details, :body)
 
+      proposed_body_text = remove_html_elements(proposed_body_text, ".gem-c-attachment-link .gem-c-attachment-link__attribute:nth-of-type(2)")
```
kevindew: Ah yes this rings a bell about what you were saying about a generic class name. This seems reasonable. Might be an idea to put this in a variable eg file_size_selector = ".gem-c-attachment-link .gem-c-attachment-link__attribute:nth-of-type(2)" since we use it twice and it's meaning is quite obscure.

spec/lib/whitehall_importer/integrity_checker_spec.rb
```diff
@@ -92,6 +92,45 @@
...
+        links: {
+          primary_publishing_organisation: edition_with_inline_attachment.tags["primary_publishing_organisation"].to_a,
+          organisations: edition_with_inline_attachment.tags["organisations"].to_a + edition_with_inline_attachment.tags["primary_publishing_organisation"].to_a,
+        },
+      }
+
+      stub_publishing_api_has_item(publishing_api_item_with_inline_attachment)
+
+      integrity_check = WhitehallImporter::IntegrityChecker.new(edition_with_inline_attachment)
+      expect(integrity_check.valid?).to be true
```
kevindew: This test seems pretty heavy in the amount of things going on. I think it's a sign that there are perhaps too many things going on in this class. Maybe it's an idea to move the body checking part of the integrity checker to a nested class that we can test individually with body text input? That might make it easier to cover testing things like dealing with different HTML or spacing which I don't think we really test against.

lib/whitehall_importer/create_file_attachment_revision.rb
```diff
@@ -96,11 +96,16 @@ class AttachmentFileDecorator < SimpleDelegator
...
       @original_filename = original_filename
     end
 
     def content_type
       nil
     end
+
+    def tempfile
+      @tmp_file
+    end
```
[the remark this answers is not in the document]
brucebolt: I did initially use `PDF::Reader.new(file).page_count` but encountered a test failure (below) and also a 500 error when uploading an attachment.

```  1) Upload file attachment 
     Failure/Error:
       within(".gem-c-attachment") do
         expect(page).to have_content(@title)
         expect(page).to have_content(metadata)
       end
     
     Capybara::ElementNotFound:
       Unable to find css ".gem-c-attachment"
     # ./spec/features/editing_file_attachments/upload_file_attachment_spec.rb:43:in `then_i_can_see_previews_of_the_attachment'
     # ./spec/features/editing_file_attachments/upload_file_attachment_spec.rb:9:in `block (2 levels) in <top (required)>'```

File: lib/govspeak_document/payload_options.rb [hunk 4]
```diff
@@ -38,7 +38,7 @@ def attachment_attributes(attachment_revision)
     attributes = file_attachment_attributes(attachment_revision, edition.document)
 
     attributes.merge(
-      url: attachment_revision.asset_url,
+      url: attachment_revision.asset_url.nil? ? "http://placeholder" : attachment_revision.asset_url,
```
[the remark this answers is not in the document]
brucebolt: Ah, yes we did speak about this.  I've updated the commit to use the `in_app_html` method in the integrity check.

lib/whitehall_importer/integrity_checker.rb
brucebolt: Done

spec/lib/whitehall_importer/integrity_checker_spec.rb
brucebolt: I agree, testing this proved to be the most complex part of this PR.  I've nested a new class, and simplified the tests.

lib/whitehall_importer/integrity_checker.rb
```diff
@@ -138,5 +113,55 @@ def default_image?(proposed_image_payload, publishing_api_image, attribute)
         proposed_image_payload.empty? &&
         publishing_api_image[attribute] == "placeholder"
     end
+
+    class ContentIntegrityChecker < IntegrityChecker
```
kevindew: Suggestion that we move this into an integirty_checker directory

Also that we only deal with the body in here, then we can avoid the need to subclass and be simpler to test.

brucebolt: Good idea.  Have moved this and also broken the commit down into two (one to move the code, the other to add the file size ignoring).

brucebolt:
@kevindew I've made the changes we discussed this morning and added a test for running the integrity checker against a document with attachment.  I've also done some manual testing against the three documents we previously couldn't import.

lib/whitehall_importer/integrity_checker/body_check.rb
```diff
@@ -0,0 +1,29 @@
+module WhitehallImporter
+  class IntegrityChecker::BodyCheck
+    attr_reader :proposed_payload, :publishing_api_content
+
+    def initialize(proposed_payload, publishing_api_content)
+      @proposed_payload = proposed_payload
+      @publishing_api_content = publishing_api_content
```
kevindew: Rather than being passed in hashes that we have to pluck the body field from (with dig) can we pass in the body as strings? It'll save a lot of set-up in tests.

lib/whitehall_importer/integrity_checker/body_check.rb
```diff
@@ -0,0 +1,29 @@
...
+      publishing_api_body_text = publishing_api_content.dig(:details, :body)
+
+      content_publisher_file_size_selector = ".gem-c-attachment-link .gem-c-attachment-link__attribute:nth-of-type(2)"
+      whitehall_file_size_selector = ".attachment-inline .file-size, .gem-c-attachment-link .gem-c-attachment-link__attribute:nth-of-type(2)"
+      proposed_body_text = remove_html_elements(proposed_body_text, content_publisher_file_size_selector)
+      publishing_api_body_text = remove_html_elements(publishing_api_body_text, whitehall_file_size_selector)
+      Sanitize.clean(publishing_api_body_text).squish == Sanitize.clean(proposed_body_text).squish
+    end
+
+    def remove_html_elements(body, selector)
```
kevindew: This probably wants to be flagged as a private method.

lib/whitehall_importer/integrity_checker/body_check.rb
```diff
@@ -0,0 +1,29 @@
...
+    end
+
+    def body_text_matches?
+      proposed_body_text = proposed_payload.dig(:details, :body)
+      publishing_api_body_text = publishing_api_content.dig(:details, :body)
+
+      content_publisher_file_size_selector = ".gem-c-attachment-link .gem-c-attachment-link__attribute:nth-of-type(2)"
+      whitehall_file_size_selector = ".attachment-inline .file-size, .gem-c-attachment-link .gem-c-attachment-link__attribute:nth-of-type(2)"
+      proposed_body_text = remove_html_elements(proposed_body_text, content_publisher_file_size_selector)
+      publishing_api_body_text = remove_html_elements(publishing_api_body_text, whitehall_file_size_selector)
```
kevindew: I feel like this code would be easier to understand if we it was within a method that was called something like `remove_attachment_file_size`

lib/whitehall_importer/integrity_checker/body_check.rb
```diff
@@ -0,0 +1,29 @@
+module WhitehallImporter
+  class IntegrityChecker::BodyCheck
```
kevindew: I wonder if this should be BodyTextCheck? I'm a bit confused as we seem to mix up whether we describe the field as body or body text and it'd be simpler if we used the same phrasing consistently.

lib/whitehall_importer/integrity_checker/body_check.rb
```diff
@@ -0,0 +1,29 @@
...
+    attr_reader :proposed_payload, :publishing_api_content
+
+    def initialize(proposed_payload, publishing_api_content)
+      @proposed_payload = proposed_payload
+      @publishing_api_content = publishing_api_content
+    end
+
+    def body_text_matches?
```
kevindew: As this class is focussed on body we don't really need to prefix this method with body_text

I think calling it something like "sufficently_similar?" similar might be enough to convey what it does.

spec/lib/whitehall_importer/integrity_checker/body_check_spec.rb
```diff
@@ -0,0 +1,52 @@
+RSpec.describe WhitehallImporter::IntegrityChecker::BodyCheck do
+  describe "#content_problems" do
```
kevindew: This no longer represents the method name under test

spec/lib/whitehall_importer/integrity_checker/body_check_spec.rb
```diff
@@ -0,0 +1,52 @@
...
+      }
+    end
+
+    it "retuns no problems if the proposed payload matches" do
+      integrity_check = WhitehallImporter::IntegrityChecker::BodyCheck.new(proposed_payload, publishing_api_content)
+      expect(integrity_check.body_text_matches?).to be true
+    end
+
+    it "returns no problems even if there is a mismatch in inline atttachment URL filesize" do
+      proposed_payload[:details][:body] = "<p>Some other text</p><p><span class=\"gem-c-attachment-link\"><a class=\"govuk-link\" href=\"filename.pdf\" target=\"_blank\">Test File</a> (<span class=\"gem-c-attachment-link__attribute\"><abbr title=\"Portable Document Format\" class=\"gem-c-attachment-link__abbr\">PDF</abbr></span>, <span class=\"gem-c-attachment-link__attribute\">391 KB</span>, <span class=\"gem-c-attachment-link__attribute\">9 pages</span>)</span></p>"
```
kevindew: This would probably be more readable if we were able to format it without the slashes.

I think you can do something like:

```
body = %(
  <p>
    <span class="gem-c-attachment-link">
      <a class="govuk-link" href="filename.pdf" target="_blank">Test File</a>
      (<span class="gem-c-attachment-link__attribute"><abbr title="Portable Document Format" class="gem-c-attachment-link__abbr">PDF</abbr></span>,
      <span class="gem-c-attachment-link__attribute">391 KB</span>,
      <span class="gem-c-attachment-link__attribute">9 pages</span>)
    </span>
  </p>
}
```

spec/lib/whitehall_importer/integrity_checker/body_check_spec.rb
```diff
@@ -0,0 +1,52 @@
...
+    end
+
+    it "returns a problem when the body text doesn't match" do
+      proposed_payload[:details][:body] = "Some text"
+      publishing_api_content[:details][:body] = "Some different text"
+
+      integrity_check = WhitehallImporter::IntegrityChecker::BodyCheck.new(proposed_payload, publishing_api_content)
+      expect(integrity_check.body_text_matches?).to be false
+    end
+  end
```
kevindew: Now that we have a test here should we be testing a few situations that aren't covered by tests. For example how it ignores consecutive whitespace and ignores HTML differences?

spec/lib/whitehall_importer/integrity_checker_spec.rb
```diff
@@ -90,6 +90,57 @@
...
+          schema_name: edition.document_type.publishing_metadata.schema_name,
+          details: {
+            body: GovspeakDocument.new(edition.contents["body"], edition).payload_html,
+          },
+          links: {
+            primary_publishing_organisation: edition.tags["primary_publishing_organisation"].to_a,
+            organisations: edition.tags["organisations"].to_a + edition.tags["primary_publishing_organisation"].to_a,
+          },
+        }
+      end
```
kevindew: This test looks to have a lot of setup outside of the code that we are testing. I'm guessing all we're really testing is that it doesn't blow up without an attachment. 

Do errors occur if we don't have the tags?

I wonder if we could use a method in this test to setup the Publishing API stub/item as we repeat that block of around ~10-15 lines about 3 times (and then also [modify it](https://github.com/alphagov/content-publisher/blob/9770b5d51c988ea461501827c411a64d6c2ea35d/spec/lib/whitehall_importer/integrity_checker_spec.rb#L50-L52) in other places) - I'd guess we mostly only change one or two aspects so may to pass in an edition and a hash of overrides.

lib/whitehall_importer/integrity_checker/body_check.rb
brucebolt: Done.  This has made the tests much simpler to set up.

lib/whitehall_importer/integrity_checker/body_check.rb
brucebolt: Good idea, have moved it to a new private method.

lib/whitehall_importer/integrity_checker/body_check.rb
brucebolt: I agree, name changed to `BodyTextCheck`.

lib/whitehall_importer/integrity_checker/body_check.rb
brucebolt: Changed name to `sufficiently_similar?`

spec/lib/whitehall_importer/integrity_checker/body_check_spec.rb
brucebolt: Yes, this looks much nicer.  Have updated both examples.

spec/lib/whitehall_importer/integrity_checker/body_check_spec.rb
brucebolt: Added two further tests.

spec/lib/whitehall_importer/integrity_checker_spec.rb
brucebolt: We do need the tags (there's a stubbing inconsistency if we don't).  I've moved the set up of the Publishing API response into a method of its own, which can be modified by passing in a hash.

kevindew: Great - the code has definitely benefitted from adding that method.

I just had a play myself to see if we can remove the orgs set-up and I think it's doable with a couple of tweaks: https://github.com/alphagov/content-publisher/commit/33d67067f354beba7d016adf80116692f63c8d52

brucebolt: Thanks, I've taken the approach you've suggested.

spec/lib/whitehall_importer/integrity_checker/body_text_check_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+    expect(integrity_check.sufficiently_similar?).to be true
+  end
+
+  it "does not return a problem when the body text whitespace does not match" do
+    integrity_check = WhitehallImporter::IntegrityChecker::BodyTextCheck.new("Some text", "Some     text")
+    expect(integrity_check.sufficiently_similar?).to be true
+  end
+
+  it "does not return a problem when the HTML does not match" do
+    integrity_check = WhitehallImporter::IntegrityChecker::BodyTextCheck.new("<b>Some text</b>", "Some text")
```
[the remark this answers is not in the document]
brucebolt: Ah, a rebase failure.  Have added this back in.

spec/lib/whitehall_importer/integrity_checker/body_text_check_spec.rb
```diff
@@ -0,0 +1,46 @@
+RSpec.describe WhitehallImporter::IntegrityChecker::BodyTextCheck do
```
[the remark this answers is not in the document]
brucebolt: Done

spec/lib/whitehall_importer/integrity_checker/body_text_check_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+  it "retuns no problems if the proposed payload matches" do
+    integrity_check = WhitehallImporter::IntegrityChecker::BodyTextCheck.new("Some text", "Some text")
```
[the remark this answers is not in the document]
brucebolt: Ah, yes.  Rebased and now updated to `described_class`.

spec/lib/whitehall_importer/integrity_checker/body_text_check_spec.rb
```diff
@@ -0,0 +1,46 @@
+RSpec.describe WhitehallImporter::IntegrityChecker::BodyTextCheck do
+  it "retuns no problems if the proposed payload matches" do
```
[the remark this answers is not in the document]
brucebolt: Done

PR #1768, opened 2020-02-12
Support multiple levels before selecting a document type

leenagupte:
Trello: https://trello.com/c/R4z8JyhT

# What's changed and why?
Document type selection only supported two-levels of selection, a supertype and a document-type. 

To be able to support a publications, we need to be able to support multiple levels of selection.

To do that, the navigational parts of the document_type, and supertype config have been merged into its own config file. 
Having a single config for document type selection allows us to iterate the feature and support multiple levels of selection, depending on the format.

This should invisible to users as the look and behaviour in the frontend hasn't changed.

Authored with: @ChrisBAshton @GDSNewt


spec/factories/document_type_factory.rb
```diff
@@ -30,7 +30,6 @@
 
     after(:build) do |document_type|
       DocumentType.all << document_type
-      Supertype.all.first.document_types << document_type
```
ChrisBAshton: You have no idea how pleased I am to see this 🎉

config/schemas/document_type_selection.json
```diff
@@ -0,0 +1,80 @@
...
+    "options": {
+      "type": "object",
+      "required": [
+        "label",
+        "type"
+      ],
+      "additionalProperties": false,
+      "properties": {
+        "label": {
+          "$ref": "#/definitions/label"
```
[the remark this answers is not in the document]
leenagupte: This needs to be removed.

spec/models/document_type_selection_spec.rb
```diff
@@ -49,4 +49,83 @@
         .to raise_error(RuntimeError, "Document type selection unknown_document_type not found")
     end
   end
+
+  describe "SelectionOption" do
+    describe ".id" do
+      it "returns id of the option" do
+        option = {
+          "foo" => nil,
```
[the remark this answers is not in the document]
leenagupte: I think we've agreed that I just need to indent `type` under `news`. I'll do that and fix up.

app/interactors/new_document/document_type_selection_interactor.rb
```diff
@@ -0,0 +1,84 @@
...
+           :redirect_url,
+           :document,
+           :needs_refining,
+           :current_selection,
+           :previous_selection,
+           to: :context
+
+
+  def call
+    find_previous_selection
```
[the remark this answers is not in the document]
leenagupte: It would be more correct to rename `current_selection` to `selection`. "page" and "next_page" would probably make more sense as names, but felt wrong.

app/interactors/new_document/document_type_selection_interactor.rb
```diff
@@ -0,0 +1,84 @@
+# frozen_string_literal: true
```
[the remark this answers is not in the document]
leenagupte: I've already rebased. Just need to push up the changes.

app/interactors/new_document/document_type_selection_interactor.rb
```diff
@@ -0,0 +1,84 @@
...
+
+
+  def call
+    find_previous_selection
+    check_for_issues
+    find_current_selection
+    check_for_subtypes
+    find_redirect_url
+
+    if create_document?
```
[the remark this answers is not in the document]
leenagupte: Hmmm. But I don't want to fail. For example, it's valid to not create a document if it's managed elsewhere.

app/interactors/new_document/document_type_selection_interactor.rb
```diff
@@ -0,0 +1,84 @@
...
+           :current_selection,
+           :previous_selection,
+           to: :context
+
+
+  def call
+    find_previous_selection
+    check_for_issues
+    find_current_selection
+    check_for_subtypes
```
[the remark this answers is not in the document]
leenagupte: Do you mean combine `check_for_subtypes` and `find_redirect_url`?

app/interactors/new_document/document_type_selection_interactor.rb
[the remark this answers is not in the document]
benthorner: I agree. I'm saying we should use the interactor `fail` mechanism in the earlier methods to avoid conditionals later on. The example I linked to should help.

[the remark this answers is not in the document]
leenagupte: I don't understand. There are three possible successful outcomes, only one of which necessitates creating a document. So where would you put the fail? 

I didn't put a guard around `redirect_url` and `needs_refining` because at worst they would only store a nil value in the context, not touch the db.

app/interactors/new_document/document_type_selection_interactor.rb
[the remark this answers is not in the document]
leenagupte: I thought about doing something like that. But the `_path` helper methods are not available in the interactors, without importing `url_helpers`, so I decided to let the controller continue to do the routing.

config/schemas/document_type_selection.json
```diff
@@ -0,0 +1,80 @@
...
+      "type": "string",
+      "enum": [
+        "root",
+        "news",
+        "guidance",
+        "transparency",
+        "policy",
+        "document-collection",
+        "corporate-information",
+        "not-sure",
```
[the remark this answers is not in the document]
leenagupte: I'm changing the YAML to match your suggestion of:

```
- id: root
  options:
    - id: news
      type: parent
```
So I'll be able to validate the option ids by the enum as well.

spec/features/workflow/create_document_spec.rb
```diff
@@ -14,22 +14,26 @@ def given_i_am_on_the_home_page
...
+    choose I18n.t("document_type_selections.news_story.label")
     click_on "Continue"
   end
 
   def and_i_fill_in_the_contents
     stub_any_publishing_api_put_content
+    base_path = Edition.last.document_type.path_prefix + "/a-title"
+    content_id = Document.last.content_id
+    stub_publishing_api_has_lookups(base_path => content_id)
+    stub_publishing_api_has_links("content_id" => content_id)
```
kevindew: Nice work fixing these, such a bummer they creep through the stack. I think you might be able to make these a bit simpler.

From looking at the code I think `stub_publishing_api_has_lookups({})` might suffice rather than needing to work out the base path. Which could mean has_links ends up as: `stub_publishing_api_has_links(content_id: Document.last.content_id)`

spec/models/document_type_selection/option_spec.rb
```diff
@@ -0,0 +1,59 @@
+require "json"
```
kevindew: I don't think we need to `require "json"` does it not autoload correctly? I see it's already in spec/models/document_type_spec.rb and now spreading so be good to work out if it's essential.

spec/models/document_type_selection/option_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+
+RSpec.describe DocumentTypeSelection::Option do
+  describe ".managed_elsewhere?" do
```
kevindew: As this is an instance method it should be described with `#managed_elsewhere?` - see http://www.betterspecs.org/#describe

spec/models/document_type_selection/option_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+
+    it "returns the full url if the hostname is provided" do
+      option = {
+        "id" => "foo",
+        "type" => "managed_elsewhere",
+        "hostname" => "whitehall-admin",
+        "path" => "/bar",
+      }
+
+      expect(described_class.new(option).managed_elsewhere_url).to eq("#{whitehall_host}/bar")
```
kevindew: I was wondering where whitehall_host came from as I missed the let. As the variable is only used in this test it might be a simpler read being within the test.

You might also find that cutting plek out and writing the exact URL out gives the easiest read. Eg:
```
expect(described_class.new(option).managed_elsewhere_url)
  .to eq("https://whitehall-admin.test.gov.uk/bar")
```

spec/models/document_type_selection_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+      end
+    end
+
+    it "finds the corresponding object for every string id in the options" do
+      document_type_selections.flat_map { |d| d["options"] }.each do |option|
+        if option["type"] == "parent"
+          expect(described_class.find(option["id"]))
+            .to be_a(described_class)
+        end
+      end
```
kevindew: Rather than using the YAML file should we being using the DocumentTypeSelection class? and I also wonder if we can get this to be a single expectation rather than a loop (I guess this has about 35 expectations currently?

An idea would be 

```
all_options = described_class.all.flat_map(&:options)
expect { all_options.each { |option| described_class.find(option.id) } }
  .not_to raise_error(DocumentTypeSelection::NotFoundError)
```

spec/models/document_type_selection_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+    end
+
+    it "only allows unique document type selections" do
+      ids = document_type_selections.pluck("id")
+      expect(ids.size).to eq(ids.uniq.size)
+    end
```
kevindew: This looks good - I wonder if we can do this instead through the schema but again happy to check post merge.

spec/models/document_type_selection_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+  end
+
+  describe ".all" do
+    it "creates a DocumentTypeSelection for each one in the YAML" do
+      expect(described_class.all.count).to eq(document_type_selections.count)
+    end
+  end
+
+  describe ".find" do
```
kevindew: Similar point about instance methods prefixed with #

spec/models/document_type_selection_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+    it "returns the hash of the corresponding DocumentTypeSelection" do
```
kevindew: hash seems to be the wrong word here - I wonder if this test description was wrote before some changes.

spec/models/document_type_selection_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+      expect(described_class.find("news")).to be_a(described_class)
```
kevindew: Should we also assert that it's the right item?

spec/models/document_type_selection_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+    end
+
+    it "raises a RuntimeError when there is no corresponding entry for the id" do
```
kevindew: This (and the expectation) should be updated for the custom exception.

spec/models/document_type_selection_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+        .to raise_error(RuntimeError, "Document type selection unknown_document_type not found")
+    end
+  end
+
+  describe ".parent" do
+    it "returns nil if we pass it 'root'" do
+      expect(described_class.find("root").parent).to be_nil
+    end
+
+    it "returns a DocumentTypeSelection for the parent if it exists" do
```
kevindew: This description doesn't match that the test returns a string, but my earlier suggestion does suggest changing the type to be a DocumentTypeSelection

spec/models/document_type_selection_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+      expect(described_class.find("news").parent).to eq("root")
+    end
+  end
+
+  describe ".find_option" do
+    it "returns the requested option" do
+      option = described_class.find("news").find_option("news_story")
+      expect(option.id).to eq("news_story")
+      expect(option.type).to eq("document_type")
```
kevindew: Not sure you need this second expectation as that seems beyond what is being tested.

spec/requests/new_document_spec.rb
```diff
@@ -1,64 +1,58 @@
...
-    end
-
-    it "redirects to guidance page when \"I'm not sure\" option is selected" do
-      supertype = Supertype.all.find { |type| type.id == "not-sure" }
-      get choose_document_type_path, params: { supertype: supertype.id }
-      expect(response).to redirect_to(guidance_url)
+    it "shows the page for the selected document type" do
+      get new_document_path, params: { type: "news" }
+      expect(response).to have_http_status(:ok)
+      expect(response.body).to have_content("News story")
```
kevindew: I think this text is defined in an i18n file so consistency would dictate we assert based on the I18n key rather than the text directly.

spec/requests/new_document_spec.rb
```diff
@@ -1,64 +1,58 @@
...
+      expect(response).to redirect_to(guidance_url)
+    end
+
+    it "asks the user to refine their selection when the document type has subtypes" do
+      post new_document_path, params: { type: "root", selected_option_id: "news" }
+
+      expect(response).to redirect_to(new_document_path(type: "news"))
+      follow_redirect!
+      expect(response.body).to have_content("News story")
     end
```
kevindew: I think rather than having 3 tests here for the different types of redirect we only need to have one since the controller we're testing only has logic for one type of redirect_url and we're testing how those are set in the interactor.  I'm guessing though that this came about when we had multiple conditions in the controller when it definitely made sense to be here. 

But yeah I think now it probably makes most sense for it to be "it redirects users to a destination on a successful request"

Edit: If you go with the suggestion I've made about dealing with redirect destination in controller we'd want to keep the 3 test approach.

app/interactors/new_document/select_interactor.rb
```diff
@@ -0,0 +1,71 @@
...
+  delegate :params,
+           :user,
+           :redirect_url,
+           :document_type_selection,
+           to: :context
+
+  def call
+    find_document_type_selection
+    check_for_issues
+    find_redirect_url
```
kevindew: I've been having a bit of a think about what my thoughts are on what should go in the call method.

I think it's a reasonable shout to not use context.fail as I agree with your stance that this isn't a failing scenario.
I did think it was a pretty reasonable shout about redirect_url though I am wondering if it might actually be fine to not have to do that (and could mean losing the helper include which feels like a bit of a smell)

I also think we don't say quite enough in the call method, so my suggestion would be that we do something like:

```
find_selection
check_for_issues
return unless selected_option.document_type?
create_document
create_timeline_entry
```

I know that reintroduces the conditional that was a point of contention but I think it's ok as it is a conditional path after all. 

I guess that'd mean the controller ends up with:

```
    result = NewDocument::SelectInteractor.call(params: params, user: current_user)
    issues, document, document_type_selection, selected_option =
      result.to_h.values_at(:issues, :document, :document_type_selection, :selected_option)

    if issues
      flash.now["requirements"] = { "items" => issues.items }
      render :show,
             assigns: { issues: issues, document_type_selection: document_type_selection },
             status: :unprocessable_entity
    else
      destination = if document
                      content_path(document)
                    elsif selected_option.managed_elsewhere?
                      selected_option.managed_elsewhere_url
                    else
                      new_document_path(type: selected_option.id)
                    end
      redirect_to destination
    end
```
Which seems good to me - apologies if this is eerily similar to how you had this before.

Minor suggestions as part of that would be that `find_selection` comprises setting `context.document_type_selection` and `context.selected_option` (it feels a little unusual to me that selected option is an instance variable rather than on context); You would need to set a context.document to have that managed; `select_option.document_type?` is suggested instead of `create_document?` to avoid `create_document? create_document` and consistency `with managed_elsewhere?`.

app/interactors/new_document/select_interactor.rb
```diff
@@ -0,0 +1,71 @@
...
+  end
+
+  def create_timeline_entry(document)
+    TimelineEntry.create_for_status_change(entry_type: :created,
+                                           status: document.current_edition.status)
+  end
+
+  def default_tags
+    user.organisation_content_id ? { primary_publishing_organisation: [user.organisation_content_id] } : {}
+  end
```
kevindew: Minor but I'd probably move these tags into the create_document method since it's already so short

spec/models/document_type_selection_spec.rb
leenagupte: At the moment there should only be 4 expectations.

This test was written before the options were modelled, so `option.id` wouldn't have been available.

spec/models/document_type_selection_spec.rb
leenagupte: This isn't an instance method.

spec/models/document_type_selection_spec.rb
leenagupte: Oops! Good spot!

spec/models/document_type_selection/option_spec.rb
leenagupte: I like Plek because I don't think I should have to find out or care what the url is. 
I'll move assignment of `whitehall_host` into the test though.

app/interactors/new_document/select_interactor.rb
leenagupte: That's pretty much what the controller looked like before. I'll remove the last commit and revert this change.

app/interactors/new_document/select_interactor.rb
leenagupte: I'm leaving as is, otherwise the line gets long:

```
default_tags = user.organisation_content_id ? { primary_publishing_organisation: [user.organisation_content_id] } : {}
```

Or you end up with an even longer conditional"
```
default_tags = if user.organisation_content_id
                           primary_publishing_organisation: [user.organisation_content_id]
                         else
                           {}
                         end
```

spec/requests/new_document_spec.rb
leenagupte: I'm going with the suggestion about dealing with the redirect in the controller. It was my preferred option as it felt wrong to have to include `url_helpers` in the interactor.

app/views/new_document/show.html.erb
```diff
@@ -0,0 +1,30 @@
+<% content_for :browser_title, t("document_type_selections.#{@document_type_selection.id}.label") %>
+<% content_for :back_link, render_back_link(href: new_document_path(type: @document_type_selection.parent)) %>
```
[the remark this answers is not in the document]
leenagupte: As suggested in a previous comment, I'm adding the safe navigation operator.

spec/models/document_type_selection_spec.rb
leenagupte: Also we'd need to put a guard around the `expect` or pluck out the options of type `document_type_selection` because they are the only ones that should pass this test.

leenagupte: It would work if the first line changed to:
```
all_options = described_class.all.flat_map(&:options).select(&:document_type_selection?)
```

leenagupte:
@kevindew @ChrisBAshton Thanks for your comments. I've fixed nearly all of them, and responded to the others.

app/views/new_document/show.html.erb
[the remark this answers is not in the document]
kevindew: Doesn't that mean we get a back link with an empty url though for root? 

I wonder if we actually want to have a back link to documents for root, I had assumed we'd want no link but that might be inconsistent with the rest of the app.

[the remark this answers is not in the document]
leenagupte: I've added a conditional to the back link so that it renders the link to the index page if we're on the "root" selection page.

spec/models/document_type_selection_spec.rb
kevindew: Ooops - methods below (parent and find_option)

app/views/new_document/show.html.erb
```diff
@@ -0,0 +1,30 @@
+<% content_for :browser_title, t("document_type_selections.#{@document_type_selection.id}.label") %>
+<% content_for :back_link, render_back_link(href: @document_type_selection.parent.present? ? new_document_path(type:  @document_type_selection.parent.id)  : root_path) %>
```
kevindew: Could separate this onto two lines:

```
<% back_href = @document_type_selection.parent ? new_document_path(type:  @document_type_selection.parent.id)  : root_path %>
<% content_for :back_link, render_back_link(href: back_href) %>
```

I don't think you need the present.

app/models/document_type_selection.rb
```diff
@@ -0,0 +1,37 @@
...
+      end
+    end
+  end
+
+  def parent
+    self.class.all.find do |document_type_selection|
+      document_type_selection.options.map(&:id).include?(id)
+    end
+  rescue NotFoundError
+    nil
```
kevindew: I don't think this method can raise an error (as it doesn't use the find defined above)

leenagupte:
@kevindew Thanks, I've addressed your comments and fixed up.

PR #1769, opened 2020-02-12
Disable rubocop FrozenStringLiteral cop

benthorner:
This was added in preparation for Ruby 3, which is no longer going
to enforce frozen string literals.

https://github.com/rubocop-hq/rubocop/issues/7197


benthorner:
@kevindew yeah, it's all benign - just a few `.freeze` calls added to constants here and there.

PR #1772, opened 2020-02-12
Return document import from WhitehallImporter::Sync

emmabeynon:
We expect this class to return a document import so that in the Whitehall
Document Import Job we can check that if its Whitehall Migration has finished.

### DIFF
--- a/lib/whitehall_importer/sync.rb
+++ b/lib/whitehall_importer/sync.rb
@@ -23,6 +23,7 @@ module WhitehallImporter
       MigrateAssets.call(whitehall_import)
 
       whitehall_import.update!(state: "completed")
+      whitehall_import
     end
   end
 end
--- a/spec/lib/whitehall_importer/sync_spec.rb
+++ b/spec/lib/whitehall_importer/sync_spec.rb
@@ -42,8 +42,8 @@ RSpec.describe WhitehallImporter::Sync do
     end
 
     it "returns a completed WhitehallMigration::DocumentImport" do
-      described_class.call(whitehall_migration_document_import)
-      expect(whitehall_migration_document_import).to be_completed
+      document_import = described_class.call(whitehall_migration_document_import)
+      expect(document_import).to be_completed
     end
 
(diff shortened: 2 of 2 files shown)


kevindew:
@emmabeynon can you explain a bit more why this is needed. I'm a bit confused because the variable we're returning is the same one used as input, so I'm not sure why a caller would need to assign to the output?

emmabeynon:
@kevindew Sure, I found that without returning the document import from Sync, when we call `document_import.whitehall_migration` we get `#<NoMethodError: undefined method `whitehall_migration' for true:TrueClass>`.  This error was caught by the job, but had the unfortunate side effect of the wrong state being assigned (because [here](https://github.com/alphagov/content-publisher/blob/master/app/jobs/whitehall_document_import_job.rb#L39)), so ended up being a bit of wild goose chase figuring out where things went wrong!

kevindew:
> @kevindew Sure, I found that without returning the document import from Sync, when we call `document_import.whitehall_migration` we get `#<NoMethodError: undefined method `whitehall_migration' for true:TrueClass>`. This error was caught by the job, but had the unfortunate side effect of the wrong state being assigned (because [here](https://github.com/alphagov/content-publisher/blob/master/app/jobs/whitehall_document_import_job.rb#L39)), so ended up being a bit of wild goose chase figuring out where things went wrong!

Ah ha Gotcha! Sure can see why this makes sense.

I'd suggest though that we actually resolve this nicer by removing code. Looking at: https://github.com/alphagov/content-publisher/blob/98b5b056019e8ca12f8e6c56929bc0df5447e30a/app/jobs/whitehall_document_import_job.rb#L14-L15
we don't actually need to be doing the document_import assigns and can simplify the code without them.

Then once we did that we could remove https://github.com/alphagov/content-publisher/blob/98b5b056019e8ca12f8e6c56929bc0df5447e30a/lib/whitehall_importer/import.rb#L44

emmabeynon:
@kevindew Makes sense.  I'm closing this in favour of https://github.com/alphagov/content-publisher/pull/1775

PR #1777, opened 2020-02-13
Add published at data

1pretz1:
First PR to populate `published_at` column for editions when publishing an edition, we also move when publishing times are set to be before the syncing with publishing-api.

Trello:
https://trello.com/c/ShhyTqqU/1459-create-store-and-backfill-publishedat-on-editions

### DIFF
--- a/app/services/publish_draft_edition_service.rb
+++ b/app/services/publish_draft_edition_service.rb
@@ -11,10 +11,10 @@ class PublishDraftEditionService < ApplicationService
 
     publish_assets
     associate_with_government
+    set_published_at
     publish_current_edition
     supersede_live_edition
     set_new_live_edition
-    set_first_published_at
   rescue GdsApi::BaseError => e
     GovukError.notify(e)
     raise
@@ -73,9 +73,12 @@ private
     document.reload_live_edition
   end
 
-  def set_first_published_at
+  def set_published_at
+    current_time = Time.current
+    edition.published_at = current_time
+
     return if document.first_published_at
 
-    document.update!(first_published_at: Time.current)
+    document.update!(first_published_at: current_time)
   end
 end
--- a/spec/services/publish_draft_edition_service_spec.rb
+++ b/spec/services/publish_draft_edition_service_spec.rb
@@ -24,6 +24,13 @@ RSpec.describe PublishDraftEditionService do
         expect(edition).to be_live
       end
 
+      it "sets the document first publishing time" do
+        freeze_time do
+          expect { described_class.call(edition, user, with_review: true) }
+            .to change(edition.document, :first_published_at).to(Time.current)
+        end
+      end
+
       it "can specify if edition is reviewed" do
         described_class.call(edition, user, with_review: false)
         expect(edition).to be_published_but_needs_2i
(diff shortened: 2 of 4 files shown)


spec/services/publish_draft_edition_service_spec.rb
```diff
@@ -31,15 +37,23 @@
...
+      let(:current_edition) { document.current_edition }
+
       it "supersedes the live edition" do
-        document = create(:document, :with_current_and_live_editions)
-        current_edition = document.current_edition
         live_edition = document.live_edition
 
         described_class.call(current_edition, user, with_review: true)
         expect(document.live_edition).to eq(current_edition)
         expect(live_edition).to be_superseded
```
kevindew: while you're in here it's a bit weird that we set a variable of live_edition on line 44 and then don't use it on 47. I'd suggest removing line 44 and switch line 48 to use document.live_edition too.

Do we need to set current_edition let? if not I'd suggest we could reference that off document for consistency with document.live_edition

spec/services/publish_draft_edition_service_spec.rb
```diff
@@ -31,15 +37,23 @@
...
       end
+
+      it "doesn't overwrite when the document was first published" do
+        first_published_at = document.reload.first_published_at
+        described_class.call(current_edition, user, with_review: true)
+
+        expect(document.first_published_at).to eq(first_published_at)
```
kevindew: I think you can get around the reload by doing 

```
        expect { described_class.call(current_edition, user, with_review: true) }
          .not_to change(document, :first_published_at)
```

app/services/publish_draft_edition_service.rb
```diff
@@ -70,7 +70,9 @@ def set_new_live_edition
     edition.save!
   end
 
-  def set_first_published_at
+  def set_published_at
```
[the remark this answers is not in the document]
1pretz1: Good point, done!

spec/services/publish_draft_edition_service_spec.rb
1pretz1: The reason we're setting live_edition in a variable beforehand is because after the class is run the live edition changes to the current edition, meaning if we referenced `document.live_edition` in the expectation -`expect(live_edition).to be_superseded` then it would use the newly published edition (current_edition) where as we want to assert the previously published edition if that makes sense?

> Do we need to set current_edition let? if not I'd suggest we could reference that off document for consistency with document.live_edition

Nope, current_edition now references from document!

spec/services/publish_draft_edition_service_spec.rb
1pretz1: Good call, that is a lot nicer!

1pretz1:
Thanks for the review Kev, it's ready to be looked at again.

spec/services/publish_draft_edition_service_spec.rb
```diff
@@ -16,12 +16,18 @@
...
+          described_class.call(edition, user, with_review: true)
+        end
 
-        described_class.call(edition, user, with_review: true)
         expect(publish_request).to have_been_requested
         expect(edition.document.live_edition).to eq(edition)
         expect(edition).to be_published
         expect(edition).to be_live
+        expect(edition.published_at).to eq(time)
+        expect(edition.document.first_published_at).to eq(time)
```
kevindew: Sorry to have not brought this up in the last review but I don't think this is the right place for this test as we've got a lot of expectations here and the "it" doesn't really capture this.

I'd suggest you have a separate it for example:

```
it "sets the document first publishing time" do
  freeze_time do
    expect { described_class.call(document.current_edition, user, with_review: true) }
      .to change(document, :first_published_at).to(Time.current)
  end
end
```

Then it's probably worth adding a similar expectation outside of any contexts (like `it "calls the PublishAssetsService"` for instance) which asserts that edition has a published_at timestamp set to current time.

app/services/publish_draft_edition_service.rb
```diff
@@ -73,9 +73,12 @@ def set_new_live_edition
     document.reload_live_edition
   end
 
-  def set_first_published_at
+  def set_published_at
+    current_time = Time.current
+    edition.update!(published_at: current_time)
```
kevindew: Rather than running an `update!` that calls an instant query we can just assign the attribute with `edition.published_at = current_time` and save an unnecessary query (we're going to run a save later.

spec/services/publish_draft_edition_service_spec.rb
1pretz1: Sure, I've added a separate test which tests the editions published_at timestamp instead of piggy backing on another test!

app/services/publish_draft_edition_service.rb
1pretz1: Good point, done!

PR #1778, opened 2020-02-13
Backfill published_at column for edition

1pretz1:
This backfills the published_at column for previously published
editions, we use when the published status was created as a means to
work this time out as this should reflect the actual publishing time.

This relies on:
https://github.com/alphagov/content-publisher/pull/1777

Trello:
https://trello.com/c/ShhyTqqU/1459-create-store-and-backfill-publishedat-on-editions

### DIFF
--- a/db/migrate/20200218161942_backfill_published_at_for_editions.rb
+++ b/db/migrate/20200218161942_backfill_published_at_for_editions.rb
@@ -1,0 +1,7 @@
+class BackfillPublishedAtForEditions < ActiveRecord::Migration[6.0]
+  def up
+    Status.where(state: %w[published published_but_needs_2i])
+          .order(:created_at)
+          .find_each { |s| s.edition.update!(published_at: s.created_at) unless s.edition.published_at }
+  end
+end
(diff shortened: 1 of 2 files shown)


db/migrate/20200213145521_backfill_published_at_for_editions.rb
```diff
@@ -0,0 +1,6 @@
+class BackfillPublishedAtForEditions < ActiveRecord::Migration[6.0]
+  def up
+    Status.where(state: %w[published published_but_needs_2i])
+          .find_each { |s| s.edition.update!(published_at: s.created_at) unless s.edition.published_at }
```
kevindew: This should have ordering on it (as per the trello card) to ensure that it only catches the first publish status of an Edition - they can have multiple based on either clearing a withdrawal or an unreviewed piece of content gaining a review.

1pretz1: Ah ha, you're absolutely correct! Sorry I went rogue and thought the ordering was superfluous so removed it, have reverted it back to your suggestion exactly.

1pretz1:
Ready for review no.2 @kevindew

PR #1780, opened 2020-02-14
Don't mark assets as absent in an ensure

kevindew:
This is a little bit of a cleanup on the `AssetCleanupJob` where we marked assets as absent in an ensure block, however this would get called for all types of exceptions.

It also removes a bunch of duplication in the tests and makes their expectations a bit clearer to read.

### DIFF
--- a/app/jobs/asset_cleanup_job.rb
+++ b/app/jobs/asset_cleanup_job.rb
@@ -37,10 +37,12 @@ private
   end
 
   def delete_asset(asset)
-    GdsApi.asset_manager.delete_asset(asset.asset_manager_id)
-  rescue GdsApi::HTTPNotFound
-    logger.warn("No asset to delete for id #{asset.asset_manager_id}")
-  ensure
+    begin
+      GdsApi.asset_manager.delete_asset(asset.asset_manager_id)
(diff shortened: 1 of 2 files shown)


app/jobs/asset_cleanup_job.rb
```diff
@@ -37,10 +37,12 @@ def clean_file_attachment_assets
...
-    GdsApi.asset_manager.delete_asset(asset.asset_manager_id)
-  rescue GdsApi::HTTPNotFound
-    logger.warn("No asset to delete for id #{asset.asset_manager_id}")
-  ensure
+    begin
+      GdsApi.asset_manager.delete_asset(asset.asset_manager_id)
+    rescue GdsApi::HTTPNotFound
+      logger.warn("No asset to delete for id #{asset.asset_manager_id}")
+    end
+
```
leenagupte: Nitpick, I think commit message is missing a comma after "ensure".

kevindew: Thank you - I've reworded the commit and will merge when it's green.

leenagupte: The commit message is much better! Thanks!

PR #1782, opened 2020-02-14
Basic UI for viewing Whitehall migration status

GDSNewt:
Have set up the following for the migration UI:

- routes
- controller
- request specs
- placeholder views which display some migration data

This PR is for the initial setup, the next PR will contain the full data that we wish to display.

Example:
Screenshot 2020-02-21 at 11 54 27

Trello card: https://trello.com/c/yFGFseiN

### DIFF
--- a/app/controllers/whitehall_migration_controller.rb
+++ b/app/controllers/whitehall_migration_controller.rb
@@ -1,0 +1,19 @@
+class WhitehallMigrationController < ApplicationController
+  before_action { authorise_user!(User::DEBUG_PERMISSION) }
+
+  def show
+    @whitehall_migration = WhitehallMigration.find(params[:migration_id])
+  end
+
+  def documents
+    whitehall_migration = WhitehallMigration.find(params[:migration_id])
+    @documents = whitehall_migration.document_imports
+  end
+
+  def document
+    @document = WhitehallMigration::DocumentImport.find_by!(
+      id: params[:document_import_id],
+      whitehall_migration_id: params[:migration_id],
+    )
+  end
+end
--- a/app/views/whitehall_migration/show.html.erb
+++ b/app/views/whitehall_migration/show.html.erb
@@ -1,0 +1,14 @@
+<h1>Whitehall Migration: <%= @whitehall_migration.id %> </h1>
+<ul>
+    <li>Time started: <%= @whitehall_migration.created_at.strftime("At %T on %d/%m/%Y") %></li>
+    <li>Time finished: <%= @whitehall_migration.finished_at.nil? ? "Unfinished" : @whitehall_migration.finished_at %></li>
+    <li>Number of documents: <%= @whitehall_migration.document_imports.count %></li>
+</ul>
+
+<h3>Document Import Status</h3>
+
+<ul>
+    <% @whitehall_migration.document_imports.states.each_value do |state| %>
+        <li><%= "#{state}:" %> <%= @whitehall_migration.document_imports.where(state: "#{state}").count%></li>
+    <% end %>
+</ul>
--- a/lib/tasks/import.rake
+++ b/lib/tasks/import.rake
@@ -1,31 +1,37 @@
-# frozen_string_literal: true
-
 require "gds_api/whitehall_export"
 
 namespace :import do
   desc "Import all documents matching an organisation, document type and optional list of document subtypes from Whitehall Publisher, e.g. import:whitehall_migration[\"cabinet-office\",\"news_article\",\"news_story,press_release\"]"
   task :whitehall_migration, %i[organisation_slug document_type document_subtypes] => :environment do |_, args|
+    include Rails.application.routes.url_helpers
+
     organisation_content_id = GdsApi.publishing_api.lookup_content_id(
       base_path: "/government/organisations/#{args.organisation_slug}",
-    )
+     )
     document_subtypes = args.document_subtypes ? args.document_subtypes.split(",") : []
     whitehall_migration = WhitehallImporter::CreateMigration.call(
       organisation_content_id, args.document_type, document_subtypes
     )
 
-    documents_to_import = WhitehallMigration::DocumentImport.where(whitehall_migration_id: whitehall_migration.id).count
+    documents_to_import = WhitehallMigration::DocumentImport.where(whitehall_migration: whitehall_migration).count
     puts "Identified #{documents_to_import} documents to import"
+
+    puts whitehall_migration_url(whitehall_migration, host: Plek.new.external_url_for("content-publisher"))
   end
 
   desc "Import a single document from Whitehall Publisher using Whitehall's internal document ID e.g. import:whitehall_document[123]"
   task :whitehall_document, [:document_id] => :environment do |_, args|
+    include Rails.application.routes.url_helpers
+
     whitehall_import = WhitehallMigration::DocumentImport.create!(
       whitehall_document_id: args.document_id,
       whitehall_migration: WhitehallMigration.create!,
       state: "pending",
-    )
+     )
 
     WhitehallDocumentImportJob.perform_later(whitehall_import)
     puts "Added whitehall document with ID:#{args.document_id} to the import queue"
+
+    puts whitehall_migration_url(whitehall_import.whitehall_migration_id, host: Plek.new.external_url_for("content-publisher"))
   end
 end
--- a/spec/lib/tasks/import_spec.rb
+++ b/spec/lib/tasks/import_spec.rb
@@ -2,7 +2,7 @@ RSpec.describe "Import tasks" do
   include ActiveJob::TestHelper
 
   describe "import:whitehall_migration" do
-    let(:whitehall_migration_document_import) { build(:whitehall_migration_document_import) }
+    let(:whitehall_migration) { create(:whitehall_migration) }
 
     before do
       allow($stdout).to receive(:puts)
@@ -11,7 +11,7 @@ RSpec.describe "Import tasks" do
         "/government/organisations/cabinet-office" => "96ae61d6-c2a1-48cb-8e67-da9d105ae381",
       )
       allow(WhitehallImporter::CreateMigration).to receive(:call)
-                                               .and_return(whitehall_migration_document_import)
+                                               .and_return(whitehall_migration)
     end
 
     it "calls WhitehallImporter::create_migration with correct arguments when subtype is specified" do
--- a/spec/requests/whitehall_migration_spec.rb
+++ b/spec/requests/whitehall_migration_spec.rb
@@ -1,0 +1,53 @@
+RSpec.describe "Whitehall Migration" do
+  let(:whitehall_migration) { create(:whitehall_migration) }
+  let(:document_import) { create(:whitehall_migration_document_import, whitehall_migration: whitehall_migration) }
+  let(:debug_permission_user) { create(:user, permissions: [User::DEBUG_PERMISSION]) }
+
+  it_behaves_like "requests that return status",
+                  "when a user without debug permissions looks at a whitehall migration",
+                  status: :forbidden,
+                  routes: { whitehall_migration_path: %i[get],
+                          whitehall_migration_documents_path: %i[get] } do
+    before { login_as(create(:user)) }
+
+    let(:route_params) { [whitehall_migration] }
+  end
+
+  it_behaves_like "requests that return status",
+                  "when a user without debug permissions looks at a whitehall migration document",
+                  status: :forbidden,
+                  routes: { whitehall_migration_document_path: %i[get] } do
+    before { login_as(create(:user)) }
+
+    let(:route_params) do
+      [whitehall_migration, document_import]
+    end
+  end
+
+  describe "GET /whitehall-migration/:migration_id" do
+    it "returns success" do
+      login_as(debug_permission_user)
+      get whitehall_migration_path(whitehall_migration)
+
+      expect(response).to have_http_status(:ok)
+    end
+  end
+
+  describe "GET /whitehall-migration/:migration_id/documents" do
+    it "returns success" do
+      login_as(debug_permission_user)
+      get whitehall_migration_documents_path(whitehall_migration)
+
+      expect(response).to have_http_status(:ok)
+    end
+  end
+
+  describe "GET /whitehall-migration/:migration_id/documents/:document_import_id" do
+    it "returns success" do
+      login_as(debug_permission_user)
+      get whitehall_migration_document_path(whitehall_migration, document_import)
+
+      expect(response).to have_http_status(:ok)
+    end
+  end
+end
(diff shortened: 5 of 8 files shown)


brucebolt:
This is looking good.

There's a few things I've been manually querying that could be useful in the UI:
- A tally of documents in each status per migration run
- A tally of error log messages per migration run
- A tally of integrity check errors per migration run
- Some more information about the document (e.g. a link to Whitehall, or at least the live item or content item)

I'm sure there's more things I've been querying and missed off this list.  Hopefully I'll remember them next time we run a migration.

lib/tasks/import.rake
```diff
@@ -15,6 +15,7 @@ namespace :import do
 
     documents_to_import = WhitehallMigration::DocumentImport.where(whitehall_migration_id: whitehall_migration.id).count
     puts "Identified #{documents_to_import} documents to import"
+    puts "#{Plek.new.external_url_for("content-publisher")}/whitehall-migration/#{whitehall_migration.id}"
```
kevindew: Cool! - It'll be neat having a link to click after starting a migration.

We can use the Rails route helpers rather than writing our own route. This can be done by importing Rails.application.routes.url_helpers and then using a method like `whitehall_migration_url(whitehall_migration, host: Plek.new.external_url_for("content-publisher"))`

spec/requests/whitehall_migration_spec.rb
```diff
@@ -0,0 +1,61 @@
...
+
+    it "returns a forbidden status for user without debug permission" do
+       login_as(create(:user))
+       document_import = create(:whitehall_migration_document_import, whitehall_migration_id: whitehall_migration.id)
+       get whitehall_migration_document_path(whitehall_migration.id, document_import.id)
+
+       expect(response).to have_http_status(:forbidden)
+       expect(response.body).to have_content("Sorry, you don't seem to have the #{User::DEBUG_PERMISSION} permission for this app")
+    end
+  end
```
kevindew: Nice job putting together these request tests. 

I think we can use an existing helper to handle the permission cases more concisely: https://github.com/alphagov/content-publisher/blob/6506cabb5109eabbc14a4048340be27891044823/spec/requests/file_attachments_spec.rb#L17-L26

You could do something like: 

```
  it_behaves_like "requests that return status",
                  "when a user without debug permissions looks at a whitehall migration",
                  status: :forbidden,
                  routes: { whitehall_migration: %i[get],
                            whitehall_migration_documents: %i[get] } do
    before { login_as(create(:user)) }
    let(:route_params) { [whitehall_migration] }
  end
  
  it_behaves_like "requests that return status",
                  "when a user without debug permissions looks at a whitehall migration document",
                  status: :forbidden,
                  routes: { whitehall_migration_document: %i[get] } do
    before { login_as(create(:user)) }
    let(:route_params) do
      document = create(:whitehall_migration_document_import, whitehall_migration: whitehall_migration)
      [whitehall_migration, document]
    end
  end  
```

spec/requests/whitehall_migration_spec.rb
```diff
@@ -0,0 +1,61 @@
...
+
+  describe "GET /whitehall-migration/:migration_id/documents/:document_import_id" do
+    it "returns success" do
+      document_import = create(:whitehall_migration_document_import, whitehall_migration_id: whitehall_migration.id)
+      get whitehall_migration_document_path(whitehall_migration.id, document_import.id)
```
kevindew: you should be ok to pass these in as objects rather than needing to call `.id` on each of them - same on other routes.

spec/requests/whitehall_migration_spec.rb
```diff
@@ -0,0 +1,61 @@
...
+       get whitehall_migration_documents_path(whitehall_migration.id)
+
+       expect(response).to have_http_status(:forbidden)
+       expect(response.body).to have_content("Sorry, you don't seem to have the #{User::DEBUG_PERMISSION} permission for this app")
+    end
+  end
+
+  describe "GET /whitehall-migration/:migration_id/documents/:document_import_id" do
+    it "returns success" do
+      document_import = create(:whitehall_migration_document_import, whitehall_migration_id: whitehall_migration.id)
```
kevindew: Similarly here you can do this factory with `document_import = create(:whitehall_migration_document_import, whitehall_migration: whitehall_migration)`

Though since the factory [can create a whitehall migration](https://github.com/alphagov/content-publisher/blob/6506cabb5109eabbc14a4048340be27891044823/spec/factories/whitehall_migration_document_import_factory.rb#L8)

you could change this test to be:
```
document_import = create(:whitehall_migration_document_import)
get whitehall_migration_document_path(document_import.whitehall_migration, document_import)
```

app/views/whitehall_migration/show.html.erb
```diff
@@ -0,0 +1,14 @@
+<h1>Whitehall Migration numero <%= @whitehall_migration.id %> </h1>
+<li>Time started: <%= @whitehall_migration.created_at.strftime("At %T on %m/%d/%Y") %></li>
+<li>Time finished: <%= @whitehall_migration.finished_at.nil? ? "UNFINISHED!!!!!" : @whitehall_migration.finished_at %></li>
+<li>Number of documents: <%= @whitehall_migration.document_imports.count %></li>
+
+<h3>Document Import status</h3>
+
+<% @states.each do |state| %>
```
kevindew: Rather than needing to pass in a variable in for this you can pull it off the document_imports object with `whitehall_migration.document_imports.states` eg:

```
irb(main):009:0> wm.document_imports.states
=> {"pending"=>"pending", "imported"=>"imported", "import_aborted"=>"import_aborted", "import_failed"=>"import_failed", "sync_failed"=>"sync failed", "completed"=>"completed"}
```

app/controllers/whitehall_migration_controller.rb
```diff
@@ -0,0 +1,24 @@
+class WhitehallMigrationController < ApplicationController
+  before_action { authorise_user!(User::DEBUG_PERMISSION) }
+
+  def show
+    @whitehall_migration = whitehall_migration
```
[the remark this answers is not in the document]
Rosa-Fox: 👍

spec/requests/whitehall_migration_spec.rb
Rosa-Fox: 👍

app/views/whitehall_migration/show.html.erb
Rosa-Fox: 👍

app/controllers/whitehall_migration_controller.rb
```diff
@@ -0,0 +1,24 @@
...
+  end
+
+  def documents
+    @documents = whitehall_migration.document_imports
+  end
+
+  def document
+    @document = whitehall_migration
+                .document_imports
+                .find(params[:document_import_id])
```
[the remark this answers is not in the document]
Rosa-Fox: 👍

app/views/whitehall_migration/show.html.erb
```diff
@@ -0,0 +1,10 @@
+<h1>Whitehall Migration numero <%= @whitehall_migration.id %> </h1>
+<li>Time started: <%= @whitehall_migration.created_at.strftime("At %T on %m/%d/%Y") %></li>
+<li>Time finished: <%= @whitehall_migration.finished_at.nil? ? "UNFINISHED!!!!!" : @whitehall_migration.finished_at %></li>
+<li>Number of documents: <%= @whitehall_migration.document_imports.count %></li>
```
kevindew: This list seems to be missing the `<ul>` elements - same for the list below.

spec/requests/whitehall_migration_spec.rb
```diff
@@ -0,0 +1,46 @@
+RSpec.describe "Whitehall Migration" do
+  let(:whitehall_migration) { create(:whitehall_migration) }
+  let(:document) { create(:whitehall_migration_document_import, whitehall_migration: whitehall_migration) }
```
kevindew: It'd be good to call this a document_import rather than a document since document is already a word with quite a lot of meaning in Content Publisher.

spec/requests/whitehall_migration_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+  it_behaves_like "requests that return status",
+                  "when a user with debug permissions looks at a whitehall migration document",
+                  status: :success,
+                  routes: { whitehall_migration_document_path: %i[get] } do
+    before { login_as(create(:user, permissions: [User::DEBUG_PERMISSION])) }
+
+    let(:route_params) do
+      [whitehall_migration, document]
+    end
+  end
```
kevindew: We don't tend to have any precedent for using these `it_behaves_like "requests that return status",` for successful scenarios. 

It'd be more consistent with the app if we had regular request tests for success scenario. We also have precedent of asserting on status ok rather than success - for example: https://github.com/alphagov/content-publisher/blob/0ba258d25eba53016416e41e28c347dee96433e0/spec/requests/access_limit_spec.rb#L30-L34

GDSNewt:
> This is looking good - it looks like the git commit history needs cleaning up still and we have a few instances of throwaway text (Documents!, numero, UNFINISHED!!!!) that'd be good to not have in the permanent history of the app.

We have crushed the commits and taken care of the throwaway text

spec/requests/whitehall_migration_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+      get whitehall_migration_path(whitehall_migration)
+
+      expect(response).to have_http_status(:ok)
+    end
+  end
+
+  describe "GET /whitehall-migration/:migration_id/documents" do
+    it "returns success" do
+      login_as(debug_permission_user)
+      create(:whitehall_migration_document_import, whitehall_migration_id: whitehall_migration.id)
```
kevindew: I don't think we need this as we can still show the document lists without any documents.

spec/requests/whitehall_migration_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+      get whitehall_migration_documents_path(whitehall_migration.id)
```
kevindew: We can just pass in` whitehall_migration` here without the need for an id call

spec/requests/whitehall_migration_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+
+      expect(response).to have_http_status(:ok)
+    end
+  end
+
+  describe "GET /whitehall-migration/:migration_id/documents/:document_import_id" do
+    it "returns success" do
+      login_as(debug_permission_user)
+      document_import = create(:whitehall_migration_document_import, whitehall_migration_id: whitehall_migration.id)
```
kevindew: I don't think we need to create this either as we already have a document_import from the let at the start of this test.

spec/requests/whitehall_migration_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+      get whitehall_migration_document_path(whitehall_migration.id, document_import.id)
```
kevindew: Same comment as above about not needing the .id calls

lib/tasks/import.rake
```diff
@@ -27,5 +31,7 @@ namespace :import do
 
     WhitehallDocumentImportJob.perform_later(whitehall_import)
     puts "Added whitehall document with ID:#{args.document_id} to the import queue"
+
+    puts whitehall_migration_url(whitehall_import.whitehall_migration_id, host: Plek.new.external_url_for("content-publisher")).to_s
```
kevindew: Do we need the to_s here? (and in the example above)

spec/requests/whitehall_migration_spec.rb
Rosa-Fox: ah thought we changed this... think it got broken in the rebase, have fixed.

spec/requests/whitehall_migration_spec.rb
Rosa-Fox: 👍

spec/requests/whitehall_migration_spec.rb
Rosa-Fox: 👍

spec/requests/whitehall_migration_spec.rb
Rosa-Fox: 👍

Rosa-Fox:
Thanks @kevindew. Have made the changes... the build is passing locally but failing on CI... I will ask you about it tomorrow most likely.

lib/tasks/import.rake
```diff
@@ -1,10 +1,12 @@
-# frozen_string_literal: true
+ # frozen_string_literal: true
```
kevindew: This line has crept in after being removed in https://github.com/alphagov/content-publisher/pull/1769

lib/tasks/import.rake
```diff
@@ -13,12 +15,15 @@ namespace :import do
       organisation_content_id, args.document_type, document_subtypes
     )
 
-    documents_to_import = WhitehallMigration::DocumentImport.where(whitehall_migration_id: whitehall_migration.id).count
+    documents_to_import = WhitehallMigration::DocumentImport.where(whitehall_migration_id: whitehall_migration).count
```
kevindew: s/whitehall_migration_id/whitehall_migration/

spec/lib/tasks/import_spec.rb
```diff
@@ -1,8 +1,9 @@
 RSpec.describe "Import tasks" do
   include ActiveJob::TestHelper
+  include Rails.application.routes.url_helpers
```
kevindew: I don't think think this will be doing anything

lib/tasks/import.rake
```diff
@@ -1,10 +1,12 @@
-# frozen_string_literal: true
+ # frozen_string_literal: true
 
 require "gds_api/whitehall_export"
 
 namespace :import do
   desc "Import all documents matching an organisation, document type and optional list of document subtypes from Whitehall Publisher, e.g. import:whitehall_migration[\"cabinet-office\",\"news_article\",\"news_story,press_release\"]"
   task :whitehall_migration, %i[organisation_slug document_type document_subtypes] => :environment do |_, args|
+    include Rails.application.routes.url_helpers
```
kevindew: You may want to put this at the top following `namespace :import do` so that you don't need to include this in two separate tasks (I'm assuming that works but could be wrong)

lib/tasks/import.rake
brucebolt: Removed this line by rebasing against master

lib/tasks/import.rake
brucebolt: Updated.  Also tested running the rake test manually to ensure a useable URL is printed.

spec/lib/tasks/import_spec.rb
brucebolt: Removed, which had no negative effect.

lib/tasks/import.rake

brucebolt: Reverted, as discussed in person (due to unrelated test failures)

PR #1788, opened 2020-02-18
Make modals less verbose for screen readers

alex-ju:
In [#1300](https://github.com/alphagov/govuk_publishing_components/pull/1300) we enabled `aria-label` on modal dialogue component to make it less verbose. Because in Content Publisher we generate the modal dialogue content dynamically we'll need to set this property via JavaScript as part of the `modalWorkflow` logic.

[Trello card](https://trello.com/c/xZwFwpul)

### DIFF
--- a/app/assets/javascripts/modal/modal-workflow.js
+++ b/app/assets/javascripts/modal/modal-workflow.js
@@ -26,6 +26,7 @@ ModalWorkflow.prototype.render = function (response) {
 
 ModalWorkflow.prototype.renderSuccess = function (result) {
   this.$multiSectionViewer.showDynamicSection(result.body)
+  this.setModalAriaLabel(this.getSectionHeadingText())
   this.overrideActions(this.performAction)
   this.initComponents()
 }
@@ -34,6 +35,20 @@ ModalWorkflow.prototype.renderError = function (result) {
   window.Raven.captureException(result)
   console.error(result)
   this.$multiSectionViewer.showStaticSection('error')
+  this.setModalAriaLabel(this.getSectionHeadingText())
+}
+
+ModalWorkflow.prototype.getSectionHeadingText = function () {
+  var sectionHeading = this.$multiSectionViewer.querySelector('h1')
+  if (sectionHeading) {
+    return sectionHeading.innerText
+  }
+}
+
+ModalWorkflow.prototype.setModalAriaLabel = function (ariaLabel) {
+  if (ariaLabel) {
+    this.$modal.querySelector('dialog').setAttribute('aria-label', ariaLabel)
+  }
 }
 
 ModalWorkflow.prototype.initComponents = function () {
--- a/app/views/components/_multi_section_viewer.html.erb
+++ b/app/views/components/_multi_section_viewer.html.erb
@@ -11,9 +11,9 @@
   id: id,
   data: data_attributes,
   aria: aria_attributes do %>
+  <%= tag.section class: "app-c-multi-section-viewer__section js-dynamic-section" %>
+
   <% sections.each do |section| %>
     <%= tag.section section[:content], class: "app-c-multi-section-viewer__section", id: section[:id] %>
   <% end %>
-
-  <%= tag.div class: "app-c-multi-section-viewer__section js-dynamic-section" %>
 <% end %>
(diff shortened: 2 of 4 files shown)


app/assets/javascripts/modal/modal-workflow.js
```diff
@@ -26,6 +26,7 @@ ModalWorkflow.prototype.render = function (response) {
 
 ModalWorkflow.prototype.renderSuccess = function (result) {
   this.$multiSectionViewer.showDynamicSection(result.body)
+  this.setModalLabel()
```
ChrisBAshton: I'd be tempted to rename this to `setModalAriaLabel` so it's clear we're not dealing with `<label>` elements.

app/assets/javascripts/modal/modal-workflow.js
```diff
@@ -34,6 +35,14 @@ ModalWorkflow.prototype.renderError = function (result) {
   window.Raven.captureException(result)
   console.error(result)
   this.$multiSectionViewer.showStaticSection('error')
+  this.setModalLabel()
```
ChrisBAshton: Looks like we're missing a corresponding test for the 'render error aria-label' state.

app/assets/javascripts/modal/modal-workflow.js
```diff
@@ -34,6 +35,14 @@ ModalWorkflow.prototype.renderError = function (result) {
...
+}
+
+ModalWorkflow.prototype.setModalLabel = function () {
```
ChrisBAshton: I'd also be tempted to have this take a parameter:

```suggestion
ModalWorkflow.prototype.setModalLabel = function (ariaLabel) {
```

It would bring the `setModalLabel` calls into line with the `show(Static|Dynamic)Section` calls, insofar as they take a parameter, making it less stateful and more obvious what it's doing. Would end up looking something like `this.setModalLabel(this.getSectionHeading())`.

app/assets/javascripts/modal/modal-workflow.js
```diff
@@ -34,6 +35,14 @@ ModalWorkflow.prototype.renderError = function (result) {
...
+  let sectionHeading = this.$multiSectionViewer.querySelector('h1')
```
ChrisBAshton: I think I've asked this before, but can't remember the answer - is there a reason we're not using `const`?
And just to double check, this will get transpiled to `var`, right?

alex-ju: No, that should be a `var` – no transpilation at the moment. Good spot, will update.

app/assets/javascripts/modal/modal-workflow.js
alex-ju: Thanks for the suggestion, Chris. It makes sense to aim for stateless, will update the code.

alex-ju:
@ChrisBAshton I tried to address all your comments – good for a second review whenever you have time.

app/assets/javascripts/modal/modal-workflow.js
```diff
@@ -26,6 +26,8 @@ ModalWorkflow.prototype.render = function (response) {
 
 ModalWorkflow.prototype.renderSuccess = function (result) {
   this.$multiSectionViewer.showDynamicSection(result.body)
+  var headingText = this.getSectionHeadingText()
```
ChrisBAshton: I think it would be better to avoid creating a temporary variable here - the function name should be descriptive enough.

```js
  this.setModalAriaLabel(this.getSectionHeadingText())
```

alex-ju: I rely too much on garbage collector 😄 . I agree is clear enough if we use them as suggested. Updated.

alex-ju:
> Only thing that strikes me is we're guarding against one DOM query failing (`if (this.$multiSectionViewer.querySelector('h1'))`, but not the other (`this.$modal.querySelector('dialog')`). Do we expect `h1` to sometimes be missing, but `dialog` to always be present? If not I'd stick to one approach (though I'm not sure which I'd prefer!).

The difference between them is that the `<dialog>` element is part of the component, so we have control over it, whereas the `<h1>` is retrieved via fetch from a view. In theory all of our templates should provide an `<h1>` but if for some reason it doesn't have one I'd like to avoid failing with a hard error.

PR #1793, opened 2020-02-19
Make contacts code more resilient

1pretz1:
Now when there are no contacts for an edition the embed is ignored, the
same is true for when there is no matching contacts for an embed.

Previously the code would throw an error at either of these two
scenarios listed above as it tried to look for a content-id on nil.

Trello:
https://trello.com/c/OzdqkkW2/1455-fix-documents-that-failed-the-import-with-an-array-error

### DIFF
--- a/spec/lib/whitehall_importer/embed_body_references_spec.rb
+++ b/spec/lib/whitehall_importer/embed_body_references_spec.rb
@@ -10,6 +10,22 @@ RSpec.describe WhitehallImporter::EmbedBodyReferences do
       expect(govspeak_body).to eq("[Contact:#{content_id}]")
     end
 
+    it "ignores contact embed when there is no contacts" do
+      govspeak_body = described_class.call(
+        body: "[Contact:123]",
+        contacts: [],
+      )
+      expect(govspeak_body).to be_empty
+    end
+
+    it "ignores contact embed when there is no matching contacts" do
+      govspeak_body = described_class.call(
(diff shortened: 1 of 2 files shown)


spec/lib/whitehall_importer/embed_body_references_spec.rb
```diff
@@ -10,6 +10,22 @@
       expect(govspeak_body).to eq("[Contact:#{content_id}]")
     end
 
+    it "ignores contact embed when there is no contacts" do
+      govspeak_body = described_class.call(
+        body: "[Contact:123]",
+        contacts: [],
+      )
+      expect(govspeak_body).to eq("")
```
brucebolt: Instead of `eq("")` you could use `be_empty`, which looks nicer

1pretz1: `be_empty` is certainly nicer!

1pretz1:
I've updated the specs, ready for another gander @brucebolt !

To summarise our offline conversation - We should keep the behaviour of not aborting when a contact is embedded but doesn't exist in the import. The reason being when it comes to checking the integrity of the body text, it will pass if the contact genuinely doesn't exist in whitehall but will fail if the contact does exist and the import just hasn't included it. 

Throwing an error at the embed point would mean we wouldn't be able to decipher which of these situations has happened and so we could be rejecting a valid document. We'll need to bookmark that text comparer...

PR #1796, opened 2020-02-19
Only allow unique options in a document type selection

leenagupte:
Trello: https://trello.com/c/tbxRLYMP

This is to prevent the same radio button appearing multiple times on a document type selection page.

### DIFF
--- a/spec/models/document_type_selection_spec.rb
+++ b/spec/models/document_type_selection_spec.rb
@@ -36,6 +36,13 @@ RSpec.describe DocumentTypeSelection do
       ids = document_type_selections.pluck("id")
       expect(ids.size).to eq(ids.uniq.size)
     end
+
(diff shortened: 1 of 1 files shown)


spec/models/document_type_selection_spec.rb
```diff
@@ -36,6 +36,13 @@
       ids = document_type_selections.pluck("id")
       expect(ids.size).to eq(ids.uniq.size)
     end
+
+    it "only allows unique options in each document_type_selection" do
+      document_type_selections.each do |document_type_selection|
+        ids = document_type_selection["options"].pluck("id")
+        expect(ids.size).to eq(ids.uniq.size)
```
benthorner: Minor: could we get a better error message if we lost the `size`?

leenagupte: Have added an error message to the test.

PR #1800, opened 2020-02-20
Add "What Managing Editors can do" information page

brucebolt:
Adds a page informing publishers what Managing Editors can do, and lets the user know if they have this permission.

Also added a sidebar link and adjusted the footer, as shown in the Trello card and linked Google document.

User without managing editor permission:
Screenshot 2020-02-20 at 10 33 00

User with managing editor permission:
Screenshot 2020-02-20 at 10 32 47

Trello card: https://trello.com/c/nwjWvsJR


app/views/publisher_information/what_managing_editors_can_do.html.erb
```diff
@@ -0,0 +1,13 @@
+<% content_for :browser_title, t("publisher_information.what_managing_editors_can_do.title") %>
+<div class="govuk-grid-row">
+  <%= render partial: "publication_info_nav", locals: {current_page: managing_editors_path } %>
+  <div class="govuk-grid-column-two-thirds">
+    <h1 class="govuk-heading-l page-title"><%= t("publisher_information.what_managing_editors_can_do.title") %></h1>
```
alex-ju: I know this pattern is not introduced with this PR, but ideally we don't add CSS classes only for testing purpose (e.g. `page-title`). I'd rather use `h1` as a selector since we must have one on every page.

app/views/publisher_information/_publication_info_nav.html.erb
```diff
@@ -3,25 +3,25 @@
     aria_label: "Publisher information",
     items: [
       {
-        label: "What the Beta can and cannot do",
+        label: t("publisher_information.beta_capabilities.title"),
```
alex-ju: 👍

app/views/publisher_information/what_managing_editors_can_do.html.erb
brucebolt: Good point.  I've added a change that removes the `page-title` class.

brucebolt:
Thanks for the review @alex-ju.  I've made your proposed change regarding the `page-title` class, and asked a question on the Trello card regarding the headings.

app/views/publisher_information/_publisher_info_nav.html.erb
```diff
@@ -3,23 +3,29 @@
     aria_label: "Publisher information",
     items: [
       {
-        label: "What the Beta can and cannot do",
+        label: t("publisher_information.beta_capabilities.title"),
```
kevindew: I believe these are intentionally not using the translation key, as per https://github.com/alphagov/content-publisher/blob/master/docs/adr/0004-editing-microcopy.md section on link and button labels.

In the past we've also consciously chosen not to use translation keys as variables that can be re-used to reduce duplication. This is because we don't know if when someone edits them they want to edit one place or both. Though could be convinced it's the right thing in this case. Though if we do so we should change the test to use the translation key rather than the strings directly.

All essentially minor concerns, especially as we have very little editing of copy, but just pointing out where this may be drifting from what is consistent.

app/views/publisher_information/beta_capabilities.html.erb
```diff
@@ -1,8 +1,8 @@
 <% content_for :browser_title, t("publisher_information.beta_capabilities.title") %>
 <div class="govuk-grid-row">
-  <%= render partial: "publication_info_nav", locals: {current_page: beta_capabilities_path } %>
+  <%= render partial: "publisher_info_nav", locals: {current_page: beta_capabilities_path } %>
```
kevindew: Great catch 👍

app/views/publisher_information/what_managing_editors_can_do.html.erb
```diff
@@ -0,0 +1,13 @@
+<% content_for :browser_title, t("publisher_information.what_managing_editors_can_do.title") %>
+<div class="govuk-grid-row">
+  <%= render partial: "publisher_info_nav", locals: {current_page: managing_editors_path } %>
```
kevindew: We seem to be missing a space between the { and current_page. Looks like this is on the other ones too if you wouldn't mind?

config/locales/en/publisher_information/what_managing_editors_can_do.yml
```diff
@@ -0,0 +1,31 @@
...
+
+        [Find out more about withdrawing documents.](https://www.gov.uk/guidance/content-design/gov-uk-content-retention-and-withdrawal-archiving-policy)
+
+        ## History mode
+
+        Managing Editors can change whether or not a document gets history mode.
+
+        They can also edit documents that are already in history mode.
+
+        [Find out more about history mode.](https://www.gov.uk/guidance/content-design/gov-uk-content-retention-and-withdrawal-archiving-policy)
```
kevindew: I can see why you went this route in order to interpolate the variable, though I'm a bit concerned that this ends up different to all the other publisher information pages. 

I'm wondering if the simplest way for consistency is to put all the other publisher information content in locale files like this?

app/views/publisher_information/_publisher_info_nav.html.erb
brucebolt: I've removed the commit that made these changes to leave the links as static text

app/views/publisher_information/what_managing_editors_can_do.html.erb
brucebolt: Added a space to all the views where this was missing

config/locales/en/publisher_information/what_managing_editors_can_do.yml
brucebolt: I've moved all the existing publisher information body text into locales, which has made this consistent

app/views/publisher_information/what_managing_editors_can_do.html.erb
kevindew: ⭐️

PR #1802, opened 2020-02-20
Display data UI migration

Rosa-Fox:
UI to display information resulting from a Whitehall migration.

- On the `whitehall_migration#documents` page, for unsuccessfully imported documents, the link to view an individual document import currently displays the `id` of the `document_import`. We want it to display the `content_id` but at the moment the import script is not importing `content_ids`. This will be addressed in a following PR.

- To anyone reviewing... Please review the code as per usual and also let us know of any additional information that you think would be useful to display.

Show page:
whitehall-migration-show

Document imports page for successfully imported documents:
document-imports-completed

Document imports page for unsuccessfully imported documents:
document-imports-import-failed

Document page:
document-import

### DIFF
--- a/app/controllers/whitehall_migration_controller.rb
+++ b/app/controllers/whitehall_migration_controller.rb
@@ -9,7 +9,7 @@ class WhitehallMigrationController < ApplicationController
     @whitehall_migration = WhitehallMigration.find(params[:migration_id])
     scope = @whitehall_migration.document_imports
     scope = scope.where(state: params[:state]) if params[:state]
-    @document_imports = scope.order(:id).page(params.fetch(:page, 1)).per(25)
+    @document_imports = scope.order(:id).page(params.fetch(:page, 1)).per(50)
   end
 
   def document_import
--- a/app/controllers/whitehall_migration_controller.rb
+++ b/app/controllers/whitehall_migration_controller.rb
@@ -7,8 +7,9 @@ class WhitehallMigrationController < ApplicationController
 
   def document_imports
     @whitehall_migration = WhitehallMigration.find(params[:migration_id])
+    @state = params[:state] if WhitehallMigration::DocumentImport.states.keys.include?(params[:state])
     scope = @whitehall_migration.document_imports
-    scope = scope.where(state: params[:state]) if params[:state]
+    scope = scope.where(state: @state) if @state
     @document_imports = scope.order(:id).page(params.fetch(:page, 1)).per(50)
   end
 
--- a/app/controllers/whitehall_migration_controller.rb
+++ b/app/controllers/whitehall_migration_controller.rb
@@ -5,14 +5,14 @@ class WhitehallMigrationController < ApplicationController
     @whitehall_migration = WhitehallMigration.find(params[:migration_id])
   end
 
-  def documents
+  def document_imports
     @whitehall_migration = WhitehallMigration.find(params[:migration_id])
     scope = @whitehall_migration.document_imports
     scope = scope.where(state: params[:state]) if params[:state]
     @document_imports = scope.order(:id).page(params.fetch(:page, 1)).per(25)
   end
 
-  def document
+  def document_import
     @document_import = WhitehallMigration::DocumentImport.find_by!(
       id: params[:document_import_id],
       whitehall_migration_id: params[:migration_id],
--- a/app/controllers/whitehall_migration_controller.rb
+++ b/app/controllers/whitehall_migration_controller.rb
@@ -13,7 +13,7 @@ class WhitehallMigrationController < ApplicationController
   end
 
   def document
-    @document = WhitehallMigration::DocumentImport.find_by!(
+    @document_import = WhitehallMigration::DocumentImport.find_by!(
       id: params[:document_import_id],
       whitehall_migration_id: params[:migration_id],
     )
--- a/app/controllers/whitehall_migration_controller.rb
+++ b/app/controllers/whitehall_migration_controller.rb
@@ -9,7 +9,7 @@ class WhitehallMigrationController < ApplicationController
     @whitehall_migration = WhitehallMigration.find(params[:migration_id])
     scope = @whitehall_migration.document_imports
     scope = scope.where(state: params[:state]) if params[:state]
-    @documents = scope.order(:id).page(params.fetch(:page, 1)).per(25)
+    @document_imports = scope.order(:id).page(params.fetch(:page, 1)).per(25)
   end
 
   def document
--- a/app/controllers/whitehall_migration_controller.rb
+++ b/app/controllers/whitehall_migration_controller.rb
@@ -6,8 +6,10 @@ class WhitehallMigrationController < ApplicationController
   end
 
   def documents
-    whitehall_migration = WhitehallMigration.find(params[:migration_id])
-    @documents = whitehall_migration.document_imports
+    @whitehall_migration = WhitehallMigration.find(params[:migration_id])
+    scope = @whitehall_migration.document_imports
+    scope = scope.where(state: params[:state]) if params[:state]
+    @documents = scope.order(:id).page(params.fetch(:page, 1)).per(25)
   end
 
   def document
--- a/app/views/whitehall_migration/_document_imports_list.html.erb
+++ b/app/views/whitehall_migration/_document_imports_list.html.erb
@@ -1,0 +1,40 @@
+<% items = document_imports.map do |document_import|
+  view_links = []
+
+  if document_import.content_id
+    url = "#{Plek.find('whitehall-admin')}/government/admin/by-content-id/#{document_import.content_id}"
+    view_links << link_to("Whitehall", url, target: "_blank")
+  end
+
+  if document_import.document
+    view_links << link_to("Content Publisher",
+                          document_path(document_import.document),
+                          target: "_blank")
+  end
+
+  if document_import.completed?
+    view_links << link_to("Draft GOV.UK",
+                          edition_preview_url(document_import.document.current_edition),
+                          target: "_blank")
+  end
+
+  if document_import.completed? && document_import.document.live_edition
+    view_links << link_to("Live GOV.UK",
+                          edition_public_url(document_import.document.live_edition),
+                          target: "_blank")
+  end
+
+  {
+    link: {
+      text: document_import.current_title ? document_import.current_title : document_import.id.to_s,
+      path: whitehall_migration_document_path(document_import.whitehall_migration_id, document_import),
+    },
+    metadata: {
+      state: document_import.state.to_s.titleize,
+      error_message: document_import.error_log ? "Error log: #{document_import.error_log}" : ""
+    },
+    subtext: view_links.any? ? "View: #{view_links.join(" ")}".html_safe : nil,
+  }
+  end
+%>
+<%= render "govuk_publishing_components/components/document_list", items: items %>
--- a/app/views/whitehall_migration/_document_imports_list.html.erb
+++ b/app/views/whitehall_migration/_document_imports_list.html.erb
@@ -35,6 +35,5 @@
     },
     subtext: view_links.any? ? "View: #{view_links.join(" ")}".html_safe : nil,
   }
-  end
-%>
+end %>
 <%= render "govuk_publishing_components/components/document_list", items: items %>
--- a/app/views/whitehall_migration/_document_imports_list.html.erb
+++ b/app/views/whitehall_migration/_document_imports_list.html.erb
@@ -27,7 +27,7 @@
   {
     link: {
       text: document_import.current_title ? document_import.current_title : document_import.id.to_s,
-      path: whitehall_migration_document_path(document_import.whitehall_migration_id, document_import),
+      path: whitehall_migration_document_import_path(document_import.whitehall_migration_id, document_import),
     },
     metadata: {
       state: document_import.state.to_s.titleize,
--- a/app/views/whitehall_migration/document.html.erb
+++ b/app/views/whitehall_migration/document.html.erb
@@ -1,2 +1,27 @@
-<h1>Document</h1>
-<%= @document.id %>
+<% content_for :back_link, render_back_link(href: whitehall_migration_documents_path(migration_id: @document_import.whitehall_migration, state: @document_import.state)) %>
+<% content_for :title, "Document Import: #{@document_import.id}" %>
+<%= render "govuk_publishing_components/components/summary_list", {
+  items: [
+    {
+      field: "ID",
+      value: @document_import.id
+    },
+    {
+      field: "Created at",
+      value: format_time_on_date(@document_import.created_at)
+    },
+    {
+      field: "Updated at",
+      value: format_time_on_date(@document_import.updated_at)
+    },
+    {
+      field: "State",
+      value: @document_import.state.titleize
+    },
+    {
+      field: "Error log",
+      value: @document_import.error_log
+    }
+  ],
+  borderless: true
+} %>
--- a/app/views/whitehall_migration/document_imports.html.erb
+++ b/app/views/whitehall_migration/document_imports.html.erb
@@ -16,13 +16,13 @@
 <%
   pages = {}
   pages[:previous_page] = {
-    url: whitehall_migration_documents_path(migration_id: @whitehall_migration, page: @document_imports.prev_page, state: params['state']),
+    url: whitehall_migration_document_imports_path(migration_id: @whitehall_migration, page: @document_imports.prev_page, state: params['state']),
     title: "Previous page",
     label: "#{@document_imports.prev_page} of #{@document_imports.total_pages}"
   } if @document_imports.prev_page
 
   pages[:next_page] = {
-    url: whitehall_migration_documents_path(migration_id: @whitehall_migration, page: @document_imports.next_page, state: params['state']),
+    url: whitehall_migration_document_imports_path(migration_id: @whitehall_migration, page: @document_imports.next_page, state: params['state']),
     title: "Next page",
     label: "#{@document_imports.next_page} of #{@document_imports.total_pages}"
   } if @document_imports.next_page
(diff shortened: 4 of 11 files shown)


kevindew:
> * @kevindew would it be better to put construct the ‘items’ data to be passed into the components within the views, controllers or somewhere else? At the moment govuk_publishing_components/components/document_list are in the controllers and govuk_publishing_components/components/summary_list in the views.

In app convention for Content Publisher has tended to be that we have a lot of the logic in the view. Your idea of a partial sounds good though if it gets hairy.

> * @kevindew  How in depth should the feature specs be? There are some initial specs but I wasn’t sure how much of the specific data to test for, given that feature specs are slowish and its an internal tool

I'd probably not bother with them myself. As we already have request tests we should already have the code being executed so we'll have coverage and be covered for syntax errors. As it's an internal tool we're probably not too concerned about semantic errors.

kevindew:
From a quick play around [misusing the component system](https://github.com/alphagov/content-publisher/tree/show-import) I've put this together for each document import type:

Screenshot 2020-02-24 at 12 37 53

Code for this is at: https://github.com/alphagov/content-publisher/blob/6a99619359fd2ed2d7219a155d12dfaa13152d90/app/views/components/_demo_document_import.html.erb

app/controllers/whitehall_migration_controller.rb
```diff
@@ -6,8 +6,14 @@ def show
...
 
   def documents
-    whitehall_migration = WhitehallMigration.find(params[:migration_id])
-    @documents = whitehall_migration.document_imports
+    state_param = params[:state]
+    @whitehall_migration = WhitehallMigration.find(params[:migration_id])
+    if state_param
+      filter = @whitehall_migration.document_imports.where(state: state_param)
+      @documents = filter.order(:id).page(params[:page]).per(25)
+    else
```
GDSNewt: Do we even show documents without a state anymore? From what I can see the UI only allows users to see a random selection and by 'state' 🤔

app/views/whitehall_migration/document.html.erb
```diff
@@ -1,2 +1,27 @@
...
+      value: @document.whitehall_migration.finished_at.nil? ? "unfinished" : format_time_on_date(@document.whitehall_migration.finished_at)
+    },
+    {
+      field: "Status",
+      value: @document.state
+    },
+    {
+      field: "Error log",
+      value: @document.error_log
+    }
```
kevindew: We'll likely want to also include integrity check problems here.

May even want to dump things like payload, integrity check problems payload and the list of asset imports. Though they can come later.

app/views/whitehall_migration/documents.html.erb
```diff
@@ -1,6 +1,38 @@
...
-<% @documents.each do |document| %>
-  <li>ID: <%= document.id %></li>
-  <li>Status: <%= document.state %></li>
-<% end %>
+<% states = ["imported", "sync_failed", "completed"] %>
+
+<% items = @documents.map do |document_import|
+    {
+      link: {
+        text: states.include?(document_import.state) ? document_import.document.current_edition.title : document_import.id.to_s,
```
kevindew: You may want to put a method of current_title on document_import that does the whole document.current_edition.title lookup and returns nil if there is no document. You could then use the presence of document_import.current_title to decide what to show rather than needing to do this state check.

app/views/whitehall_migration/documents.html.erb
```diff
@@ -1,6 +1,38 @@
...
+        path: whitehall_migration_document_path(document_import.whitehall_migration_id, document_import),
+      },
+      metadata: {
+        status: document_import.state.to_s,
+        error_message: document_import.error_log ? "Error log: #{document_import.error_log}" : ""
+      },
+      subtext: states.include?(document_import.state) ? "View: <a href=\"#{Plek.find('whitehall-admin')}/government/admin/by-content-id/#{document_import.content_id}\">Whitehall</a> View: <a href=\"#{Plek.find('content-publisher')}/documents/#{document_import.content_id}:en\">Content Publisher</a>".html_safe : "View: <a href=\"#{Plek.find('whitehall-admin')}/government/admin/by-content-id/#{document_import.content_id}\">Whitehall</a>".html_safe,
```
kevindew: I'd suggest setting up these links before hand and then injecting them.

For example:

```
<%
  items = @documents.map do |document_import|
    view_links = []
    if document_import.content_id
      url = "#{Plek.find('whitehall-admin')}/government/admin/by-content-id/#{document_import.content_id}"
      view_links << link_to("Whitehall", url, target: "_blank")
    end
    
    if document_import.document
      view_links << link_to("Content Publisher",
                            document_path(document_import.document), 
                            target: "_blank")
    end
    
    if document_import.completed?
      view_links << link_to("Draft GOV.UK",
                            edition_preview_url(document_import.document.current_edition)
                            target: "_blank")
    end

    if document_import.completed? && document_import.document.live_edition
      view_links << link_to("Live GOV.UK",
                            edition_public_url(document_import.document.live_edition)
                            target: "_blank")
    end
    
    # ...
    
      subtext: view_links.any? ? "View: #{view_links.map(&:html_safe).join(" ")}" : nil,
    }
  end
%>
```

I also snuck in the concept of draft and live gov.uk links above as I realised not everything will be on live GOV.UK

app/views/whitehall_migration/documents.html.erb
```diff
@@ -1,6 +1,38 @@
...
+    }
+  end
+%>
+
+<%= render "govuk_publishing_components/components/document_list", items: items %>
```
kevindew: Do we want a fallback for when there are no items?

app/views/whitehall_migration/documents.html.erb
```diff
@@ -1,6 +1,38 @@
-<h1>Documents</h1>
+<% content_for :back_link, render_back_link(href: whitehall_migration_path(@whitehall_migration)) %>
+<% content_for :title, "Document Imports" %>
```
kevindew: The title should probably reflect the filter in usage

app/views/whitehall_migration/show.html.erb
```diff
@@ -1,14 +1,49 @@
...
+<%= render "govuk_publishing_components/components/document_list", items: items %>
+
+<% random_items = @whitehall_migration.document_imports.completed.order(Arel.sql("RANDOM()")).first(10).map do |document|
+    {
+      link: {
+        text: document.document.current_edition.title
+      },
+      subtext: "View: <a href=\"#{Plek.find('whitehall-admin')}/government/admin/by-content-id/#{document.content_id}\">Whitehall</a> View: <a href=\"#{Plek.find('content-publisher')}/documents/#{document.content_id}:en\">Content Publisher</a>".html_safe
+    }
+  end
```
kevindew: I feel like we should have a partial that we can pass an array of document_imports into renders the document_list component. This would avoid the need for repetition here and on the documents view

app/views/whitehall_migration/show.html.erb
```diff
@@ -1,14 +1,49 @@
...
+%>
+
+<%= render "govuk_publishing_components/components/document_list", items: items %>
+
+<% random_items = @whitehall_migration.document_imports.completed.order(Arel.sql("RANDOM()")).first(10).map do |document|
+    {
+      link: {
+        text: document.document.current_edition.title
```
kevindew: It'd be a little bit nicer to read if the items we iterated were called document_import.

app/views/whitehall_migration/show.html.erb
```diff
@@ -1,14 +1,49 @@
...
+        text: "#{state.titleize}: #{@whitehall_migration.document_imports.where(state: state).count}",
+        path: whitehall_migration_documents_path(migration_id: @whitehall_migration, state: state),
+      },
+    }
+  end
+%>
+
+<%= render "govuk_publishing_components/components/document_list", items: items %>
+
+<% random_items = @whitehall_migration.document_imports.completed.order(Arel.sql("RANDOM()")).first(10).map do |document|
```
kevindew: I think this should be inside the conditional about whether there are any completed documents

lib/tasks/import.rake
```diff
@@ -16,13 +15,12 @@ namespace :import do
     documents_to_import = WhitehallMigration::DocumentImport.where(whitehall_migration: whitehall_migration).count
     puts "Identified #{documents_to_import} documents to import"
 
-    puts whitehall_migration_url(whitehall_migration, host: Plek.new.external_url_for("content-publisher"))
+    puts whitehall_migration_url(whitehall_migration, host: Plek.new.external_url_for("content-publisher")).to_s
```
kevindew: I'm not sure this needs to_s whitehall_migration_url already returns a string. I couldn't see anything in the commit message regarding why this was needed.

spec/requests/whitehall_migration_spec.rb
```diff
@@ -36,7 +36,8 @@
   describe "GET /whitehall-migration/:migration_id/documents" do
     it "returns success" do
       login_as(debug_permission_user)
-      get whitehall_migration_documents_path(whitehall_migration)
+      create(:whitehall_migration_document_import, whitehall_migration_id: whitehall_migration.id)
```
kevindew: I'm not sure you need this, what happens without it?

spec/requests/whitehall_migration_spec.rb
```diff
@@ -45,7 +46,8 @@
   describe "GET /whitehall-migration/:migration_id/documents/:document_import_id" do
     it "returns success" do
       login_as(debug_permission_user)
-      get whitehall_migration_document_path(whitehall_migration, document_import)
+      document_import = create(:whitehall_migration_document_import, whitehall_migration_id: whitehall_migration.id)
```
kevindew: We shouldn't need to do this as we already have a document_import variable defined in a let at the top of this spec.

lib/tasks/import.rake
GDSNewt: If memory serves correctly I think this was an automatic change by the linter, I believe it works without to_s

app/controllers/whitehall_migration_controller.rb
Rosa-Fox: Hi, they can be viewed without a state ( aka http://content-publisher.dev.gov.uk/whitehall-migration/3/documents) but we don't have a link to view all without state. I don't think we ever explicitly had a link to do so, it was just that we hadn't added the filtering yet

app/views/whitehall_migration/document.html.erb
```diff
@@ -1,2 +1,27 @@
-<h1>Document</h1>
-<%= @document.id %>
+<% content_for :back_link, render_back_link(href: whitehall_migration_documents_path(@document.whitehall_migration)) %>
+<% content_for :title, "Document Import: #{@document.id}" %>
```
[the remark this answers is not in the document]
Rosa-Fox: Thanks, have done. Will later rename the routes/controllers to be `document_imports` to avoid confusion between the two.

app/views/whitehall_migration/documents.html.erb
Rosa-Fox: changes to `items ||= []`

app/views/whitehall_migration/show.html.erb
Rosa-Fox: I am not sure if I understand... the `document_list` needs to take `items`. The `items` are constructed from some form of `document_imports` but they all vary a lot in how the values within `items` are generated... so I don't think one partial would fit all very easily. The `render "govuk_publishing_components/components/document_list"` already feels like we are using a partial anyway?

app/views/whitehall_migration/document.html.erb
Rosa-Fox: @kevindew that would be good. Sounds like one for a future PR?

app/views/whitehall_migration/document.html.erb
```diff
@@ -1,2 +1,27 @@
...
+      field: "ID",
+      value: @document_import.id
+    },
+    {
+      field: "Created at",
+      value: format_time_on_date(@document_import.created_at)
+    },
+    {
+      field: "Updated at",
+      value: @document_import.whitehall_migration.updated_at.nil? ? "unfinished" : format_time_on_date(@document_import.whitehall_migration.updated_at)
```
kevindew: I think we should the document imports updated_at rather than the migrations one, I don't think we need a conditional for unfinished here (the current condition here would never be nil once persisted to the db)

app/views/whitehall_migration/document.html.erb
```diff
@@ -1,2 +1,27 @@
...
+    },
+    {
+      field: "State",
+      value: @document_import.state
```
kevindew: Do we want to titleize on this state to be consistent with other parts of the UI?

app/views/whitehall_migration/documents.html.erb
```diff
@@ -1,6 +1,61 @@
-<h1>Documents</h1>
+<% content_for :back_link, render_back_link(href: whitehall_migration_path(@whitehall_migration)) %>
+<% params[:state] ? title = "Document Imports: #{params[:state].titleize}" : title = "Document Imports" %>
```
kevindew: It'd be more common to write this as <% title = params[:state] ? "Document Imports: #{params[:state].titleize}" : "Document Imports" %>

app/views/whitehall_migration/documents.html.erb
```diff
@@ -1,6 +1,61 @@
...
+                          target: "_blank")
+  end
+
+  {
+    link: {
+      text: document_import.current_title ? document_import.current_title : document_import.id.to_s,
+      path: whitehall_migration_document_path(document_import.whitehall_migration_id, document_import),
+    },
+    metadata: {
+      state: document_import.state.to_s,
```
kevindew: Do we need to titleize here too?

app/views/whitehall_migration/show.html.erb
```diff
@@ -1,14 +1,48 @@
...
+
+<%= render "govuk_publishing_components/components/summary_list", {
+  items: [
+    {
+      field: "Time started",
+      value: format_time_on_date(@whitehall_migration.created_at)
+    },
+    {
+      field: "Time finished",
+      value: @whitehall_migration.finished_at.nil? ? "unfinished" : @whitehall_migration.finished_at
```
kevindew: looks like we're missing the format_time_on_date here?

app/views/whitehall_migration/show.html.erb
```diff
@@ -1,14 +1,48 @@
...
+    },
+    {
+      field: "Number of documents",
```
kevindew: should this be "Number of document imports" given our change in terminology.

app/controllers/whitehall_migration_controller.rb
```diff
@@ -5,13 +5,15 @@ def show
...
   end
 
-  def documents
-    whitehall_migration = WhitehallMigration.find(params[:migration_id])
-    @documents = whitehall_migration.document_imports
+  def document_imports
+    @whitehall_migration = WhitehallMigration.find(params[:migration_id])
+    scope = @whitehall_migration.document_imports
+    scope = scope.where(state: params[:state]) if params[:state]
+    @document_imports = scope.order(:id).page(params.fetch(:page, 1)).per(25)
```
kevindew: Minor - assume it got missed somewhere but we planned to up this to 50.

app/views/whitehall_migration/_document_imports_list.html.erb
```diff
@@ -0,0 +1,40 @@
...
+      text: document_import.current_title ? document_import.current_title : document_import.id.to_s,
+      path: whitehall_migration_document_import_path(document_import.whitehall_migration_id, document_import),
+    },
+    metadata: {
+      state: document_import.state.to_s.titleize,
+      error_message: document_import.error_log ? "Error log: #{document_import.error_log}" : ""
+    },
+    subtext: view_links.any? ? "View: #{view_links.join(" ")}".html_safe : nil,
+  }
+  end
```
kevindew: Indentation seems to have slipped a little in this block.

app/views/whitehall_migration/document_imports.html.erb
```diff
@@ -0,0 +1,31 @@
+<% content_for :back_link, render_back_link(href: whitehall_migration_path(@whitehall_migration)) %>
+<% title = params[:state] ? "Document Imports: #{params[:state].titleize}" : "Document Imports" %>
+<% content_for :title, title %>
+
+<% if @document_imports.any? %>
+<%= render "document_imports_list", document_imports: @document_imports %>
+<% else %>
+  <p class="govuk-body">
+    There are no document imports
+    <% if params[:state] %>
```
kevindew: Would probably be prudent to check this is an expected state given this is user input.

```suggestion
    <% if WhitehallMigration::DocumentImport.states.keys.include(params[:state]) %>
```

app/views/whitehall_migration/show.html.erb
```diff
@@ -1,14 +1,40 @@
...
-</ul>
-
-<h3>Document Import Status</h3>
-
-<ul>
-    <% @whitehall_migration.document_imports.states.each_value do |state| %>
-        <li><%= "#{state}:" %> <%= @whitehall_migration.document_imports.where(state: "#{state}").count%></li>
-    <% end %>
-</ul>
+
```
kevindew: Probably want to kill this empty line

app/views/whitehall_migration/document_imports.html.erb
Rosa-Fox: This gives the same result regardless of the lookup, so not sure if it is worth the query?

app/views/whitehall_migration/document_imports.html.erb
```diff
@@ -0,0 +1,31 @@
...
+<%
+  pages = {}
+  pages[:previous_page] = {
+    url: whitehall_migration_document_imports_path(migration_id: @whitehall_migration, page: @document_imports.prev_page, state: params['state']),
+    title: "Previous page",
+    label: "#{@document_imports.prev_page} of #{@document_imports.total_pages}"
+  } if @document_imports.prev_page
+
+  pages[:next_page] = {
+    url: whitehall_migration_document_imports_path(migration_id: @whitehall_migration, page: @document_imports.next_page, state: params['state']),
```
kevindew: You'll want to switch params[:state] to @state here and on line 19

Rosa-Fox: Thanks, have amended!

PR #1804, opened 2020-02-21
Extract tag fields into config

ChrisBAshton:
Moves tag field configuration out of YAML - which was too flexible and allowed configuring tags in a way we wouldn't want them to be configured - and into prescriptive models instead. This follows the same "Tell-Don't-Ask" approach as we've used for content fields in #1705, and removes some blockers on moving publishing API dependencies into our [bulk-data mechanism](https://github.com/alphagov/content-publisher/blob/master/docs/adr/0015-bulk-data-loading.md).

It also gives us the freedom to easily refer to content or tag fields from different UI screens. For example, here is the result of adding `primary_publishing_organisation` to the [`contents` array](https://github.com/alphagov/content-publisher/blob/7b18bb70df6ed347d269dbd88fbc7e2eb5296c0b/config/document_types.yml#L8-L11):

screenshot of edit screen with additional 'primary publishing organisation' input

## New 'API down' screen

As a result of moving to granular views, error handling has had to change. This does mean the UX will be slightly different if the API goes down.

Before:

Screenshot 2020-03-05 at 10 43 45

After:

Screenshot 2020-03-05 at 10 43 02

We may want to consider shortening the error message, or moving error handling into a presenter, but this might be more work than we expect. We don't think this is a blocking issue for merge - see full discussion in https://github.com/alphagov/content-publisher/pull/1804#discussion_r387047128.

Trello: https://trello.com/c/BOPwzj7L/1399-extract-tag-fields-into-config


lib/requirements/tag_checker.rb
```diff
@@ -19,18 +18,5 @@ def pre_update_issues(params)
     def pre_publish_issues
       pre_update_issues(edition.tags.symbolize_keys)
     end
```
kevindew: It seems like this should iterate through tags for pre_publish and pre_preview issues

spec/factories/tag_field_factory.rb
```diff
@@ -1,19 +1,21 @@
 FactoryBot.define do
-  factory :tag_field, class: "DocumentType::TagField" do
+  factory :tag_field, class: "DocumentType::MultiTagField" do
```
kevindew: I'm not sure we need this since we don't seem to need it for contents fields.

spec/factories/tag_field_factory.rb
```diff
@@ -1,19 +1,21 @@
 FactoryBot.define do
-  factory :tag_field, class: "DocumentType::TagField" do
+  factory :tag_field, class: "DocumentType::OrganisationsTagField" do
```
benthorner: Thanks for pointing this out in the commit - I missed it when I was reading through. For now, we should dispense with the factory and use the same approach we did for [the original PR on this work](https://github.com/alphagov/content-publisher/pull/1705/commits/b7033fee13ab2c1f4e96ab69e4c99d15c7882893#diff-7ebb5ee02d495684dcb59a514e54bb8aR19).

app/views/tags/tags/_primary_publishing_organisation.html.erb
```diff
@@ -0,0 +1,9 @@
+<% if edition.access_limit.present? %>
```
benthorner: We should call the directory `fields/`, since they're intended to form part of the wider body of field (that just happen to be shown only on this page at the moment).

app/views/tags/edit.html.erb
```diff
@@ -13,18 +13,17 @@
...
+                      else
+                        "tags/tags/default"
+                      end %>
+        <%= render template, edition: @edition, tag_field: tag_field do %>
+          <%= render "tags/inputs/#{tag_field.type}_input",
             tag_field: tag_field,
             tags: params[:tags] || @edition.tags,
             issues: @issues %>
-        </div>
+          <% end %>
```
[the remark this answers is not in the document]
ChrisBAshton: This does feel like a lot of duplication even with a partial, especially where things like `id` can be retrieved dynamically via `tag_field.id` rather than duplicating the ID in the model and the view. But I do like that we'd only have one view-rendering approach for all field types. I'll follow your suggestion and see what it looks like.

app/views/tags/tags/_primary_publishing_organisation.html.erb
ChrisBAshton: Can I just confirm you'd like this PR to implement this structure?

```diff
  views/
    content/
-       edit/
-           _body_field.html.erb
     edit.html.erb
+   fields/
+       edit/
+           _body_field.html.erb
+           _primary_publishing_organisation_field.html.erb
+       show/
+           _body_field.html.erb
+           _primary_publishing_organisation_field.html.erb
    documents/
        show/
-          contents/
-              _body_field.html.erb
        _actions.html.erb
        ...etc
    tags/
-      tags/
+      types/
            _multi_tag_input.html.erb
            _single_tag_input.html.erb
     edit_api_down.html.erb
     edit.html.erb
```

I.e. consolidate content and tag fields under one `fields/` directory, with their 'edit' view under a `edit/` subdirectory and their rendered view under a `show/` subdirectory? And `fields/edit/_primary_publishing_organisation_field.html.erb` referencing the `tags/types/_single_tag_input.html.erb` partial?

benthorner: Thanks for elaborating - it's good to be clear 👍. I'd say "yes", except that I don't see the need for `_single_tag_input.html.erb`, if it's only used once - why have indirection without motivation?

ChrisBAshton: Agreed 👍

app/models/document_type/multi_tag_field.rb
```diff
@@ -0,0 +1,13 @@
+class DocumentType::MultiTagField
+  def id
+    document_type.pluralize
+  end
```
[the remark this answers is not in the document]
ChrisBAshton: Have now removed this 👍 looked into removing `document_type` but would mean hardcoding the document type in multiple places in the views (e.g. in both 'show' and 'edit' contexts) and in multiple places in the tests. Think it's cleaner overall if we keep it in.

lib/requirements/tag_checker.rb
ChrisBAshton: I've now added methods for pre_publish and pre_update.

pre_preview_issues isn't currently a thing when it comes to tags, so it felt a little odd adding it. I can see it coming in useful later though, so have added some methods for it that return no issues at the moment.

app/views/documents/show/_contents.html.erb
```diff
@@ -1,7 +1,7 @@
 <% contents = @edition.document_type.contents.map { |field|
   {
     field: t_doctype_field(@edition, "#{field.id}.label"),
-    value: render("documents/show/contents/#{field.id}_field", edition: @edition)
+    value: render("fields/show/#{field.id}", edition: @edition)
```
kevindew: Similar comment about the switch of this location as the edit fields.

app/views/fields/edit/_primary_publishing_organisation.html.erb
```diff
@@ -1,3 +1,9 @@
+<% if edition.access_limit.present? %>
+  <%= render "govuk_publishing_components/components/warning_text", {
+    text: t("tags.edit.organisation_warning"),
+  } %>
+<% end %>
+
```
kevindew: This looks good 👍

lib/bulk_data/government_repository.rb
```diff
@@ -46,7 +46,7 @@ def populate_cache(older_than: nil)
     end
 
     def cache_age
-      Time.current - Cache.written_at(CACHE_KEY)
+      Time.zone.now - Cache.written_at(CACHE_KEY)
```
kevindew: I think this is already fixed in master and needs a rebase.

app/views/tags/edit.html.erb
```diff
@@ -13,14 +13,9 @@
...
-          } %>
-        <% end %>
-
         <div id="<%= tag_field.id %>">
-          <%= render "tags/tags/#{tag_field.type}_input",
+          <%= render "fields/edit/#{tag_field.id}",
+            edition: @edition,
             tag_field: tag_field,
             tags: params[:tags] || @edition.tags,
             issues: @issues %>
```
kevindew: I don't think this should have tag_field or tags passed in as it looks like a coupling of this to the tag view and a [deviation from the approach used for content](https://github.com/alphagov/content-publisher/blob/ab212f07bfcdcf10678a33479e8d7a5aa24994dc/app/views/content/edit.html.erb#L21-L23).

Can this match the interface used for content?

app/models/document_type/role_appointments_field.rb
```diff
@@ -0,0 +1,41 @@
...
+    role_appointments
+      .each_with_object(roles: [], people: []) do |appointment_id, memo|
+        response = GdsApi.publishing_api.get_links(appointment_id).to_hash
+
+        roles = response.dig("links", "role") || []
+        people = response.dig("links", "person") || []
+
+        memo[:roles] = (memo[:roles] + roles).uniq
+        memo[:people] = (memo[:people] + people).uniq
+      end
```
kevindew: Nice job getting this stuff in here 👍

spec/support/field_examples.rb
```diff
@@ -0,0 +1,33 @@
+module FieldExamples
```
kevindew: As field is quite a generic word you might want to call this DocumentTypeFieldExamples

spec/support/field_examples.rb
```diff
@@ -0,0 +1,33 @@
...
+  # an example calling this is expected to provided an instance of a
+  # DocumentType::*Field
+  #
+  # For example:
+  #   it_behaves_like "document type field", DocumentType::PrimaryPublishingOrganisationField.new
+  #
+  shared_examples "document type field" do |field|
```
kevindew: Rather than passing in the field object this looks like one of the rare cases where the one-liner syntax for RSpec is probably most applicable. 

In that instance it'd be like

```
shared_examples "document type field" do
  it { is_expected.to respond_to(:id).with(0).arguments }
  # etc
end

# and called like:


RSpec.describe DocumentType::PrimaryPublishingOrganisationField do
  it_behaves_like "document type field"
```

The disadvantage of passing the initialised object in, as you've done, is that you end up with the same object under test for each test. Not too big a deal given the scenario but is something of a smell.

app/views/content/edit.html.erb
```diff
@@ -19,7 +19,7 @@
     ) do |f| %>
 
   <% @edition.document_type.contents.each do |field| %>
-    <%= render "content/edit/#{field.id}_field", edition: @edition, issues: @issues %>
+    <%= render "fields/edit/#{field.id}", edition: @edition, issues: @issues %>
```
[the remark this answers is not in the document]
ChrisBAshton: We agreed to make this change in https://github.com/alphagov/content-publisher/pull/1804#discussion_r383788368.

ChrisBAshton:
This is ready for re-review, pending outcome of the `fields/` directory discussion.

app/models/document_type/organisations_field.rb
```diff
@@ -16,4 +16,8 @@ def document_type
   def pre_update_issues(_edition, _params)
     Requirements::CheckerIssues.new
   end
+
+  def pre_publish_issues(edition)
+    pre_update_issues(edition, edition.tags.symbolize_keys)
```
benthorner: Minor: this seems a little premature (and a little complex), given that we don't have a definition of pre-publish issues for most of these fields; it may be that they differ, like they do for [the body field](https://github.com/alphagov/content-publisher/blob/master/app/models/document_type/body_field.rb#L32). Can we explicitly return no issues to indicate they have none defined?

app/views/fields/edit/_primary_publishing_organisation.html.erb
```diff
@@ -1,3 +1,9 @@
...
 <%= render "components/autocomplete", {
   id: "#{tag_field.id}-field",
   name: "tags[#{tag_field.id}][]",
```
benthorner: Can we also look at making the naming consistent with [content fields](https://github.com/alphagov/content-publisher/blob/master/app/views/content/edit/_summary_field.html.erb#L22)? These don't have a `content[]` wrapper, which helps make them 'screen-agnostic'.

spec/support/document_type_field_examples.rb
```diff
@@ -0,0 +1,16 @@
+module DocumentTypeFieldExamples
```
benthorner: Can you help me understand the problem we're trying to solve here? It looks a bit like we're mandating which methods someone writes, but this should be self-evident by tests failing if such methods were missing. There does seem to be a duplication issue, so perhaps a little inheritance would clarify the interface and DRY things up. I don't think we should be adding meta-programming type tests, though.

app/views/documents/show/_tags.html.erb
```diff
@@ -23,7 +18,7 @@
     ),
     items: tags
   } %>
-<% rescue GdsApi::BaseError => e %>
+<% rescue ActionView::Template::Error => e %>
```
benthorner: I'm pretty sure we shouldn't do this, as it'll confuse a future developer who makes a typo in the view. This is definitely the reason we don't catch the API error at the controller level - we would also swallow any other problem with rendering the view.

Another way we could approach this would be to do the API calls in the individual views. We can still have a `multi_tag` partial for the actual rendering, so the result is reasonably DRY. Can we try that out and see how it looks?

To be properly compatible across views, the partials really ought to catch their own exceptions. However, we may as well save ourselves the effort, as we're intending to apply our local data caching approach to the linkables for tags (exception becomes redundant).

spec/features/editing_tags/edit_tags_spec.rb
```diff
@@ -16,19 +17,21 @@
   end
 
   def given_there_is_an_edition
-    document_type = build(:document_type, tags: [multi_tag_field, single_tag_field])
-
-    tag_linkables = [initial_tag, tag_to_select_1, tag_to_select_2]
-    stub_publishing_api_has_linkables(tag_linkables, document_type: "world_location")
-    stub_publishing_api_has_linkables(tag_linkables, document_type: "organisation")
+    news_story_tags = DocumentType.find("news_story").tags
```
benthorner: This seems to be a new problem, as our content fields got covered implicitly by the format specs, which no longer touch this page. Is there a way we could do this that doesn't rely on a document type? A single document type is unlikely to have all of the possible tags, so ideally we want the spec to cover all kinds; one approach would be to list them all explicitly here.

However, it's a bit strange to have a feature spec that only makes use of some of its setup. It may be unclear why all of the news tags are necessary to someone reading this in future. One approach might be to make the spec exercise all the different tags, which means we're thoroughly testing them at the integration level, albeit with more complexity. Another approach could be to use the view spec for the tags edit page, to check each one renders with appropriate elements.

Although we seem to be pushing for the latter approach with [the work on content fields](https://github.com/alphagov/content-publisher/blob/master/spec/features/editing_content/edit_edition_spec.rb), it turns out other specs [sneakily ensure we test all the content fields](https://github.com/alphagov/content-publisher/blob/master/spec/features/workflow/create_document_spec.rb#L35). So it does look like testing all the tags via this feature is the way to go.

app/models/document_type/primary_publishing_organisation_field.rb
```diff
@@ -0,0 +1,31 @@
+class DocumentType::PrimaryPublishingOrganisationField
+  def id
+    "primary_publishing_organisation"
+  end
+
+  def payload(edition)
+    { links: { primary_publishing_organisation: edition.tags["primary_publishing_organisation"] } }
```
benthorner: Minor: could we use the `id` here and similarly in other methods/models? We did this for [content fields](https://github.com/alphagov/content-publisher/blob/master/app/models/document_type/body_field.rb#L9), where the field is store in a JSON blob (so the naming is dictated by the code here), which applies to all tags.

spec/support/document_type_field_examples.rb
ChrisBAshton: Fair 👍 I didn't think our tests were testing 'real world' usage of tags; I think I was misremembering what `topics { false }` was doing. So I anticipated us forgetting to implement, say, `pre_preview_issues` in one of our field models, and all of our tests continuing to pass.

I've just tried removing this meta-programming test and then commenting out `pre_preview_issues` for `TopicalEventsField`, and tests do indeed break, so it should be safe to remove this. I think the DRY-ing up I add later in the PR also helps with a lot of this.

benthorner: @ChrisBAshton I had the same thought about (lack of) coverage. I wrote another comment about how we could do this more explicitly (if we want to) with the approach we take to fix locale coverage.

spec/features/editing_tags/edit_tags_spec.rb
ChrisBAshton: Have now changed this to `all_tags = DocumentType.all.flat_map(&:tags).uniq(&:class)` 👍

app/views/documents/show/_tags.html.erb
ChrisBAshton: The API calls are already happening in the individual views. The `GdsApi::BaseError` isn't caught in those views (as we're catching them at this higher level `_tags.html.erb` partial) so Rails throws a `ActionView::Template::Error` error instead.

I could move the `rescue` into each individual view, which I think is what you are suggesting, but there will be a change of behaviour. Instead of using the summary list component to render a "Error connecting to the API" message in its callback block, we'd be rendering lots of individual "Error connecting to the API" messages in each 'row' of the summary list.
I actually think this would be more useful - if a little ugly - as if some of the GdsApi requests succeed and some time out, we will at least get the information for the ones that succeeded.

Shall I proceed with moving the error handling to the individual views?

ChrisBAshton:
I've now reverted the movement of content field views into the `fields/` dir, and created the tag field views under their own `tags/show` and `tags/edit` subdirs. I believe I've addressed all of your comments apart from the GdsApi exception handling, which I'm currently not clear on how we would solve.

I'm away tomorrow but can pick this up on Thursday, or someone else is welcome to get it over the line in the meantime 👍

app/views/documents/show/_tags.html.erb
benthorner: Ah yes, thanks for pointing out the subtlety here! I remember we had a similar issue when this got implemented originally. It's likely they will all succeed or not, so the original concern was about not being repetitive. I had a chat with @kevindew yesterday and we agreed the individual messages would be OK, although we could look at making them shorter.

Go for it 👍

ChrisBAshton: Done 👍

app/views/tags/edit/_organisations.html.erb
```diff
@@ -0,0 +1,5 @@
+<%= render "tags/edit/multi_tag",
```
kevindew: This view (and similar) should have their filename suffixed with `_field` to be consistent with the content fields.

app/views/tags/show/_organisations.html.erb
```diff
@@ -0,0 +1 @@
+<%= render "tags/show/multi_tag", service: "organisation",  values: edition.tags["organisations"].to_a %>
```
kevindew: I'm assuming this is the view shown on the document summary? For consistency with contents this should be in `app/views/documents/show/tags` and have a `_field` suffix in the name - same for similar.

spec/models/document_type/organisations_field_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+                        organisations: org_ids,
+                      })
+      payload = described_class.new.payload(edition)
+      expect(payload[:links][:organisations]).to eq(org_ids)
+    end
+  end
+
+  describe "#updater_params" do
+    it "returns a hash of the organisations" do
+      edition = build :edition
```
kevindew: Oh no - have you caught this bad habit from @benthorner? 😉

spec/support/schemas/document_type.json
```diff
@@ -59,7 +59,14 @@
...
       "items": {
-        "$ref": "#/definitions/tag_field"
+        "type": "string",
+        "enum": [
+          "primary_publishing_organisation",
+          "organisations",
+          "role_appointments",
+          "topical_events",
+          "world_locations"
+        ]
```
kevindew: This is cleaner 👍 

One thing I wonder is whether we need to keep this list of allowed ones here - as it's something that needs to be kept in sync. Do we initialise every document type in tests? If so I'd assume the tests would blow up anyway if they had one outside this list

spec/models/document_type/organisations_field_spec.rb
ChrisBAshton: The lack of parentheses you mean? I was just following the convention of the file 🤷‍♂

spec/support/schemas/document_type.json
ChrisBAshton: Have just tested by adding a made up tag field to publications, and indeed most of the tests blow up with `uninitialized constant DocumentType::FooBarField`. I'll remove this array!

spec/models/document_type/organisations_field_spec.rb
benthorner: #therightconvention

PR #1808, opened 2020-02-24
Include change history in Publishing API payload

brucebolt:
Now that we store a full `change_history` for an edition, we need to send this full history (along with the `first_published_date` and `public_updated_at`) to Publishing API.

For the first edition, we always sent a fixed change note.  For subsequent major editions, the user entered change note is used.

This change also moves `Payload` out of the `PreviewDraftEditionService` namespace, into `PublishingApiPayload` since this class is used in multiple contexts, not just when dealing with the preview of draft editions.

Not to be merged until #1806 is merged and deployed.

Trello card: https://trello.com/c/KBFWEkKG

### DIFF
--- a/lib/publishing_api_payload.rb
+++ b/lib/publishing_api_payload.rb
@@ -22,8 +22,9 @@ class PublishingApiPayload
       links: links,
       access_limited: access_limited,
       auth_bypass_ids: auth_bypass_ids,
+      public_updated_at: history.public_updated_at,
     }
-    payload[:change_note] = edition.change_note if edition.major?
+    payload[:first_published_at] = history.first_published_at if history.first_published_at.present?
 
     document_type.contents.each do |field|
       payload.deep_merge!(field.payload(edition))
@@ -43,6 +44,10 @@ class PublishingApiPayload
   end
 
 private
+
+  def history
+    @history ||= History.new(edition)
+  end
 
   def access_limited
     return {} unless edition.access_limit
@@ -79,7 +84,10 @@ private
   end
 
   def details
-    details = { political: edition.political? }
+    details = {
+      political: edition.political?,
+      change_history: history.change_history,
+    }
 
     if document_type.lead_image? && edition.lead_image_revision.present?
       details[:image] = image
--- a/lib/publishing_api_payload.rb
+++ b/lib/publishing_api_payload.rb
@@ -1,4 +1,4 @@
-class PreviewDraftEditionService::Payload
+class PublishingApiPayload
   PUBLISHING_APP = "content-publisher".freeze
 
   attr_reader :edition, :document_type, :publishing_metadata, :republish
--- a/spec/lib/publishing_api_payload/history_spec.rb
+++ b/spec/lib/publishing_api_payload/history_spec.rb
@@ -1,0 +1,119 @@
+RSpec.describe PublishingApiPayload::History do
+  describe "#public_updated_at" do
+    it "returns current time if a major change is published" do
+      freeze_time do
+        first_edition = build(:edition)
+        second_edition = build(:edition, document: first_edition.document, update_type: "major")
+
+        expect(described_class.new(second_edition).public_updated_at).to eq(Time.zone.now)
+      end
+    end
+
+    it "returns most recent change history time if a minor change is published" do
+      first_edition = build(:edition, first_published_at: "2020-02-22 11:00:00")
+      second_edition = build(:edition, document: first_edition.document, update_type: "minor")
+
+      expect(described_class.new(second_edition).public_updated_at).to eq("2020-02-22 11:00:00")
+    end
+  end
+
+  describe "#first_published_at" do
+    it "returns backdated date if edition is backdated" do
+      edition = build(:edition, first_published_at: "2020-02-22 11:00:00", backdated_to: "2020-02-20 09:00:00")
+
+      expect(described_class.new(edition).first_published_at).to eq("2020-02-20 09:00:00")
+    end
+
+    it "returns first published date if edition is not backdated" do
+      edition = build(:edition, first_published_at: "2020-02-22 11:00:00")
+
+      expect(described_class.new(edition).first_published_at).to eq("2020-02-22 11:00:00")
+    end
+  end
+
+  describe "#change_history" do
+    let(:first_edition) { build(:edition, first_published_at: "2020-02-22 11:00:00") }
+    let(:first_change_note) do
+      {
+        note: "First published.",
+        public_timestamp: "2020-02-22 11:00:00",
+      }
+    end
+
+    it "appends first change note to end of history array" do
+      expect(described_class.new(first_edition).change_history).to eq([first_change_note])
+    end
+
+    it "adds change note to start of array for major changes" do
+      freeze_time do
+        second_edition = build(:edition, document: first_edition.document, number: 2, update_type: "major", change_note: "Some changes")
+
+        expected_change_note = {
+          note: "Some changes",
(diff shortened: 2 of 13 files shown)


spec/lib/publishing_api_payload_spec.rb
```diff
@@ -18,6 +18,70 @@
...
+          public_updated_at: Time.current,
+        }
+
+        expect(payload).to match a_hash_including(payload_hash)
+        expect(payload).not_to include(:first_published_at)
+      end
+    end
+
+    it "includes a first_published_at and public_updated_at timestamp for published editions" do
+      freeze_time do
```
kevindew: I wonder if you can avoid the need for freeze time here by setting fixed dates? I assume we're more testing it passes dates on rather than it is the current time.

lib/publishing_api_payload.rb
```diff
@@ -22,7 +22,9 @@ def payload
       links: links,
       access_limited: access_limited,
       auth_bypass_ids: auth_bypass_ids,
+      public_updated_at: history.public_updated_at,
     }
+    payload[:first_published_at] = history.first_published_at if edition.published?
```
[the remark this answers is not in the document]
brucebolt: A nil value is not valid (according to govuk-content-schemas) and an unrelated test fails due to an invalid schema without this guard clause.

spec/factories/metadata_revision_factory.rb
```diff
@@ -2,6 +2,7 @@
   factory :metadata_revision do
     update_type { "major" }
     document_type_id { build(:document_type, path_prefix: "/prefix").id }
+    change_history { [] }
```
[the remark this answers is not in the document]
brucebolt: Have removed this

spec/lib/publishing_api_payload_spec.rb
```diff
@@ -18,6 +18,70 @@
...
+        edition = build(:edition, document_type: document_type)
+
+        payload = described_class.new(edition).payload
+
+        payload_hash = {
+          public_updated_at: Time.current,
+        }
+
+        expect(payload).to match a_hash_including(payload_hash)
+        expect(payload).not_to include(:first_published_at)
```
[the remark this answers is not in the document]
brucebolt: Removed this, we're already testing it elsewhere

spec/lib/publishing_api_payload_spec.rb
```diff
@@ -18,6 +18,70 @@
       expect(payload).not_to include(:first_published_at)
     end
 
+    it "includes a public_updated_at timestamp for draft editions" do
+      freeze_time do
+        document_type = build(:document_type)
+        edition = build(:edition, document_type: document_type)
```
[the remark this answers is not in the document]
brucebolt: Yes, we don't need a document_type, so I've removed that

spec/lib/publishing_api_payload_spec.rb
brucebolt: I've removed the freeze_time from all the tests apart from those where we expect the current time

brucebolt:
Thanks for the review @kevindew.  I've moved the History class out into a namespaced file and added several unit tests for this class that cover multiple scenarios.

lib/publishing_api_payload.rb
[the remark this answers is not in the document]
kevindew: Oh bummer hmm this should probably check that history.first_published_at is set then since we'd want to set this when an edition is backdated.

spec/lib/publishing_api_payload/history_spec.rb
```diff
@@ -0,0 +1,110 @@
...
+
+    it "returns first published date if edition is not backdated" do
+      edition = build(:edition, first_published_at: "2020-02-22 11:00:00")
+
+      expect(described_class.new(edition).first_published_at).to eq("2020-02-22 11:00:00")
+    end
+  end
+
+  describe "#change_history" do
+    let(:edition) { build(:edition, first_published_at: "2020-02-22 11:00:00") }
```
kevindew: Might be clearer to name this as first_edition

Might also be useful to have a let of `initial_change_note` since you reference the same one a few times in tests.

spec/lib/publishing_api_payload/history_spec.rb
```diff
@@ -0,0 +1,110 @@
...
+
+    it "does not add change note to start of array for minor changes" do
+      second_edition = build(:edition, document: edition.document, number: 2, update_type: "minor")
+
+      expected_change_note = {
+        note: "First published.",
+        public_timestamp: "2020-02-22 11:00:00",
+      }
+
+      expect(described_class.new(second_edition).change_history.first).to eq(expected_change_note)
```
kevindew: I wasn't sure if this test quite captured the scenario as it only tests one of the change notes.

Might be better if the assertion was more along the lines of:

`expect(described_class.new(second_edition).change_history).to eq([expected_change_note])`

spec/lib/publishing_api_payload/history_spec.rb
```diff
@@ -0,0 +1,110 @@
...
+          public_timestamp: "2020-02-23 09:00:00",
+        },
+        {
+          note: "First published.",
+          public_timestamp: "2020-02-22 11:00:00",
+        },
+      ]
+
+      expect(described_class.new(second_edition).change_history).to eq(expected_change_notes)
+    end
```
kevindew: Be useful to also have a test that change notes set before a backdating aren't shown.

spec/lib/publishing_api_payload_spec.rb
```diff
@@ -18,6 +18,62 @@
...
+            public_timestamp: Time.zone.parse("2020-02-20 09:30:00"),
+          },
+          {
+            note: "First published.",
+            public_timestamp: Time.zone.parse("2020-02-10 09:30:00"),
+          },
+        ],
+      }
+
+      expect(payload[:details]).to match a_hash_including(expected_details_hash)
```
kevindew: I don't think you need to test these bits here since we're covering a bunch of this in the history object test.

We can probably do something like:

```
it "delegates to PublishingApiPayload::History to populate change_history" do
  history = instance_double(
    PublishingApiPayload::History, 
    change_history: [{ note: "note", public_timestamp: Time.current }],
    public_updated_at: Time.current,
    first_published_at: Time.current,
  )
  allow(PublishingApiPayload::History).to receive(:new).and_return(history)
  payload = described_class.new(build(:edition)).payload
  expect(payload[:details][:change_history]).to match(history.change_history)
end
```

spec/lib/publishing_api_payload/history_spec.rb
brucebolt: Good points.  I've changed the name of `edition` to `first_edition` and added `initial_change_note`.

spec/lib/publishing_api_payload/history_spec.rb
brucebolt: Made that change, and also made a similar change for other tests where we expect an array to be returned by `change_history`.

spec/lib/publishing_api_payload/history_spec.rb
brucebolt: Added a test for this

spec/lib/publishing_api_payload_spec.rb
brucebolt: Sure.  I've replaced the test with the one you've proposed, and ensured the `PublishingApiPayload::History` spec covers the situations we were previously testing here.

brucebolt:
Thanks for the review @kevindew.  I've made all the proposed minor changes.

Issue #1676, opened 2020-01-21
Our document types config is inconsistent


benthorner:
The [document_types.yml](https://github.com/alphagov/content-publisher/blob/566d2bff24ca13877a94eb67776e036a067b0b4e/config/document_types.yml) file is pretty inconsistent: it's a mixture of English (e.g. `hint_govspeak`), flags (e.g. `images: true`), abstractions (e.g. `contents: [...]`) and metadata (e.g. `path_prefix: ...`). We're currently doing work to mitigate some of this inconsistency, by [extracting bits of English](https://trello.com/c/AuzTdQVt/1100-use-i18n-to-manage-copy-for-each-document-type) and [using a more fine-grained type system to support  "flexible fields"](https://trello.com/c/u9kRKVsX/1324-extract-title-and-summary-into-classes-views-and-config).

After this work, the `document_types.yml` file should look something like this:

```
- id: news_story
  path_prefix: /government/news
  images: true
  publishing_metadata:
    schema_name: news_article
    rendering_app: government-frontend
  contents:
    - title_and_base_path
    - summary
    - body
  tags:
    - id: primary_publishing_organisation
      type: single_tag
      document_type: organisation
    - id: organisations
      type: multi_tag
      document_type: organisation
    - id: role_appointments
      type: multi_tag
      document_type: role_appointment
    - id: topical_events
      type: multi_tag
      document_type: topical_event
    - id: world_locations
      type: multi_tag
      document_type: world_location
```

Some of the obvious next steps are tidying-up tag fields and metadata. This will make it easier to talk about (and implement) further iterations to the config file.

```
- id: news_story
  images: true
  metadata:
    schema_name: news_article
    rendering_app: government-frontend
    path_prefix: /government/news
  contents:
    - title_and_base_path
    - summary
    - body
  tags:
    - primary_publishing_organisation
    - organisations
    - role_appointments
    - topical_events
    - world_locations
```

Although there is only one remaining anomaly (`images: true`) in the above example, as part of the work on attachments, we expect to need additional config:

   - Config to optionally switch on the "attachment list" feature, where a document can have an ordered/curated/featured list of file/external/HTML attachments.
   - Config to modify the existing "file attachments" feature, where a user can only insert inline/link snippets into their body content, for Publication document types.
   - Config to modify the upcoming external/HTML attachment features, so that neither of them is available for existing (news) document types.
[...]

PR #1809, opened 2020-02-24
Add new assertion for edition features

benthorner:
https://trello.com/c/85gkiigT/1397-make-disabled-features-more-robust-lead-images

This introduces a new 'assert_edition_feature' method and applies it to
lead image actions in order to render/return a 404 response when the user
tries to perform a lead image action on an edition that does not support
that feature, as specified in its document type.

Screenshot 2020-02-24 at 14 40 15

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -17,7 +17,7 @@ class ApplicationController < ActionController::Base
     Rails.logger.warn(e.message)
 
     if rendering_context == "modal"
-      render nothing: true, status: :bad_request
+      raise ActionController::BadRequest
     elsif e.edition.first? && e.edition.discarded?
       redirect_to documents_path
     else
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -25,6 +25,10 @@ class ApplicationController < ActionController::Base
     end
   end
 
+  rescue_from EditionAssertions::FeatureError do |e|
+    raise ActionController::RoutingError, e.message
+  end
+
   rescue_from BulkData::LocalDataUnavailableError do |error|
     GovukError.notify(error)
 
--- a/lib/edition_assertions.rb
+++ b/lib/edition_assertions.rb
@@ -14,11 +14,25 @@ module EditionAssertions
     end
   end
 
+  class FeatureError < Error
+    def initialize(edition, assertion)
+      super(edition, assertion || "edition supports feature")
+    end
+  end
+
   def assert_edition_state(edition, options = {}, &block)
     return if block.call(edition)
 
     assertion = options[:assertion]
     assertion ||= block.to_s if block.to_s =~ /&:/ # &:editable?
     raise StateError.new(edition, assertion)
+  end
+
+  def assert_edition_feature(edition, options = {}, &block)
+    return if block.call(edition.document_type)
+
+    assertion = options[:assertion]
+    assertion ||= block.to_s if block.to_s =~ /&:/ # &:lead_image?
+    raise FeatureError.new(edition, assertion)
   end
 end
--- a/spec/requests/lead_image_spec.rb
+++ b/spec/requests/lead_image_spec.rb
@@ -18,6 +18,14 @@ RSpec.describe "Lead Image" do
       let(:route_params) { [edition.document, "image_id"] }
     end
 
+    it_behaves_like "requests that return status",
+                    "choosing lead image not associated with edition",
+                    status: :not_found,
+                    routes: { choose_lead_image_path: %i[post] } do
+      let(:edition) { create :edition, document_type: document_type }
+      let(:route_params) { [edition.document, image_revision.image_id] }
+    end
+
     it "redirects to document summary with an alert" do
       edition = create(:edition,
                        document_type: document_type,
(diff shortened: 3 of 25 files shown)


benthorner:
It's hard to check this works properly in a modal, so I temporarily added the feature assertion to the image edit action, to see it handles the response OK.

Screenshot 2020-02-24 at 14 46 25

lib/edition_assertions.rb
```diff
@@ -14,11 +14,23 @@ def initialize(edition, assertion)
...
 
     assertion = options[:assertion]
     assertion ||= block.to_s if block.to_s =~ /&:/ # &:editable?
     raise StateError.new(edition, assertion)
   end
+
+  def assert_edition_feature(edition, &block)
+    return if block.call(edition.document_type)
+
+    raise FeatureError.new(edition, block.to_s)
```
kevindew: I'm wondering here if we're only expecting this to be called with `&:lead_image` style blocks? I'm basing that on us not doing the `if block.to_s =~ /&:/` check and so we'd end up with "#<Proc:0x00007fd" style output (if I understand right) for a customised block.

So this makes me think maybe we actually want to just change &block to be a symbol of the feature name and not worry about blocks and instead do `edition.document_type.public_send(feature)`.

app/controllers/application_controller.rb
```diff
@@ -17,7 +17,7 @@ class ApplicationController < ActionController::Base
     Rails.logger.warn(e.message)
 
     if rendering_context == "modal"
-      render nothing: true, status: :bad_request
+      head :bad_request
```
kevindew: Oh good catch - I feel like we might want to instead do `raise ActionController::BadRequest`

spec/requests/lead_image_spec.rb
```diff
@@ -18,6 +18,14 @@
       let(:route_params) { [edition.document, "image_id"] }
     end
 
+    it_behaves_like "requests that return status",
+                    "choosing lead image not associated with edition",
+                    status: :not_found,
+                    routes: { choose_lead_image_path: %i[post] } do
+      let(:edition) { create :edition, document_type: document_type }
+      let(:route_params) { [edition.document, image_revision.image_id] }
+    end
```
kevindew: Nice catch

lib/edition_assertions.rb
benthorner: This is actually the approach I was assuming and started with. I wrote in the commit how this was awkward, because I wanted to make the assertion method robust for different arguments; at the moment, we only support `:lead_image`. It seemed simpler and more consistent to directly tie the use of the assertion to the invocation of a method.

Alternatives I thought of:

   - Have a `ASSERTABLE_FEATURES` constant and raise if outside of this
   - Check the class of the 'flag' and raise if it's not a boolean

I think this is a symptom of postponing https://github.com/alphagov/content-publisher/issues/1676, which is fine, as all the alternatives are relatively simple.

Do you have any preferences or other suggestions?

app/controllers/application_controller.rb
```diff
@@ -25,6 +25,11 @@ class ApplicationController < ActionController::Base
     end
   end
 
+  rescue_from EditionAssertions::FeatureError do |e|
+    Rails.logger.warn(e.message)
+    render "errors/not_found", status: :not_found
```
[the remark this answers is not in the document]
benthorner: Ah, that's where I'd seen it before! Updated 👍

app/controllers/application_controller.rb
benthorner: I think we should keep this as-is, so that both branches of the conditional have a similar level of abstraction. Another reason I'd like to keep it is that we're already doing some handling by logging a warning and checking the context, so I'd expect us to do all the handling here. Is there a counter-argument here?

lib/edition_assertions.rb
```diff
@@ -14,11 +14,23 @@ def initialize(edition, assertion)
     end
   end
 
+  class FeatureError < Error
+    def initialize(edition, assertion)
+      super(edition, "edition.document_type #{assertion}")
```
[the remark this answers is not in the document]
benthorner: Sure. I've rewritten this method as discussed, although I've avoided modifying the assertion text, as the fact that it's a "FeatureError" should provide the necessary context.

lib/edition_assertions.rb
benthorner: Updated as discussed.

benthorner:
@kevindew this should be ready for another look.

app/controllers/application_controller.rb
kevindew: The counter argument would be that it's a consistent way to handle bad requests across the app when we're not concerned about a particular view. If anyone happens to get this response rendered to their browser they'll just get a blank screen.

lib/edition_assertions.rb
```diff
@@ -14,11 +14,25 @@ def initialize(edition, assertion)
     end
   end
 
+  class FeatureError < Error
+    def initialize(edition, assertion)
+      super(edition, assertion || "edition supports feature")
```
kevindew: This sounds positive but don't we want this to be negative as we raise this when the document type doesn't support a feature rather than it does? 

I notice the state one also seems to be positive so I'm not sure if I've missed something.

benthorner: Might help to point out these get interpolated into `"Assertion failed: #{assertion}"`.

app/controllers/application_controller.rb
benthorner: True. I thought we were being deliberately inconsistent here, given the modal context, but that's not the case. Updated.

lib/edition_assertions.rb
kevindew: Oh yes, that does make sense.

PR #1810, opened 2020-02-24
Enable Group 1 Publication formats for pre-release users

leenagupte:
Trello: https://trello.com/c/KI9c4Uue

## What's changed?
Allows users with `pre_release_features` to select and create documents in group 1 publications:

- correspondence
- independent report
- foi release
- corporate report
- transparency
- research and analysis
- promotional material
- impact assessment
- policy paper

### User with "pre_release_features"
Screenshot 2020-02-24 at 14 52 39

Screenshot 2020-02-24 at 14 53 19

Screenshot 2020-02-24 at 14 53 51

### Everyone else
Screenshot 2020-02-24 at 14 49 53

Screenshot 2020-02-24 at 14 50 33

### DIFF
--- a/app/models/document_type_selection/option.rb
+++ b/app/models/document_type_selection/option.rb
@@ -1,7 +1,7 @@
 class DocumentTypeSelection::Option
   include InitializeWithHash
 
-  attr_reader :id, :type, :hostname, :path
+  attr_reader :id, :type, :hostname, :path, :pre_release
 
   def document_type_selection?
     type == "document_type_selection"
@@ -18,4 +18,6 @@ class DocumentTypeSelection::Option
   def managed_elsewhere_url
     hostname ? Plek.new.external_url_for(hostname) + path : path
   end
+
+  alias_method :pre_release?, :pre_release
 end
--- a/app/models/document_type_selection/option.rb
+++ b/app/models/document_type_selection/option.rb
@@ -1,7 +1,7 @@
 class DocumentTypeSelection::Option
   include InitializeWithHash
 
-  attr_reader :id, :type, :hostname, :path, :pre_release
+  attr_reader :id, :type, :hostname, :path
 
   def document_type_selection?
     type == "document_type_selection"
@@ -19,5 +19,9 @@ class DocumentTypeSelection::Option
     hostname ? Plek.new.external_url_for(hostname) + path : path
   end
 
-  alias_method :pre_release?, :pre_release
+  def pre_release?
+    return false unless document_type?
+
+    DocumentType.find(id).pre_release?
+  end
 end
--- a/config/document_types.yml
+++ b/config/document_types.yml
@@ -54,3 +54,263 @@
     - id: world_locations
       type: multi_tag
       document_type: world_location
+
+- id: correspondence
+  label: Correspondence
+  path_prefix: /government/publications
+  images: false
+  publishing_metadata:
+    schema_name: publication
+    rendering_app: government-frontend
+  contents:
+    - title_and_base_path
+    - summary
+    - body
+  tags:
+    - id: primary_publishing_organisation
+      type: single_tag
+      document_type: organisation
+    - id: organisations
+      type: multi_tag
+      document_type: organisation
+    - id: role_appointments
+      type: multi_tag
+      document_type: role_appointment
+    - id: topical_events
+      type: multi_tag
+      document_type: topical_event
+    - id: world_locations
+      type: multi_tag
+      document_type: world_location
+
+- id: corporate_report
+  label: Corporate report
+  path_prefix: /government/publications
+  images: false
+  check_path_conflict: true
+  publishing_metadata:
+    schema_name: publication
+    rendering_app: government-frontend
+  contents:
+    - title_and_base_path
+    - summary
+    - body
+  tags:
+    - id: primary_publishing_organisation
+      type: single_tag
+      document_type: organisation
+    - id: organisations
+      type: multi_tag
+      document_type: organisation
+    - id: role_appointments
+      type: multi_tag
+      document_type: role_appointment
+    - id: topical_events
+      type: multi_tag
+      document_type: topical_event
+    - id: world_locations
+      type: multi_tag
+      document_type: world_location
+
+- id: foi_release
+  label: FOI Release
+  path_prefix: /government/publications
+  images: false
+  check_path_conflict: true
+  publishing_metadata:
+    schema_name: publication
+    rendering_app: government-frontend
+  contents:
+    - title_and_base_path
+    - summary
+    - body
+  tags:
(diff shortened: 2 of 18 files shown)


leenagupte:
> * feature test for the new schema

Cheers. I saw the acceptance criteria for the feature tests, but wasn't sure what they referred too!

I tried to DRY up the yaml, but I ended up breaking every test. I don't think it's possible with the YAML in the format it is now.

I think the format would have to change from:

```
- id: correspondence
  label: Correspondence
  pre_release: true
  path_prefix: /government/publications
  images: false
.
.
.
```
to:
```
correspondence:
  label: Correspondence
  pre_release: true
  path_prefix: /government/publications
  images: false
```
But if we do that, then the schema becomes complex, so I left it out so we'd have time to think through the options.

benthorner:
> But if we do that, then the schema becomes complex, so I left it out so we'd have time to think through the options.

Oh dear, you're right! I had my Docker experience in mind when writing that. It's still possible, and the simplest approach would be to nest the whole array in a `document_types` key, which means the top-level YAML becomes a hash (like we need it to be).

config/document_types.yml
```diff
@@ -27,12 +26,12 @@
       type: multi_tag
       document_type: world_location
 
-- id: press_release
-  label: Press release
-  path_prefix: /government/news
-  images: true
+publication: &publication
+  pre_release: true
```
kevindew: It seems a shame that we have to manage this pre_release flag in two places (here and document_type_selection).

Could we have it so a document type selection is able to look up the corresponding document_type and question whether that is pre release rather than needing to maintain two lists?

leenagupte: Yeah. That is something that occurred to me too, after the functionality had to be duplicated  for DocumentTypes.

I'll take a look and see what I come up with.

spec/views/documents/index/_filters.html.erb_spec.rb
```diff
@@ -23,4 +23,31 @@
...
+
+    before do
+      stub_publishing_api_has_linkables([], document_type: "organisation")
+
+      allow(DocumentType)
+        .to receive(:all)
+        .and_return([pre_release_document_type, document_type])
+    end
+
+    it "shows lists document types when the user has pre_release_features permissions" do
```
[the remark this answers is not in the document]
leenagupte: 🤦‍♂

leenagupte:
> One thing I wasn't sure whether it was in the scope of this PR/story was whether pre-release document types would be excluded from the [edition filter](https://github.com/alphagov/content-publisher/blob/61f8f5c5cc538ae75d85bb95ac258b53dbd36f12/lib/edition_filter.rb) ?
> 

Is edition filter what's used to populate the search results? I'm not sure. I don't think that was part of this card.

kevindew:
> > One thing I wasn't sure whether it was in the scope of this PR/story was whether pre-release document types would be excluded from the [edition filter](https://github.com/alphagov/content-publisher/blob/61f8f5c5cc538ae75d85bb95ac258b53dbd36f12/lib/edition_filter.rb) ?
> 
> Is edition filter what's used to populate the search results? I'm not sure. I don't think that was part of this card.

Ah yes it doesn't seem to be mentioned on the card. We should do it though - maybe put in another story to do so as it should be a small self contained task that is easy for someone to pick up.

leenagupte:
@kevindew @benthorner Thanks for your comments. They've been addressed. This is ready for a re-review

app/models/document_type_selection/option.rb
```diff
@@ -18,4 +18,8 @@ def managed_elsewhere?
   def managed_elsewhere_url
     hostname ? Plek.new.external_url_for(hostname) + path : path
   end
+
+  def pre_release?
+    DocumentType.find(id).pre_release? if document_type?
```
kevindew: Nice job 👍 

Since this method should only return a bool should we do either:

```
return false unless document_type?
DocumentType.find(id).pre_release?
```

or:

```
document_type? ? DocumentType.find(id).pre_release? : false
```

config/locales/en/document_types.yml
```diff
@@ -24,6 +24,116 @@ en:
...
+          guidance:
+            title: Creating a title
+            body_govspeak: This does not have to be the full publication title. The title must be unique and specific, and make clear what the content offers users in their words.
+        summary:
+          guidance:
+            title: Writing a summary
+            body_govspeak: The summary explains the main point of the content and it should end with a full stop. Keep it short, possibly under 160 characters, and avoid repeating the first line of the body.
+        body:
+          guidance:
+            title: Writing the details for a policy paper
```
kevindew: I think this should say corporate report

spec/features/formats/publication_spec.rb
```diff
@@ -0,0 +1,31 @@
+RSpec.describe "Publication format" do
```
kevindew: I wonder if should mention the word schema in here. I assume we've already got precedence in news article schema so could well be a problem for another day. Could be "format using publication schema"

spec/features/formats/publication_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+  include TopicsHelper
```
kevindew: I don't think we actually use the methods from the helper.

spec/features/formats/publication_spec.rb
leenagupte: Yeah, I followed the convention set in the news article test:
```
RSpec.describe "News article format" do
```

spec/views/documents/index/_filters.html.erb_spec.rb
```diff
@@ -23,4 +23,31 @@
       end
     end
   end
+
+  describe "Document type select" do
+    let(:pre_release_document_type) { build(:document_type, pre_release: true) }
+    let(:document_type) { build(:document_type) }
```
benthorner: Do we need this? I can't see it being used anywhere.

spec/models/document_type_selection/option_spec.rb
```diff
@@ -56,4 +56,23 @@
       expect(described_class.new(option).document_type?).to be true
     end
   end
+
+  describe "#pre_release?" do
+    it "returns true when the option is an existing document type and a pre-release feature" do
+      option = {
+        "id" => "policy_paper",
```
benthorner: I think we should do some stubbing here, as this test will start failing when we release publications, but I believe we want to retain the functionality for future document types.

leenagupte:
@kevindew @benthorner Thanks for your comments again. I've made the suggested changes and fixed up.

spec/features/formats/publication_spec.rb
kevindew: Sure, can fix them both in another day. Not too big a deal.

PR #1815, opened 2020-02-25
Add instructions for running a migration from Whitehall

brucebolt:
These include instructions for running the migration locally without needing to replicate data.

Trello card: https://trello.com/c/W6Ay4rYb

### DIFF
--- a/docs/import-from-whitehall.md
+++ b/docs/import-from-whitehall.md
@@ -1,0 +1,129 @@
+# Importing documents from Whitehall to Content Publisher
+
+There are two ways to import documents from Whitehall:
+- [a single document](#single-document-import)
+- [documents of a specific type or subtypes from a single publishing organisation](#import-documents-of-a-specific-type-and-subtypes-from-a-single-publishing-organisation)
+
+## Running an import
+
+### Single document import
+
+This will import a single document and all its editions from Whitehall into Content Publisher.
+
+```
+rake import:whitehall_document[1234]
+```
+
+Where `1234` is the Whitehall document ID.  After running the rake task, a URL will be printed to the console which allows the status of import to be examined.
+
+Note that the ID in Whitehall's internal URLs is the edition number (e.g. for `/government/admin/publications/1034451` the edition is `1034451`).  The document ID can be obtained on a Whitehall Rails console, for example:
+
+```
+irb> URI.parse("https://whitehall-admin.integration.publishing.service.gov.uk/government/admin/publications/1038279").path.split("/")[4].then { |id| Edition.find(id).document_id }
+=> 412144
+```
+
+### Import documents of a specific type and subtypes from a single publishing organisation
+
+To import a single document type (e.g. news articles) from a single publisher (e.g. NDA):
+
+```
+rake import:whitehall_migration["nuclear-decommissioning-authority","news_article"]
+```
+
+To import multiple document subtypes (e.g. news stories and press releases) from a single publisher (e.g. NDA):
+
+```
+rake import:whitehall_migration["nuclear-decommissioning-authority","news_article","news_story,press_release"]
+```
+
+There is no limit to the number of subtypes (additional values are added comma separated).
+
+It is also possible to import a single document subtype (e.g. press releases) from a single publisher (e.g. NDA):
+
+```
+rake import:whitehall_migration["nuclear-decommissioning-authority","news_article","press_release"]
+```
+
+After running the rake task, a URL will be printed to the console which allows the status of import to be examined.
+
+## Running a migration locally using govuk-docker
+
+Using [govuk-docker](https://github.com/alphagov/govuk-docker), you can run an import locally without needing to replicate data.  In order to do this, you will need to override the [Plek](https://github.com/alphagov/plek) URL for Whitehall and Publishing API to point to integration.  Additionally, you will require [bearer tokens](#getting-bearer-tokens).
+
+Outside the GDS office, you must be connected to the GDS Developer VPN.  Inside the GDS office, you must be connected to the Brattain network (not GovWifi).  This is due to IP address restrictions on the applications that will be connected to.
+
+### Getting bearer tokens
+
+Tokens can be obtained by creating a new API user in Signon.  These allow access to the relevant remote applications without needing to enter a username and password.
+
+1. Navigate to https://signon.integration.publishing.service.gov.uk/api_users
+1. Click 'Create API user'
+1. Enter a name (e.g. your name followed by 'Whitehall Import to Content Publisher Migration Testing') and an email address
+1. Once the user is created, click 'Add application token'
+1. Select 'Whitehall', then add the 'Export data' permission
+1. Copy the Whitehall token - you will not be able to access this again
+1. Click 'Add application token' again
+1. Select 'Publishing API'
+1. Copy the Publishing API token - you will not be able to access this again
+
+These tokens can then be set as environment variables:
+
+```
+export whitehall_bearer_token=<>
+export publishing_api_bearer_token=<>
+```
+
+Replace `<>` with the bearer tokens you have previously obtained.
+
+### Overriding Plek URLs
+
+By default, Plek will point to your local development environment when running rake tasks in govuk-docker.  Therefore, you will need to override the Plek URLs.
+
+E.g., for integration:
+
+```
+export publishing_api_url=https://publishing-api.blue.integration.govuk.digital
+export whitehall_url=https://whitehall-admin.integration.publishing.service.gov.uk
+```
+
+These environment variables will be passed into govuk-docker in later steps.
+
+### Setting up Sidekiq
+
+In order to run a migration locally, Sidekiq must be running when the rake tasks are executed.  As jobs are run independently of the rake task, Plek overrides and bearer tokens are also required here.
+
+```
+govuk-docker run -e PLEK_SERVICE_PUBLISHING_API_URI=$publishing_api_url -e PUBLISHING_API_BEARER_TOKEN=$publishing_api_bearer_token -e PLEK_SERVICE_PUBLISHING_API_URI=$publishing_api_urlPLEK_SERVICE_WHITEHALL_ADMIN_URI=$whitehall_url -e WHITEHALL_BEARER_TOKEN=$whitehall_bearer_token content-publisher-worker bundle exec sidekiq -C config/sidekiq.yml
+```
+
+### Running a migration
+
+Either a single document or multiple document import can be run, as detailed in previous sections.
+
+For example:
+
+```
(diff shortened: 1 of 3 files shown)


docs/import-from-whitehall.md
```diff
@@ -0,0 +1,110 @@
...
+rake import:whitehall_document[1234]
+```
+
+Where `1234` is the Whitehall document ID.  After running the rake task, a URL will be printed to the console which allows the status of import to be examined.
+
+Note that the ID in Whitehall's internal URLs is the edition number (e.g. for `/government/admin/publications/1034451` the edition is `1034451`).  The document ID can be obtained on a Rails console, using:
+
+```
+Edition.find(1034451).document.id
+```
```
kevindew: Might be good to reference that you have to be using a Whitehall Rails console and it needs to be in the environment. We could also give a command that takes the whole URL rather than giving people instructions of what to pull out the URL. For example:

```
irb(main):005:0> URI.parse("https://whitehall-admin.integration.publishing.service.gov.uk/government/admin/publications/1038279").path.split("/")[4].then { |id| Edition.find(id).document_id }
=> 412144
```

docs/import-from-whitehall.md
```diff
@@ -0,0 +1,110 @@
...
+
+```
+rake import:whitehall_migration["nuclear-decommissioning-authority","news_article","press_release"]
+```
+
+After running the rake task, a URL will be printed to the console which allows the status of import to be examined.
+
+## Running a migration locally using govuk-docker
+
+Using [govuk-docker](https://github.com/alphagov/govuk-docker), you can run an import locally without needing to replicate data.  In order to do this, you will need to override the [Plek](https://github.com/alphagov/plek) URL for Whitehall and Publishing API to point to staging.  Additionally, you will require the bearer tokens for these applications from [govuk-secrets](https://github.com/alphagov/govuk-secrets).
```
kevindew: Is there a particular reason we suggest staging rather than integration?

docs/import-from-whitehall.md
```diff
@@ -0,0 +1,110 @@
...
+Tokens can be obtained from govuk-secrets.  These allow access to the relevant remote applications without needing to enter a username and password.
+
+```
+cd govuk-secrets/puppet_aws
+bundle exec rake 'eyaml:edit[staging,apps]'
+```
+
+The values you need to obtain are:
+- `govuk::apps::content_publisher::whitehall_bearer_token` - the Whitehall bearer token
+- `govuk::apps::content_publisher::publishing_api_bearer_token` - the Publishing API bearer token
```
kevindew: We shouldn't really be pulling out these tokens as these are the production ones and we should be really cautious that these could accidentally be leaked if used as part of dev.

What I'd suggest we should do is that rather than pulling out tokens, developers should be creating their own tokens from signon. This can be done by going to: https://signon.integration.publishing.service.gov.uk/api_users and going to create new API user.

docs/import-from-whitehall.md
```diff
@@ -0,0 +1,110 @@
...
+
+### Setting up Sidekiq
+
+In order to run a migration locally, Sidekiq must be running when the rake tasks are executed.  As jobs are run independently of the rake task, Plek overrides and bearer tokens are also required here.
+
+```
+govuk-docker run -e PLEK_SERVICE_PUBLISHING_API_URI=https://publishing-api.staging.publishing.service.gov.uk -e PUBLISHING_API_BEARER_TOKEN=<> -e PLEK_SERVICE_WHITEHALL_ADMIN_URI=https://whitehall-admin.staging.publishing.service.gov.uk -e WHITEHALL_BEARER_TOKEN=<> content-publisher-worker bundle exec sidekiq -C config/sidekiq.yml
+```
```
kevindew: This looks really helpful 👍

docs/import-from-whitehall.md
```diff
@@ -0,0 +1,110 @@
...
+
+Replace `<>` with the bearer tokens you have previously obtained.
```
kevindew: I wonder if it'd be a bit clearer if we used example vars like $whitehall_bearer_token and $publishing_api_bearer_token.

Could also tell users how to export them first with 

```
export whitehall_bearer_token=<your token>
export publishing_api_bearer_token=<your token>
```

and then the longer commands could be pasted verbatim

docs/import-from-whitehall.md
```diff
@@ -0,0 +1,110 @@
...
+
+Where `1234` is the Whitehall document ID.  After running the rake task, a URL will be printed to the console which allows the status of import to be examined.
+
+Note that the ID in Whitehall's internal URLs is the edition number (e.g. for `/government/admin/publications/1034451` the edition is `1034451`).  The document ID can be obtained on a Rails console, using:
+
+```
+Edition.find(1034451).document.id
+```
+
+### Import documents of a specific type (and optional subtypes) from a single publishing organisation
```
kevindew: I feel like we should suggest that people pass in the subtypes by default rather than saying it's optional.

This seems to have caught Emma out (see https://trello.com/c/nibnOHkx/1538-dont-export-government-responses-from-whitehall) where I've suggested we might want to do more with defaults.

docs/import-from-whitehall.md
```diff
@@ -0,0 +1,110 @@
...
+
+### Running a migration
+
+Either a single document or multiple document import can be run, as detailed in previous sections.
+
+For example:
+
+```
+govuk-docker-run rake import:whitehall_migration["nuclear-decommissioning-authority","news_article"] env PLEK_SERVICE_PUBLISHING_API_URI=https://publishing-api.staging.publishing.service.gov.uk PUBLISHING_API_BEARER_TOKEN=<> PLEK_SERVICE_WHITEHALL_ADMIN_URI=https://whitehall-admin.staging.publishing.service.gov.uk WHITEHALL_BEARER_TOKEN=<>
+```
```
kevindew: There's a few inconsistencies with this command with the previous one that seem like something that could catch someone out. We need to be in a specific directory in this example and do environment variables differently.

Might be good to switch it to be consistent, for example:

```
govuk-docker run -e PLEK_SERVICE_PUBLISHING_API_URI=https://publishing-api.staging.publishing.service.gov.uk -e PUBLISHING_API_BEARER_TOKEN=<> -e PLEK_SERVICE_WHITEHALL_ADMIN_URI=https://whitehall-admin.staging.publishing.service.gov.uk -e WHITEHALL_BEARER_TOKEN=<> bundle exec rake import:whitehall_migration["nuclear-decommissioning-authority","news_article"]
```

docs/import-from-whitehall.md
```diff
@@ -0,0 +1,110 @@
...
+
+Replace `<>` with the bearer tokens you have previously obtained.
+
+In order to access the URL printed at the end of the rake task, the Content Publisher application will need to be running locally:
+
+```
+govuk-docker-up
```
kevindew: This one is also directory specific could switch it to `govuk-docker up content-publisher-app`

docs/import-from-whitehall.md
```diff
@@ -0,0 +1,110 @@
+# Importing documents from Whitehall to Content Publisher
+
+There are two ways to import documents from Whitehall:
+- a single document
+- documents of a specific type or subtypes from a single publishing organisation
```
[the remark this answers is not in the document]
brucebolt: Yes, I've added internal links

docs/import-from-whitehall.md
brucebolt: Updated the text above to say it must be a Whitehall Rails console.  Also changed the code to parse a URL.

docs/import-from-whitehall.md
brucebolt: I was previously having trouble accessing Publishing API in integration.

However, it appears I should've been using https://publishing-api.blue.integration.govuk.digital rather than https://publishing-api.integration.publishing.service.gov.uk.  Have updated the document to now use integration.

docs/import-from-whitehall.md
brucebolt: That's a good point.  I've change this to give instructions for creating an API user and getting that user's tokens.

docs/import-from-whitehall.md
brucebolt: Updated throughout.  Also setting the URLs in the same way to reduce the length of the commands.

docs/import-from-whitehall.md
brucebolt: I've removed the words optional.  We can update this if any changes are made to the optional nature of the subtypes.

docs/import-from-whitehall.md
brucebolt: Changed this to be consistent with the other commands.

docs/import-from-whitehall.md
brucebolt: Changed this one too.

brucebolt:
Thank you for the review @kevindew.  I've addressed all your comments, and worked out how to get the import working against the integration environment.

docs/import-from-whitehall.md
```diff
@@ -0,0 +1,136 @@
...
+export publishing_api_url=https://publishing-api.blue.integration.govuk.digital
+export whitehall_url=https://whitehall-admin.integration.publishing.service.gov.uk
+```
+
+For staging:
+
+```
+export publishing_api_url=https://publishing-api.staging.publishing.service.gov.uk
+export whitehall_url=https://whitehall-admin.staging.publishing.service.gov.uk
+```
```
kevindew: Does the `blue.staging.govuk.digital` work for this? it seems rather weird to be using the carrenza URL here and not sure if it's intended to be supported or not. Life might be simpler if we can use the govuk.digital one in case it's a mistake that this different one is available.

docs/import-from-whitehall.md
```diff
@@ -0,0 +1,136 @@
...
+
+In order to access the URL printed at the end of the rake task, the Content Publisher application will need to be running locally:
+
+```
+govuk-docker up content-publisher-app
+```
+
+### Accessing a Rails console
+
+You may wish to open a Rails console that allows you to perform tasks on Whitehall documents (e.g. unlock or lock a document in order to test the import task multiple times).
```
kevindew: You might want to link to https://github.com/alphagov/whitehall/blob/master/docs/migration_to_content_publisher/locked-documents.md for the "unlock or lock documents" since that is potentially unfamiliar terminology.

docs/import-from-whitehall.md
brucebolt: Unfortunately not.  This can be seen by visiting https://publishing-api.staging.publishing.service.gov.uk/healthcheck which gives a 200, whereas http://publishing-api.blue.staging.govuk.digital/healthcheck times out.

docs/import-from-whitehall.md
brucebolt: Added a link to this doc

docs/import-from-whitehall.md
kevindew: If you use the https equivalent of the latter it works.

brucebolt: Ah, I didn't see that.  Have updated.

PR #1818, opened 2020-02-26
Store the cumulative change history of all document editions as each new edition is created 

gclssvglx:
We have previously created a JSON column in which to store the change history of all editions of a document.  We have also added validation of the fields and structure of the JSON data to be stored.  

We now need to ensure that as each new edition is created - the change history is updated with the superseded edition's change note(s) - if appropriate - and store them with the newly created edition.

This change adds that functionality by introducing a service object (`CreateNextEditionService`) and updating the existing code where required.

Trello: https://trello.com/c/xMRotBoK/1460-create-a-changehistory-json-column-on-metadatarevisions-and-populate-it-when-new-editions-are-added

### DIFF
--- a/app/interactors/editions/create_interactor.rb
+++ b/app/interactors/editions/create_interactor.rb
@@ -4,13 +4,11 @@ class Editions::CreateInteractor < ApplicationInteractor
            :edition,
            :next_edition,
            :discarded_edition,
-           :next_revision,
            to: :context
 
   def call
     Edition.transaction do
       find_and_lock_edition
-      create_next_revision
       create_next_edition
       create_timeline_entry
       preview_next_edition
@@ -21,38 +19,15 @@ private
 
   def find_and_lock_edition
     context.edition = Edition.lock.find_current(document: params[:document])
-
-    assert_edition_state(edition, assertion: "can create new edition") do
-      edition.live || edition.discarded?
-    end
-  end
-
-  def create_next_revision
-    updater = Versioning::RevisionUpdater.new(edition.revision, user)
-
-    updater.assign(change_note: "",
-                   update_type: "major",
-                   proposed_publish_time: nil)
-
-    context.next_revision = updater.next_revision
+    context.discarded_edition = Edition.find_by(document: edition.document,
+                                                number: edition.number + 1)
+    assert_edition_state(edition, assertion: "can create new edition") { edition.live }
   end
 
   def create_next_edition
-    edition.update!(current: false)
-
-    context.discarded_edition = Edition.find_by(
-      document: edition.document,
-      number: edition.number + 1,
-    )
-
-    context.next_edition = discarded_edition ||
-      Edition.new(document: edition.document,
-                  number: edition.document.next_edition_number,
-                  created_by: user)
-
-    EditDraftEditionService.call(next_edition, user, current: true, revision: next_revision)
-    AssignEditionStatusService.call(next_edition, user: user, state: :draft)
-    next_edition.save!
+    context.next_edition = CreateNextEditionService.call(current_edition: edition,
+                                                         user: user,
+                                                         discarded_edition: discarded_edition)
   end
 
   def create_timeline_entry
--- a/app/services/create_next_edition_service.rb
+++ b/app/services/create_next_edition_service.rb
@@ -1,0 +1,51 @@
+class CreateNextEditionService < ApplicationService
+  def initialize(current_edition:, user:, discarded_edition: nil)
+    @current_edition = current_edition
+    @user = user
+    @discarded_edition = discarded_edition
+  end
+
+  def call
+    raise "Can only create a next edition from a live edition" unless current_edition.live
+
+    current_edition.update!(current: false)
+    EditDraftEditionService.call(next_edition, user, current: true, revision: next_revision)
+    AssignEditionStatusService.call(next_edition, user: user, state: :draft)
+    next_edition.save!
+    next_edition
+  end
+
+private
+
+  attr_reader :current_edition, :user, :discarded_edition
+
+  def next_edition
+    @next_edition ||= begin
+      document = current_edition.document
+      discarded_edition || Edition.new(document: document,
+                                       number: document.next_edition_number,
+                                       created_by: user)
+    end
+  end
+
+  def next_revision
+    updater = Versioning::RevisionUpdater.new(current_edition.revision, user)
+    updater.assign(change_note: "",
+                   update_type: "major",
+                   proposed_publish_time: nil,
+                   change_history: change_history)
+    updater.next_revision
+  end
+
+  def change_history
+    if !current_edition.major? || current_edition.change_note.empty? || current_edition.first?
+      return current_edition.change_history
+    end
+
+    current_edition.change_history.prepend(
+      "id" => SecureRandom.uuid,
+      "note" => current_edition.change_note,
+      "public_timestamp" => current_edition.published_at.rfc3339,
+    )
+  end
+end
--- a/spec/interactors/editions/create_interactor_spec.rb
+++ b/spec/interactors/editions/create_interactor_spec.rb
@@ -41,20 +41,21 @@ RSpec.describe Editions::CreateInteractor do
       let(:live_edition) { create(:edition, :published) }
       let(:params) { { document: live_edition.document.to_param } }
 
-      let!(:edition) do
+      let!(:discarded_edition) do
         create(:edition,
                state: "discarded",
                current: false,
                document: live_edition.document)
       end
 
-      it "resumes the edition" do
-        next_edition = described_class
-          .call(params: params, user: user)
-          .next_edition
-
-        expect(next_edition.number).to eq edition.number
-        expect(next_edition).to eq edition.reload
+      it "delegates to the CreateNextEditionService" do
+        expect(CreateNextEditionService)
+          .to receive(:call)
+          .with(current_edition: live_edition,
+                user: user,
+                discarded_edition: discarded_edition)
+          .and_call_original
+        described_class.call(params: params, user: user)
       end
 
       it "creates a timeline entry" do
@@ -68,18 +69,16 @@ RSpec.describe Editions::CreateInteractor do
       end
     end
 
-    context "when the edition is live" do
-      let(:edition) { create(:edition, live: true, number: 2) }
+    context "when there is not a discarded edition" do
+      let(:edition) { create(:edition, :published, number: 2) }
       let(:params) { { document: edition.document.to_param } }
 
-      it "creates a new edition" do
-        next_edition = described_class
-          .call(params: params, user: user)
-          .next_edition
-
-        expect(next_edition).not_to eq edition.reload
-        expect(next_edition.number).to eq 3
-        expect(next_edition.created_by).to eq user
+      it "delegates to the CreateNextEditionService" do
+        expect(CreateNextEditionService)
+          .to receive(:call)
+          .with(current_edition: edition, user: user, discarded_edition: nil)
+          .and_call_original
+        described_class.call(params: params, user: user)
       end
 
       it "creates a timeline entry" do
--- a/spec/services/create_next_edition_service_spec.rb
+++ b/spec/services/create_next_edition_service_spec.rb
@@ -1,0 +1,145 @@
+RSpec.describe CreateNextEditionService do
+  describe ".call" do
+    let(:user) { create :user }
+
+    before do
+      populate_default_government_bulk_data
+    end
+
+    it "aborts if the current edition isn't live" do
+      current_edition = create(:edition, live: false)
+
+      expect { described_class.call(current_edition: current_edition, user: user) }
+        .to raise_error("Can only create a next edition from a live edition")
+    end
+
+    it "returns a new current edition" do
+      current_edition = create(:edition, :published, number: 2)
+
+      next_edition = described_class.call(current_edition: current_edition,
+                                          user: user)
+
+      expect(current_edition).not_to be_current
+      expect(next_edition).to be_current
+    end
+
+    it "updates the edition number of the new current edition" do
+      current_edition = create(:edition, :published, number: 2)
+
+      next_edition = described_class.call(current_edition: current_edition,
+                                          user: user)
+
+      expect(next_edition.number).to eq 3
+      expect(next_edition.created_by).to eq user
+    end
+
+    it "returns a draft edition" do
+      current_edition = create(:edition, :published)
+
+      next_edition = described_class.call(current_edition: current_edition,
+                                          user: user)
+
+      expect(next_edition).to be_draft
+    end
+
+    it "resets change_note, update_type and propoposed_publish_time values" do
+      current_edition = create(:edition,
+                               :published,
+                               change_note: "note",
+                               update_type: "minor",
+                               proposed_publish_time: Time.zone.now)
+
+      next_edition = described_class.call(current_edition: current_edition,
+                                          user: user)
+
+      expect(next_edition.change_note).to be_empty
+      expect(next_edition.update_type).to eq("major")
+      expect(next_edition.proposed_publish_time).to be_nil
+    end
+
+    it "appends the change note details when there is a change note and a major change" do
+      current_edition = create(:edition,
+                               :published,
+                               number: 2,
+                               published_at: Date.yesterday.noon,
+                               change_note: "note",
+                               update_type: "major")
+
+      next_edition = described_class.call(current_edition: current_edition,
+                                          user: user)
+
+      expect(next_edition.change_history.first).to match a_hash_including(
+        "note" => "note",
+        "public_timestamp" => Date.yesterday.noon.rfc3339,
+      )
+    end
+
+    it "does not update the change history when the current edition is not a major change" do
+      current_edition = create(:edition,
+                               :published,
+                               number: 2,
+                               published_at: Date.yesterday.noon,
+                               change_note: "note",
+                               update_type: "minor")
+
+      next_edition = described_class.call(current_edition: current_edition,
+                                          user: user)
+
+      expect(next_edition.change_history).to eq(current_edition.change_history)
+    end
+
+    it "does not update the change history when the current edition lacks a change note" do
+      current_edition = create(:edition,
+                               :published,
+                               number: 2,
+                               change_note: "")
+
+      next_edition = described_class.call(current_edition: current_edition,
+                                          user: user)
+
+      expect(next_edition.change_history).to be_empty
+    end
+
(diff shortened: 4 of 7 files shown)


app/services/create_next_edition_service.rb
```diff
@@ -0,0 +1,51 @@
...
+    updater = Versioning::RevisionUpdater.new(current_edition.revision, user)
+    updater.assign(change_note: "",
+                   update_type: "major",
+                   proposed_publish_time: nil,
+                   change_history: change_history)
+    updater.next_revision
+  end
+
+  def change_history
+    if !current_edition.major? || current_edition.change_note.empty? || current_edition.first?
```
1pretz1: Instead of `!current_edition.major?` you could just do `current_edition.minor?`

spec/services/create_next_edition_service_spec.rb
```diff
@@ -0,0 +1,118 @@
...
+    it "returns a draft edition" do
+      current_edition = create(:edition, :published)
+
+      next_edition = described_class.call(current_edition: current_edition,
+                                          user: user)
+
+      expect(next_edition.draft?).to be(true)
+    end
+
+    it "resets change_note, update_type, and propoposed_publish_time values" do
```
1pretz1: Is there an extra comma just after update_type?

spec/services/create_next_edition_service_spec.rb
```diff
@@ -0,0 +1,118 @@
...
+      current_edition = create(:edition, :published)
+
+      next_edition = described_class.call(current_edition: current_edition,
+                                          user: user)
+
+      expect(next_edition.change_note).to eq("")
```
1pretz1: You can use `be_empty` here which reads a little nicer -> `expect(next_edition.change_note).to be_empty`

spec/services/create_next_edition_service_spec.rb
```diff
@@ -0,0 +1,118 @@
...
+    end
+
+    it "aborts if the current edition isn't live" do
+      current_edition = create(:edition, live: false)
+
+      expect { described_class.call(current_edition: current_edition, user: user) }
+        .to raise_error("Can only create a next edition from a live edition")
+    end
+
+    it "updates the documents current edition" do
```
kevindew: this test name doesn't seem to correlate with what's being done. It might be better phrased as `it "returns a new current edition" and then assert that both next edition is current and the previous current edition is not.

spec/services/create_next_edition_service_spec.rb
```diff
@@ -0,0 +1,118 @@
...
+      next_edition = described_class.call(current_edition: current_edition,
+                                          user: user)
+
+      expect(next_edition.change_history.first).to match a_hash_including(
+        "note" => "note",
+        "public_timestamp" => Date.yesterday.noon.rfc3339,
+      )
+    end
+
+    it "maintains the change history when the current edition is not a major change" do
```
1pretz1: Should we mention in the `it` how the edition contains a change note too as it's a little surprising from reading the it statement that the edition also contains a change note when it's a minor change? The example above mentions both `it "appends the change note details when there is a change note and a major change" do`

spec/services/create_next_edition_service_spec.rb
```diff
@@ -0,0 +1,118 @@
...
+                               :published,
+                               number: 2,
+                               published_at: Date.yesterday.noon,
+                               change_note: "note",
+                               update_type: "minor")
+
+      next_edition = described_class.call(current_edition: current_edition,
+                                          user: user)
+
+      expect(next_edition.change_history).to eq([])
```
1pretz1: You can also use `be_empty` for empty arrays too which may read a little nicer, I've noticed a few expectations this could apply to but I'll just leave this one comment!

spec/services/create_next_edition_service_spec.rb
```diff
@@ -0,0 +1,118 @@
...
+        live_edition = create(:edition, :published)
+        discarded_edition = create(:edition,
+                                   state: "discarded",
+                                   current: false,
+                                   document: live_edition.document)
+
+        next_edition = described_class.call(current_edition: live_edition,
+                                            user: user, discarded_edition: discarded_edition)
+
+        expect(next_edition.state).to eq "draft"
```
1pretz1: You could do `expect(next_edition).to be_draft` here to tap into some rspec wizardry

spec/services/create_next_edition_service_spec.rb
```diff
@@ -0,0 +1,118 @@
...
+      current_edition = create(:edition, :published)
+
+      next_edition = described_class.call(current_edition: current_edition,
+                                          user: user)
+
+      expect(next_edition.draft?).to be(true)
+    end
+
+    it "resets change_note, update_type, and propoposed_publish_time values" do
+      current_edition = create(:edition, :published)
```
kevindew: It would be good to set change, note update type and proposed publish time on this so that it's clear that something actually changed

spec/interactors/editions/create_interactor_spec.rb
```diff
@@ -69,7 +69,7 @@
     end
 
     context "when the edition is live" do
-      let(:edition) { create(:edition, live: true, number: 2) }
+      let(:edition) { create(:edition, live: true, number: 2, published_at: Time.zone.now) }
```
1pretz1: Could you just do `create(:edition, :published, number: 2)` here? As the published trait sets live to true and populates the relevant timestamps like `published_at`

spec/services/create_next_edition_service_spec.rb
```diff
@@ -0,0 +1,118 @@
...
+                                          user: user)
+
+      expect(next_edition.change_history).to eq([])
+    end
+
+    it "maintains the change history when current edition lacks a change note" do
+      current_edition = create(:edition,
+                               :published,
+                               number: 2,
+                               published_at: Date.yesterday.noon,
```
kevindew: I don't think you need the published_at attribute here.

spec/services/create_next_edition_service_spec.rb
```diff
@@ -0,0 +1,118 @@
...
+                                          user: user)
+
+      expect(next_edition.change_history).to eq([])
+    end
+
+    context "when the current edition is the first edition" do
+      it "doesn't append the change note details" do
+        current_edition = create(:edition,
+                                 :published,
+                                 published_at: Date.yesterday.noon,
```
kevindew: similar regarding published_at

spec/services/create_next_edition_service_spec.rb
kevindew: I think this test would be clearer if it didn't assign to next_edition.

Eg:

```
  described_class.call(current_edition: live_edition,
                       user: user, 
                       discarded_edition: discarded_edition)
  expect(discarded_edition).to be_draft
```

app/interactors/editions/create_interactor.rb
```diff
@@ -21,38 +19,15 @@ def call
...
-    updater = Versioning::RevisionUpdater.new(edition.revision, user)
-
-    updater.assign(change_note: "",
-                   update_type: "major",
-                   proposed_publish_time: nil)
-
-    context.next_revision = updater.next_revision
+    context.discarded_edition = Edition.find_by(document: edition.document,
+                                                number: edition.number + 1)
+    assert_edition_state(edition, assertion: "can create new edition") { edition.live }
```
kevindew: We should probably reference somewhere (in this commit message or a separate commit) why we removed the `discarded?` conditional as it's not clear we thought that was done in error.

spec/interactors/editions/create_interactor_spec.rb
kevindew: I'd wonder if this test and the one above about discarded now need adapting as they're logic is no longer covered by this test and should be covered by the service unit tests (I'm thinking we may have missed a test that the number is changed).

We may just need to assert here that the service is called.

app/services/create_next_edition_service.rb
kevindew: With this we should be cautious about our terminology. As update_type is an enum and not a boolean we should stick to one meaning throughout. So we probably do best to decide _does change history apply to only major changes?_ or _does it not apply to minor changes?_ otherwise we end up using different words in different places to describe the same thing.

gclssvglx: @1pretz1 While I prefer `current_edition.minor?` for its clarity, I also take @kevindew 's point in that this, as an enum.  So, it could be 'extended' at some point - potentially, breaking this code an obscure way.  As the card states in the update criteria: _"The update type is major"_ - and if we're all ok with it - I'd prefer to leave this as `!current_edition.major?`, at least for now.  The only other resolution to this I feel would be to make `update_type` a boolean.

spec/services/create_next_edition_service_spec.rb
gclssvglx: @1pretz1 It's use like this is known as the [Oxford comma](https://www.grammarbook.com/punctuation/commas.asp).  Personally, I don't feel it matters too much unless you have this situation "one, two, three and four, and five" where it adds clarity.  It's a matter of taste I guess - I'm happy to remove it.

spec/services/create_next_edition_service_spec.rb
gclssvglx: @1pretz1 Ah, nice!  Yes, I like this much better.  Thanks!

spec/services/create_next_edition_service_spec.rb
gclssvglx: Thanks Peter!  Please see comment on previous instance above.  Have replaced all `eq([])` with `be_empty`.

spec/services/create_next_edition_service_spec.rb
gclssvglx: @1pretz1 Expelliarmus - I'm convinced that RSpec has more wizardry than Hogwarts!  Initially, I tried `expect(next_edition.state).to be_draft` but that caused a failure with `expected "draft" to respond to 'draft?'`.  Then I noticed the typo (`.state`), so then `expect(next_edition).to be_draft` and it works!  You're gonna have to explain this one to me.

spec/services/create_next_edition_service_spec.rb
gclssvglx: @1pretz1 Thanks, yes - this is a little clunky.  Have updated this to `it "does not update the change history when the current edition is not a major change" do` and the following test (with a similar issue) to `it "does not update the change history when the current edition lacks a change note" do`.

spec/services/create_next_edition_service_spec.rb
gclssvglx: @kevindew Yes, that makes more sense - thanks.  Will fix.

spec/services/create_next_edition_service_spec.rb
gclssvglx: @kevindew Yup, makes sense.  Will fix.

spec/services/create_next_edition_service_spec.rb
gclssvglx: @kevindew Thanks!  Will fix.

spec/services/create_next_edition_service_spec.rb
gclssvglx: @kevindew Thanks!  Will fix.

spec/services/create_next_edition_service_spec.rb
gclssvglx: @kevindew Thanks, yes - that does look clearer and cleaner.  Will fix.

app/interactors/editions/create_interactor.rb
gclssvglx: @kevindew Thanks - have updated the commit message on this.

spec/interactors/editions/create_interactor_spec.rb
gclssvglx: @1pretz1 Thanks - good spot, have updated accordingly for now.  But may change due to @kevindew 's comment.  Will leave this outstanding for now.

app/services/create_next_edition_service.rb
```diff
@@ -0,0 +1,51 @@
...
+  end
+
+private
+
+  attr_reader :current_edition, :user, :discarded_edition
+
+  def next_edition
+    @next_edition ||= begin
+      document = current_edition.document
+      discarded_edition || Edition.new(document: document,
```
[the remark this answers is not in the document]
gclssvglx: @1pretz1 Thanks, good spot.  It also looks much nicer!  Will fix.

app/services/create_next_edition_service.rb
```diff
@@ -0,0 +1,54 @@
...
+  def next_edition
+    @next_edition ||= begin
+      document = current_edition.document
+
+      return discarded_edition if discarded_edition.present?
+
+      Edition.new(document: document,
+                  number: document.next_edition_number,
+                  created_by: user)
+    end
```
kevindew: This looks to have gone a bit wrong following its most recent change. If I recall correctly Pete's suggestion was to avoid any queries for document so doing the document assign before the return doesn't achieve that. 

It's also feels rather strange doing a return within a begin block. It looks like you're trying to escape the begin block whereas the return would terminate the method.

I'd suggest either doing:

```
def next_edition
  return discarded_edition if discarded_edition

  @next_edition ||= Edition.new(document: current_edition.document,
                                number: current_edition.document.next_edition_number,
                                created_by: user)
end
```

or switching it back to how it was before. Myself, I'm not concerned about the extra query because in the circumstances we expect this to be used [a document will already be cached](https://github.com/alphagov/content-publisher/blob/708dcd9552a63f6e884711032f9eb3b347958787/app/controllers/errors_controller.rb#L3)

app/services/create_next_edition_service.rb
kevindew: Sounds good - I know having a boolean of either "major_change" or "minor_change" was unpopular at the time of building the modelling.

spec/services/create_next_edition_service_spec.rb
```diff
@@ -0,0 +1,121 @@
...
+      expect(next_edition.current).to be(true)
+    end
+
+    it "returns a draft edition" do
+      current_edition = create(:edition, :published)
+
+      next_edition = described_class.call(current_edition: current_edition,
+                                          user: user)
+
+      expect(next_edition.draft?).to be(true)
```
kevindew: The follow the trend of Pete's suggestion you'd probably want to change this to `expect(next_edition).to be_draft`

spec/services/create_next_edition_service_spec.rb
```diff
@@ -0,0 +1,121 @@
...
+                               change_note: "note",
+                               update_type: "minor",
+                               proposed_publish_time: Time.zone.now)
+
+      next_edition = described_class.call(current_edition: current_edition,
+                                          user: user)
+
+      expect(next_edition.change_note).to be_empty
+      expect(next_edition.update_type).to eq("major")
+      expect(next_edition.proposed_publish_time).to eq(nil)
```
kevindew: you can change this to `be_nil`

spec/services/create_next_edition_service_spec.rb
```diff
@@ -0,0 +1,121 @@
...
+                               :published,
+                               number: 2,
+                               published_at: Date.yesterday.noon,
+                               change_note: "note",
+                               update_type: "minor")
+
+      next_edition = described_class.call(current_edition: current_edition,
+                                          user: user)
+
+      expect(next_edition.change_history).to be_empty
```
kevindew: should this actually be ` expect(next_edition.change_history).to eq(current_edition.change_history)` as I guess we don't care that it's empty more that it matches.

spec/services/create_next_edition_service_spec.rb
```diff
@@ -0,0 +1,121 @@
...
+      it "doesn't append the change note details" do
+        current_edition = create(:edition,
+                                 :published,
+                                 change_note: "note")
+
+        next_edition = described_class.call(current_edition: current_edition,
+                                            user: user)
+
+        expect(next_edition.change_history).to be_empty
+      end
```
kevindew: Similar points about the eq change_history on this and above test

app/services/create_next_edition_service.rb
gclssvglx: @kevindew Thanks - have revered to previous version.

app/services/create_next_edition_service.rb
gclssvglx: @kevindew @1pretz1 - Thanks very much for your comments - sounds like we stick with the current `!current_edition.major?`.

spec/services/create_next_edition_service_spec.rb
gclssvglx: @kevindew **Hogwarts** strikes again!  Have made all the matcher changes.  Turns out this also works...
```
      expect(current_edition).not_to be_current
      expect(next_edition).to be_current
```

spec/services/create_next_edition_service_spec.rb
gclssvglx: See previous comment.

spec/services/create_next_edition_service_spec.rb
gclssvglx: See previous comment.

spec/services/create_next_edition_service_spec.rb
gclssvglx: See previous comment.

spec/interactors/editions/create_interactor_spec.rb
gclssvglx: @kevindew Have re-jigged things - hopefully in the right way.

gclssvglx:
@kevindew @1pretz1 Thanks for all the comments and help on this.  Hopefully, we're close!

spec/services/create_next_edition_service_spec.rb
1pretz1: Ah ha okay didn't know comma's can be used in that way, my grammar is awful tbf so It doesn't surprise me that I don't know this! 😆

spec/services/create_next_edition_service_spec.rb
1pretz1: I believe I found why this works and it's due to statuses state enum being delegated to the edition -> https://github.com/alphagov/content-publisher/blob/master/app/models/edition.rb#L56. Not as wizardly as I first though...

kevindew: Any object that has a question mark method can be used with a be_ expectation. For example if you had object x that responded to :test? you could do expect(x).to be_test.

It's pretty magical - it's defined here: https://github.com/rspec/rspec-expectations/blob/master/lib/rspec/matchers/built_in/be.rb#L181 (not that I understand much of that). Also looks like it checks plural too so could do expect(x).to be_tests

PR #1819, opened 2020-02-26
Add reorderable list component

alex-ju:
Add reorderable list component to be used for reordering attachments in Publications.

### Desktop

<table>
<tr><th>JavaScript enabled</th><th>JavaScript disabled</th></tr>
<tr><td valign="top">

reorder-js-enabled-desktop

</td><td valign="top">

content-publisher dev gov uk_component-guide_reorderable_list_default_preview_

</td></tr>
</table>

### Mobile
<table>
<tr><th>JavaScript enabled</th><th>JavaScript disabled</th></tr>
<tr><td valign="top">

reorder-js-enabled-mobile

</td><td valign="top">

content-publisher dev gov uk_component-guide_reorderable_list_default_preview_(iPhone 6_7_8 Plus)

</td></tr>
</table>

### Screen reader instructions
Examples using VoiceOver
<table>
<tr><th>JavaScript enabled</th><th>JavaScript disabled</th></tr>
<tr><td valign="top">

> Move Budget 2018 up, button, list 5 items.

</td><td valign="top">

> Order index for Budget 2019, edit text, list 5 items

</td></tr>
</table>

[Trello card](https://trello.com/c/H3T7JWtS)

### DIFF
--- a/app/assets/javascripts/components/reorderable-list.js
+++ b/app/assets/javascripts/components/reorderable-list.js
@@ -12,7 +12,8 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
 
     this.sortable = window.Sortable.create(this.$module, { // eslint-disable-line new-cap
       chosenClass: 'app-c-reorderable-list__item--chosen',
-      dragClass: 'app-c-reorderable-list__item--drag'
+      dragClass: 'app-c-reorderable-list__item--drag',
+      onSort: this.updateOrderIndexes.bind(this)
     })
 
     if (typeof window.matchMedia === 'function') {
@@ -48,6 +49,7 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
     var previousItem = item.previousElementSibling
     if (item && previousItem) {
       item.parentNode.insertBefore(item, previousItem)
+      this.updateOrderIndexes()
     }
     // if triggered by keyboard preserve focus on button
     if (e.detail === 0) {
@@ -64,6 +66,7 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
     var nextItem = item.nextElementSibling
     if (item && nextItem) {
       item.parentNode.insertBefore(item, nextItem.nextElementSibling)
+      this.updateOrderIndexes()
     }
     // if triggered by keyboard preserve focus on button
     if (e.detail === 0) {
@@ -73,6 +76,13 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
         e.target.previousElementSibling.focus()
       }
     }
+  }
+
+  ReorderableList.prototype.updateOrderIndexes = function () {
+    var $orderInputs = this.$module.querySelectorAll('.app-c-reorderable-list__actions input')
+    $orderInputs.forEach(function (input, index) {
+      input.setAttribute('value', index + 1)
+    })
   }
 
   Modules.ReorderableList = ReorderableList
--- a/app/assets/javascripts/components/reorderable-list.js
+++ b/app/assets/javascripts/components/reorderable-list.js
@@ -7,6 +7,8 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
 
   ReorderableList.prototype.start = function ($module) {
     this.$module = $module[0]
+    this.$upButtons = this.$module.querySelectorAll('.js-reorderable-list-up')
+    this.$downButtons = this.$module.querySelectorAll('.js-reorderable-list-down')
 
     this.sortable = window.Sortable.create(this.$module, { // eslint-disable-line new-cap
       chosenClass: 'app-c-reorderable-list__item--chosen',
@@ -18,6 +20,16 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
     } else {
       this.sortable.option('disabled', true)
     }
+
+    var boundOnClickUpButton = this.onClickUpButton.bind(this)
+    this.$upButtons.forEach(function (button) {
+      button.addEventListener('click', boundOnClickUpButton)
+    })
+
+    var boundOnClickDownButton = this.onClickDownButton.bind(this)
+    this.$downButtons.forEach(function (button) {
+      button.addEventListener('click', boundOnClickDownButton)
+    })
   }
 
   ReorderableList.prototype.setupResponsiveChecks = function () {
@@ -29,6 +41,38 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
 
   ReorderableList.prototype.checkMode = function () {
     this.sortable.option('disabled', !this.mediaQueryList.matches)
+  }
+
+  ReorderableList.prototype.onClickUpButton = function (e) {
+    var item = e.target.closest('.app-c-reorderable-list__item')
+    var previousItem = item.previousElementSibling
+    if (item && previousItem) {
+      item.parentNode.insertBefore(item, previousItem)
+    }
+    // if triggered by keyboard preserve focus on button
+    if (e.detail === 0) {
+      if (item !== item.parentNode.firstElementChild) {
+        e.target.focus()
+      } else {
+        e.target.nextElementSibling.focus()
+      }
+    }
+  }
+
+  ReorderableList.prototype.onClickDownButton = function (e) {
+    var item = e.target.closest('.app-c-reorderable-list__item')
+    var nextItem = item.nextElementSibling
+    if (item && nextItem) {
+      item.parentNode.insertBefore(item, nextItem.nextElementSibling)
+    }
+    // if triggered by keyboard preserve focus on button
+    if (e.detail === 0) {
+      if (item !== item.parentNode.lastElementChild) {
+        e.target.focus()
+      } else {
+        e.target.previousElementSibling.focus()
+      }
+    }
   }
 
   Modules.ReorderableList = ReorderableList
--- a/app/assets/javascripts/components/reorderable-list.js
+++ b/app/assets/javascripts/components/reorderable-list.js
@@ -1,0 +1,18 @@
+//= require sortablejs/Sortable.js
+window.GOVUK = window.GOVUK || {}
+window.GOVUK.Modules = window.GOVUK.Modules || {};
+
+(function (Modules) {
+  function ReorderableList () { }
+
+  ReorderableList.prototype.start = function ($module) {
+    this.$module = $module[0]
+
+    this.sortable = window.Sortable.create(this.$module, { // eslint-disable-line new-cap
+      chosenClass: 'app-c-reorderable-list__item--chosen',
+      dragClass: 'app-c-reorderable-list__item--drag'
+    })
+  }
+
+  Modules.ReorderableList = ReorderableList
+})(window.GOVUK.Modules)
--- a/app/assets/javascripts/components/reorderable-list.js
+++ b/app/assets/javascripts/components/reorderable-list.js
@@ -12,6 +12,23 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
       chosenClass: 'app-c-reorderable-list__item--chosen',
       dragClass: 'app-c-reorderable-list__item--drag'
     })
+
+    if (typeof window.matchMedia === 'function') {
+      this.setupResponsiveChecks()
+    } else {
+      this.sortable.option('disabled', true)
+    }
+  }
+
+  ReorderableList.prototype.setupResponsiveChecks = function () {
+    var tabletBreakpoint = '40.0625em' // ~640px
+    this.mediaQueryList = window.matchMedia('(min-width: ' + tabletBreakpoint + ')')
+    this.mediaQueryList.addListener(this.checkMode.bind(this))
+    this.checkMode()
+  }
+
+  ReorderableList.prototype.checkMode = function () {
+    this.sortable.option('disabled', !this.mediaQueryList.matches)
   }
 
   Modules.ReorderableList = ReorderableList
--- a/app/assets/stylesheets/components/_reorderable-list.scss
+++ b/app/assets/stylesheets/components/_reorderable-list.scss
@@ -1,0 +1,109 @@
+.app-c-reorderable-list {
+  @include govuk-font(19, bold);
+
+  list-style-type: decimal;
+  margin-bottom: govuk-spacing(6);
+  margin-top: 0;
+  padding-left: govuk-spacing(4);
+  position: relative;
+
+  .govuk-form-group {
+    margin-bottom: 0;
+  }
+}
+
+.app-c-reorderable-list__item {
+  margin-bottom: govuk-spacing(3);
+
+  &:after {
+    background: $govuk-border-colour;
+    content: '';
+    display: block;
+    height: 1px;
+    left: 0;
+    position: absolute;
+    width: 100%;
+  }
+}
+
+.app-c-reorderable-list__wrapper {
+  display: block;
+
+  @include govuk-media-query($from: desktop) {
+    display: inline-flex;
+    width: 100%;
+  }
+}
+
+.app-c-reorderable-list__content,
+.app-c-reorderable-list__actions {
+  margin-bottom: govuk-spacing(2);
+}
+
+.app-c-reorderable-list__content {
+  @include govuk-media-query($from: desktop) {
+    flex: 0 1 auto;
+    min-width: 70%;
+  }
+}
+
+.app-c-reorderable-list__title {
+  margin: 0;
+}
+
+.app-c-reorderable-list__description {
+  @include govuk-font(16, regular);
+  margin: 0;
+}
+
+.app-c-reorderable-list__actions {
+  display: block;
+
+  @include govuk-media-query($from: desktop) {
+    flex: 1 0 auto;
+    text-align: right;
+  }
+
+  .gem-c-button {
+    display: none;
+  }
+}
+
+.js-enabled {
+  .app-c-reorderable-list__actions {
+    .govuk-form-group {
+      display: none;
+    }
+
+    .gem-c-button {
+      display: inline-block;
+      margin-left: govuk-spacing(3);
+      width: 80px;
+    }
+
+    .gem-c-button:first-of-type {
+      margin-left: 0;
+
+      @include govuk-media-query($from: desktop) {
+        margin-left: govuk-spacing(3);
+      }
+    }
+  }
+
+  .app-c-reorderable-list__item:first-child .gem-c-button:first-of-type,
+  .app-c-reorderable-list__item:last-child .gem-c-button:last-of-type {
+    display: none;
+
+    @include govuk-media-query($from: desktop) {
+      display: inline-block;
+      visibility: hidden;
+    }
+  }
+
+  .app-c-reorderable-list__item:first-child .gem-c-button:last-of-type {
+    margin-left: 0;
+    @include govuk-media-query($from: desktop) {
+      margin-left: govuk-spacing(3);
+    }
+  }
+}
--- a/app/assets/stylesheets/components/_reorderable-list.scss
+++ b/app/assets/stylesheets/components/_reorderable-list.scss
@@ -26,6 +26,28 @@
   }
 }
 
+.app-c-reorderable-list__item--chosen {
+  background-color: govuk-colour('light-grey');
+  outline: 2px dotted $govuk-border-colour;
+
+  &:after {
+    content: none;
+  }
+}
+
+.app-c-reorderable-list__item--drag {
+  background-color: govuk-colour('white');
+  list-style-type: none;
+
+  .app-c-reorderable-list__actions {
+    visibility: hidden;
+  }
+
+  &:after {
+    content: none;
+  }
+}
+
 .app-c-reorderable-list__wrapper {
   display: block;
 
@@ -70,6 +92,13 @@
 }
 
 .js-enabled {
+
+  .app-c-reorderable-list__item {
+    @include govuk-media-query($from: desktop) {
+      cursor: move;
+    }
+  }
+
   .app-c-reorderable-list__actions {
     .govuk-form-group {
       display: none;
@@ -102,6 +131,7 @@
 
   .app-c-reorderable-list__item:first-child .gem-c-button:last-of-type {
     margin-left: 0;
+
     @include govuk-media-query($from: desktop) {
       margin-left: govuk-spacing(3);
     }
--- a/app/views/components/_reorderable_list.html.erb
+++ b/app/views/components/_reorderable_list.html.erb
@@ -1,0 +1,41 @@
+<%
+  items ||= []
+  data_attributes ||= {}
+  data_attributes[:module] = "reorderable-list"
+%>
+
+<%= tag.ol class: "app-c-reorderable-list", data: data_attributes do %>
+  <% items.each do |item| %>
+    <%= tag.li class: "app-c-reorderable-list__item" do %>
+      <%= tag.div class: "app-c-reorderable-list__wrapper" do %>
+        <%= tag.div class: "app-c-reorderable-list__content" do %>
+          <%= tag.p item[:title], class: "app-c-reorderable-list__title" %>
+          <%= tag.p item[:description], class: "app-c-reorderable-list__description" %>
+        <% end %>
+        <%= tag.div class: "app-c-reorderable-list__actions" do %>
+          <%= render "govuk_publishing_components/components/input", {
+            label: {
+              text: sanitize("Order index<span class='govuk-visually-hidden'> for #{item[:title]}</span>")
+            },
+            name: "#{order_input_name}[#{item[:order_input_index]}]",
+            type: "number",
+            value: index + 1,
+            width: 2
+          } %>
+          <%= render "govuk_publishing_components/components/button", {
+            text: "Up",
+            type: "button",
+            aria_label: "Move #{item[:title]} up",
+            secondary_quiet: true
+          } %>
+          <%= render "govuk_publishing_components/components/button", {
+            text: "Down",
+            type: "button",
+            aria_label: "Move #{item[:title]} down",
+            secondary_quiet: true
+          } %>
+        <% end %>
+      <% end %>
+    <% end %>
+  <% end %>
+<% end %>
--- a/app/views/components/_reorderable_list.html.erb
+++ b/app/views/components/_reorderable_list.html.erb
@@ -5,7 +5,7 @@
 %>
 
 <%= tag.ol class: "app-c-reorderable-list", data: data_attributes do %>
-  <% items.each do |item| %>
+  <% items.each_with_index do |item, index| %>
     <%= tag.li class: "app-c-reorderable-list__item" do %>
       <%= tag.div class: "app-c-reorderable-list__wrapper" do %>
         <%= tag.div class: "app-c-reorderable-list__content" do %>
--- a/app/views/components/_reorderable_list.html.erb
+++ b/app/views/components/_reorderable_list.html.erb
@@ -26,12 +26,14 @@
             text: "Up",
             type: "button",
             aria_label: "Move #{item[:title]} up",
+            classes: "js-reorderable-list-up",
             secondary_quiet: true
           } %>
           <%= render "govuk_publishing_components/components/button", {
             text: "Down",
             type: "button",
             aria_label: "Move #{item[:title]} down",
+            classes: "js-reorderable-list-down",
             secondary_quiet: true
           } %>
         <% end %>
--- a/app/views/components/docs/reorderable_list.yml
+++ b/app/views/components/docs/reorderable_list.yml
@@ -1,0 +1,31 @@
+name: Reorderable list
+description: A list of items that can be reordered
+body: |
+  List items can be reordered by drag and drop or by using the up/down buttons.
+  On small viewports the drag and drop feature is disabled to prevent being triggered
+  when scrolling on touch devices.
+
+  This component uses SortableJS - a JavaScript library for drag and drop interactions.
+  When JavaScript is disabled a set of inputs will be shown allowing users to provide
+  an order index for each item.
+accessibility_criteria: |
+  Buttons in this component must:
+
+  * be keyboard focusable
+  * inform the user about which item they operate on
+  * preserve focus after interacting with them
+examples:
+  default:
+    data:
+      order_input_name: attachments_order
+      items:
+        - title: "Budget 2018"
+          description: "PDF, 2.56MB, 48 pages"
+        - title: "Budget 2018 (web)"
+          description: "HTML attachment"
+        - title: "Impact on households: distributional analysis to accompany Budget 2018"
+          description: "PDF, 592KB, 48 pages"
+        - title: "Table 2.1: Budget 2018 policy decisions"
+          description: "MS Excel Spreadsheet, 248KB"
+        - title: "Table 2.2: Measures announced at Autumn Budget 2017 or earlier that will take effect from November 2018 or later (£ million)"
+          description: "MS Excel Spreadsheet, 248KB"
--- a/app/views/components/docs/reorderable_list.yml
+++ b/app/views/components/docs/reorderable_list.yml
@@ -29,3 +29,27 @@ examples:
           description: "MS Excel Spreadsheet, 248KB"
         - title: "Table 2.2: Measures announced at Autumn Budget 2017 or earlier that will take effect from November 2018 or later (£ million)"
           description: "MS Excel Spreadsheet, 248KB"
+  within_form:
+    embed: |
+      <form>
+        <%= component %>
+        <button class="govuk-button" type="submit">Save order</button>
+      </form>
+    data:
+      order_input_name: attachments_order_in_form
+      items:
+        - title: "Budget 2018"
+          description: "PDF, 2.56MB, 48 pages"
+          order_input_index: "217ab88f-74a9-41eb-ab47-428f54461c00"
+        - title: "Budget 2018 (web)"
+          description: "HTML attachment"
+          order_input_index: "217ab88f-74a9-41eb-ab47-428f54461c01"
+        - title: "Impact on households: distributional analysis to accompany Budget 2018"
+          description: "PDF, 592KB, 48 pages"
+          order_input_index: "217ab88f-74a9-41eb-ab47-428f54461c02"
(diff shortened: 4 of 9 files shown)


app/assets/stylesheets/components/_reorderable-list.scss
```diff
@@ -0,0 +1,128 @@
...
+    min-width: 70%;
+  }
+}
+
+.app-c-reorderable-list__title {
+  margin: 0;
+}
+
+.app-c-reorderable-list__description {
+  @include govuk-font(16, regular);
```
benthorner: This confused me slightly, because we're making a 'global' font setting and then undoing it here. Would it be simpler to style the title and indices explicitly instead?

app/assets/javascripts/components/reorderable-list.js
```diff
@@ -8,10 +8,30 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
...
+
+    if (typeof window.matchMedia === 'function') {
+      this.setupResponsiveChecks()
+    } else {
+      this.sortable.option("disabled", false);
+    }
+  }
+
+  ReorderableList.prototype.setupResponsiveChecks = function () {
+    this.mql = window.matchMedia('(min-width: 40.0625em)')
```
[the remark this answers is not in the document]
benthorner: Perhaps we could query the device instead? Especially as a narrow width window on my desktop is still usable for drag and drop.

app/assets/stylesheets/components/_reorderable-list.scss
```diff
@@ -0,0 +1,128 @@
...
+  }
+
+  .gem-c-button {
+    display: none;
+  }
+}
+
+.js-enabled {
+  .app-c-reorderable-list__actions {
+    .govuk-form-group {
```
benthorner: This is a bit surprising. Could we refer refer to the group (the order index) by name somehow?

app/assets/stylesheets/components/_reorderable-list.scss
```diff
@@ -0,0 +1,128 @@
...
+    .gem-c-button:first-of-type {
+      margin-left: 0;
+
+      @include govuk-media-query($from: desktop) {
+        margin-left: govuk-spacing(3);
+      }
+    }
+  }
+
+  .app-c-reorderable-list__item:first-child .gem-c-button:first-of-type,
```
benthorner: Same point here about having a way to refer to these buttons more explicitly.

app/assets/stylesheets/components/_reorderable-list.scss
```diff
@@ -0,0 +1,128 @@
...
+  .app-c-reorderable-list__item:last-child .gem-c-button:last-of-type {
+    display: none;
+
+    @include govuk-media-query($from: desktop) {
+      display: inline-block;
```
benthorner: Minor: do we need this if we're setting it for `.gem-c-button`?

app/assets/stylesheets/components/_reorderable-list.scss
```diff
@@ -0,0 +1,128 @@
...
+
+      @include govuk-media-query($from: desktop) {
+        margin-left: govuk-spacing(3);
+      }
+    }
+  }
+
+  .app-c-reorderable-list__item:first-child .gem-c-button:first-of-type,
+  .app-c-reorderable-list__item:last-child .gem-c-button:last-of-type {
+    display: none;
```
benthorner: Could you help me understand what this is for?

app/assets/stylesheets/components/_reorderable-list.scss
```diff
@@ -0,0 +1,128 @@
...
+    .gem-c-button {
+      display: inline-block;
+      margin-left: govuk-spacing(3);
+      width: 80px;
+    }
+
+    .gem-c-button:first-of-type {
+      margin-left: 0;
+
+      @include govuk-media-query($from: desktop) {
```
benthorner: I'm struggling to interpret this sequence of styles. I think the purpose of this is to ensure we don't have a left margin when the buttons have wrapped onto their own line. Could we achieve the same thing more explicitly by only setting a `margin-left` on the `last-of-type` (and perhaps a `margin-right` on the content to the left)?

app/assets/javascripts/components/reorderable-list.js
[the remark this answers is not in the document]
alex-ju: a `MediaQueryList` object. we need to store it so we can use it [in `checkMode` function, a few lines below](https://github.com/alphagov/content-publisher/pull/1819/files/67635252b5b937886efb2c3f6f2e04b0759142ed..6ede0ebd3576cad8ef30fdb56af579636ce9dd1e#diff-ddc0e2bfcf373620a17549ad2593fe83R30).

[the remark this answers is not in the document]
alex-ju: This is the value we use to detect the mobile breakpoint. We can store it into a const to give it name, but we're only using it once in the code.

app/assets/javascripts/components/reorderable-list.js
```diff
@@ -8,10 +8,30 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
   ReorderableList.prototype.start = function ($module) {
     this.$module = $module[0]
 
-    this.Sortable = new window.Sortable.create(this.$module, {
+    this.sortable = new window.Sortable.create(this.$module, {
       chosenClass: "app-c-reorderable-list__item--chosen",
       dragClass: "app-c-reorderable-list__item--drag"
     })
+
+    if (typeof window.matchMedia === 'function') {
```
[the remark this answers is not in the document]
alex-ju: This is our way of doing feature detection on media queries. If we want to be defensive, if not available we can disable the drag and drop indeed. Will update and record the decision.

app/assets/javascripts/components/reorderable-list.js
```diff
@@ -7,17 +7,27 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
 
   ReorderableList.prototype.start = function ($module) {
     this.$module = $module[0]
+    this.$upButtons = this.$module.querySelectorAll('.app-c-reorderable-list__actions button:first-of-type')
+    this.$downButtons = this.$module.querySelectorAll('.app-c-reorderable-list__actions button:last-of-type')
 
-    this.sortable = new window.Sortable.create(this.$module, {
-      chosenClass: "app-c-reorderable-list__item--chosen",
-      dragClass: "app-c-reorderable-list__item--drag"
+    this.sortable = new window.Sortable.create(this.$module, { // eslint-disable-line new-cap
```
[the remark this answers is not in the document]
alex-ju: Indeed, will rebase. Haven't got my linter set-up correctly.

app/views/components/_reorderable_list.html.erb
```diff
@@ -0,0 +1,38 @@
...
+    <%= tag.li class: "app-c-reorderable-list__item" do %>
+      <%= tag.div class: "app-c-reorderable-list__wrapper" do %>
+        <%= tag.div class: "app-c-reorderable-list__content" do %>
+          <%= tag.p item[:title], class: "app-c-reorderable-list__title" %>
+          <%= tag.p item[:description], class: "app-c-reorderable-list__description" %>
+        <% end %>
+        <%= tag.div class: "app-c-reorderable-list__actions" do %>
+          <%= render "govuk_publishing_components/components/input", {
+            label: {
+              text: sanitize("Order index<span class='govuk-visually-hidden'> for #{item[:title]}</span>")
```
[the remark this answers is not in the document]
alex-ju: Yes, `aria-label` is a good alternative – will have to update the component to allow it.

app/assets/javascripts/components/reorderable-list.js
```diff
@@ -7,17 +7,27 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
 
   ReorderableList.prototype.start = function ($module) {
     this.$module = $module[0]
+    this.$upButtons = this.$module.querySelectorAll('.app-c-reorderable-list__actions button:first-of-type')
```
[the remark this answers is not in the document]
alex-ju: I would've normally used a `.js-reorderable-list-up-button` class for querying the elements via JavaScript (as we do in all components) but components from the library don't allow custom classes to be added.

app/assets/stylesheets/components/_reorderable-list.scss
alex-ju: The global font setting is at the list level and it's required to style the numbers – it also implicitly applies to list items. We're using this as an exception for the description indeed.

app/assets/stylesheets/components/_reorderable-list.scss
alex-ju: The advantage of using CSS pseudo-selectors in this way is that we don't have to add JavaScript to manipulate the DOM and update classes when reordering elements.

app/assets/stylesheets/components/_reorderable-list.scss
alex-ju: We need it to override the `display: none;` used until `$desktop` breakpoint, on L117.

app/assets/stylesheets/components/_reorderable-list.scss
alex-ju: The up button for the first item and the down button for the last item are being hidden using different techniques on mobile compared to desktop. On mobile we use `display: none` as we don't want them to be shown, nor to occupy layout space (which helps us aligning them to the left). On desktop we use `visibility: hidden` which hides the button but takes up the same space as before which helps us keeping the up/down buttons aligned (more precisely the down button for the first item).

alex-ju:
> We need to think about the data this element will submit. Firstly, the `order_index` inputs will need to be kept in-sync when elements move on the page. Secondly, we need to know which index goes with each element. There are a couple of ways we could do this. I think the easiest option is for each item to have an `original_order_index[]`, so we can work out how the items have moved.

I was hoping that we can save users from some JavaScript being served and computations being done after each move by exposing a reversed attribution `input-name[new_index]=old_index.
I'll try and exemplify how I envisioned the order submission would work:

no-js: before reorder
```sh
reorderable-list[0]=1
reorderable-list[1]=2
reorderable-list[2]=3
```

no-js: after reorder (users manually replacing the 2 with 1 in the first input and vice-versa for the second)
```sh
reorderable-list[0]=2 #the second item will get index 0
reorderable-list[1]=1 #first item will get index 1
reorderable-list[2]=3 #third item remains the same
```

js enabled: before reorder (after script is being executed)
```
reorderable-list[]=1
reorderable-list[]=2
reorderable-list[]=3
```

js enabled: after reorder
```
reorderable-list[]=2
reorderable-list[]=1
reorderable-list[]=3
```

Now looking at it may work fine, but what I'm not capturing here is how do proceed when we have custom item IDs. I'm trying to figure out what a new set of inputs storing `original_order_index` would involve (both frontend and backend) – maybe we can catch-up on this.

> We seem to lose the focus states for the Up/Down buttons on click and on "Enter". I think there may be a general point there about indicating the element that was moved.

Great point. I've updated the code to preserve the focus.

> There's something weird going on with the drag/drop, which I can demo in person: sometimes the item 'jumps' downward and then returns to its current position.

Thanks for demo-ing this bug. I've updated the code to fix this.

> There's a slight inconsistency in the total space of the buttons for the first/last items. Probably not a big deal, but I thought I'd mention it, since we were concerned about it in the first commit.

Updated the code to fix these inconsistencies.

> My screen reader is seeing the first 'Down' button as a list of 5 buttons. I had a quick Google but can't immediately see the reason for this - I'm guessing it's reading down the 'column'.

It's a standard way for screen readers to let you know this is an element in a list of 5 items – valuable information when you're reordering.

> When using my screen reader I found the 'Move' help text confusing, since it sounds like it's part of the title. Can we make the label just `Move Up` and "`aria-describedby`" the title?

Replied in an inline comment below; I consider the `aria-label` suggestion as being more practical that `aria-describedby` – regardless it requires an update on the button component.

app/assets/stylesheets/components/_reorderable-list.scss
alex-ju: Yes, it can be done that way too, not convinced it's more explicit thought since we'd only want the `margin-right` on the content if there are buttons on the right-hand side.

app/assets/javascripts/components/reorderable-list.js
```diff
@@ -34,5 +44,23 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
     }
   }
 
+  ReorderableList.prototype.onClickUpButton = function (e) {
+    var item = this.closest('.app-c-reorderable-list__item')
+    var previousItem = item.previousElementSibling
```
ChrisBAshton: Nice - I hadn't come across these DOM properties before. 🎉

app/views/components/_reorderable_list.html.erb
```diff
@@ -0,0 +1,39 @@
+<%
+  items ||= []
+  data_attributes ||= {}
+  data_attributes[:module] = "reorderable-list"
+  input_name ||= "reorderable-list-#{SecureRandom.hex(4)}"
```
ChrisBAshton: One other suggestion that has just occurred to me - should the component take an `error_message`/`error_items` property like [other components in govuk_publishing_components](https://github.com/alphagov/govuk_publishing_components/blob/master/app/views/govuk_publishing_components/components/_input.html.erb)?

I imagine if a user enters an unparseable input on one of the inputs, we should highlight the specific input in a red error state and explain what is wrong.

I'm a little hesitant to suggest this, as I know @benthorner was keen we don't bog ourselves down in validation. I suppose the alternative would be to just skip over any offending inputs and put them at the back of the queue, so to speak, though once we start adding that logic we're not far off the complexity we'd get from adding 'proper' validation.

app/views/components/_reorderable_list.html.erb
```diff
@@ -0,0 +1,39 @@
...
+<%= tag.ol class: "app-c-reorderable-list", data: data_attributes do %>
+  <% items.each do |item| %>
+    <%= tag.li class: "app-c-reorderable-list__item" do %>
+      <%= tag.div class: "app-c-reorderable-list__wrapper" do %>
+        <%= tag.div class: "app-c-reorderable-list__content" do %>
+          <%= tag.p item[:title], class: "app-c-reorderable-list__title" %>
+          <%= tag.p item[:description], class: "app-c-reorderable-list__description" %>
+        <% end %>
+        <%= tag.div class: "app-c-reorderable-list__actions" do %>
+          <%= render "govuk_publishing_components/components/input", {
```
ChrisBAshton: Side note: is it worth using `<input type=”text” inputmode=”numeric” pattern="[0-9]*">` here as per https://technology.blog.gov.uk/2020/02/24/why-the-gov-uk-design-system-team-changed-the-input-type-for-numbers/?

ChrisBAshton:
One last comment from me!

Been thinking about how best to represent the ordering. I think the backend can be kept very simple (no need to pass a `original_order_index`, or indeed any index) and frontend needs to render only one input:

```erb
<% items = [{ id: 43, type: "FileAttachment"}, { id: 37, type: "FileAttachment" }] %>
<% # ^ example output from backend %>
<% items.each_with_index do |item, index| %>
  <input type="text" name="order_for_item_#{item.id}" value="#{index}" />
<% end %>
```

NoJS version: user can edit the value of the input, and submit the form. JS version: user can drag-and-drop (or use buttons) to move the element around; the JS should update the `value` of all inputs automatically. Then form submission would be something like:

```ruby
def reordered_items
  items.sort do |a, b|
    params["order_for_item_#{a.id}"].to_f <=> params["order_for_item_#{b.id}"].to_f
  end
end
```

Although it's nice submitting an input array (as that shows some relationship between the inputs), I think this approach is simpler as it means not having to maintain two separate arrays for original and new ordering.

benthorner:
@ChrisBAshton thanks for the suggestion. I can see a couple of problems we would encounter when we come to implementing it, which make me think we should still with the two-array approach:

   - Each item would be a different parameter. This means we would have to manually extract each one from the params by name, instead of a single array. It's also a bit of a smell that a group of related values aren't actually grouped, except by a common name.
   - The ID of featured attachments is not unique, since it depends on a combination of the database ID and the type of attachment model we're referring to, as mentioned [on the card about reordering](https://trello.com/c/7NXYdxZV/1566-user-can-reorder-featured-attachments). We did explore this approach, but believe it's simpler to focus the data around the order, not IDs.

ChrisBAshton:
We wouldn't necessarily have to use `.id` - we could pass some combination of id and attachment type to the frontend. So long as it's globally unique - and reversible - the frontend doesn't need to know how we've arrived at the ID, we're only using it ephemerally and can change our implementation of the ID over time if needed.

I do agree about the code smell 👍 I've thought more about your proposed implementation - I think what you're proposing is this:

```erb
<%
  items = [
    { id: 43, type: "FileAttachment", order_index: 1, original_order_index: 1 },
    { id: 37, type: "FileAttachment", order_index: 2, original_order_index: 2 }
  ] %>
<% # ^ example output from backend %>

<% items.each do |item| %>
  <input type="hidden" name="original_order[]" value="#{item.original_order_index}" />
  <input type="text" name="new_order[]" value="#{item.order_index}" />
<% end %>
```

...on form submission, we'd have something like:

```ruby
def reordered_items
  # params["original_order"] # ["1", "2", "3"] # I don't think we'll need to use this!
  # params["new_order"] # ["2", "1", "3"]

  params["new_order"].map do |old_index|
    items[old_index.to_i]
  end
end
```

This, too, can be simplified I think 🤔 

```erb
<% items = [{ id: 43, type: "FileAttachment"}, { id: 37, type: "FileAttachment" }] %>
<% # ^ example output from backend %>
<% items.each_with_index do |item, index| %>
  <input type="text" name="new_order[]" value="#{index}" />
<% end %>
```

Form submission:

```ruby
def reordered_items
  # params["new_order"] # ["2", "1", "3"]

  params["new_order"].map do |old_index|
    items[old_index.to_i]
  end
end
```

Either way, I don't think the backend needs to pass additional array(s) of orderings. They can be inferred by the order of the `items` array and some clever manipulation of the form DOM.

ChrisBAshton:
Chatted IRL with Ben, I now see the issue 😅 

```html
* ('foo' param added to disambiguate each input, since their values can change)

<input foo="bar" val="1">
<input foo="baz" val="2">
<input foo="tom" val="3">

<!-- ...reordered with JavaScript -->
<input foo="tom" val="3">
<input foo="bar" val="1">
<input foo="baz" val="2">

<!-- Submit -->
...would give form params of: [3, 1, 2]
...which could be mapped to: [tom, bar, baz]
...great!

But the noJS version wouldn't use ordering, it would involve changing the values:

<input foo="bar" val="2">
<input foo="baz" val="3">
<input foo="tom" val="1">

<!-- Submit -->
...would give form params of: [2, 3, 1]
...which would be mapped to: [baz, tom, bar]
...and NOT [tom, bar, baz] as we wanted!
```

Not sure why that took so long to click!

And to clarify, the suggested implementation is:

```erb
<% items = [{ id: 43, type: "FileAttachment"}, { id: 37, type: "FileAttachment" }] %>
<% # ^ example output from backend %>

<% items.each_with_index do |item, index| %>
  <input type="hidden" name="original_order[]" value="#{index}" />
  <input type="text" name="new_order[]" value="#{index}" />
<% end %>
```

(i.e. we will not pass any extra params from the backend. Rather, it is the frontend that will pass the ordering params to the backend).

app/views/components/_reorderable_list.html.erb
```diff
@@ -0,0 +1,38 @@
...
+  input_name ||= "reorderable-list-#{SecureRandom.hex(4)}"
+%>
+
+<%= tag.ol class: "app-c-reorderable-list", data: data_attributes do %>
+  <% items.each do |item| %>
+    <%= tag.li class: "app-c-reorderable-list__item" do %>
+      <%= tag.div class: "app-c-reorderable-list__wrapper" do %>
+        <%= tag.div class: "app-c-reorderable-list__content" do %>
+          <%= tag.p item[:title], class: "app-c-reorderable-list__title" %>
+          <%= tag.p item[:description], class: "app-c-reorderable-list__description" %>
```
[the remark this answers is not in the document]
alex-ju: Most of the app-level components in Content Publisher use helper functions to generate markup. Because we don't have a strict set of rules around how markup should be generated components in govuk_publishing_components are written in a mix of ways. If you pick [the accordion component](https://github.com/alphagov/govuk_publishing_components/blob/master/app/views/govuk_publishing_components/components/_accordion.html.erb) you'll find the same helpers (aside from one tag, to be fair). I have a note to start discussions around this within the community but it got deprioritised a few times in the past year.

I personally prefer using helpers as they tend to be easier to read and more forgiving when not providing options (by not generating empty attributes, which otherwise would require lots of conditional statements).

app/views/components/docs/reorderable_list.yml
```diff
@@ -0,0 +1,30 @@
...
+
+  This component uses SortableJS - a JavaScript library for drag and drop interactions.
+  When JavaScript is disabled a set of inputs will be shown allowing users to provide
+  an order index for each item.
+accessibility_criteria: |
+  Buttons in this component must:
+
+  * be keyboard focusable
+  * inform the user about which item they operate on
+  * preserve focus after interacting with them
```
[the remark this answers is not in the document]
alex-ju: All of our components are meant to be build using progressive enhancement.

app/assets/javascripts/components/reorderable-list.js
```diff
@@ -12,6 +12,26 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
       chosenClass: 'app-c-reorderable-list__item--chosen',
       dragClass: 'app-c-reorderable-list__item--drag'
     })
+
+    if (typeof window.matchMedia === 'function') {
+      this.setupResponsiveChecks()
```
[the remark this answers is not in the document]
alex-ju: Hmm.. this doesn't apply something really, it just adds a listener to check for a breakpoint. Probably `checkMode` could be renamed as it doesn't only check the media query, but also disable the feature.

app/views/components/_reorderable_list.html.erb
alex-ju: Good question about this. I didn't want to introduce it in the component before discussing how/if we're going to support this.

app/views/components/_reorderable_list.html.erb
alex-ju: Most probably very few/if nay people may benefit from this. Happy to add it if you think it's useful, but will have to update the input component to support these attributes.

alex-ju:
@ChrisBAshton Thank you for the review and for trying to find a way to only use one input. I've been 'seduced' by the same drive to reduce the number of inputs, but it seems like it just doesn't fit with the requirements.

benthorner:
@alex-ju I think @kevindew is planning to comment, so this is just a pre-emptive warning. After talking with @kevindew, we think there's a better approach we can take to solving the ordering approach in the backend - sorry for so many iterations! The quick summary is:

   - introduce an extra UUID into the ordering data we store (and rename `id`/`type` to `attachment_X`)
   - have a single params *hash* that's set like `attachments[uuid]=<order_index>` to represent the reordering

So each item/attachment still needs to arguments set, but we're proposing to use them differently.

kevindew:
Ah yes it was taking me a while to get the notes down. I've just had a chat with Ben about the parameters here and I'm summarising it here. 

## Two parameter fields

We talked about the need for two parameter fields, the working understanding was that we would need two input fields for example:

```
<div class="reorderable-attachment">
  <input type="hidden" name="previous_ordering[]" value="0" />
  <input type="text" name="new_ordering[]" value="2" />
</div>
```

Where this would look up the item in the file attachment ordering at item 0 and then move that item to be at position 2.

We've identified that this could be resolved with passing the id in as part of the params. For example:

```
<div class="reorderable-attachment">
  <input type="text" name="ordering[0]" value="2" />
</div>
```

Where the ordering hash could be looped through and the item at 0 could then be moved to position 2.

## Using the ordering as an id

We then talked about the risks in using the ordering as index to look up which items should have its position moved. We talked in a particular about a concurrent editing situation where for example: An editor asks a slack room for someone to move a CSV to the top of the list of featured attachments. If both Kevin & Ben load the page before the other has submitted their edits the first person (Ben) to submit the form will achieve the ordering as expected, whereas when Kevin submits it the data submitted on current ordering will be out of date and will undo the changes Ben made.

To resolve this we talked about what could be used as an id and came up with perhaps using a UUID. The effect of this could be changing the featured attachment array to something like:

```
[
  { id: "696d2cfb-b97a-479f-952a-abcef0c1c88e",
    attachment_type: "FileAttachment",
    attachment_id: 123 },
  { id: "217ab88f-74a9-41eb-ab47-428f54461c02",
    attachment_type: "HtmlAttachment",
    attachment_id: 456 }
]
```

and then changing the input approach to something like:

```
<div class="reorderable-attachment">
  <input type="text" name="ordering[696d2cfb-b97a-479f-952a-abcef0c1c88e]" value="1" />
</div>
```

So I'm guessing we'd end up with something along these lines to do the ordering (where the infinity is used to push any items missed from submission to the end of the list and doing nothing can have them at the start): 

```
id_1, id_2, id_3 = 3.times.map { SecureRandom.uuid }
file_attachments = [{ id: id_1 }, { id: id_2 }, { id: id_3 }]
submission = { id_1 => "1", id_2 => "0" }
ordering = submission.sort { |a, b| a.last.to_i <=> b.last.to_i }.map(&:first)
sorted = file_attachments.sort_by { |a| ordering.index(a[:id]) || Float::INFINITY }
```

alex-ju:
> To resolve this we talked about what could be used as an id and came up with perhaps using a UUID.

Thank you for the comment @kevindew. Coping with concurrency was what I was thinking when I referred to 'custom item IDs' above, and in previous discussions. What I wasn't sure about was how permissive the index is in terms of character types (e.g `inputname[696d2cfb-b97a-479f-952a-abcef0c1c88e]` which contains `-`) but I imagine it's flexible enough. I updated the code to support custom indexes again, only that now both inputs will share the same index.

app/assets/javascripts/components/reorderable-list.js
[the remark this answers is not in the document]
kevindew: Yeah I think something extra that tells us what this means would be useful. I was able to google a px converter and work out that it's probably 641px.

Maybe something like:
```
var tabletUp = (min-width: 40.0625em) // 641px ?
```

variable as Ben is suggesting is more for intent based documentation than re-use.

app/assets/javascripts/components/reorderable-list.js
```diff
@@ -0,0 +1,84 @@
...
+    this.mql = window.matchMedia('(min-width: 40.0625em)')
+    this.mql.addListener(this.checkMode.bind(this))
+    this.checkMode()
+  }
+
+  ReorderableList.prototype.checkMode = function () {
+    if (this.mql.matches) {
+      this.sortable.option('disabled', false)
+    } else {
+      this.sortable.option('disabled', true)
```
kevindew: You can probably shorten this to `this.sortable.option('disabled', this.mql.matches)`

app/assets/javascripts/components/reorderable-list.js
```diff
@@ -0,0 +1,84 @@
...
+  ReorderableList.prototype.onClickUpButton = function (e) {
+    var item = e.target.closest('.app-c-reorderable-list__item')
+    var previousItem = item.previousElementSibling
+    if (item && previousItem) {
+      item.parentNode.insertBefore(item, previousItem)
+      this.updateOrderIndexes()
+    }
+    // if triggered by keyboard preserve focus on button
+    if (e.detail === 0) {
+      e.target.focus()
```
kevindew: There seems to be a case here where if you move an item up to the top with keyboard this will apply a focus to a hidden element - and after that the tab index seems a little strange. Same situation happens on down too.

app/assets/javascripts/components/reorderable-list.js
```diff
@@ -0,0 +1,84 @@
...
+    }
+  }
+
+  ReorderableList.prototype.updateOrderIndexes = function () {
+    var $orderInputs = this.$module.querySelectorAll('.app-c-reorderable-list__actions input')
```
kevindew: Is there a more precise selector we can use on these?

app/views/components/_reorderable_list.html.erb
```diff
@@ -0,0 +1,39 @@
+<%
+  items ||= []
+  data_attributes ||= {}
+  data_attributes[:module] = "reorderable-list"
+  order_input_name ||= "reorderable-list-order-#{SecureRandom.hex(4)}"
```
kevindew: Since names don't have to be unique I don't think we really benefit from the random hex here

app/views/components/docs/reorderable_list.yml
```diff
@@ -0,0 +1,54 @@
...
+    embed: |
+      <form>
+        <%= component %>
+        <button class="govuk-button" type="submit">Save order</button>
+      </form>
+    data:
+      order_input_name: new_order
+      items:
+        - title: "Budget 2018"
+          description: "PDF, 2.56MB, 48 pages"
```
kevindew: I've talked to Simon that doing these trio of attributes is not going to be very easy as this is data that is wrapped up in the attachment component. So we talked about simplifying them. It shouldn't matter to you much since we're just passing a string in here though.

app/assets/javascripts/components/reorderable-list.js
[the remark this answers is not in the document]
kevindew: Are you sure this should be a new call? looking at the docs the use this without new: https://github.com/SortableJS/Sortable#usage (It seems quite unusual to be calling new on what looks like a class method)

app/assets/javascripts/components/reorderable-list.js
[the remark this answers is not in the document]
alex-ju: Updated as suggested.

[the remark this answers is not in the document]
alex-ju: Updated to store the value in a `tabletBreakpoint` variable, for intent-based documentation.

app/assets/javascripts/components/reorderable-list.js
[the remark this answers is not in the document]
alex-ju: If it's not a new call, multiple instances of this component on the same page will share the same instance of SortableJS – I assume we don't want this feature.

app/assets/javascripts/components/reorderable-list.js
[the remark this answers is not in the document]
alex-ju: In this specific case, it won't work as the `:first-of-type` pseudo selector only works on different element types. What would be the advantage of using the `.gem-c-button` class instead?

Sure, if we're not in a rush with this will start again the conversation around allowing classes to be passed in for components.

app/assets/stylesheets/components/_reorderable-list.scss
alex-ju: It seems like I missed this comment. We can refer to the input by name, but not to the group (which is the outermost wrapper of the input component).

app/assets/stylesheets/components/_reorderable-list.scss
```diff
@@ -0,0 +1,136 @@
...
+    content: '';
+    display: block;
+    height: 1px;
+    left: 0;
+    position: absolute;
+    width: 100%;
+  }
+
+  @include govuk-media-query($from: desktop) {
+    cursor: move;
```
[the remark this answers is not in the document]
alex-ju: Good spot. Moved it in the `js-enabled` block.

app/assets/javascripts/components/reorderable-list.js
alex-ju: Indeed, only the other way around (`!this.mediaQueryList.matches`) – I wasn't sure if the intent will remain clear. Let me know if you want me to shorten it like this.

app/assets/javascripts/components/reorderable-list.js
alex-ju: Yes, it's an edge case that I haven't captured in the code. We will need to add more logic to cope with these two cases (first visible up button and last visible down button).

app/assets/javascripts/components/reorderable-list.js
alex-ju: We can, but given we control the markup within the component I wouldn't add complexity to this query. What's the concern around it?

app/views/components/_reorderable_list.html.erb
alex-ju: Yep, a leftover. More than that, I think this attribute should be required (no default value provided) so I removed this line.

alex-ju:
> I've added a bunch of comments. I also was a bit confused about whether some of the existing ones were pending response/resolution?

Thank you, @kevindew. I tried to reply to them all.

app/assets/javascripts/components/reorderable-list.js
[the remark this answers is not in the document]
alex-ju: Opened https://github.com/alphagov/govuk_publishing_components/pull/1344 for this.

app/views/components/_reorderable_list.html.erb
alex-ju: Opened https://github.com/alphagov/govuk_publishing_components/pull/1345 for this

alex-ju: Opened https://github.com/alphagov/govuk_publishing_components/pull/1345 for this

app/assets/javascripts/components/reorderable-list.js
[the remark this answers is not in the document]
alex-ju: As we checked together, the create function generates a new instance already, so I've updated our code to remove the `new` operator.

app/assets/javascripts/components/reorderable-list.js
alex-ju: Replaced this logic with a one-liner

alex-ju:
@ChrisBAshton, sure I will. I don't expect to get this merge anytime sooner, as it needs refactoring after the requested changes in govuk_publishing_components.

alex-ju:
Updated screenshots, updated the code to profit from the changes introduced in `govuk_publishing_components` for [buttons](https://github.com/alphagov/govuk_publishing_components/pull/1344) and [inputs](https://github.com/alphagov/govuk_publishing_components/pull/1345). Unfortunately, I've only got support in using custom classes for `js-` selectors (which cleans the JavaScript code a bit), but not for styling (which leaves us with things like `gem-c-button:first-of-type` in our code). I'll try and raise the custom classes topic again in community discussions. I hope I addressed and/or reply to all queries and suggestions.

PR #1820, opened 2020-02-26
Introduce cop for TimeZone

JonathanHallam:
It was raised in a previous pr (#1748 (comment)) that we should have a linter to look at TimeZone. In that discussion it was stated that we should prefer `Time.current` over `Time.zone.now` but this is not a thing that rubocop supports. Instead I've implemented the opposite in order to ensure that we have a consistent style.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -36,6 +36,10 @@ Rails/DynamicFindBy:
 Rails/SaveBang:
   Enabled: true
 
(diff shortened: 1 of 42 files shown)


kevindew:
This looks a good idea to me - I don't mind too much whether we use Time.zone.now or Time.current as long as it's consistent.

It's worth shouting to see if this is something anyone else in the team has concerns about

PR #1825, opened 2020-02-27
Generic forbidden response handling

kevindew:
Trello: https://trello.com/c/HILuWHOt/1483-users-without-pre-release-permissions-cant-interact-with-publications-documents

This renames the documents/forbidden view to reflect that this is used for access_limited. It then creates a generic forbidden response for the errors controller. 

As there are no built in exceptions in Rails that generate a forbidden response this creates a AppliciationController::Forbidden error that is registered in the rescue responses config.

This work has been done so that we have something generic to use for when a pre_release format is accessed.

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -113,6 +113,7 @@ Rails.application.routes.draw do
 
   scope via: :all do
     match "/400" => "errors#bad_request"
+    match "/403" => "errors#forbidden"
     match "/404" => "errors#not_found"
     match "/422" => "errors#unprocessable_entity"
     match "/500" => "errors#internal_server_error"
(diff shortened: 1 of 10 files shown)


config/routes.rb
```diff
@@ -114,6 +114,7 @@
   scope via: :all do
     match "/400" => "errors#bad_request"
     match "/404" => "errors#not_found"
+    match "/403" => "errors#forbidden"
```
leenagupte: Nitpick: This should probably be in numerical order.

kevindew: D'oh! the shame

kevindew:
> It looks good. My main question is how and where do we test this?

I'd suggest putting a test in requests/error_spec.rb that's where we have a similar one for local_data_unavailable

kevindew:
This has now been merged via https://github.com/alphagov/content-publisher/pull/1826

PR #1826, opened 2020-02-27
Restrict access to pre release

leenagupte:
Trello: https://trello.com/c/HILuWHOt
Builds on: #1825 

## What's changed and why?
Publications were added in: #1810, but the ability to create new publications, or filter by publications on the index page was restricted to users who have the "pre_release_features" permission.

This restriction has been expanded to viewing and editing any publications documents as well. This includes everything from viewing the document summary page, to being able to tag the document etc. Users without the correct permission are shown a generic 403 Forbidden page.

This is so that we don't end up with users being in the position of being able to do some parts of pre-release format work based off a link but not being able to create one or find it in the index.

Co-authored with @kevindew

### DIFF
--- a/spec/requests/errors_spec.rb
+++ b/spec/requests/errors_spec.rb
@@ -7,6 +7,14 @@ RSpec.describe "Errors" do
     end
   end
 
+  describe "/403" do
+    it "returns a forbidden response" do
+      get "/403"
+      expect(response).to have_http_status(:forbidden)
+      expect(response.body).to include(I18n.t!("errors.forbidden.title"))
+    end
+  end
+
   describe "/404" do
     it "returns a not found response" do
       get "/404"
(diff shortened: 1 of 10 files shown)


kevindew:
This looks good to me 👍 

I feel like we should get someone else to review it since 3 of the 4 commits are mine.

spec/requests/errors_spec.rb
```diff
@@ -60,4 +60,16 @@
       expect(response.body).to include(I18n.t!("errors.local_data_unavailable.title"))
     end
   end
+
+  it "forbids users without pre-release permission from accessing pre-release documents" do
```
benthorner: This test is good. I found it hard to review, though, because this request spec is inconsistent with all the others, which indicate the route they're testing using a `describe "METHOD /route"` block. 

I think would find this spec easier to understand if it were in a `describe` block that indicated the request being made. Since it could be anything, perhaps something like `describe "ANY /pre-release-document-path"`.

PR #1828, opened 2020-02-27
Move check for pending document import before importing

emmabeynon:
For https://trello.com/c/gao9twxQ/1524-skip-import-if-document-import-is-not-in-pending-state

We originally checked if a document import is in a pending state within the
WhitehallImporter::Import class. However we've found that it's beneficial to do
this check before the class is called, as in situations where a document has
successfully been imported but the job is being retried as a result of an error
in the syncing process, the job errors since the document import is no longer
pending.

### DIFF
--- a/spec/jobs/whitehall_document_import_job_spec.rb
+++ b/spec/jobs/whitehall_document_import_job_spec.rb
@@ -22,10 +22,15 @@ RSpec.describe WhitehallDocumentImportJob do
     )
   end
 
-  it "calls WhitehallImporter::Import" do
+  it "calls WhitehallImporter::Import if the document import is pending" do
     expect(WhitehallImporter::Import).to receive(:call)
                                      .with(whitehall_migration_document_import)
     described_class.perform_now(whitehall_migration_document_import)
+  end
+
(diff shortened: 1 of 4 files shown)


spec/jobs/whitehall_document_import_job_spec.rb
```diff
@@ -22,12 +22,18 @@
...
-  it "calls WhitehallImporter::Import" do
+  it "calls WhitehallImporter::Import if the document import is pending" do
     expect(WhitehallImporter::Import).to receive(:call)
                                      .with(whitehall_migration_document_import)
     described_class.perform_now(whitehall_migration_document_import)
   end
 
+  it "does not call WhitehallImporter::Import if the document import is not pending" do
+    expect(WhitehallImporter::Import).not_to receive(:call)
+                                     .with(imported_document_import)
```
brucebolt: I'm not sure the `with` is strictly necessary here, since we don't want `WhitehallImporter::Import` to be called at all

emmabeynon: Fair point 👍

PR #1831, opened 2020-02-28
Exclude pre-release publications from edition filter results

JonathanHallam:
Trello - https://trello.com/c/wzzVIihM/1521-exclude-prerelease-publications-from-edition-filter-results

As it stands, users without the pre_release_features permission are able to see editions with the pre release flag when they visit content-publisher. This will give the false impression that these are things that can be edited. This adds the functionality to remove these editions from the list of things that such users can see.

### DIFF
--- a/lib/edition_filter.rb
+++ b/lib/edition_filter.rb
@@ -22,6 +22,7 @@ class EditionFilter
                    .preload(revision_joins, :status, :document, :last_edited_by)
     scope = access_limited_scope(scope)
     scope = filtered_scope(scope)
+    scope = non_pre_release_type_scope(scope)
     scope = ordered_scope(scope)
     scope.page(page).per(per_page)
   end
@@ -89,5 +90,12 @@ private
     else
       scope
     end
+  end
+
+  def non_pre_release_type_scope(scope)
+    return scope if user.has_permission?(User::PRE_RELEASE_FEATURES_PERMISSION)
+
+    types = DocumentType.all.reject(&:pre_release?).map(&:id)
+    scope.where("metadata_revisions.document_type_id": types)
   end
 end
--- a/spec/lib/edition_filter_spec.rb
+++ b/spec/lib/edition_filter_spec.rb
@@ -203,6 +203,23 @@ RSpec.describe EditionFilter do
         expect(editions).to eq([edition])
       end
     end
+
+    context "when the edition's document type is tagged as pre release" do
+      let!(:edition1) { create(:edition) }
+      let!(:edition2) { create(:edition, document_type: build(:document_type, :pre_release)) }
+
+      it "includes the edition for users with pre_release_features permission" do
+        pre_release_user = build(:user)
+        editions = described_class.new(pre_release_user).editions
+        expect(editions).to match_array([edition1, edition2])
+      end
+
+      it "excludes the edition for users without pre_release_features permission" do
+        user = build(:user, permissions: [])
+        editions = described_class.new(user).editions
+        expect(editions).to match_array([edition1])
+      end
+    end
   end
 
   describe "#filter_params" do
--- a/spec/factories/document_type_factory.rb
+++ b/spec/factories/document_type_factory.rb
@@ -37,5 +37,9 @@ FactoryBot.define do
     trait :with_lead_image do
       lead_image { true }
     end
+
+    trait :pre_release do
+      pre_release { true }
+    end
   end
 end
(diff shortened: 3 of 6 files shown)


spec/lib/edition_filter_spec.rb
```diff
@@ -203,6 +203,24 @@
...
+        edition1 = create(:edition)
+        edition2 = create(:edition, :pre_release)
+        pre_release_user = build(:user)
+        editions = described_class.new(pre_release_user).editions
+        expect(editions).to match_array([edition1, edition2])
+      end
+
+      it "excludes the edition for users without pre_release_features permission" do
+        edition1 = create(:edition)
+        _edition2 = create(:edition, :pre_release)
```
benthorner: We shouldn't need to define a variable if it's not used.

lib/edition_filter.rb
```diff
@@ -90,4 +91,11 @@ def ordered_scope(scope)
       scope
     end
   end
+
+  def pre_release_document_type_scope(scope)
+    return scope if user.has_permission?(User::PRE_RELEASE_FEATURES_PERMISSION)
+
+    ids = scope.map { |edition| edition.id unless edition.document_type.pre_release? }.compact
```
benthorner: This is going to be inefficient because it involves loading all of the documents into memory, and then paginating them to get back to a smaller set. We want to preserve the original property of this code where all the processing is done in the DB, which avoids shipping loads of redundant data back to the app and keeps page load times low.

We should be able to do this by figuring out allowed set of document type IDs (ones that aren't pre-release) and doing a `where` query on those.

lib/edition_filter.rb
```diff
@@ -22,6 +22,7 @@ def editions
                    .preload(revision_joins, :status, :document, :last_edited_by)
     scope = access_limited_scope(scope)
     scope = filtered_scope(scope)
+    scope = pre_release_document_type_scope(scope)
```
benthorner: I think we actually mean the inverse here: the scope actually returns the non-pre-release ones.

spec/factories/edition_factory.rb
```diff
@@ -209,5 +209,9 @@
     trait :not_political do
       system_political { false }
     end
+
+    trait :pre_release do
+      document_type { build(:document_type, :pre_release, path_prefix: "/prefix") }
```
benthorner: Could you help me understand why we're setting a `path_prefix` here? It looks like it's redundant.

spec/factories/edition_factory.rb
```diff
@@ -209,5 +209,9 @@
     trait :not_political do
       system_political { false }
     end
+
+    trait :pre_release do
```
benthorner: We probably don't need a trait for functionality that's only used once (/in one test). As a more minor point: it does seem a bit weird to say the edition is pre-release, since we're actually referring to the document type.

spec/lib/edition_filter_spec.rb
JonathanHallam: So it is being used but indirectly. The point here is that there are editions with pre release features that are being excluded. If I were to remove this variable then this test would pass without any of the code changes.

JonathanHallam: Have removed the variable declaration

lib/edition_filter.rb
JonathanHallam: Thanks for your help with this, was struggling with it yesterday!

lib/edition_filter.rb
JonathanHallam: Have renamed

spec/factories/edition_factory.rb
JonathanHallam: It was, I'd seen it used in a lot of other places and assumed it was mandatory, gone now

spec/factories/edition_factory.rb
JonathanHallam: Have removed the trait and moved the functionality into the test.

lib/edition_filter.rb
```diff
@@ -22,6 +22,7 @@ def editions
                    .preload(revision_joins, :status, :document, :last_edited_by)
     scope = access_limited_scope(scope)
     scope = filtered_scope(scope)
+    scope = remove_pre_release_document_type_scope(scope)
```
benthorner: Minor: these methods seem to be phrased in "noun-form" - `remove` is a verb - so `non_pre_release_types_scope` would be more consistent.

lib/edition_filter.rb
```diff
@@ -90,4 +91,11 @@ def ordered_scope(scope)
       scope
     end
   end
+
+  def remove_pre_release_document_type_scope(scope)
+    return scope if user.has_permission?(User::PRE_RELEASE_FEATURES_PERMISSION)
+
+    types = DocumentType.all.reject(&:pre_release?).map(&:id)
```
benthorner: Much cleaner 👍

lib/edition_filter.rb
JonathanHallam: cool, will do!

PR #1841, opened 2020-03-02
Create featured file attachments

leenagupte:
Trello: https://trello.com/c/RTw12Wxq

# What's changed and why?

Allow the user to upload a feature file attachment.

Publications, don't just allow inline attachments, linked to from within the body text of a document, they also allow a list of "featured" documents to be added to them. Featured documents are usually displayed along with some metadata about them.

This example has both: https://www.gov.uk/government/publications/the-uks-approach-to-trade-negotiations-with-the-us

To be able to support Publishers creating publications documents, we need to be support adding featured attachments to documents.

Until publications are ready to be released, adding a featured attachment is only available to users with "pre_release" permissions.

# Expected changes

## New attachment page
Screenshot 2020-03-02 at 10 44 41
Screenshot 2020-03-02 at 10 45 15

## Attachment index page
Screenshot 2020-03-03 at 12 37 35

Screenshot 2020-03-02 at 10 44 27

## Document summary page

Screenshot 2020-03-03 at 12 56 47

# Assumptions
The following features are not covered by this change, they will be added as part of a separate piece of work:

* Featured attachments are not sent as part of the publishing-api payload
* Existing attachments are not listed on the summary page
* The attachment index page does not have any "actions" e.g. delete
* The new attachment page and the upload attachment modal are very similar but have not been consolidated here.

### DIFF
--- a/app/controllers/file_attachments_controller.rb
+++ b/app/controllers/file_attachments_controller.rb
@@ -6,6 +6,11 @@ class FileAttachmentsController < ApplicationController
     assert_edition_state(@edition, &:editable?)
   end
 
+  def new
+    @edition = Edition.find_current(document: params[:document])
+    assert_edition_state(@edition, &:editable?)
+  end
+
   def show
     @edition = Edition.find_current(document: params[:document])
     assert_edition_state(@edition, &:editable?)
--- a/app/controllers/file_attachments_controller.rb
+++ b/app/controllers/file_attachments_controller.rb
@@ -48,7 +48,7 @@ class FileAttachmentsController < ApplicationController
         ),
       }
 
-      render :index,
+      render params[:wizard] == "new" ? :new : :index,
              assigns: { edition: edition,
                         issues: issues },
              status: :unprocessable_entity
--- a/app/controllers/file_attachments_controller.rb
+++ b/app/controllers/file_attachments_controller.rb
@@ -52,6 +52,8 @@ class FileAttachmentsController < ApplicationController
              assigns: { edition: edition,
                         issues: issues },
              status: :unprocessable_entity
+    elsif params[:wizard] == "new"
+      redirect_to featured_attachments_path(edition.document, attachment_revision.file_attachment)
     else
       redirect_to file_attachment_path(edition.document, attachment_revision.file_attachment)
     end
--- a/app/views/documents/show.html.erb
+++ b/app/views/documents/show.html.erb
@@ -51,6 +51,10 @@
   <div class="govuk-grid-column-two-thirds">
     <%= render "documents/show/contents" %>
 
+    <% if @edition.document_type.attachments.featured? && current_user.has_permission?(User::PRE_RELEASE_FEATURES_PERMISSION) %>
+      <%= render "documents/show/featured_attachments" %>
+    <% end %>
+
     <% if @edition.document_type.lead_image? %>
       <%= render "documents/show/lead_image" %>
     <% end %>
--- a/app/views/file_attachments/new.html.erb
+++ b/app/views/file_attachments/new.html.erb
@@ -7,6 +7,7 @@
       file_attachments_path(@edition.document),
       multipart: true,
     ) do %>
+      <%= hidden_field_tag(:wizard, params[:wizard]) %>
 
       <%= render_govspeak(t("file_attachments.new.description_govspeak")) %>
 
--- a/app/views/file_attachments/new.html.erb
+++ b/app/views/file_attachments/new.html.erb
@@ -1,0 +1,40 @@
+<% content_for :back_link, render_back_link(href: featured_attachments_path(@edition.document)) %>
+<% content_for :title, t("file_attachments.new.title", title: @edition.title_or_fallback) %>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <%= form_tag(
+      file_attachments_path(@edition.document),
+      multipart: true,
+    ) do %>
+
+      <%= render_govspeak(t("file_attachments.new.description_govspeak")) %>
+
+      <%= render "govuk_publishing_components/components/inset_text", {
+        text: render_govspeak(
+          t("file_attachments.new.callout_govspeak")
+        ),
+      } %>
+
+      <%= render "govuk_publishing_components/components/file_upload", {
+        label: {
+          text: t("file_attachments.new.attachment_file.heading"),
+        },
+        name: "file"
+      } %>
+
+      <%= render "govuk_publishing_components/components/input", {
+        label: {
+          text: t("file_attachments.new.attachment_title.heading"),
+          bold: true
+        },
+        name: "title",
+        value: params[:title],
+        hint: t("file_attachments.new.attachment_title.hint_text"),
+        error_items: @issues&.items_for(:file_attachment_title),
+      } %>
+
+      <%= render "govuk_publishing_components/components/button", { text: "Save and continue" } %>
+    <% end %>
+  </div>
+</div>
--- a/spec/features/editing_file_attachments/upload_file_attachment_spec.rb
+++ b/spec/features/editing_file_attachments/upload_file_attachment_spec.rb
@@ -1,10 +1,23 @@
 RSpec.feature "Upload file attachment", js: true do
-  scenario do
+  scenario "inline attachment" do
     given_there_is_an_edition
     when_i_go_to_edit_the_edition
     and_i_go_to_insert_an_attachment
-    and_i_upload_a_file_attachment
-    then_i_can_see_previews_of_the_attachment
+    and_i_select_a_file_to_upload
+    and_i_upload_the_file_attachment
+    then_i_can_see_the_attachment
+    and_i_can_see_an_action_to_preview_the_attachment
+    and_i_see_the_timeline_entry
+  end
+
+  scenario "featured attachment" do
+    given_there_is_an_edition_that_allows_featured_attachments
+    when_i_visit_the_summary_page
+    and_i_go_to_change_an_attachment
+    and_i_go_to_add_an_attachment
+    and_i_select_a_file_to_upload
+    and_i_save_the_file_attachment
+    then_i_can_see_the_attachment
     and_i_see_the_timeline_entry
   end
 
@@ -13,9 +26,18 @@ RSpec.feature "Upload file attachment", js: true do
     @edition = create(:edition, document_type: document_type)
   end
 
+  def given_there_is_an_edition_that_allows_featured_attachments
+    document_type = build(:document_type, :with_body, attachments: "featured")
+    @edition = create(:edition, document_type: document_type)
+  end
+
   def when_i_go_to_edit_the_edition
     visit document_path(@edition.document)
     click_on "Change Content"
+  end
+
+  def when_i_visit_the_summary_page
+    visit document_path(@edition.document)
   end
 
   def and_i_go_to_insert_an_attachment
@@ -23,7 +45,15 @@ RSpec.feature "Upload file attachment", js: true do
     click_on "Attachment"
   end
 
-  def and_i_upload_a_file_attachment
+  def and_i_go_to_change_an_attachment
+    click_on "Change Attachments"
+  end
+
+  def and_i_go_to_add_an_attachment
+    click_on "Upload file attachment"
+  end
+
+  def and_i_select_a_file_to_upload
     @attachment_filename = "13kb-1-page-attachment.pdf"
     @title = "A title"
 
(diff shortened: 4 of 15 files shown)


app/controllers/file_attachments_controller.rb
```diff
@@ -47,10 +47,17 @@ def create
         ),
       }
 
-      render :index,
-             assigns: { edition: edition,
-                        issues: issues },
-             status: :unprocessable_entity
+      if params[:wizard] == "new"
+        render :new,
```
benthorner: Minor: could we make this a ternary to remove the duplication?

spec/requests/file_attachments_spec.rb
```diff
@@ -101,6 +101,18 @@
         .to redirect_to(file_attachment_path(edition.document, file_attachment))
     end
 
+    it "redirects to attachments index view when featured attachment is created successfully" do
```
benthorner: I think this test goes in a different commit, as it relates to the 'success' case.

app/views/attachments/index.html.erb
```diff
@@ -0,0 +1,32 @@
...
+<% content_for :title, t("attachments.index.title", title: @edition.title_or_fallback) %>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <p class="govuk-body">
+      <%= t("attachments.index.upload.description_govspeak") %>
+    </p>
+
+    <%= render "govuk_publishing_components/components/button", {
+      text: t("attachments.index.upload.title"),
```
benthorner: Can we double-check this against the prototype/design? I think we're also missing an `else` clause for when we have no featured attachments.

spec/features/editing_file_attachments/upload_file_attachment_spec.rb
```diff
@@ -3,8 +3,21 @@
...
     when_i_go_to_edit_the_edition
     and_i_go_to_insert_an_attachment
+    and_i_select_a_file
     and_i_upload_a_file_attachment
     then_i_can_see_previews_of_the_attachment
+    and_i_can_see_a_link_to_the_attachment
+    and_i_see_the_timeline_entry
+  end
+
+  scenario do
```
benthorner: Can we add a description to the scenario to distinguish what they're doing? [Images are a good example of this.](https://github.com/alphagov/content-publisher/blob/master/spec/features/editing_images/edit_image_spec.rb)

spec/features/editing_file_attachments/upload_file_attachment_spec.rb
```diff
@@ -13,17 +26,34 @@ def given_there_is_an_edition
     @edition = create(:edition, document_type: document_type)
   end
 
+  def given_there_is_an_edition_with_a_pre_release_document_type
```
benthorner: Minor: we can reduce the future work to remove the pre-release flag by not mentioning the 'pre-release' aspect, since it's only temporary and tangential to the feature.

spec/features/editing_file_attachments/upload_file_attachment_spec.rb
```diff
@@ -3,8 +3,21 @@
     given_there_is_an_edition
     when_i_go_to_edit_the_edition
     and_i_go_to_insert_an_attachment
+    and_i_select_a_file
     and_i_upload_a_file_attachment
     then_i_can_see_previews_of_the_attachment
+    and_i_can_see_a_link_to_the_attachment
```
benthorner: We should reconsider these steps, as they're actually referring to the two different kinds of insertion methods for an inline file attachment, which doesn't make sense for featured file attachments at this point in the workflow. Can we have two distinct steps - something like `then_i_can_insert_the_attachment_inline` and `then_i_can_see_the_attachment`.

spec/features/editing_file_attachments/upload_file_attachment_spec.rb
```diff
@@ -3,8 +3,21 @@
...
+    and_i_see_the_timeline_entry
+  end
+
+  scenario do
+    given_there_is_an_edition_with_a_pre_release_document_type
+    when_i_visit_the_summary_page
+    and_i_go_to_change_an_attachment
+    and_i_go_to_upload_an_attachment
+    and_i_select_a_file
+    and_i_save_a_file_attachment
```
benthorner: Minor: I think this should be a "the" here, as we're talking about a specific attachment at this point. I wonder if there's a way we could combine some of these steps, as some of them aren't doing very much and just make the file/scenario longer - it's more noticeable now there are two scenarios.

spec/requests/attachments_spec.rb
```diff
@@ -0,0 +1,13 @@
+RSpec.describe "Attachments" do
```
benthorner: This comment is rather subjective, so I'll just write it and leave any changes up to you. It takes a lot of time to run the tests already - especially request and feature tests - and I'm worried about incrementally making that worse. It's not something we've really documented; personally, I only create new test files when I'm missing coverage for some reason - it just feels like extra effort to me.

app/views/documents/show.html.erb
```diff
@@ -51,6 +51,10 @@
   <div class="govuk-grid-column-two-thirds">
     <%= render "documents/show/contents" %>
 
+    <% if @edition.document_type.pre_release? && current_user.has_permission?(User::PRE_RELEASE_FEATURES_PERMISSION) %>
```
benthorner: Minor: we only need the user-level permission check here; we're doing [other work to make the rendering specific to the document type](https://trello.com/c/dzFScHVI/1498-disable-inline-file-attachments-when-featured-attachments-are-available).

app/views/file_attachments/new.html.erb
```diff
@@ -0,0 +1,41 @@
...
+      <%= hidden_field_tag(:wizard, params[:wizard]) %>
+
+      <%= render "govuk_publishing_components/components/file_upload", {
+        label: {
+          text: t("file_attachments.new.attachment_file.heading"),
+        },
+        name: "file"
+      } %>
+
+      <%= render "govuk_publishing_components/components/input", {
```
benthorner: I think this is supposed to be two-thirds width.

app/views/documents/show/_attachments.html.erb
```diff
@@ -0,0 +1,12 @@
...
+  id: "attachments",
+  title: t("documents.show.attachments.title"),
+  borderless: true,
+  edit: (
+    if @edition.editable?
+      { href: attachments_path(@edition.document),
+        data_attributes: { gtm: "edit-tags" } }
+    end
+  ),
+  block: t("documents.show.attachments.no_attachments")
```
benthorner: We should probably remove this until we do the overall rendering work - it was a surprise to see it's still here when I've uploaded a featured attachment.

leenagupte:
@benthorner Thanks for your review. I've received confirmation from Polly and Simon that the inline model can be updated, so I'm going to dismiss your review until then as the code is going to change quite a lot.

spec/features/editing_file_attachments/upload_file_attachment_spec.rb
leenagupte: I'll rename this, but I'll need to keep `pre_ release` set to true in the document type

app/views/documents/show.html.erb
leenagupte: I'm going to leave this check in for now, otherwise we'll be in the confusing situation where, until the "other work" is completed, people with pre_release features will see an attachments section on news and press releases.

I'll add an action to the other card to replace this check.

leenagupte:
> We're missing some pre-release checks for the new pages. It does seem a bit overkill sometimes, but we generally "fully" protect a new feature. I've dug out an example PR that shows how we were enforcing the permission for withdrawing a document.

@benthorner I don't think this needs to be added until the conditional added as part of https://trello.com/c/dzFScHVI/1498-disable-inline-file-attachments-when-featured-attachments-are-available is implemented, as the [application controller](https://github.com/alphagov/content-publisher/blob/master/app/controllers/application_controller.rb#L54-L58) already restricts access to all pages associated with a pre_release document type

Unless your suggestion is to add the restriction to the GET actions, in case the user hacks the URL for news stories?

benthorner:
@leenagupte that would be a valid scenario where the pre-release flag would be used. You are right that the work on that upcoming card would make pre-release protection redundant, so it does seem like quite a small risk and ultimately wouldn't cause any issue for the user, since the capabilities of the workflows are (currently) the same.

I reckon we could leave it, but @kevindew might have another perspective.

leenagupte:
I've tagged this as "Do not merge" as some clarification is needed about the changes to the modal.

spec/requests/attachments_spec.rb
leenagupte: I'll change this to a view test.

spec/features/editing_file_attachments/upload_file_attachment_spec.rb
leenagupte: This step no longer exists.

spec/features/editing_file_attachments/upload_file_attachment_spec.rb
leenagupte: This step no longer exists.

spec/features/editing_file_attachments/upload_file_attachment_spec.rb
```diff
@@ -35,17 +63,19 @@ def and_i_upload_a_file_attachment
...
+    @metadata = "PDF, 13 KB, 1 page"
 
     within(".gem-c-attachment") do
       expect(page).to have_content(@title)
-      expect(page).to have_content(metadata)
+      expect(page).to have_content(@metadata)
     end
+  end
 
+  def and_i_can_see_previews_of_the_attachment
```
benthorner: I think we should be more accurate with the naming here. This is only checking a single type of preview now: the inline version. I think it's important we try and keep the original meaning of the step, so maybe use a different step for each scenario, or rename this one to be about the inline preview specifically.

app/views/file_attachments/index.html.erb
```diff
@@ -36,7 +41,7 @@
           error_items: @issues&.items_for(:file_attachment_title),
           } %>
 
-        <%= render "govuk_publishing_components/components/button", { text: "Upload" } %>
+        <%= render "govuk_publishing_components/components/button", { text: "Save and continue" } %>
```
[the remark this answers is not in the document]
leenagupte: There is no design to refer to. Simon and Polly are going to get back to me.

spec/features/editing_file_attachments/upload_file_attachment_spec.rb
```diff
@@ -1,10 +1,21 @@
...
+    and_i_can_see_previews_of_the_attachment
+    and_i_see_the_timeline_entry
+  end
+
+  scenario "featured attachment" do
+    given_there_is_an_edition_that_allows_featured_attachments
+    when_i_visit_the_summary_page
+    and_i_go_to_change_an_attachment
+    and_i_go_to_add_an_attachment
+    and_i_upload_a_file_attachment
```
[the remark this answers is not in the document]
leenagupte: `and_i_upload_a_file_attachment` is being used in both scenarios. That's why it's separate.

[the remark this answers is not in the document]
benthorner: Ah yes, sorry - I remember that from my first review now!

spec/features/editing_file_attachments/upload_file_attachment_spec.rb
leenagupte: So, I need to revert this line, because the preview does work for both inline and featured attachments, however the test fails. So I need to investigate a bit more.

leenagupte: I'm getting my previews confused.

leenagupte:
@kevindew As agreed, I've removed the changes to the inline attachment upload modal.

app/controllers/attachments_controller.rb
```diff
@@ -0,0 +1,6 @@
+class AttachmentsController < ApplicationController
```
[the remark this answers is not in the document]
leenagupte: Sure. I called it Attachments because that's what the route is called, and I figured that inline attachments would start to be listed here.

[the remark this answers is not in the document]
kevindew: Gotcha - my understanding is that we intend for this controller to only be used for functionality related to featured attachments - which probably makes sense since the convergence of these things is super confusing.

app/views/attachments/_attachment.html.erb
```diff
@@ -0,0 +1,4 @@
+<%= render "components/attachment_meta", {
+  attachment: file_attachment_attributes(attachment, document),
+  actions: [],
+} %>
```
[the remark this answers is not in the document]
leenagupte: I was following the pattern set by the `_file_attachment` partial.
Also, the `actions` will be added to later stories, which could make the `index` view very long.

leenagupte:
> I also think we should cover the new routes with basic request tests

The request tests are being added in: https://github.com/alphagov/content-publisher/pull/1863

leenagupte:
@kevindew thanks. I've addressed your comments and fixed up.

leenagupte:
@kevindew As discussed yesterday, the basic request tests have been added and fixed up.

PR #1842, opened 2020-03-02
Add new 'attachments' attribute to document types

benthorner:
https://trello.com/c/dzFScHVI/1498-disable-inline-file-attachments-when-featured-attachments-are-available

This does some of the initial work on the card to unblock the next ones.

Previously we only supported (inline) file attachments for news, but we
will have a different workflow for publications, which have (inline)
featured attachments. This adds a new attribute to clearly indicate in what
form attachments are available for a document type.

### DIFF
--- a/config/document_types.yml
+++ b/config/document_types.yml
@@ -2,6 +2,7 @@
 news_article: &news_article
   path_prefix: /government/news
   lead_image: true
+  attachments: inline_file_only
   publishing_metadata:
     schema_name: news_article
     rendering_app: government-frontend
@@ -30,6 +31,7 @@ publication: &publication
   pre_release: true
   path_prefix: /government/publications
   lead_image: false
+  attachments: featured
   publishing_metadata:
     schema_name: publication
     rendering_app: government-frontend
--- a/spec/factories/document_type_factory.rb
+++ b/spec/factories/document_type_factory.rb
@@ -12,6 +12,7 @@ FactoryBot.define do
     tags { [] }
     guidance { [] }
     topics { false }
+    attachments { "inline_file_only" }
 
     contents do
       [DocumentType::SummaryField.new]
--- a/app/models/document_type.rb
+++ b/app/models/document_type.rb
@@ -34,6 +34,10 @@ class DocumentType
 
   alias_method :pre_release?, :pre_release
 
+  def attachments
+    ActiveSupport::StringInquirer.new(@attachments)
+  end
+
   class TagField
     include InitializeWithHash
     attr_reader :id, :type, :document_type, :hint
--- a/app/models/document_type.rb
+++ b/app/models/document_type.rb
@@ -1,7 +1,7 @@
 class DocumentType
   include InitializeWithHash
 
-  attr_reader :contents, :id, :managed_elsewhere, :publishing_metadata, :label,
+  attr_reader :contents, :id, :publishing_metadata, :label,
               :path_prefix, :tags, :topics
 
   def self.find(id)
--- a/app/models/document_type.rb
+++ b/app/models/document_type.rb
@@ -2,9 +2,7 @@ class DocumentType
   include InitializeWithHash
 
   attr_reader :contents, :id, :managed_elsewhere, :publishing_metadata, :label,
-              :path_prefix, :tags, :lead_image, :topics, :pre_release
-
-  alias_method :lead_image?, :lead_image
+              :path_prefix, :tags, :topics
 
   def self.find(id)
     item = all.find { |document_type| document_type.id == id }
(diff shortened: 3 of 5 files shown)


config/document_types.yml
```diff
@@ -30,6 +31,7 @@ publication: &publication
   pre_release: true
   path_prefix: /government/publications
   lead_image: false
+  attachments: featured_and_inline
```
leenagupte: Doesn't this contradict: https://trello.com/c/dzFScHVI/1498-disable-inline-file-attachments-when-featured-attachments-are-available ?

benthorner:
@leenagupte we did consider having two attributes. The conclusion we came to was that it's bizarre to allow both features to be "on" or "true", when that isn't something we support. I did [have a plan for improving how we represent features in general](https://github.com/alphagov/content-publisher/issues/1676), but we don't have much evidence or motivation to do any major work right now. In the meantime, this seemed like the simplest approach that represents what we're trying to say.

I'm deliberately not doing much in this PR, as the aim is purely to introduce this attribute to unblock some of the following cards; the work is quite sequential at the moment. If it helps to see a pattern, we've got another "setting the stage" type card coming up: https://trello.com/c/HRUXGp3K/1518-support-metadata-for-featured-file-attachments-db-modelling

config/document_types.yml
benthorner: Can you help me understand what you mean here?

leenagupte: So it sounds like Publications can only have featured attachments, and not inline ones, so this attribute will need to be changed.

benthorner: Hopefully this makes more sense following our chat: publications can have inline attachments, but the available actions are very different. Using the word `inline` twice does now seem confusing. I've update this to just say `featured`. 

We should be getting design docs on what these two modes mean, and I've had a chat to make sure we're using this terminology in that documentation. The places we use this conditional in the codebase will reflect the design doc, and be another way of defining what it means in executable code. We could have our own, separate doc in the app. I'm worried that would get out-of-date with the others, especially as we've barely started the work to implement the features; so I'd prefer not to add more documentation if possible.

benthorner:
@leenagupte thanks for the suggestion earlier. I've updated the PR and replied to your comment. Could you take another look? Also happy to chat about it if that's easier.

config/document_types.yml
```diff
@@ -31,7 +31,7 @@ publication: &publication
   pre_release: true
   path_prefix: /government/publications
   lead_image: false
-  attachments: featured_and_inline
+  attachments: featured
```
leenagupte: This is a better name.

I think the document type could just say "I have featured" attachments, and the rules in the code could determine what that means.

For example, the document type config doesn't need to configure that it can't "insert" via the inline attachments modal because it has feature attachments.  Does that make sense?

P.S. I assume this is commit is going to be fixed up? "foo" 🤨

benthorner: Ah, that's basically what I wrote in the original card (`featured_attachments: true`). The problem I had is that the insert/inline attachment actions and screens are very different between the news and publications, and both can't be used at the same time. Representing that with a single boolean smells because `featured_attachments: false` actually means turning something else on; representing with two booleans (`featured_attachments` and `inline_file_attachments`) also smells, because this means we can write config with both set to `true`, and would want a validation to stop that. Since we don't really have time to redesign our config to represent this better, I thought an enum/string was the best compromise.

Oops - fixed the commit!

benthorner: Does that help clarify the approach? Some of this is also written on the card.

leenagupte: So as far as I am aware documents can have both "inline" and "featured" attachments, because that is how the attachment ends being rendered.

To me, if you said that "publications" don't have inline attachments, what I hear is: "Attachments can't be inserted to the body of the text using the drop-down menu in the markdown editor in the body field".

The fact that for publications, the "inline" attachments can only be selected and inserted via the markdown editor, and not uploaded and inserted doesn't change that they are still rendered inside the body text.

So when I said, that the code not the config determines the behaviour, that's what I meant. That we can say all document types allow inline attachments, but if featured attachments is true, then attachments can't be uploaded via the markdown editor, only inserted. But if featured attachments is false/nil, then attachments can be both uploaded and inserted via the markdown editor.

benthorner:
@leenagupte thanks for the discussion earlier. I think we agreed there's a lot of detailed differences between inline attachments in the two modes we're looking to support, and that it's difficult to capture that succinctly. I've also added an extra commit to replace the conditional in the API payload generator, so it uses the new attribute.

Could you take another look?

spec/factories/document_type_factory.rb
```diff
@@ -12,6 +12,7 @@
     tags { [] }
     guidance { [] }
     topics { false }
+    attachments { "inline_file_only" }
```
leenagupte: Nice!

PR #1844, opened 2020-03-02
Remove `lint` Rake task

issyl0:
- This `lint` task is not very useful as it hides the directories that
  it operates on, and we're trying to move away from wrappers.
- Instead, I think we can leave people to run `bundle exec rubocop` on
  their own - it's less typing than `bundle exec rake lint` and
  `rubocop --help` works for options.

https://trello.com/c/IEUuFOVa/129-remove-the-lint-rake-task-from-apps-that-have-it
as part of the developer tooling group.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -4,4 +4,4 @@
 require_relative "config/application"
 
 Rails.application.load_tasks
-task default: %i(jasmine:ci lint brakeman)
+task default: %i(jasmine:ci brakeman)
--- a/lib/tasks/lint.rake
+++ b/lib/tasks/lint.rake
@@ -1,7 +1,0 @@
-desc "lint Ruby, FactoryBot, Sass and Javascript"
-task lint: :environment do
-  sh "bundle exec rubocop --format clang"
-  sh "bundle exec rake factorybot:lint RAILS_ENV='test'"
-  sh "bundle exec scss-lint app/assets/stylesheets"
-  sh "yarn run lint"
-end


kevindew:
@issyl0 this task encapsulates all the linting tasks of Content Publisher. So running rubocop is not a direct replacement. 

I'm a bit confused by: 
> This lint task is not very useful as it hides the directories that it operates on, and we're trying to move away from wrappers.

As this is information configured in rubocop?

issyl0:
> @issyl0 this task encapsulates all the linting tasks of Content Publisher. So running rubocop is not a direct replacement.

Oops. Obviously my "do the same thing across all the repos that contain `lib/tasks/lint.rake`" was inaccurate here. I based it on "all of the apps I've manually checked just use this for running `rubocop` and its predecessor `govuk-lint-ruby`".

I'll reinstate the file and just remove the Rubocop line.

> > This lint task is not very useful as it hides the directories that it operates on, and we're trying to move away from wrappers.
> 
> As this is information configured in rubocop?

That's a follow-up goal of the developer tools epic (bullet point 4): https://trello.com/c/nEHfEd1H/127-standardize-linting-infra-rake-tasks-jenkinsfile-ci. I'd love some help with all of this, there are _a lot_ of apps.

As it is now, `govuk-jenkinslib` [specifies directories to run the linter on](https://github.com/alphagov/govuk-jenkinslib/blob/master/vars/govuk.groovy#L594-L608) and we should be consistent with those. My point is that running `bundle exec rubocop app spec lib` is more explicit than `bundle exec rake lint` assuming people don't look at the Rake task first to see what directories it runs against.

Is this OK? What would you like to see better? Would you like to see the follow-up goal complete before this one?

issyl0:
You know what, I'm going to close this PR as it's not worth just removing one line given it's so engrained into your workflows. We can do the other things on that dev tools card later!

kevindew:
> > @issyl0 this task encapsulates all the linting tasks of Content Publisher. So running rubocop is not a direct replacement.
> 
> Oops. Obviously my "do the same thing across all the repos that contain `lib/tasks/lint.rake`" was inaccurate here. I based it on "all of the apps I've manually checked just use this for running `rubocop` and its predecessor `govuk-lint-ruby`".

Sorry, only after I'd put my comment did I learn this was a blanket change with a generic message, which now makes some of the content make more sense. It might be useful to flag when a PR message is part of a wide batch for context.

> I'll reinstate the file and just remove the Rubocop line.

Thanks - It'd be good to keep rubocop in there as well so that this covers all app linting.

> > > This lint task is not very useful as it hides the directories that it operates on, and we're trying to move away from wrappers.
> > 
> > 
> > As this is information configured in rubocop?
> 
> That's a follow-up goal of the developer tools epic (bullet point 4): https://trello.com/c/nEHfEd1H/127-standardize-linting-infra-rake-tasks-jenkinsfile-ci. I'd love some help with all of this, there are _a lot_ of apps.
> 
> As it is now, `govuk-jenkinslib` [specifies directories to run the linter on](https://github.com/alphagov/govuk-jenkinslib/blob/master/vars/govuk.groovy#L594-L608) and we should be consistent with those. My point is that running `bundle exec rubocop app spec lib` is more explicit than `bundle exec rake lint` assuming people don't look at the Rake task first to see what directories it runs against.
> 
> Is this OK? What would you like to see better? Would you like to see the follow-up goal complete before this one?

I think we should aim to have govuk-jenkinslib stop running the lint only on specific directories, it seems pretty weird that this point of configuration is in jenkinslib. We have that switched off in Content Publisher: https://github.com/alphagov/content-publisher/blob/d03f2cbb984fcff76208ba85ec80c205ad8c302c/Jenkinsfile#L17 and I think should work towards that being the default.

To avoid not configuring these directories in each app this info could be set in govuk-rubocop.

PR #1848, opened 2020-03-03
Use consistent describe blocks in request specs

benthorner:
https://trello.com/c/HILuWHOt

Previously the 'errors' spec was inconsistent with all other request
specs, and with itself: half of the file used a consistent format

    describe "METHOD /path-under-test"

while the other half used freetext. I found this confusing when
reviewing a change to this file [[1]], as the request(s) under test was
unclear. This modifies the 'errors' spec to use a consistent format
around each test, using just enough freetext to indicate the range
of paths being described (even if they're not actually covered by
the test).

[1]: https://github.com/alphagov/content-publisher/pull/1826#discussion_r385275352

### DIFF
--- a/spec/requests/errors_spec.rb
+++ b/spec/requests/errors_spec.rb
@@ -39,7 +39,7 @@ RSpec.describe "Errors" do
     end
   end
 
-  describe "bypassing user access checks" do
+  describe "GET /any-path-for-a-document" do
     it "returns a not found response when a document doesn't exist" do
       get document_path("document-that-does-not-exist")
 
@@ -47,8 +47,8 @@ RSpec.describe "Errors" do
     end
   end
 
-  describe "local data unavailable" do
-    it "returns service unavailable" do
+  describe "ANY /document-path-using-local-data" do
+    it "returns service unavailable when local data is unavailable" do
       # This is a somewhat contrived example, hopefully this can be replaced
       # when there's a simple way to trigger this error
       expect(Edition).to receive(:find_current)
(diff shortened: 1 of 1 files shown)


spec/requests/errors_spec.rb
```diff
@@ -39,16 +39,16 @@
...
       get document_path("document-that-does-not-exist")
 
       expect(response).to have_http_status(:not_found)
     end
   end
 
-  describe "local data unavailable" do
-    it "returns service unavailable" do
+  describe "ANY /document-path-using-local-data" do
+    it "returns service unavailable when locale data is unavailable" do
```
leenagupte: Should this be "local" rather than "locale"?

spec/requests/errors_spec.rb
```diff
@@ -61,15 +61,17 @@
     end
   end
 
-  it "forbids users without pre-release permission from accessing pre-release documents" do
-    pre_release_document_type = build(:document_type, pre_release: true)
-    edition = create(:edition, document_type: pre_release_document_type)
-    user = build(:user, permissions: %w(signin))
+  describe "ANY /pre-release-document-path" do
+    it "returns forbidden when the document type is pre-release" do
+      pre_release_document_type = build(:document_type, pre_release: true)
```
leenagupte: Not blocking: A `pre_release` [trait](https://github.com/alphagov/content-publisher/blob/master/spec/factories/document_type_factory.rb#L41) was added to the document_type factory after this test was added. I think most of the other tests were updated to use this trait, but it looks like this test was missed. Could it be updated as well?

spec/requests/errors_spec.rb
benthorner: Oops - updated 👍.

spec/requests/errors_spec.rb
benthorner: Good spot! Also updated.

PR #1849, opened 2020-03-03
Add removed_at to removals

emmabeynon:

This adds a `removed_at` column to Removal and sets this value when creating a Removal in the remove content rake tasks.  At this stage we are not making this a required value for a Removal; we will need to backfill existing Removals before doing so.

### DIFF
--- a/spec/lib/tasks/remove_spec.rb
+++ b/spec/lib/tasks/remove_spec.rb
@@ -12,13 +12,16 @@ RSpec.describe "Remove tasks" do
     it "delegates to RemoveDocumentService" do
       note = "The reason the edition is being removed"
 
-      ClimateControl.modify NOTE: note do
-        Rake::Task["remove:gone"].invoke(edition.content_id)
-      end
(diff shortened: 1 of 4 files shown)


spec/lib/tasks/remove_spec.rb
```diff
@@ -12,13 +12,16 @@
...
       note = "The reason the edition is being removed"
 
-      ClimateControl.modify NOTE: note do
-        Rake::Task["remove:gone"].invoke(edition.content_id)
-      end
-
-      expect(RemoveDocumentService).to have_received(:call) do |removed_edition, removal|
-        expect(removed_edition).to eq(edition)
-        expect(removal.explanatory_note).to eq(note)
+      travel_to(Time.zone.now) do
```
kevindew: you can achieve the same as this with `freeze_time do`

emmabeynon: Changed 👍

PR #1850, opened 2020-03-03
Backfill removed_at value for Removals

emmabeynon:

In anticipation of making `removed_at` a required attribute for a Removal
record, we backfill the value for existing Removals. Since we set the
`removed_at` value to the current time when a Removal is created, it makes
sense to set `removed_at` for existing Removals to their `created_at` value.

### DIFF
--- a/db/migrate/20200303143426_backfill_removals_removed_at.rb
+++ b/db/migrate/20200303143426_backfill_removals_removed_at.rb
@@ -1,0 +1,9 @@
+class BackfillRemovalsRemovedAt < ActiveRecord::Migration[6.0]
+  disable_ddl_transaction!
+
+  class Removal < ApplicationRecord; end
+
+  def up
+    Removal.find_each { |removal| removal.update!(removed_at: removal.created_at) }
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2020_03_03_115207) do
+ActiveRecord::Schema.define(version: 2020_03_03_143426) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"


db/migrate/20200303143426_backfill_removals_removed_at.rb
```diff
@@ -0,0 +1,13 @@
+class BackfillRemovalsRemovedAt < ActiveRecord::Migration[6.0]
+  disable_ddl_transaction!
+
+  class Removal < ApplicationRecord
+    has_one :status, as: :details
+  end
+
+  def up
+    Removal.all.each { |removal| removal.update!(removed_at: removal.created_at) }
```
1pretz1: Super tiny comment, you can use `find_each` here instead of `all.each` if you so wish which reads a little nicer!

db/migrate/20200303143426_backfill_removals_removed_at.rb
```diff
@@ -0,0 +1,13 @@
+class BackfillRemovalsRemovedAt < ActiveRecord::Migration[6.0]
+  disable_ddl_transaction!
+
+  class Removal < ApplicationRecord
+    has_one :status, as: :details
```
kevindew: This line isn't necessary by the way - so you can inline the class definition.

db/migrate/20200303143426_backfill_removals_removed_at.rb
kevindew: It's also more performant to do find_each as it pulls out in queries of 1000 - though not too big an issue as we probably have 5.

You can do `Removal.where(removed_at: nil)` so you're only replacing items that aren't already set (though I guess it's unlikely there are any given we have to do the removals)

db/migrate/20200303143426_backfill_removals_removed_at.rb
```diff
@@ -0,0 +1,13 @@
...
+  end
+
+  def up
+    Removal.all.each { |removal| removal.update!(removed_at: removal.created_at) }
+  end
+
+  def down; end
```
kevindew: Minor, you don't need to define this.

emmabeynon: Yeah I didn't bother with `find_each` since there are only 6 records.

db/migrate/20200303143426_backfill_removals_removed_at.rb
emmabeynon: Yeah I didn't bother with find_each since there are only 6 records.

db/migrate/20200303143426_backfill_removals_removed_at.rb
emmabeynon: Removed

PR #1855, opened 2020-03-04
Import change history

1pretz1:
We now want to generate change history for editions we import.  

Further info in the commits ->

Trello:
https://trello.com/c/MeyxpGHG/1465-import-publishedat-and-changehistory-for-whitehall-content

### DIFF
--- a/lib/whitehall_importer/change_history.rb
+++ b/lib/whitehall_importer/change_history.rb
@@ -1,0 +1,50 @@
+module WhitehallImporter
+  class ChangeHistory
+    attr_reader :entries
+
+    def initialize(whitehall_export)
+      @entries = build_entries(whitehall_export["editions"])
+    end
+
+    def for(edition_number)
+      entries.each_with_object([]) do |item, change_history|
+        change_history << item["change_history_entry"] if item["edition_number"] < edition_number
+      end
+    end
+
+  private
+
+    def build_entries(editions)
+      change_history_entries = editions.map.with_index do |edition, index|
+        edition_number = index + 1
+
+        next if edition["minor_change"] || edition_number == 1
+        next if edition_hasnt_been_published?(edition)
+
+        history_event(edition, edition_number)
+      end
+
+      change_history_entries.compact.reverse
+    end
+
+    def edition_hasnt_been_published?(edition)
+      %w[submitted rejected draft].include?(edition["state"]) && edition["unpublishing"].blank?
+    end
+
+    def history_event(edition, edition_number)
+      publish_event = EditionHistory.new(edition["revision_history"]).last_state_event("published")
+
+      raise AbortImportError, "Edition has a major change but no change note" if edition["change_note"].blank?
+      raise AbortImportError, "Edition has a major change but no publish event" if publish_event.blank?
+
+      {
+        "edition_number" => edition_number,
+        "change_history_entry" => {
+          "id" => SecureRandom.uuid,
+          "note" => edition["change_note"],
+          "public_timestamp" => publish_event["created_at"],
+        },
+      }
+    end
+  end
+end
--- a/lib/whitehall_importer/create_edition.rb
+++ b/lib/whitehall_importer/create_edition.rb
@@ -1,15 +1,17 @@
 module WhitehallImporter
   class CreateEdition
-    attr_reader :document_import, :current, :whitehall_edition, :edition_number, :user_ids
+    attr_reader :document_import, :whitehall_edition, :change_history, :current, :edition_number, :user_ids
 
     def self.call(*args)
       new(*args).call
     end
 
-    def initialize(document_import:, whitehall_edition:, current: true, edition_number: 1, user_ids: {})
+    def initialize(document_import:, whitehall_edition:, change_history:,
+                   current: true, edition_number: 1, user_ids: {})
       @document_import = document_import
-      @current = current
       @whitehall_edition = whitehall_edition
+      @change_history = change_history
+      @current = current
       @edition_number = edition_number
       @user_ids = user_ids
     end
@@ -27,8 +29,10 @@ module WhitehallImporter
                   create_removed_edition
                 else
                   state = MigrateState.call(whitehall_edition["state"], whitehall_edition["force_published"])
-                  status = build_status(state)
-                  create_edition(status: status, current: current, edition_number: edition_number)
+                  revision = create_revision(edition_number)
+                  status = build_status(revision, state)
+                  create_edition(status: status, current: current,
+                                 edition_number: edition_number, revision: revision)
                 end
 
       create_revision_history(edition)
@@ -40,8 +44,8 @@ module WhitehallImporter
 
   private
 
-    def revision
-      @revision ||= CreateRevision.call(document_import, whitehall_edition)
+    def create_revision(edition_number)
+      CreateRevision.call(document_import, whitehall_edition, change_history.for(edition_number))
     end
 
     def history
@@ -54,24 +58,30 @@ module WhitehallImporter
 
     def split_unpublished_edition
       unpublishing_event = history.last_unpublishing_event!
+      removed_revision = create_revision(edition_number)
       create_edition(
-        status: build_status("removed", build_removal),
+        status: build_status(removed_revision, "removed", build_removal),
         current: false,
         edition_number: edition_number,
         last_event: unpublishing_event,
+        revision: removed_revision,
       )
 
+      current_revision = create_revision(edition_number + 1)
+      migrated_state = MigrateState.call(whitehall_edition["state"], whitehall_edition["force_published"])
       create_edition(
-        status: build_status(MigrateState.call(whitehall_edition["state"], whitehall_edition["force_published"])),
+        status: build_status(current_revision, migrated_state),
         edition_number: edition_number + 1,
         current: true,
         create_event: history.next_event!(unpublishing_event),
+        revision: current_revision,
       )
     end
 
     def create_removed_edition
-      removed_status = build_status("removed", build_removal)
-      create_edition(status: removed_status, current: current, edition_number: edition_number)
+      revision = create_revision(edition_number)
+      removed_status = build_status(revision, "removed", build_removal)
+      create_edition(status: removed_status, current: current, edition_number: edition_number, revision: revision)
     end
 
     def check_only_in_english
@@ -83,9 +93,11 @@ module WhitehallImporter
     end
 
     def create_withdrawn_edition
-      create_edition(status: build_status("published"),
+      revision = create_revision(edition_number)
+      create_edition(status: build_status(revision, "published"),
                      current: current,
-                     edition_number: edition_number).tap { |edition| set_withdrawn_status(edition) }
+                     edition_number: edition_number,
+                     revision: revision).tap { |edition| set_withdrawn_status(edition) }
     end
 
     def set_withdrawn_status(edition)
@@ -99,7 +111,7 @@ module WhitehallImporter
         withdrawn_at: whitehall_edition["unpublishing"]["created_at"],
       )
 
-      edition.update!(status: build_status("withdrawn", withdrawal))
+      edition.update!(status: build_status(edition.revision, "withdrawn", withdrawal))
     end
 
     def create_scheduled_edition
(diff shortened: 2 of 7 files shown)


lib/whitehall_importer/change_history.rb
```diff
@@ -0,0 +1,41 @@
...
+          "edition_number" => edition_number,
+          "change_history_entry" => {
+            "id" => SecureRandom.uuid,
+            "note" => edition["change_note"],
+            "public_timestamp" => publish_event["created_at"],
+          },
+        }
+      end
+    end
+  end
```
kevindew: This class looks like a nice encapsulation of the change history concerns.

There's a few concerns I have that there is scatterings of logic slipping into different methods perhaps unnecessarily. For example: the caller of whitehall_document_change_history needs to know that some of the entries are nil and call compact; the for method needs to know that the data is a different order.

I'd wonder if you want to aim for something a bit more compartmentalised. For example (with some variable name tweaks)

```
class ChangeHistory
  attr_reader :entries
  
  def initialize(whitehall_export)
    @entries = build_entries(whitehall_export["editions"])
  end
  
  def for(edition_number)
    entries.each_with_object([]) do |item, memo|
      memo << item["change_history_entry"] if item["edition_number"] < edition_number
    end
  end
  
private

  def build_entries(editions)
    # deals with the looping, reversing and compacting. Calls #history_entry for each item
  end

  def history_entry(edition, edition_number)
    # deals with guard clauses, exceptions and returning the hash
  end
end
```

lib/whitehall_importer/create_edition.rb
```diff
@@ -1,17 +1,19 @@
...
 
-    def initialize(document_import:, whitehall_edition:, current: true, edition_number: 1, user_ids: {})
+    def initialize(document_import:, whitehall_edition:, change_history:,
+                   current: true, edition_number: 1, user_ids: {})
       @document_import = document_import
       @current = current
       @whitehall_edition = whitehall_edition
       @edition_number = edition_number
       @user_ids = user_ids
+      @change_history = change_history
```
kevindew: Minor, and not your fault, but would be nice if these matched param ordering.

lib/whitehall_importer/create_revision.rb
```diff
@@ -16,9 +16,10 @@ def self.call(*args)
       new(*args).call
     end
 
-    def initialize(document_import, whitehall_edition)
+    def initialize(document_import, whitehall_edition, change_history = [])
```
kevindew: Do we have any advantages of this defaulting to an empty array?

lib/whitehall_importer/create_revision.rb
```diff
@@ -16,9 +16,10 @@ def self.call(*args)
...
       @document_import = document_import
       @whitehall_edition = whitehall_edition
+      @change_history = change_history
```
kevindew: Does seem a little bit of a bummer that we use the same variable name here and in create_edition to refer to different objects 🤔 I can't think of a nice solution as I think the name we have in here is correct and the most important one, nor do I really want to suggest making the other one longer like change_history_extractor or something. I guess it'll all be ok

spec/lib/whitehall_importer/change_history_spec.rb
```diff
@@ -0,0 +1,81 @@
+RSpec.describe WhitehallImporter::ChangeHistory do
```
kevindew: Looks like we're missing a `describe "#for"` block

lib/whitehall_importer/change_history.rb
```diff
@@ -0,0 +1,41 @@
...
+
+  private
+
+    def whitehall_document_change_history(whitehall_export)
+      whitehall_export["editions"].map.with_index do |edition, index|
+        publish_event = EditionHistory.new(edition["revision_history"]).last_state_event("published")
+        edition_number = index + 1
+
+        next if edition["minor_change"] || edition_number == 1
+        next if %w[submitted rejected draft].include?(edition["state"]) && edition["unpublishing"].blank?
```
kevindew: This caught me out a little - it might be better to capture this condition in a method or variable name. I'd guess this is `edition_hasnt_been_published?` ?

spec/lib/whitehall_importer/change_history_spec.rb
```diff
@@ -0,0 +1,81 @@
...
+                               revision_history: [build(:whitehall_export_revision_history_event,
+                                                        state: "published",
+                                                        created_at: Time.zone.now.beginning_of_day.rfc3339)])
+    published_edition = build(:whitehall_export_edition,
+                              state: "published",
+                              change_note: "Some great changes",
+                              revision_history: [build(:whitehall_export_revision_history_event,
+                                                       state: "published")])
+    whitehall_export = build(:whitehall_export_document,
+                             editions: [first_edition, superseded_edition, published_edition])
```
kevindew: I don't think creating this published edition actually has any impact on the test?

spec/lib/whitehall_importer/change_history_spec.rb
```diff
@@ -0,0 +1,81 @@
...
+  it "raises if an edition has a major change but no publish event" do
+    edition = build(:whitehall_export_edition,
+                    state: "published",
+                    change_note: "Some change",
+                    revision_history: [build(:whitehall_export_revision_history_event)])
+    whitehall_export = build(:whitehall_export_document,
+                             editions: [first_edition, edition])
+
+    expect { described_class.new(whitehall_export).for(2) }
+      .to raise_error(WhitehallImporter::AbortImportError, "Edition has a major change but no publish event")
```
kevindew: 👍

spec/lib/whitehall_importer/change_history_spec.rb
```diff
@@ -0,0 +1,81 @@
...
+  it "raises if an edition has a major change but no change note" do
+    edition = build(:whitehall_export_edition,
+                    state: "published",
+                    change_note: "",
+                    revision_history: [build(:whitehall_export_revision_history_event, state: "published")])
+    whitehall_export = build(:whitehall_export_document,
+                             editions: [first_edition, edition])
+
+    expect { described_class.new(whitehall_export).for(2) }
+      .to raise_error(WhitehallImporter::AbortImportError, "Edition has a major change but no change note")
```
kevindew: 👍

spec/lib/whitehall_importer/change_history_spec.rb
```diff
@@ -0,0 +1,81 @@
...
+                                             created_at: Time.zone.now.beginning_of_day.rfc3339)])
+    whitehall_export = build(:whitehall_export_document,
+                             editions: [first_edition, edition])
+
+    expect(described_class.new(whitehall_export).for(3)).to match([
+      { "id" => anything, "note" => "Some great changes", "public_timestamp" => Time.zone.now.beginning_of_day.rfc3339 },
+    ])
+  end
+
+  it "doesn't return change history for minor change editions" do
```
kevindew: You probably also want a test that it skips drafts.

spec/lib/whitehall_importer/change_history_spec.rb
```diff
@@ -0,0 +1,81 @@
+RSpec.describe WhitehallImporter::ChangeHistory do
+  let(:first_edition) do
+    build(:whitehall_export_edition,
+          state: "superseded",
+          revision_history: [build(:whitehall_export_revision_history_event,
+                                   state: "published",
+                                   created_at: Time.zone.now.last_year.rfc3339)])
```
kevindew: Do we use this at all?

spec/lib/whitehall_importer/change_history_spec.rb
```diff
@@ -0,0 +1,81 @@
...
+  end
+
+  it "returns change history for the published edition" do
```
kevindew: This test can probably be split in two where in the first we just want to test that this method returns some change history and then a second test where we test how the number we pass into for limits the results returned.

spec/lib/whitehall_importer/change_history_spec.rb
```diff
@@ -0,0 +1,81 @@
...
+                    state: "published",
+                    change_note: "Some change",
+                    revision_history: [build(:whitehall_export_revision_history_event)])
+    whitehall_export = build(:whitehall_export_document,
+                             editions: [first_edition, edition])
+
+    expect { described_class.new(whitehall_export).for(2) }
+      .to raise_error(WhitehallImporter::AbortImportError, "Edition has a major change but no publish event")
+  end
+end
```
kevindew: This test looks pretty good, a couple of cases still to cover. It does seem pretty heavy on the boilerplate needed for it. I wonder if we can reduce some of that with factory bot traits. I know we have one for published that could reduce some of the code here: https://github.com/alphagov/content-publisher/blob/bc6b81dead92587cebd65dba2d36fe27a95c4a88/spec/factories/whitehall_export/edition_factory.rb#L60-L75.

May also want one for superseded and to have published and superseded accept a published_at transient attribute.

lib/whitehall_importer/change_history.rb
1pretz1: Yep I agree with your concerns & thanks for the comment, I was struggling to brake the class up and your suggestion separates things nicely!! :)

lib/whitehall_importer/create_edition.rb
1pretz1: Yep sure, I've also made sure the ordering matches up in the import file which passes these arguments over!

lib/whitehall_importer/create_revision.rb
1pretz1: The advantage would be not having to add extra code to the create_revision_spec, think theres around 25 tests so seemed worthwhile, what do you think?

lib/whitehall_importer/create_revision.rb
1pretz1: I know I felt the same when I added this but came to the same conclusion as you. 🤷‍♂

spec/lib/whitehall_importer/change_history_spec.rb
1pretz1: Sure... added

spec/lib/whitehall_importer/change_history_spec.rb
1pretz1: Nope we don't, added it originally so the first edition would have a published at time which was before the third edition lets say, have removed it anyway as it's superfluous.

spec/lib/whitehall_importer/change_history_spec.rb
1pretz1: I've split the test up into 4, testing each point of the import for each edition. This seems a bit OTT but gives a nice test case which tests a semi realistic situation. What do you think?

spec/lib/whitehall_importer/change_history_spec.rb
1pretz1: Sure, this cleans up the tests nicely, don't know why I didn't use the factories! 🤦‍♂

spec/lib/whitehall_importer/change_history_spec.rb
1pretz1: Sure!

spec/lib/whitehall_importer/change_history_spec.rb
1pretz1: Nope it doesn't, since however I have updated the test so it makes use of the published edition...

lib/whitehall_importer/change_history.rb
1pretz1: Good point, have moved it to a method.

1pretz1:
Thankyou for the review @kevindew, found your comments super useful and the code looks a lump better than first time around! Ready for another review.

lib/whitehall_importer/create_revision.rb
kevindew: Yup sounds a good reason 👍

spec/lib/whitehall_importer/change_history_spec.rb
```diff
@@ -0,0 +1,86 @@
...
+          { "id" => anything, "note" => "Some fantastic changes", "public_timestamp" => Time.zone.now.beginning_of_week.rfc3339 },
+        ])
+      end
+
+      it "returns more change history for the fourth edition" do
+        expect(described_class.new(whitehall_export).for(4)).to match([
+          { "id" => anything, "note" => "Some great changes", "public_timestamp" => Time.zone.now.beginning_of_day.rfc3339 },
+          { "id" => anything, "note" => "Some fantastic changes", "public_timestamp" => Time.zone.now.beginning_of_week.rfc3339 },
+        ])
+      end
```
kevindew: I don't think you've really got these tests named correctly to capture the functionality of the method. They seem to be focused on very circumstantial cases.

tests I'd suggest are:

(outside of this context) it "doesn't include the first edition in change history"
it "returns the change history for the editions preceding the given edition number" 
it "returns all the change history for a number greater than the number of editions"

1pretz1: You're correct, I've amended them accordingly!

1pretz1:
Ready for another look @kevindew!

spec/lib/whitehall_importer/change_history_spec.rb
kevindew: I think the thing we miss by combining those tests is any test that checks that if there are say 5 change history events and we request edition 3 we don't get back all 5 results.

1pretz1: Damn I only saw this comment after I saw the approve which I merged! You make a good point which I didn't consider, I'll create another PR to add this test!

PR #1856, opened 2020-03-04
Maintain removed_at in Remove Document Service

emmabeynon:

In cases where we attempt to remove an edition that has already been removed
(for example, if the explanation needs to be updated) then we need to ensure
that the existing `removed_at` date is kept the same. This modifies the Remove
Document Service to do this.  We also update the Remove Document Service to
send the `removed_at` date to the Publishing API as the `unpublished_at` value.

### DIFF
--- a/app/services/remove_document_service.rb
+++ b/app/services/remove_document_service.rb
@@ -9,6 +9,7 @@ class RemoveDocumentService < ApplicationService
     Document.transaction do
       edition.document.lock!
       check_removeable
+      set_removed_at
       unpublish_edition
       update_edition_status
       create_timeline_entry
@@ -45,6 +46,14 @@ private
       status: edition.status,
       details: removal,
     )
+  end
+
+  def set_removed_at
+    removal.removed_at = if edition.removed?
+                           edition.status.details.removed_at
+                         else
+                           Time.zone.now
+                         end
   end
 
   def check_removeable
--- a/app/services/remove_document_service.rb
+++ b/app/services/remove_document_service.rb
@@ -29,6 +29,7 @@ private
       explanation: removal.explanatory_note,
       alternative_path: removal.alternative_url,
       locale: edition.locale,
+      unpublished_at: removal.removed_at,
     )
   end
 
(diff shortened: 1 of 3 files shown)


spec/services/remove_document_service_spec.rb
```diff
@@ -78,13 +79,40 @@
...
+        stub_publishing_api_unpublish(
+          edition.content_id,
+          body: {
+            alternative_path: "/path",
+            explanation: "explanation",
+            locale: edition.locale,
+            type: "redirect",
+            unpublished_at: existing_removal.removed_at,
+          },
+        )
```
kevindew: I think for this stub we either want to assert it was requested or not have it defined, as it's already covered by the `stub_any_publishing_api_unpublish` in the before block - I don't really think we need it since we're already checking removed_at is sent in other test.

spec/services/remove_document_service_spec.rb
```diff
@@ -78,13 +79,40 @@
...
           },
         )
         described_class.call(edition, removal)
         expect(request).to have_been_requested
       end
     end
 
+    context "when a removal already exists" do
+      it "maintains the existing removed_at time" do
+        existing_removed_at = 2.days.ago
```
kevindew: There's a good chance this could end up flaky due to the annoyance of date precisions at Ruby and DB level. You can probably resolve this by adding a .noon on the end.

app/services/remove_document_service.rb
```diff
@@ -47,6 +49,10 @@ def create_timeline_entry
     )
   end
 
+  def update_removed_at
+    removal.removed_at = edition.status.details.removed_at
```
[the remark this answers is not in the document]
emmabeynon: Done 👍

spec/services/remove_document_service_spec.rb
emmabeynon: Done 👍

spec/services/remove_document_service_spec.rb
emmabeynon: Removed 👍

emmabeynon:
Thanks @kevindew your comments should be addressed now.

PR #1861, opened 2020-03-04
Persist content_ids for document_imports on import

Rosa-Fox:
[Trello](https://trello.com/c/ACUVDeXD/1523-export-contentid-from-whitehall-into-content-publisher)

- Will enable the 'View on Whitehall' links in the migration UI to work (as they rely on having the content ID to form the URL path).
- It will also make it easier to find a relationship between the document_import and the whitehall document.

Merge after: https://github.com/alphagov/whitehall/pull/5417

### DIFF
--- a/lib/whitehall_importer/create_migration.rb
+++ b/lib/whitehall_importer/create_migration.rb
@@ -45,6 +45,7 @@ module WhitehallImporter
         record.document_imports.create!(
           whitehall_document_id: document["document_id"],
           state: "pending",
+          content_id: document["content_id"],
         )
       end
     end
--- a/spec/factories/whitehall_export/index_document_factory.rb
+++ b/spec/factories/whitehall_export/index_document_factory.rb
@@ -4,13 +4,7 @@ FactoryBot.define do
 
     sequence(:document_id)
 
-    document_information {
-      {
-        locales: %w(en),
-        sub_types: %w(news_story),
-        lead_organisations: %w(96ae61d6-c2a1-48cb-8e67-da9d105ae381),
-      }.stringify_keys
-    }
+    content_id { SecureRandom.uuid }
 
     initialize_with { attributes.stringify_keys }
   end
--- a/spec/lib/whitehall_importer/create_migration_spec.rb
+++ b/spec/lib/whitehall_importer/create_migration_spec.rb
@@ -36,6 +36,11 @@ RSpec.describe WhitehallImporter::CreateMigration do
           .to change { WhitehallMigration::DocumentImport.pending.count }.by(110)
       end
 
+      it "sets the content_id on the WhitehallMigration::DocumentImport" do
+        described_class.call("123", "news_article")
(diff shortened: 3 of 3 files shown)


spec/factories/whitehall_export/index_document_factory.rb
```diff
@@ -4,6 +4,8 @@
 
     sequence(:document_id)
 
+    content_id { SecureRandom.uuid }
```
brucebolt: Based on the change in https://github.com/alphagov/whitehall/pull/5417/files#diff-fe39f00137564632291307d578535423R15, the `content_id` will need to be nested under `document_information` (as it is in Whitehall).

lib/whitehall_importer/create_migration.rb
```diff
@@ -45,6 +45,7 @@ def create_document_imports(page, record)
         record.document_imports.create!(
           whitehall_document_id: document["document_id"],
           state: "pending",
+          content_id: document["content_id"],
```
brucebolt: As per my comment in the `index_document_factory`, this will need to be obtained from the `document_information`:
```suggestion
          content_id: document["document_information"]["content_id"],
```

spec/factories/whitehall_export/index_document_factory.rb
```diff
@@ -4,13 +4,10 @@
...
 
-    document_information {
-      {
-        locales: %w(en),
-        sub_types: %w(news_story),
-        lead_organisations: %w(96ae61d6-c2a1-48cb-8e67-da9d105ae381),
-      }.stringify_keys
-    }
+    content_id { SecureRandom.uuid }
+    locales { %w(en) }
```
brucebolt: Since they are not being exported from Whitehall now, `locales`, `sub_types` and `lead_organisations` can all be removed from this factory.

spec/lib/whitehall_importer/create_migration_spec.rb
```diff
@@ -36,6 +36,11 @@
           .to change { WhitehallMigration::DocumentImport.pending.count }.by(110)
       end
 
+      it "sets the content_id on the WhitehallMigration::DocumentImport" do
+        described_class.call("123", "news_article")
+        expect(WhitehallMigration::DocumentImport.last.content_id).not_to eq(nil)
```
brucebolt: Could this be updated to check the correct ID is set, e.g.
```
expect(WhitehallMigration::DocumentImport.last.content_id).to eq(whitehall_export_page_1.last[:content_id])
```

spec/factories/whitehall_export/index_document_factory.rb
Rosa-Fox: 👍

spec/lib/whitehall_importer/create_migration_spec.rb
Rosa-Fox: 👍

Rosa-Fox:
Thanks @brucebolt. Have amended.

PR #1863, opened 2020-03-05
Restrict attachment actions

leenagupte:
Trello: https://trello.com/c/dzFScHVI
Follows on from: #1842 
Depends on: #1841 

# What's changed?
Only allow inline attachments to be uploaded from the markdown editor if the document type allows inline attachments.
Only allow attachments to be uploaded via the Attachments section in the document summary if the document type allows featured attachments.

# Why?
This allows different types of attachments to be configured ways, and stops users from accidentally stumbling across pages or options that might confuse them.

### DIFF
--- a/app/controllers/featured_attachments_controller.rb
+++ b/app/controllers/featured_attachments_controller.rb
@@ -2,5 +2,9 @@ class FeaturedAttachmentsController < ApplicationController
   def index
     @edition = Edition.find_current(document: params[:document])
     assert_edition_state(@edition, &:editable?)
+
+    assert_edition_feature(@edition, assertion: "supports featured attachments") do
+      @edition.document_type.attachments.featured?
+    end
   end
 end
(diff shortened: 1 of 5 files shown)


app/controllers/attachments_controller.rb
```diff
@@ -2,5 +2,9 @@ class AttachmentsController < ApplicationController
   def index
     @edition = Edition.find_current(document: params[:document])
     assert_edition_state(@edition, &:editable?)
+
```
benthorner: Minor: the commit message is slightly different to the code, in that the changes don't mention pre-release.

leenagupte:
> Looks good - thanks @leenagupte 👍. I understand it took a bit of effort to pull the different pieces together. Only thing that seems to be missing is using the `attachments` attribute to show/hide the section on the doc summary, which I see depends on #1841

I'm already using the "attachments" attribute: https://github.com/alphagov/content-publisher/pull/1841/commits/4e582bd3c373593c1eb72d6963ad99fd031e1417

leenagupte:
@benthorner This was approved before a lot of the changes to https://github.com/alphagov/content-publisher/pull/1841 were made. Do you want to re-review?

PR #1866, opened 2020-03-05
Change migration view to use external Whitehall links

emmabeynon:
This changes links to view migrated documents in Whitehall to use external links.


emmabeynon:
@kevindew Yep I noticed it when trying to access the Whitehall links from a migration I ran today.

PR #1868, opened 2020-03-06
Send file attachments details in publishing-api payload

leenagupte:
Trello: https://trello.com/c/lHtCxM6V

# What's changed and why?
Send attachment and featured attachment data to the publishing-api.

As part of the RFC on Attachments we now make data about attachments available in the Content API in a way that's machine readable so that's consumers of the Content API will soon no longer have to manually parse data about attachment from the rendered HTML.

Adding this data into the payload will also featured attachments to be rendered on the frontend as this is done by the publishing-api.

# Expected changes
attachents_details

### DIFF
--- a/lib/publishing_api_payload.rb
+++ b/lib/publishing_api_payload.rb
@@ -77,6 +77,7 @@ private
     details = {
       political: edition.political?,
       change_history: history.change_history,
+      attachments: attachments,
     }
 
     if document_type.lead_image? && edition.lead_image_revision.present?
@@ -90,5 +91,13 @@ private
 
   def publication?
     publishing_metadata.schema_name == "publication"
+  end
+
+  def attachments
+    file_attachments = edition.file_attachment_revisions
+
+    file_attachments.map do |attachment|
+      FileAttachmentPayload.new(attachment, edition.document).payload
+    end
   end
 end
--- a/spec/lib/publishing_api_payload_spec.rb
+++ b/spec/lib/publishing_api_payload_spec.rb
@@ -59,6 +59,19 @@ RSpec.describe PublishingApiPayload do
       expect(payload[:details][:change_history]).to match(history.change_history)
     end
 
+    it "delegates to PublishingApiPayload::FileAttachmentPayload to populate attachments" do
+      file_attachment_revision = create(:file_attachment_revision)
+      edition = build(:edition,
+                      file_attachment_revisions: [file_attachment_revision])
+
+      payload = described_class.new(edition).payload
+      attachment_payload = PublishingApiPayload::FileAttachmentPayload
+                             .new(file_attachment_revision, edition.document)
+                             .payload
+
+      expect(payload[:details][:attachments]).to contain_exactly(attachment_payload)
+    end
+
     it "specifies an auth bypass ID for anonymous previews" do
       edition = build(:edition)
       preview_auth_bypass = instance_double(PreviewAuthBypass, auth_bypass_id: "id")
--- a/lib/publishing_api_payload/file_attachment_payload.rb
+++ b/lib/publishing_api_payload/file_attachment_payload.rb
@@ -1,0 +1,21 @@
+class PublishingApiPayload::FileAttachmentPayload
+  include Rails.application.routes.url_helpers
+  include FileAttachmentHelper
+
+  attr_reader :attachment, :document
+
+  def initialize(attachment, document)
+    @attachment = attachment
+    @document = document
+  end
+
+  def payload
+    payload = {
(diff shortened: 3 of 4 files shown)


lib/publishing_api_payload.rb
```diff
@@ -87,13 +87,17 @@ def details
...
     }
 
     if document_type.lead_image? && edition.lead_image_revision.present?
       details[:image] = image
     end
 
-    details[:documents] = [] if document_type.attachments.featured?
+    if document_type.attachments.featured?
+      details[:documents] = []
+      details[:featured_attachments] = featured_attachments
```
kevindew: I don't think we need to do this as part of this story. I've just been talking to @barrucadu about the usage of this field (as I was a bit surprised by our usage) and I think we're going to change the approach in content schemas.

lib/publishing_api_payload.rb
```diff
@@ -112,4 +116,23 @@ def roles_and_people(role_appointments)
...
+    end
+  end
+
+  def featured_attachments
+    file_attachments = edition.file_attachment_revisions
+
+    file_attachments.map do |attachment|
+      {
+        content_type: attachment.content_type,
+        content: attachment.asset.bytes,
```
kevindew: I was a bit surprised that we were sending binary or string data here 😱 - I believe the intention is this would have content_type: "text/govspeak" and content be a govspeak snippet. Though I'm hoping to change this to an id and have components directly rendered on the frontend.

lib/publishing_api_payload/file_attachment.rb
```diff
@@ -0,0 +1,22 @@
...
+  def initialize(attachment, document)
+    @attachment = attachment
+    @document = document
+  end
+
+  def payload
+    payload = {
+      attachment_type: "file",
+      locale: document.locale,
+      preview_url: file_attachment_preview_url(attachment, document),
```
kevindew: In Publishing API metadata this field is intended to be the public path to a CSV preview. So we shouldn't have a CP link. We can not fill this field for the time being.

lib/publishing_api_payload.rb
```diff
@@ -112,4 +116,23 @@ def roles_and_people(role_appointments)
         memo[:people] = (memo[:people] + people).uniq
       end
   end
+
+  def attachments
+    file_attachments = edition.file_attachment_revisions
+
+    file_attachments.map do |attachment|
+      FileAttachment.new(attachment, edition.document).payload
```
kevindew: I wonder if we want to call this FileAttachmentPayload so it feels more distinct in namespace from the global FileAttachment model?

lib/publishing_api_payload.rb
leenagupte: To be fair, I struggled to find documentation about what should go in this field. I was hoping a review would provide clarity 😄

leenagupte:
@kevindew Thanks. I've addressed your comments and fixed up.

spec/lib/publishing_api_payload_spec.rb
```diff
@@ -59,6 +59,24 @@
       expect(payload[:details][:change_history]).to match(history.change_history)
     end
 
+    it "delegates to PublishingApiPayload::FileAttachmentPayload to populate attachments" do
+      attachment = instance_double(
+        PublishingApiPayload::FileAttachmentPayload,
+        payload: {},
+      )
+
+      allow(PublishingApiPayload::FileAttachmentPayload).to receive(:new).and_return(attachment)
```
kevindew: Should this be an expect rather than allow to assert it's called?

spec/lib/publishing_api_payload_spec.rb
```diff
@@ -59,6 +59,24 @@
...
+
+      edition = build(
+        :edition,
+        file_attachment_revisions: [build(:file_attachment_revision)],
+      )
+
+      payload = described_class.new(edition).payload
+
+      expect(payload[:details][:attachments]).to match([attachment.payload])
+    end
```
kevindew: I think this test could actually be a bit shorter without using mocks - which could make it a little safer:

```
      file_attachment_revision = create(:file_attachment_revision)
      edition = build(:edition,
                      file_attachment_revisions: [file_attachment_revision])

      payload = described_class.new(edition).payload
      attachment_payload = PublishingApiPayload::FileAttachmentPayload
                             .new(file_attachment_revision, edition.document)
                             .payload

      expect(payload[:details][:attachments]).to contain_exactly(attachment_payload)
```

leenagupte:
@kevindew Thanks for your comments. I've addressed them and fixed up.

PR #1869, opened 2020-03-06
Fix public_updated_at time for published editions

1pretz1:
Previously if an edition has been previously published with a major
change the public_updated_at time being sent to the publishing-api
would've been set to the current time, this is incorrect behaviour and
as the time should be set to the published_at time of the edition.

### DIFF
--- a/spec/lib/publishing_api_payload/history_spec.rb
+++ b/spec/lib/publishing_api_payload/history_spec.rb
@@ -1,12 +1,24 @@
 RSpec.describe PublishingApiPayload::History do
   describe "#public_updated_at" do
-    it "returns current time if a major change is published" do
-      freeze_time do
-        first_edition = build(:edition)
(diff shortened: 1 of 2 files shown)


spec/lib/publishing_api_payload/history_spec.rb
```diff
@@ -9,6 +9,18 @@
       end
     end
 
+    it "returns published_at time if an edition has already been published with a major change" do
```
kevindew: Thanks for capturing this in a test. When compared in conjunction with the test preceding it the wording of that one no longer makes sense, I think we should adapt that to make explicit when this would fall back to current time.

1pretz1: Sure, have reworded the test!

1pretz1:
Thanks for the review @brucebolt! I've reworded the test as per your suggestion @kevindew and shifted the freeze block as it only concerns the class call.

PR #1871, opened 2020-03-06
Wrap Resync document service in sidekiq job with retries

leenagupte:
Trello: https://trello.com/c/aOl2VwhV

# What's changed and why?
The resync job frequently fails due to network timeout issues. Changing the service to a sidekiq job allows it to be retried.

### DIFF
--- a/spec/jobs/resync_document_job_spec.rb
+++ b/spec/jobs/resync_document_job_spec.rb
@@ -1,10 +1,21 @@
 RSpec.describe ResyncDocumentJob do
-  it "delegates to the ResyncDocumentService" do
-    document = build(:document)
+  include ActiveJob::TestHelper
 
+  let(:document) { create(:document) }
+
+  before { populate_default_government_bulk_data }
+
+  it "delegates to the ResyncDocumentService" do
     expect(ResyncDocumentService)
       .to receive(:call)
       .with(document)
     described_class.perform_now(document)
+  end
+
+  it "retries the job when an exception is raised" do
+    allow(ResyncDocumentService).to receive(:call).and_raise(GdsApi::BaseError)
+    described_class.perform_now(document)
+
+    expect(described_class).to have_been_enqueued
   end
 end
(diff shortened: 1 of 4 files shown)


spec/jobs/resync_document_job_spec.rb
```diff
@@ -0,0 +1,17 @@
+RSpec.describe ResyncDocumentJob do
+  include ActiveJob::TestHelper
+
+  before do
+    populate_default_government_bulk_data
+  end
```
kevindew:
```suggestion
  before { populate_default_government_bulk_data }
```

spec/jobs/resync_document_job_spec.rb
```diff
@@ -0,0 +1,17 @@
...
+
```
kevindew: I think you'll want a basic test that it delegates to the ResyncDocumentService

spec/jobs/resync_document_job_spec.rb
```diff
@@ -0,0 +1,17 @@
...
+  it "retries the job when an exception is raised" do
+    error = GdsApi::TimedOutException.new
+    stub_any_publishing_api_call.to_raise(error)
+
+    document = create(:document, :with_current_and_live_editions)
+    described_class.perform_now(document)
+
+    expect(described_class).to have_been_enqueued
```
kevindew: You can simplify this a bit and not need to create extra editions if you modify this to be something like:

```
    allow(ResyncDocumentService).to receive(:call).and_raise(GdsApi::BaseError)
    described_class.perform_now(build(:document))
    expect(described_class).to have_been_enqueued
```

leenagupte: I have to use `create` rather than `build` otherwise I get the following error: 

```
1) ResyncDocumentJob retries the job when an exception is raised
     Failure/Error: described_class.perform_now(build(:document))
     
     ActiveJob::SerializationError:
       Unable to serialize Document without an id. (Maybe you forgot to call save?)
     # ./spec/jobs/resync_document_job_spec.rb:8:in `block (2 levels) in <top (required)>'
     # ------------------
     # --- Caused by: ---
     # GdsApi::BaseError:
     #   GdsApi::BaseError
     #   ./app/jobs/resync_document_job.rb:9:in `perform'
```

leenagupte:
@kevindew Thanks for your comments. I've addressed them and fixed up.

PR #1875, opened 2020-03-10
Integrity check published times

1pretz1:
We want to integrity check publishing times when we import whitehall documents, the times we're concerned about checking are public_updated_at and first_published_at.

More info in the commits ->

Trello:
https://trello.com/c/aCQ4qhB3/1464-check-integrity-of-whitehall-firstpublishedat-and-publicupdatedat

### DIFF
--- a/lib/whitehall_importer/integrity_checker.rb
+++ b/lib/whitehall_importer/integrity_checker.rb
@@ -2,6 +2,17 @@ module WhitehallImporter
   class IntegrityChecker
     attr_reader :edition
 
+    def self.time_matches?(proposed_time, publishing_api_time)
+      return true if proposed_time == publishing_api_time
+
+      proposed_time = Time.zone.rfc3339(proposed_time)
+      publishing_api_time = Time.zone.rfc3339(publishing_api_time)
+
+      proposed_time.between?(publishing_api_time - 5, publishing_api_time + 5)
+    rescue ArgumentError
+      false
+    end
+
     def initialize(edition)
       @edition = edition
     end
--- a/lib/whitehall_importer/integrity_checker.rb
+++ b/lib/whitehall_importer/integrity_checker.rb
@@ -69,6 +69,14 @@ module WhitehallImporter
                                         first_public_at,
                                         proposed_first_published_at)
       end
+      proposed_public_updated_at = proposed_payload["public_updated_at"]
+      public_updated_at = publishing_api_content["public_updated_at"]
+
+      if edition.live? && !time_matches?(proposed_public_updated_at, public_updated_at)
+        problems << problem_description("public_updated_at doesn't match",
+                                        public_updated_at,
+                                        proposed_public_updated_at)
+      end
 
       problems
     end
--- a/lib/whitehall_importer/integrity_checker.rb
+++ b/lib/whitehall_importer/integrity_checker.rb
@@ -22,7 +22,7 @@ module WhitehallImporter
     end
 
     def problems
-      content_problems + image_problems + organisation_problems
+      content_problems + image_problems + organisation_problems + time_problems
     end
 
     def proposed_payload
@@ -56,6 +56,25 @@ module WhitehallImporter
       ).sufficiently_similar?
 
       problems
+    end
+
+    def time_problems
+      problems = []
+
+      proposed_first_published_at = proposed_payload["first_published_at"]
+      first_public_at = publishing_api_content["details"]["first_public_at"]
+
+      unless time_matches?(proposed_first_published_at, first_public_at)
+        problems << problem_description("our first_published_at doesn't match first_public_at",
+                                        first_public_at,
+                                        proposed_first_published_at)
+      end
+
+      problems
+    end
+
+    def time_matches?(proposed_time, publishing_api_time)
+      self.class.time_matches?(proposed_time, publishing_api_time)
     end
 
     def image_problems
--- a/spec/lib/whitehall_importer/integrity_checker_spec.rb
+++ b/spec/lib/whitehall_importer/integrity_checker_spec.rb
@@ -222,6 +222,26 @@ RSpec.describe WhitehallImporter::IntegrityChecker do
     end
   end
 
+  describe ".time_matches?" do
+    let(:time) { Time.zone.now }
+
+    it "returns true when times match" do
+      expect(described_class.time_matches?(time.rfc3339, time.rfc3339)).to eq true
+    end
+
+    it "returns true when times are sufficiently similar" do
+      expect(described_class.time_matches?((time + 4).rfc3339, time.rfc3339)).to eq true
+    end
+
+    it "returns false when times are not sufficiently similar" do
+      expect(described_class.time_matches?((time + 30).rfc3339, time.rfc3339)).to eq false
+    end
+
+    it "returns false when times are invalid" do
+      expect(described_class.time_matches?("Not a time", nil)).to eq false
+    end
+  end
+
   def default_publishing_api_item(edition, override_hash = {})
     {
       content_id: edition.content_id,
--- a/spec/lib/whitehall_importer/integrity_checker_spec.rb
+++ b/spec/lib/whitehall_importer/integrity_checker_spec.rb
@@ -26,7 +26,6 @@ RSpec.describe WhitehallImporter::IntegrityChecker do
         },
       )
     end
-
     let(:publishing_api_item) do
       default_publishing_api_item(edition,
                                   details: {
@@ -37,11 +36,11 @@ RSpec.describe WhitehallImporter::IntegrityChecker do
                                     organisations: edition.tags["organisations"].to_a + edition.tags["primary_publishing_organisation"].to_a,
                                   })
     end
+    let(:integrity_check) { described_class.new(edition) }
 
     it "returns true if there aren't any problems for edition without image or attachment" do
       stub_publishing_api_has_item(publishing_api_item)
 
-      integrity_check = described_class.new(edition)
       expect(integrity_check.valid?).to be true
     end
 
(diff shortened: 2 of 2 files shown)


kevindew:
Hey, just summarising our quick face-to-face review.

- We want to be comparing Whitehalls details->**first_public_at** rather than first_published_at. Some past about the magic first_public_at value is available here: https://github.com/alphagov/government-frontend/pull/996
- When the times don't match we want to be displaying the different values - which will likely mean using the problem description method
- We're not sure if we yet need to add the precision checking in, we want to give this a test run with NDA first and see if they have any issues first to guide us.

1pretz1:
Rebased to accommodate the json payload changes and now should be ready for another review @kevindew.

lib/whitehall_importer/integrity_checker.rb
```diff
@@ -2,6 +2,15 @@ module WhitehallImporter
   class IntegrityChecker
     attr_reader :edition
 
+    def self.time_similar?(proposed_time, publishing_api_time)
```
kevindew: I think we should probably call this matches rather than similar. I'm assuming you went for similar as we're not checking it's exact but I think match still conveys are meaning better than similar. 

A similar time to me sounds like there's a pretty big window for difference whereas we're right down to the 10 second mark here.

lib/whitehall_importer/integrity_checker.rb
```diff
@@ -2,6 +2,15 @@ module WhitehallImporter
...
+      return true if proposed_time == publishing_api_time
+
+      Time.zone.rfc3339(proposed_time).between?(Time.zone.rfc3339(publishing_api_time) - 5,
+                                                Time.zone.rfc3339(publishing_api_time) + 5)
```
kevindew: There's a lot to parse in a couple of lines here perhaps it should be broken up a bit

eg:

```suggestion
      proposed_time = Time.zone.rfc3339(proposed_time)
      publishing_api_time = Time.zone.rfc3339(publishing_api_time)
      proposed_time.between?(publishing_api_time - 5, publishing_api_time + 5)
```

lib/whitehall_importer/integrity_checker.rb
```diff
@@ -2,6 +2,15 @@ module WhitehallImporter
...
+    rescue ArgumentError
+      false
+    end
```
kevindew: I'm a bit confused why we're not testing this method in this commit? I notice the commit message says "Tests will be added in the next few commits." but there don't seem to be any explicit tests for this static public method.

I think they'd be quite useful as a way to document the 10 second window, which I think should also be documented in the commit message.

spec/lib/whitehall_importer/integrity_checker_spec.rb
```diff
@@ -146,6 +157,28 @@ def problem_message(attribute, expected, actual)
...
+    it "returns a problem when first_published_at times are invalid" do
+      publishing_api_item[:details][:first_public_at] = "Not a time"
+      stub_publishing_api_has_item(publishing_api_item)
+
+      expect(integrity_check.problems).to include(
+        problem_message("first_published_at",
+                        publishing_api_item[:details][:first_public_at],
+                        edition.document.first_published_at.as_json),
+      )
+    end
```
kevindew: I don't think you'll need this test here - should end up being covered by tests for the static method.

lib/whitehall_importer/integrity_checker.rb
```diff
@@ -67,6 +67,14 @@ def time_problems
                                         first_published_at,
                                         proposed_first_published_at)
       end
+      proposed_public_updated_at = proposed_payload["public_updated_at"]
+      public_updated_at = publishing_api_content["public_updated_at"]
+
+      unless edition.draft? && edition.first? || time_similar?(proposed_public_updated_at, public_updated_at)
```
kevindew: This is a pretty tough unless to understand - I personally stay clear of using unless with multiple things because the logic can be tough to parse. 

I'm also wondering if we have a logic issue here as I think it's not that something is draft that we care about but instead that it's not live (as submitted for review and scheduled would also have this issue).

Maybe try: `if edition.live? && time_similar?` ? This wouldn't check non first drafts but that's probably fine anyway as the live versions will be checked.

lib/whitehall_importer/integrity_checker.rb
1pretz1: Sure, I can see where you're coming from but I suppose it's a matter of opinion, you could also say that 'matches' means two things that are equal which isn't really true if the times differ slightly. To keep the PR moving I've adjusted the name as per your suggestion.

lib/whitehall_importer/integrity_checker.rb
1pretz1: Sure

spec/lib/whitehall_importer/integrity_checker_spec.rb
1pretz1: Removed!

lib/whitehall_importer/integrity_checker.rb
1pretz1: Good spot Kev, I've updated the logic to `if edition.live? && !time_similar?`  which is what I think you meant? 

This wouldn't check editions that are not live and would only raise a problem if the edition is live and the times are not similar.

lib/whitehall_importer/integrity_checker.rb
1pretz1: I originally didn't test this method as all the scenarios were covered from the tests of the instance of the class which meant testing it seemed OTT, I have since removed a test as per your suggestion and added a full test block for the method.

1pretz1:
Thanks for the review @kevindew, have made the changes you've requested.

lib/whitehall_importer/integrity_checker.rb
```diff
@@ -58,6 +58,25 @@ def content_problems
       problems
     end
 
+    def time_problems
+      problems = []
+
+      proposed_first_published_at = proposed_payload["first_published_at"]
+      first_published_at = publishing_api_content["details"]["first_public_at"]
```
kevindew: Can we add something in the commit message about why we are using first_public_at? It'd be good to have that decision documented.

spec/lib/whitehall_importer/integrity_checker_spec.rb
```diff
@@ -17,19 +17,19 @@
...
-        :edition,
-        document_type: document_type,
-        tags: {
-          primary_publishing_organisation: [SecureRandom.uuid],
-          organisations: [SecureRandom.uuid],
-        },
-      )
+      build(:edition,
+            document_type: document_type,
+            document: create(:document, first_published_at: "2020-03-11 12:00:00 +0000"),
```
kevindew: I notice in the commit message it says: "first_published_at time is returned from publishing-api in the format -"2020-03-11 12:00:00 +0000" and so the specs are using this same format for the pub-api edition." Which I wonder if this is why you've set these to these strings?

I'm a bit baffled by this though as these times are used to create Time objects in Content Publisher DB and have no relation with Publishing API and the Publishing API first_published_at time is formatted in ISO8601: https://github.com/alphagov/publishing-api/blob/6e34eb8c2593a3bfc38a28fc11de0989e927adcd/app/presenters/queries/content_item_presenter.rb#L119

lib/whitehall_importer/integrity_checker.rb
```diff
@@ -58,6 +58,25 @@ def content_problems
...
+
+      unless time_matches?(proposed_first_published_at, first_published_at)
+        problems << problem_description("first_published_at doesn't match",
```
kevindew: Might want to adjust to reflect that it doesn't match first_public_at eg "our first_published_at doesn't match first_public_at" ?

lib/whitehall_importer/integrity_checker.rb
1pretz1: Sure

lib/whitehall_importer/integrity_checker.rb
1pretz1: Sure

spec/lib/whitehall_importer/integrity_checker_spec.rb
1pretz1: I've set the times in publishing_api_item to be consistent with the times that are returned from the actual publishing_api which I think is the best thing to do as it resembles real life. 

I know this value sets a time object in ruby when creating a document but I was trying to be consistent by only using strings to set timestamps in this spec and the format - "2020-03-11 12:00:00 +0000" is what's returned when you call to_s on a Time object. The second reason I used a string is you can't use one of ruby / rails neat time helpers like `Time.zone.now.at_noon` as you need to pass a specific date as the one you are comparing against is a string in publishing_api_item. You could pass Time a specific date but it just seemed to be easier to just write this as a string, what do you think?

kevindew: Ok, I don't think you've quite captured what part of this I've found confusing. It's mostly where the commit message seems to differ from what actually happens / what we use and me trying to piece together whether you actually mean that or not.

The commit message says:

> "first_published_at time is returned from publishing-api in the format -"2020-03-11 12:00:00 +0000"

But we don't use first_published_at from the Publishing API, nor is it in that format. We seem to care about first_public_at which is in this format: "2020-01-24T16:38:00.000+00:00"

Passing strings seems fine, though you don't need quite that degree of granularity - "2020-03-11 12:00:00 +0000" produces the same output as "2020-03-11 12:00" since the time is in GMT. If you wanted it to be safe for all parts of the year you could do say: `Time.utc(2020, 3, 11).noon` or "2020-03-11 12:00 UTC"

1pretz1: Oh I see what you mean, I've updated the commit message accordingly, nice catch! 
I've also updated the times to use "2020-03-11 12:00 UTC" format.

PR #1882, opened 2020-03-11
Support "Delete file" action for featured file attachments

leenagupte:
Trello: https://trello.com/c/qpeNzbYa

# What's changed?

Adds a 'Delete attachment' link to the featured attachment preview block. Clicking delete takes the user to a confirmation page which then redirects them to the current `/attachments` page, where they can see that the attachment is gone.

# Expected changes
Screenshot 2020-03-11 at 17 15 16
Screenshot 2020-03-11 at 17 17 52
Screenshot 2020-03-11 at 17 18 02

### DIFF
--- a/app/controllers/file_attachments_controller.rb
+++ b/app/controllers/file_attachments_controller.rb
@@ -75,9 +75,13 @@ class FileAttachmentsController < ApplicationController
     result = FileAttachments::DestroyInteractor.call(params: params, user: current_user)
     attachment_revision = result.attachment_revision
 
-    redirect_to file_attachments_path(params[:document]),
-                notice: t("file_attachments.index.flashes.deleted",
-                          file: attachment_revision.filename)
+    if params[:wizard] == "featured"
+      redirect_to featured_attachments_path(params[:document])
+    else
+      redirect_to file_attachments_path(params[:document]),
+                  notice: t("file_attachments.index.flashes.deleted",
+                            file: attachment_revision.filename)
+    end
   end
 
   def edit
--- a/app/controllers/file_attachments_controller.rb
+++ b/app/controllers/file_attachments_controller.rb
@@ -31,6 +31,14 @@ class FileAttachmentsController < ApplicationController
       .find_by!(file_attachment_id: params[:file_attachment_id])
   end
 
+  def confirm_delete
+    @edition = Edition.find_current(document: params[:document])
+    assert_edition_state(@edition, &:editable?)
+
+    @attachment = @edition.file_attachment_revisions
+      .find_by!(file_attachment_id: params[:file_attachment_id])
+  end
+
   def preview
     result = FileAttachments::PreviewInteractor.call(params: params, user: current_user)
     can_preview, api_error = result.to_h.values_at(:can_preview, :api_error)
--- a/app/views/featured_attachments/_featured_attachment.html.erb
+++ b/app/views/featured_attachments/_featured_attachment.html.erb
@@ -1,7 +1,7 @@
 <% actions = [] %>
 <% actions << capture do %>
   <%= form_tag(
-    file_attachment_path(document, attachment.file_attachment_id),
+    file_attachment_path(document, attachment.file_attachment_id, wizard: "featured"),
     method: :delete,
     class: "app-inline-block",
     data: {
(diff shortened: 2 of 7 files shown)


app/views/featured_attachments/_featured_attachment.html.erb
```diff
@@ -1,4 +1,19 @@
+<% actions = [] %>
+<% actions << capture do %>
+  <%= form_tag(
+    file_attachment_path(document, attachment.file_attachment_id),
+    method: :delete,
+    class: "app-inline-block",
+    data: {
+      "modal-action": "delete",
```
benthorner: Minor: this is only applicable when this is used in a modal context. We can lose it here. It looks like it goes in a subsequent commit anyway.

app/controllers/file_attachments_controller.rb
```diff
@@ -75,7 +75,7 @@ def destroy
     result = FileAttachments::DestroyInteractor.call(params: params, user: current_user)
     attachment_revision = result.attachment_revision
 
-    redirect_to file_attachments_path(params[:document]),
+    redirect_to params[:wizard] == "featured" ? featured_attachments_path(params[:document]) : file_attachments_path(params[:document]),
                 notice: t("file_attachments.index.flashes.deleted",
```
benthorner: We should keep translations specific to the screen on which they're shown. So in this case, the translation should be `attachments.index.flashes.deleted`, or similar. Could we use a separate one here, so they can vary independently? In this case, we still haven't confirmed which flashes for attachments, so we can actually leave it out entirely until that design work is completed.

leenagupte: I'll split up the logic, so that the flash message is only shown for inline attachments.

leenagupte:
@benthorner Thanks for the comments. I've made the suggested changes. Do you want to re-review?

spec/features/editing_file_attachments/delete_file_attachment_spec.rb
```diff
@@ -15,23 +25,45 @@ def given_there_is_an_edition_with_attachments
                       file_attachment_revisions: [@attachment_revision])
   end
 
+  def given_there_is_an_edition_with_featured_attachments
+    @attachment_revision = create(:file_attachment_revision, :on_asset_manager)
```
benthorner: This shouldn't be necessary for this feature. I notice it's in the original scenario too - it should be safe to remove.

spec/features/editing_file_attachments/delete_file_attachment_spec.rb
```diff
@@ -15,23 +25,45 @@ def given_there_is_an_edition_with_attachments
...
+
+    @edition = create(:edition,
+                      document_type: build(:document_type, :with_body, attachments: "featured"),
```
benthorner: We shouldn't need the `with_body` trait here, as it only applies for scenarios involving a body field.

spec/features/editing_file_attachments/delete_file_attachment_spec.rb
leenagupte: Which bit?

spec/features/editing_file_attachments/delete_file_attachment_spec.rb
leenagupte: I followed the set up from `given_there_is_an_edition_with_attachments`. Does that mean that `:with_body` isn't needed there either?

spec/features/editing_file_attachments/delete_file_attachment_spec.rb
benthorner: Ah, sorry: `on_asset_manager`. It doesn't change the outcome from the feature's perspective - it doesn't have any Asset Manager stubs.

spec/features/editing_file_attachments/delete_file_attachment_spec.rb
benthorner: It is in that case: it's needed for scenarios involving a body field.

leenagupte: I'm not sure I understand. Where is the body field being referenced in the "inline" tests?

leenagupte:
@benthorner The tests have been updated and fixed up.

PR #1883, opened 2020-03-12
Integrity check for withdrawn and removed content

emmabeynon:
For https://trello.com/c/e8zIJIoy/1526-expand-integrity-checks-to-cover-withdrawing-and-removing

This adds the below checks for withdrawn and removed content to the Integrity Checker:

### Withdrawn content
- Unpublishing type in Publishing API is `withdrawal`
- `withdrawn_at` matches `unpublished_at` in Publishing API
- `public_explanation` matches `explanation` in Publishing API

### Removed content
- Unpublishing type in Publishing API is `gone` or `redirect`
- `public_explanation` matches `explanation` in Publishing API
- `alternative_url` matches `alternative_path` in Publishing API
- Unpublishing time is not checked

### DIFF
--- a/lib/whitehall_importer/integrity_checker.rb
+++ b/lib/whitehall_importer/integrity_checker.rb
@@ -141,6 +141,14 @@ module WhitehallImporter
         )
       end
 
+      unless check.expected_unpublishing_time?
+        problems << problem_description(
+          "unpublishing time doesn't match",
+          publishing_api_content.dig("unpublishing", "unpublished_at"),
+          check.expected_unpublishing_time,
+        )
+      end
+
       problems
     end
 
--- a/lib/whitehall_importer/integrity_checker.rb
+++ b/lib/whitehall_importer/integrity_checker.rb
@@ -141,6 +141,14 @@ module WhitehallImporter
         )
       end
 
+      unless check.expected_alternative_path?
+        problems << problem_description(
+          "unpublishing alternative path doesn't match",
+          publishing_api_content.dig("unpublishing", "alternative_path"),
+          check.expected_alternative_path,
+        )
+      end
+
       unless check.expected_unpublishing_time?
         problems << problem_description(
           "unpublishing time doesn't match",
--- a/lib/whitehall_importer/integrity_checker.rb
+++ b/lib/whitehall_importer/integrity_checker.rb
@@ -149,6 +149,10 @@ module WhitehallImporter
         )
       end
 
+      unless check.expected_explanation?
+        problems << "unpublishing explanation doesn't match"
+      end
+
       problems
     end
 
--- a/lib/whitehall_importer/integrity_checker.rb
+++ b/lib/whitehall_importer/integrity_checker.rb
@@ -130,6 +130,15 @@ module WhitehallImporter
       problems = []
       return problems unless edition.withdrawn? || edition.removed?
 
+      unless expected_state?
+        problems << problem_description(
+          "edition state isn't as expected",
+          "unpublished",
+          publishing_api_content.dig("publication_state"),
+        )
+        return problems
+      end
+
       check = UnpublishingCheck.new(edition,
                                     publishing_api_content["unpublishing"])
 
@@ -211,6 +220,11 @@ module WhitehallImporter
       attribute == "alt_text" &&
         proposed_image_payload.empty? &&
         publishing_api_image[attribute] == "placeholder"
+    end
+
+    def expected_state?
+      %w[withdrawn removed].include?(edition.state) &&
+        publishing_api_content.dig("publication_state") == "unpublished"
     end
   end
 end
--- a/lib/whitehall_importer/integrity_checker.rb
+++ b/lib/whitehall_importer/integrity_checker.rb
@@ -139,21 +139,20 @@ module WhitehallImporter
         return problems
       end
 
-      check = UnpublishingCheck.new(edition,
-                                    publishing_api_content["unpublishing"])
+      check = UnpublishingCheck.new(edition, publishing_api_unpublishing)
 
       unless check.expected_type?
         problems << problem_description(
           "unpublishing type not expected",
           check.expected_type,
-          publishing_api_content.dig("unpublishing", "type"),
+          publishing_api_unpublishing["type"],
         )
       end
 
       unless check.expected_alternative_path?
         problems << problem_description(
           "unpublishing alternative path doesn't match",
-          publishing_api_content.dig("unpublishing", "alternative_path"),
+          publishing_api_unpublishing["alternative_path"],
           check.expected_alternative_path,
         )
       end
@@ -161,7 +160,7 @@ module WhitehallImporter
       unless check.expected_unpublishing_time?
         problems << problem_description(
           "unpublishing time doesn't match",
-          publishing_api_content.dig("unpublishing", "unpublished_at"),
+          publishing_api_unpublishing["unpublished_at"],
           check.expected_unpublishing_time,
         )
       end
@@ -225,6 +224,10 @@ module WhitehallImporter
     def expected_state?
       %w[withdrawn removed].include?(edition.state) &&
         publishing_api_content.dig("publication_state") == "unpublished"
+    end
+
+    def publishing_api_unpublishing
+      publishing_api_content["unpublishing"]
     end
   end
 end
--- a/lib/whitehall_importer/integrity_checker.rb
+++ b/lib/whitehall_importer/integrity_checker.rb
@@ -22,7 +22,8 @@ module WhitehallImporter
     end
 
     def problems
-      content_problems + image_problems + organisation_problems + time_problems
+      content_problems + image_problems + organisation_problems +
+        time_problems + unpublishing_problems
     end
 
     def proposed_payload
@@ -119,6 +120,24 @@ module WhitehallImporter
           "organisations don't match",
           publishing_api_link("organisations"),
           proposed_payload.dig("links", "organisations"),
+        )
+      end
+
+      problems
+    end
+
+    def unpublishing_problems
+      problems = []
+      return problems unless edition.withdrawn? || edition.removed?
+
+      check = UnpublishingCheck.new(edition,
+                                    publishing_api_content["unpublishing"])
+
+      unless check.expected_type?
+        problems << problem_description(
+          "unpublishing type not expected",
+          check.expected_type,
+          publishing_api_content.dig("unpublishing", "type"),
         )
       end
 
--- a/lib/whitehall_importer/integrity_checker/unpublishing_check.rb
+++ b/lib/whitehall_importer/integrity_checker/unpublishing_check.rb
@@ -23,6 +23,10 @@ module WhitehallImporter
       unpublishing_time_matches?
     end
 
+    def expected_unpublishing_time
+      edition.status.details&.withdrawn_at
+    end
+
     def expected_alternative_path?
       return true unless edition.removed?
 
@@ -37,10 +41,8 @@ module WhitehallImporter
   private
 
     def unpublishing_time_matches?
-      IntegrityChecker.time_matches?(
-        unpublishing["unpublished_at"],
-        edition.status.details.withdrawn_at&.rfc3339,
-      )
+      IntegrityChecker.time_matches?(unpublishing["unpublished_at"],
+                                     expected_unpublishing_time&.rfc3339)
     end
 
     def edition_explanation_html
--- a/lib/whitehall_importer/integrity_checker/unpublishing_check.rb
+++ b/lib/whitehall_importer/integrity_checker/unpublishing_check.rb
@@ -1,0 +1,23 @@
+module WhitehallImporter
+  class IntegrityChecker::UnpublishingCheck
+    attr_reader :edition, :unpublishing
+
+    def initialize(edition, unpublishing)
+      @edition = edition
+      @unpublishing = unpublishing
+    end
+
+    def expected_type?
+      case unpublishing["type"]
+      when "gone"
+        edition.removed? && !edition.status.details.redirect?
+      when "redirect"
+        edition.removed? && edition.status.details.redirect?
+      when "withdrawal"
+        edition.withdrawn?
+      else
+        false
+      end
+    end
+  end
+end
--- a/lib/whitehall_importer/integrity_checker/unpublishing_check.rb
+++ b/lib/whitehall_importer/integrity_checker/unpublishing_check.rb
@@ -32,6 +32,11 @@ module WhitehallImporter
       unpublishing["alternative_path"] == edition.status.details.alternative_url
     end
 
+    def expected_explanation?
+      Sanitize.clean(unpublishing["explanation"]).squish ==
+        Sanitize.clean(edition_explanation_html).squish
+    end
+
   private
 
     def unpublishing_time_matches?
@@ -39,6 +44,13 @@ module WhitehallImporter
         unpublishing["unpublished_at"],
         edition.status.details.withdrawn_at&.rfc3339,
       )
+    end
+
+    def edition_explanation_html
+      explanation = edition.status.details.try(:public_explanation) ||
+        edition.status.details.try(:explanatory_note)
+
+      Govspeak::Document.new(explanation).to_html
     end
   end
 end
--- a/lib/whitehall_importer/integrity_checker/unpublishing_check.rb
+++ b/lib/whitehall_importer/integrity_checker/unpublishing_check.rb
@@ -19,5 +19,20 @@ module WhitehallImporter
         false
       end
     end
+
+    def expected_unpublishing_time?
+      return true unless edition.withdrawn?
+
+      unpublishing_time_matches?
+    end
+
+  private
+
+    def unpublishing_time_matches?
+      IntegrityChecker.time_matches?(
+        unpublishing["unpublished_at"],
+        edition.status.details.withdrawn_at&.rfc3339,
+      )
+    end
   end
 end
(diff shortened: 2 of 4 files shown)


lib/whitehall_importer/integrity_checker.rb
```diff
@@ -136,5 +191,43 @@ def default_image?(proposed_image_payload, publishing_api_image, attribute)
...
+    def removed_type_matches?
+      publishing_api_type = publishing_api_content.dig(:unpublishing, :type)
+      %w[gone redirect].include?(publishing_api_type) && edition.state == "removed"
+    end
+
+    def unpublishing_time_matches?
+      publishing_api_unpublished_at = publishing_api_content.dig(:unpublishing,
+                                                                 :unpublished_at)
+      Time.zone.iso8601(publishing_api_unpublished_at) ==
+        edition.status.details.withdrawn_at
```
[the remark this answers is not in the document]
emmabeynon: So I've found an issue with converting the Publishing API timestamp to rfc3339 here because the format seems to be different for `unpublished_at` compared with other timestamps like `public_updated_at`.  After testing on integration, `unpublished_at` is coming back in this format: `"2020-03-02T11:26:52"` whereas `public_updated_at` is coming back like this: `"2020-03-11T12:00:00Z"`.  The consequence of this is when trying to parse it e.g. `Time.zone.rfc3339("2020-03-02T11:26:52")` we get an argument error (this was tried on integration).  I'm going leave it to parse to `iso8601` for now and we may need to rethink the comparison method a bit.

lib/whitehall_importer/integrity_checker.rb
```diff
@@ -85,10 +86,64 @@ def organisation_problems
...
     end
 
+    def unpublishing_problems
+      problems = []
+      return problems unless edition.withdrawn? || edition.removed?
+
+      if edition.withdrawn? && !withdrawal_type_matches?
+        problems << unpublishing_type_problem(
+          edition.state,
+          publishing_api_content.dig(:unpublishing, :type),
```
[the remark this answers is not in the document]
emmabeynon: Done 👍

emmabeynon:
@kevindew Should be ready to review again 👍

lib/whitehall_importer/integrity_checker.rb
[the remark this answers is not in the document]
kevindew: oh shoot - it's coming back with no timezone information? that might mean we'll get the time the time wrong for the half year in BST. I'm just going to have a look at Pub API for this.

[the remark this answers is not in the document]
kevindew: I've just opened this https://github.com/alphagov/publishing-api/pull/1746

lib/whitehall_importer/integrity_checker.rb
```diff
@@ -87,10 +88,72 @@ def organisation_problems
...
+          publishing_api_unpublishing["alternative_path"],
+          edition.status.details.alternative_url,
+        )
+      end
+
+      unless check.expected_unpublishing_time?
+        problems << problem_description(
+          "unpublishing time doesn't match",
+          publishing_api_unpublishing["unpublished_at"],
+          edition.status.details.withdrawn_at,
```
kevindew: Maybe we want to ask the checker for expected_unpublishing_time it feels a bit opaque why we've used withdrawn_at here

lib/whitehall_importer/integrity_checker.rb
```diff
@@ -87,10 +88,72 @@ def organisation_problems
...
+        return problems
+      end
+
+      check = UnpublishingCheck.new(edition, publishing_api_unpublishing)
+
+      unless check.expected_type?
+        problems << unpublishing_type_problem(
+          edition.state,
+          publishing_api_unpublishing["type"],
+        )
```
kevindew: I wonder if this could be a problem description e.g: 

```
        problems << problem_description(
          "unpublshing type doesn't match", 
          publishing_api_unpublishing["type"],
          check.expected_unpublishing_type,
        )
```
?

lib/whitehall_importer/integrity_checker/unpublishing_check.rb
```diff
@@ -0,0 +1,59 @@
+module WhitehallImporter
+  class IntegrityChecker::UnpublishingCheck
+    attr_reader :edition, :publishing_api_unpublishing
+
+    def initialize(edition, publishing_api_unpublishing)
+      @edition = edition
+      @publishing_api_unpublishing = publishing_api_unpublishing
+    end
```
kevindew: Minor - since we don't have a concept of unpublishing in content publisher you might be able to shorten this to just unpublishing throughout the class.

lib/whitehall_importer/integrity_checker/unpublishing_check.rb
```diff
@@ -0,0 +1,59 @@
...
+
+  private
+
+    def withdrawal_type_matches?
+      publishing_api_unpublishing["type"] == "withdrawal" && edition.withdrawn?
+    end
+
+    def removed_type_matches?
+      publishing_api_type = publishing_api_unpublishing["type"]
+      %w[gone redirect].include?(publishing_api_type) && edition.removed?
```
kevindew: Should we be checking that redirect and gone match too? maybe like:

```
case publishing_api_unpublishing["type"]
when "gone"
  edition.removed? && !edition.status.details.redirect?
when "redirect"
  edition.removed? && edition.status.details.redirect?
else
  false
end
```

You could also put withdrawal in too and have it covered by a single method

spec/lib/whitehall_importer/integrity_checker/unpublishing_check_spec.rb
```diff
@@ -0,0 +1,160 @@
...
+    }
+  end
+
+  let(:publishing_api_redirect) do
+    {
+      "type" => "redirect",
+      "alternative_path" => "/somewhere",
+      "explanation" => "#{removed_explanation} <a href=\"https://www.gov.uk/here\">Visit here</a>",
+    }
+  end
```
kevindew: I feel like we might be setting up too many variables here that serve almost all test cases - they can often seem to form a smell when one of the tests may need to reach outside them. 

I wonder if it might be nicer doing more of the initialisation of variables inside tests/their relevant context blocks and maybe setting up functions for `publishing_api_redirect` etc.

For example:

```
def publishing_api_withdrawal(attributes = {})
  {
      type: withdrawal,
      unpublished_at: Time.zone.now.rfc3339,
      explanation: "Explanation",
  }.merge(attributes).as_json 
end
```

then in the first test you could probably get away with:

```
      it "returns true if it has a 'withdrawal' unpublishing type in Publishing API" do
        unpublishing_check = described_class.new(build(:edition, :withdrawn),
                                                 publishing_api_withdrawal)
        expect(unpublishing_check.expected_type?).to be true
      end
```

I think this suggestion may end up increasing number of lines a bit but I think the result will be simpler to follow and the tests more focused.

lib/whitehall_importer/integrity_checker.rb
```diff
@@ -87,10 +88,72 @@ def organisation_problems
...
+          publishing_api_unpublishing["unpublished_at"],
+          edition.status.details.withdrawn_at,
+        )
+      end
+
+      unless check.expected_explanation?
+        problems << problem_description(
+          "unpublishing explanation doesn't match",
+          publishing_api_unpublishing["explanation"],
+          edition_unpublishing_explanation,
```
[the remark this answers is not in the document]
emmabeynon: Sure, I was just pondering whether we'd show the HTML versions of both, or a stripped version.  May just be easier not to show them.

lib/whitehall_importer/integrity_checker/unpublishing_check.rb
emmabeynon: I'm not sure that it matters that much, we don't distinguish between 'gone' and 'redirect' in Content Publusher, and 'gone' items can have a redirect in Publishing API?

lib/whitehall_importer/integrity_checker/unpublishing_check.rb
emmabeynon: 👍

lib/whitehall_importer/integrity_checker/unpublishing_check.rb
kevindew: We do distinguish in CP with the redirect on the Removal object.

In Publishing API a gone with an alternative path isn't a redirect it'll be a gone with a link shown on the page. For example: https://www.gov.uk/government/statistics/higher-education-provider-data-finance-2018-to-2019

lib/whitehall_importer/integrity_checker/unpublishing_check.rb
emmabeynon: Done

spec/lib/whitehall_importer/integrity_checker/unpublishing_check_spec.rb
emmabeynon: I've moved the Publishing API payloads to their own methods.

lib/whitehall_importer/integrity_checker.rb
emmabeynon: I've added `#expected_unpublishing_time` to the Unpublishing Checker.

lib/whitehall_importer/integrity_checker.rb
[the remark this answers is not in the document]
emmabeynon: I've stopped passing the explanations to the problem message.

lib/whitehall_importer/integrity_checker/unpublishing_check.rb
emmabeynon: I'm now checking for gone and redirect separately.

lib/whitehall_importer/integrity_checker.rb
emmabeynon: Changed this to use `problem_description`.

lib/whitehall_importer/integrity_checker.rb
```diff
@@ -87,10 +88,72 @@ def organisation_problems
       problems
     end
 
+    def unpublishing_problems
+      problems = []
+      return problems unless edition.withdrawn? || edition.removed?
```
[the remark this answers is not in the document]
emmabeynon: This feels like a bigger check to ensure all states match what's in Publishing API.  I'd like to keep this in place to avoid generating a problem for all editions imported that aren't unpublished, and do a separate piece of work to check for states in general.

emmabeynon:
@kevindew I *think* all your comments have been addressed now.  I'll make the changes to the `unpublished_at` based on your Publishing API and do a test of this on integration, so there may be a temporary WIP commit while I'm doing that.

lib/whitehall_importer/integrity_checker.rb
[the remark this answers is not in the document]
kevindew: Ok - will there be a follow up story for this or separate PR in this story?

lib/whitehall_importer/integrity_checker.rb
```diff
@@ -87,6 +88,52 @@ def organisation_problems
...
+
+      unless expected_state?
+        problems << problem_description(
+          "edition state isn't expected",
```
kevindew: Is this missing an "as" ?

lib/whitehall_importer/integrity_checker.rb
```diff
@@ -87,6 +88,52 @@ def organisation_problems
...
+          publishing_api_content.dig("unpublishing", "type"),
+          check.expected_type,
+        )
+      end
+
+      unless check.expected_alternative_path?
+        problems << problem_description(
+          "unpublishing alternative path doesn't match",
+          publishing_api_unpublishing["alternative_path"],
+          edition.status.details.alternative_url,
```
kevindew: Might be good to have this as a method on check too for consistency with the other unpublishing checks

lib/whitehall_importer/integrity_checker.rb
```diff
@@ -135,5 +182,19 @@ def default_image?(proposed_image_payload, publishing_api_image, attribute)
         proposed_image_payload.empty? &&
         publishing_api_image[attribute] == "placeholder"
     end
+
+    def edition_unpublishing_explanation
+      edition.status.details.try(:public_explanation) ||
+        edition.status.details.try(:explanatory_note)
+    end
```
kevindew: I don't think anything uses this method anymore.

lib/whitehall_importer/integrity_checker/unpublishing_check.rb
```diff
@@ -0,0 +1,56 @@
...
+      @unpublishing = unpublishing
+    end
+
+    def expected_type?
+      unpublishing["type"] == expected_type
+    end
+
+    def expected_type
+      return "withdrawal" if edition.withdrawn?
+      return "invalid edition state" unless edition.removed?
```
kevindew: Should this be a raise rather than return? I assume we don't expect this class to be called with a non withdrawn or removed edition?

lib/whitehall_importer/integrity_checker/unpublishing_check.rb
```diff
@@ -0,0 +1,56 @@
...
+    def expected_explanation?
+      Sanitize.clean(unpublishing["explanation"]).squish ==
+        Sanitize.clean(edition_explanation_html).squish
+    end
+
+  private
+
+    def unpublishing_time_matches?
+      unpublished_at = unpublishing["unpublished_at"]
+      Time.zone.iso8601(unpublished_at) == edition.status.details.withdrawn_at
```
kevindew: I've deployed the change to Publishing API now so this should be fine for rfc3339. I think we should aim to use the method @1pretz1 has, not sure how we sequence this though as you're both super close to finishing these PRs.

spec/lib/whitehall_importer/integrity_checker_spec.rb
```diff
@@ -149,6 +232,12 @@ def problem_message(attribute, expected, actual)
       "#{attribute} doesn't match, expected: #{expected.inspect}, actual: #{actual.inspect}"
     end
 
+    def unpublishing_problem_message(publishing_api_type, expected)
+      "unpublishing type not expected, " +
+        "expected: #{publishing_api_type.inspect}, " +
+        "actual: #{expected.inspect}"
+    end
```
kevindew: We don't need this anymore.

lib/whitehall_importer/integrity_checker.rb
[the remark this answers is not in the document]
emmabeynon: I've now changed this to convert to `rfc3339`.  It's working as expected on integration.

lib/whitehall_importer/integrity_checker.rb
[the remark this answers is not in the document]
emmabeynon: I'll write up a story for it.

lib/whitehall_importer/integrity_checker/unpublishing_check.rb
emmabeynon: I don't expect it to be called with an edition that is not unpublished, possibly overly defensive code.  I can just delete it.

spec/lib/whitehall_importer/integrity_checker_spec.rb
emmabeynon: I will need to adapt the `problem_message` method in order to get rid of this.

emmabeynon: Sorted here: https://github.com/alphagov/content-publisher/pull/1883/commits/1390eaf1624423299dbccf0f9461f3f40033f261

lib/whitehall_importer/integrity_checker.rb
emmabeynon: Added in

lib/whitehall_importer/integrity_checker.rb
emmabeynon: I'd added a `#expected_alternative_path` method to the Unpublishing Check class.

lib/whitehall_importer/integrity_checker.rb
emmabeynon: Good catch - removed

lib/whitehall_importer/integrity_checker/unpublishing_check.rb
emmabeynon: Deleted

lib/whitehall_importer/integrity_checker/unpublishing_check.rb
emmabeynon: He got there before me! 😄

spec/lib/whitehall_importer/integrity_checker_spec.rb
```diff
@@ -172,7 +275,7 @@
     end
     let(:integrity_check) { described_class.new(edition) }
 
-    def problem_message(message, expected, actual)
+    def problem_description(message, expected, actual)
```
kevindew: 👍 for this renaming.

lib/whitehall_importer/integrity_checker/unpublishing_check.rb
emmabeynon: Went on a wild rebase journey with this one!  Will implement the time matching method next.

lib/whitehall_importer/integrity_checker/unpublishing_check.rb
```diff
@@ -0,0 +1,59 @@
...
+    def expected_explanation?
+      Sanitize.clean(unpublishing["explanation"]).squish ==
+        Sanitize.clean(edition_explanation_html).squish
+    end
+
+  private
+
+    def unpublishing_time_matches?
+      unpublished_at = Time.zone.rfc3339(unpublishing["unpublished_at"])
+      unpublished_at == edition.status.details.withdrawn_at
```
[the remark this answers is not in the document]
emmabeynon: All done 👍

PR #1884, opened 2020-03-12
Format times for PublishingApiPayload to rfc3339

1pretz1:
Converts time objects to rfc3339 strings for the generated payload's
timestamps. The reason we've decided to do this is that when we come to
compare times it helps if we're dealing with the same objects, e.g all
strings. An example of where we are comparing times is the
IntegrityChecker where we fetch the edition from the publishing-api and the
timestamps that are returned are strings where as the payload generates
Time objects.

Doing this will make comparing these two times in the IntegrityChecker
simpler for future pieces of work.

### DIFF
--- a/lib/publishing_api_payload.rb
+++ b/lib/publishing_api_payload.rb
@@ -22,16 +22,16 @@ class PublishingApiPayload
       links: links,
       access_limited: access_limited,
       auth_bypass_ids: auth_bypass_ids,
-      public_updated_at: history.public_updated_at,
+      public_updated_at: history.public_updated_at.rfc3339,
     }
-    payload[:first_published_at] = history.first_published_at if history.first_published_at.present?
+    payload[:first_published_at] = history.first_published_at.rfc3339 if history.first_published_at.present?
 
     fields = document_type.contents + document_type.tags
     fields.each { |f| payload.deep_merge!(f.payload(edition)) }
 
     if edition.backdated_to.present?
-      payload[:first_published_at] = edition.backdated_to
-      payload[:public_updated_at] = edition.backdated_to if edition.first?
+      payload[:first_published_at] = edition.backdated_to.rfc3339
+      payload[:public_updated_at] = edition.backdated_to.rfc3339 if edition.first?
     end
 
     if republish
--- a/lib/publishing_api_payload.rb
+++ b/lib/publishing_api_payload.rb
@@ -30,7 +30,6 @@ class PublishingApiPayload
     fields.each { |f| payload.deep_merge!(f.payload(edition)) }
 
     if edition.backdated_to.present?
-      payload[:first_published_at] = edition.backdated_to.rfc3339
       payload[:public_updated_at] = edition.backdated_to.rfc3339 if edition.first?
     end
 
--- a/spec/lib/publishing_api_payload_spec.rb
+++ b/spec/lib/publishing_api_payload_spec.rb
@@ -157,7 +157,7 @@ RSpec.describe PublishingApiPayload do
     end
 
     it "includes first_published_at if the edition has a backdated_to value" do
-      date = Time.zone.now.yesterday
+      date = Time.zone.now.yesterday.rfc3339
       revision = build(:revision, backdated_to: date)
       edition = build(:edition, revision: revision)
       payload = described_class.new(edition).payload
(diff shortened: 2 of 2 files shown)


kevindew:
Thanks for raising this @1pretz1. I've been having a ponder about this and I wonder if rather than having these returned as times it might be better to have an as_json call done in the integrity checker.

I'm just going to quickly explore that a bit and will raise a PR if it seems to work. Many apologies if I've sent you off doing work that we don't ultimately use.

gclssvglx:
Just a suggestion...

Instead of littering our code with `.rfc3339` - we could add `Time::DATE_FORMATS[:default]="%Y-%m-%dT%H:%M:%S%:z"` in an initializer.  Then call `to_s` where appropriate.  That would give us something like this...

```
irb(main):001:0> Time::DATE_FORMATS[:default]="%Y-%m-%dT%H:%M:%S%:z"
=> "%Y-%m-%dT%H:%M:%S%:z"
irb(main):002:0> Time.now.rfc3339
=> "2020-03-12T12:21:12+00:00"
irb(main):003:0> Time.now.to_s
=> "2020-03-12T12:21:19+00:00"
irb(main):004:0> Time.now     
=> 2020-03-12 12:21:24 +0000
``` 

So, we keep the Ruby default, but override the `to_s`.  I get it that this is not as clear as the explicit `.rfc3339`, but if we really want to default our times - useful.

kevindew:
> Just a suggestion...
> 
> Instead of littering our code with `.rfc3339` - we could add `Time::DATE_FORMATS[:default]="%Y-%m-%dT%H:%M:%S%:z"` in an initializer. Then call `to_s` where appropriate. That would give us something like this...
> 
> ```
> irb(main):001:0> Time::DATE_FORMATS[:default]="%Y-%m-%dT%H:%M:%S%:z"
> => "%Y-%m-%dT%H:%M:%S%:z"
> irb(main):002:0> Time.now.rfc3339
> => "2020-03-12T12:21:12+00:00"
> irb(main):003:0> Time.now.to_s
> => "2020-03-12T12:21:19+00:00"
> irb(main):004:0> Time.now     
> => 2020-03-12 12:21:24 +0000
> ```
> 
> So, we keep the Ruby default, but override the `to_s`. I get it that this is not as clear as the explicit `.rfc3339`, but if we really want to default our times - useful.

That's a really good idea 👍 

I'm not sure if this will be quite the right place for this, and the work we're doing may remove the need for it. However I definitely like the Time configuration of formats a lot and think we could benefit from this in the UI of our app where we've got these [helper functions](https://github.com/alphagov/content-publisher/blob/master/app/helpers/date_time_helper.rb#L2-L16) which might be nicer to be set as time formatter rules. Not sure I'd dare change the default though 😄

gclssvglx:
> That's a really good idea 👍
> 
> I'm not sure if this will be quite the right place for this, and the work we're doing may remove the need for it. However I definitely like the Time configuration of formats a lot and think we could benefit from this in the UI of our app where we've got these [helper functions](https://github.com/alphagov/content-publisher/blob/master/app/helpers/date_time_helper.rb#L2-L16) which might be nicer to be set as time formatter rules. Not sure I'd dare change the default though 😄

Thanks!  I totally get it that overriding the default is a *big* step - and perhaps too far!  We could provide our own - named - formats for different use cases.  For example...

```
Time::DATE_FORMATS[:api]="%Y-%m-%dT%H:%M:%S%:z"
Time::DATE_FORMATS[:ui]="%A %d %B %Y %H:%M:%S"

Time.now.to_s(:api) => "2020-03-12T13:42:40+00:00"
Time.now.to_s(:ui)   => "Thursday 12 March 2020 13:42:40"
```

kevindew:
> > That's a really good idea 👍
> > I'm not sure if this will be quite the right place for this, and the work we're doing may remove the need for it. However I definitely like the Time configuration of formats a lot and think we could benefit from this in the UI of our app where we've got these [helper functions](https://github.com/alphagov/content-publisher/blob/master/app/helpers/date_time_helper.rb#L2-L16) which might be nicer to be set as time formatter rules. Not sure I'd dare change the default though 😄
> 
> Thanks! I totally get it that overriding the default is a **big** step - and perhaps too far! We could provide our own - named - formats for different use cases if required though. For example...
> 
> ```
> Time::DATE_FORMATS[:api]="%Y-%m-%dT%H:%M:%S%:z"
> Time::DATE_FORMATS[:ui]="%A %d %B %Y %H:%M:%S"
> 
> Time.now.to_s(:api) => "2020-03-12T13:42:40+00:00"
> Time.now.to_s(:ui)  => "Thursday 12 March 2020 13:42:40"
> ```

Yup I like the idea of using that rather than the helper methods - well so long as we can come up with good names, that's always the hard parts.

kevindew:
This has now been superseded, so I'm closing it.

PR #1886, opened 2020-03-12
Remove backdating conditional for PublishingApiPayload

kevindew:
This logic is now covered by the PublishingApiPayload::History class.

### DIFF
--- a/lib/publishing_api_payload.rb
+++ b/lib/publishing_api_payload.rb
@@ -29,11 +29,6 @@ class PublishingApiPayload
     fields = document_type.contents + document_type.tags
     fields.each { |f| payload.deep_merge!(f.payload(edition)) }
 
-    if edition.backdated_to.present?
-      payload[:first_published_at] = edition.backdated_to
-      payload[:public_updated_at] = edition.backdated_to if edition.first?
-    end
-
     if republish
       payload[:update_type] = "republish"
       payload[:bulk_publishing] = true
--- a/spec/lib/publishing_api_payload_spec.rb
+++ b/spec/lib/publishing_api_payload_spec.rb
@@ -156,24 +156,6 @@ RSpec.describe PublishingApiPayload do
       expect(payload[:links][:government]).to eq [government.content_id]
     end
 
-    it "includes first_published_at if the edition has a backdated_to value" do
-      date = Time.zone.now.yesterday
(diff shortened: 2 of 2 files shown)


lib/publishing_api_payload.rb
```diff
@@ -29,11 +29,6 @@ def payload
     fields = document_type.contents + document_type.tags
     fields.each { |f| payload.deep_merge!(f.payload(edition)) }
 
-    if edition.backdated_to.present?
-      payload[:first_published_at] = edition.backdated_to
-      payload[:public_updated_at] = edition.backdated_to if edition.first?
```
1pretz1: Are we missing setting the public_updated_at time for a backdated piece of content by removing this line? 

I understand why line 33 is superfluous as it's covered by the logic in the History class but as it stands the public_updated_at for the first backdated edition will be the current time on publishing instead of whatever time the backdate was set to if I understand correctly?

kevindew: This is covered by the History class too. When we call https://github.com/alphagov/content-publisher/blob/9be2a23369c0a8885006b6104e45037333be0f50/lib/publishing_api_payload/history.rb#L8-L10
this gets the first_published_at for first edition of a document and this will be the backdated value.

1pretz1: Sorry if I'm being blind but can you point me to the code where first_published_at is being set when an edition is backdated? From what I can see this isn't set until publishing the edition, and publishing the first edition sets public_updated_at to the current time.

kevindew: Don't worry - I may have missed something.

So this public_updated_at calls change_history which returns this:

https://github.com/alphagov/content-publisher/blob/4917b67d1ffab533cc248554fb8ca3c6f3ebd9dd/lib/publishing_api_payload/history.rb#L23-L24

And that first_published_at is set by: 

https://github.com/alphagov/content-publisher/blob/4917b67d1ffab533cc248554fb8ca3c6f3ebd9dd/lib/publishing_api_payload/history.rb#L12-L16

where the backdated_at comes into play.

1pretz1: Oh yes of course, thanks for clarifying! For some reason I was thinking that first_published_at was being fetched from the DB rather than it being a method call, durrrr.

PR #1890, opened 2020-03-12
Only do realistic request errors when we need it

benthorner:
https://trello.com/c/OWZorYWA/1540-allow-attachments-to-be-downloaded

Previously we switched to 'show_exceptions' for all requests in our
tests, which means we return a realistic response instead of raising
an exception. Writing new code with this setting is painful because
it's necessary to debug/inspect the response body and manually look
for an exception/backtrace. This changes 'show_exceptions' back to
false and uses a common 'tag' pattern to enable it where needed.

https://github.com/eliotsykes/rails-testing-toolbox/blob/master/error_responses.rb

This also has the nice side-effect of improving our test performance,
locally and in Docker. I see a consistent improvement in times for
request tests of about 10s.

With: rspec spec/requests  22.40s user 3.16s system 86% cpu 29.413 total

Without: rspec spec/requests  31.52s user 3.26s system 90% cpu 38.490 total

### DIFF
--- a/config/environments/test.rb
+++ b/config/environments/test.rb
@@ -25,7 +25,7 @@ Rails.application.configure do
   config.cache_store = :null_store
 
   # Handle exceptions ourselves and return HTTP status instead of raising exceptions.
-  config.action_dispatch.show_exceptions = true
+  config.action_dispatch.show_exceptions = false
 
   # Disable request forgery protection in test environment.
   config.action_controller.allow_forgery_protection = false
--- a/spec/features/formats/news_article_spec.rb
+++ b/spec/features/formats/news_article_spec.rb
@@ -19,6 +19,8 @@ RSpec.describe "News article format" do
   def and_i_fill_in_the_form_fields
     base_path = Edition.last.document_type.path_prefix + "/a-great-title"
     stub_publishing_api_has_lookups(base_path => Document.last.content_id)
+    stub_any_publishing_api_put_content
+    stub_any_publishing_api_no_links
     fill_in "title", with: "A great title"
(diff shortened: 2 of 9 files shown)


benthorner:
@kevindew just to document a few things from our discussion. We're trying to solve a couple of problems:

   - It's laborious/confusing to debug errors that happen during a request.
   - Tests involving requests take longer to run than they need to.

I'm not set on this particular PR, although I would like a solution to be consistent across all tests (involving requests). Some of the alternatives we discussed:

   - Changing `config.consider_all_requests_local = true` to `false`, which still achieves the performance gain. This would mean all response bodies show render the page associated with their response code, which means our tests are more consistent with the production behaviour. It would mean we have to use `test.log` to debug anything.

   - Modifying the approach in this PR to (only) apply to all request tests. I'm struggling to see the advantage with this one, over what we have now: we lose the performance gains, and our test behaviour becomes fragmented. I think it would be better to stay as-is.

Hope that helps. No rush to get this in or anything - it just seemed like a good time to raise it. We could also move it to a pain card if the way forward is unclear.

kevindew:
Thanks for raising this @benthorner. I'll give this some more attention when I have some spare moments.

benthorner:
Closing this for now, as I don't think we'll have time to discuss it much further. Although it would lead to improved performance in running the tests, the approach suggested here is still a fragmented one, in that the debugging process for 'tests with realistic errors' is different. This behaviour does cause some confusion in general, so we should document the debug method.

Closing in favour of https://github.com/alphagov/content-publisher/pull/1991.

PR #1893, opened 2020-03-12
Add file attachment metadata for publications

benthorner:
https://trello.com/c/HRUXGp3K/1518-support-metadata-for-featured-file-attachments-db-modelling

This adds additional metadata fields for file attachments in order
support the extended workflow for publications. Since these fields
are inherent to the GOV.UK concept of attachments, we think they
should be represented as columns on the model - we also considered
a JSON blob alternative, but it seemed to be mostly disadvantageous.

### DIFF
--- a/app/models/file_attachment/revision.rb
+++ b/app/models/file_attachment/revision.rb
@@ -12,7 +12,17 @@ class FileAttachment::Revision < ApplicationRecord
                           foreign_key: "file_attachment_revision_id",
                           join_table: "revisions_file_attachment_revisions"
 
-  delegate :title, to: :metadata_revision
+  delegate :title,
+           :isbn,
+           :unique_reference,
+           :paper_number,
+           :unofficial?,
+           :command_paper?,
+           :act_paper?,
+           :official_document,
+           :parliamentary_session,
+           to: :metadata_revision
+
   delegate :filename,
            :asset,
            :asset_url,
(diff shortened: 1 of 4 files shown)


app/models/file_attachment/revision.rb
```diff
@@ -12,7 +12,17 @@ class FileAttachment::Revision < ApplicationRecord
...
                           join_table: "revisions_file_attachment_revisions"
 
-  delegate :title, to: :metadata_revision
+  delegate :title,
+           :isbn,
+           :unique_reference,
+           :paper_number,
+           :unofficial?,
+           :command_paper?,
+           :act_paper?,
```
ChrisBAshton: Could you help me level up my Rails knowledge and explain how `act_paper?`, `command_paper?` and `unofficial?` work?

I know Rails creates some helper methods automatically based on a model's fields, but how does it know to look under the `official_document` property for these? What would happen if we had defined `enum official_document: { isbn: "foo" }` - would that override the `isbn?` helper method?

benthorner: Sure, it is quite surprising. I copied it from our approach to revision state e.g. `revision.major?`. Docs for it seem to be pretty comprehensive...

https://api.rubyonrails.org/v5.2.4.1/classes/ActiveRecord/Enum.html

PR #1895, opened 2020-03-12
Support "Edit attachment" action for featured file attachments

leenagupte:
Trello: https://trello.com/c/lZKUDZbj

# What's changed?

Adds an "Edit attachment" link to the featured attachment preview block on the attachments index page. Clicking on "Edit attachment" takes the user to a new page where they can edit the title or replace the file.

# Expected changes
Screenshot 2020-03-12 at 16 27 02

Screenshot 2020-03-12 at 16 27 10

### DIFF
--- a/spec/features/editing_file_attachments/replace_file_attachment_file_spec.rb
+++ b/spec/features/editing_file_attachments/replace_file_attachment_file_spec.rb
@@ -1,8 +1,18 @@
 RSpec.feature "Replace a file attachment file", js: true do
-  scenario do
+  scenario "inline" do
     given_there_is_an_edition_with_an_attachment
     when_i_click_to_insert_an_attachment
     and_i_click_on_edit_file
+    and_i_upload_a_replacement_attachment_file
+    and_i_click_save
+    then_i_see_the_replacement_file
+    and_i_see_the_timeline_entry
+  end
+
+  scenario "featured" do
+    given_there_is_an_edition_with_featured_attachments
+    when_i_go_to_change_an_attachment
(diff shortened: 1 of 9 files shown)


spec/requests/featured_attachments_spec.rb
```diff
@@ -5,6 +5,22 @@
     let(:edition) { create(:edition, :published) }
   end
 
+  it_behaves_like "requests that assert edition state",
```
benthorner: Minor: this looks like it belongs in a previous commit.

spec/features/editing_file_attachments/replace_file_attachment_file_spec.rb
```diff
@@ -9,6 +9,16 @@
     and_i_see_the_timeline_entry
   end
 
+  scenario "featured" do
+    given_there_is_an_edition_with_featured_attachments
+    when_i_go_to_change_an_attachment
```
benthorner: Minor: could we combine these steps and visit the page directly? Every page we visit costs up to 1s in our test runs, so it's good to avoid unnecessary page renders if we can. (We should probably apply this in a big sweep!)

leenagupte:
> The main thing I'm concerned about is the way we've gone with the "edit" action. We have quite a few inconsistencies in featured attachments, so we decided to re-use existing file attachment pages and have a follow-up piece of work to make them consistent. Would that cause a problem for this PR?

I created a new route because the designs are noticeably different. I could move "edit" to the file_attachments_controller if you think that will make them easier to DRY-up later?

PR #1898, opened 2020-03-13
Remove organisation redirect on index page

leenagupte:
# What's changed and why?

Remove the automatic redirect that filters the index page by organisation if no filters have been selected.

This is while we fix an underlying issue with how JSON data is stored in Postgres.


1pretz1:
Should we deploy the branch straight to production to avoid the time it would take to go through the deployment pipeline?

kevindew:
Thanks for this Leena - I'm going to close this now since we've got an alternative route. Great to have had this resolve the issue for the short term.

PR #1900, opened 2020-03-13
Perform integrity check on the change_history during migration

Rosa-Fox:
[Trello](https://trello.com/c/EDTO7mb0/1466-perform-integrity-check-on-the-changehistory-during-migration)

- Update the integrity checker to ensure that the change notes we are replacing match the ones in the Publishing API

- Note to reviewer: I still need to run the import locally or on integration. This may effect whether the timestamps need to match exactly or within a range, but I wanted to open for an initial review.

### DIFF
--- a/lib/whitehall_importer/integrity_checker/change_history_check.rb
+++ b/lib/whitehall_importer/integrity_checker/change_history_check.rb
@@ -1,0 +1,47 @@
+module WhitehallImporter
+  class IntegrityChecker::ChangeHistoryCheck
+    attr_reader :live_edition, :proposed_change_history, :publishing_api_change_history
+
+    def initialize(proposed_change_history, publishing_api_change_history, live_edition:)
+      @proposed_change_history = proposed_change_history
+      @publishing_api_change_history = publishing_api_change_history
+      @live_edition = live_edition
+    end
+
+    def match?
+      return false unless history_length_matches?
+
+      if live_edition
+        history_matches?(proposed_change_history, publishing_api_change_history)
+      else
+        history_excluding_first_timestamp_matches?
+      end
+    end
+
+  private
+
+    def history_length_matches?
+      proposed_change_history.length == publishing_api_change_history.length
+    end
+
+    def history_matches?(proposed, publishing_api)
+      proposed.zip(publishing_api).all? do |proposed_history, publishing_api_history|
+        proposed_time = proposed_history["public_timestamp"]
+        publishing_api_time = publishing_api_history["public_timestamp"]
+
+        proposed_history["note"] == publishing_api_history["note"] &&
+          IntegrityChecker.time_matches?(proposed_time, publishing_api_time)
+      end
+    end
+
+    def history_excluding_first_timestamp_matches?
+      proposed_head, *proposed_tail = proposed_change_history
+      publishing_api_head, *publishing_api_tail = publishing_api_change_history
+
+      first_note_matches = proposed_head["note"] == publishing_api_head["note"]
+      remaining_history_matches = history_matches?(proposed_tail, publishing_api_tail)
+
+      first_note_matches && remaining_history_matches
+    end
+  end
+end
--- a/spec/lib/whitehall_importer/integrity_checker/change_history_check_spec.rb
+++ b/spec/lib/whitehall_importer/integrity_checker/change_history_check_spec.rb
@@ -1,0 +1,75 @@
+RSpec.describe WhitehallImporter::IntegrityChecker::ChangeHistoryCheck do
+  describe "#match?" do
+    it "returns true if the proposed change history matches the Publishing API history for a non-draft edition" do
+      proposed_change_history = change_note("First published.", Date.yesterday.noon)
+      publishing_api_change_history = change_note("First published.", Date.yesterday.noon)
+
+
+      integrity_check = described_class.new(
+        proposed_change_history,
+        publishing_api_change_history,
+        live_edition: true,
+      )
+      expect(integrity_check.match?).to be true
+    end
+
+    it "returns true if the proposed change history matches for a draft edition with mismatched time stamps for first item" do
+      proposed_change_history = change_note("First published.", Date.yesterday.noon) +
+        change_note("Updated", Time.zone.now)
+
+      publishing_api_change_history = change_note("First published.", Time.zone.now) +
+        change_note("Updated", Time.zone.now)
+
+      integrity_check = described_class.new(
+        proposed_change_history,
+        publishing_api_change_history,
+        live_edition: false,
+      )
+      expect(integrity_check.match?).to be true
+    end
+
+    it "returns false if length of change history does not match" do
+      proposed_change_history = change_note("First published.", Date.yesterday.noon)
+
+      publishing_api_change_history = change_note("First published.", Date.yesterday.noon) +
+        change_note("Updated", Time.zone.now)
+
+      integrity_check = described_class.new(
+        proposed_change_history,
+        publishing_api_change_history,
+        live_edition: true,
+      )
+      expect(integrity_check.match?).to be false
+    end
+
+    it "returns false if notes do not match" do
+      proposed_change_history = change_note("First published.", Date.yesterday.noon)
+
+      publishing_api_change_history = change_note("Updated", Date.yesterday.noon)
+
+      integrity_check = described_class.new(
+        proposed_change_history,
+        publishing_api_change_history,
+        live_edition: true,
+      )
+      expect(integrity_check.match?).to be false
+    end
+
+    it "returns false if public_timestamp does not match" do
+      proposed_change_history = change_note("First published.", Date.yesterday.noon)
+
+      publishing_api_change_history = change_note("First published.", Time.zone.now)
+
+      integrity_check = described_class.new(
+        proposed_change_history,
+        publishing_api_change_history,
+        live_edition: true,
+      )
+      expect(integrity_check.match?).to be false
+    end
+  end
+
+  def change_note(note, timestamp)
+    [{ "note" => note, "public_timestamp" => timestamp.rfc3339 }]
+  end
+end
--- a/spec/lib/whitehall_importer/integrity_checker_spec.rb
+++ b/spec/lib/whitehall_importer/integrity_checker_spec.rb
@@ -140,6 +140,12 @@ RSpec.describe WhitehallImporter::IntegrityChecker do
                                     },
                                     details: {
                                       first_public_at: first_published_at,
+                                      change_history: [
+                                        {
+                                          note: "First published.",
+                                          public_timestamp: Date.yesterday.noon,
+                                        },
+],
                                     }),
       )
 
@@ -186,6 +192,12 @@ RSpec.describe WhitehallImporter::IntegrityChecker do
                                       },
                                       details: {
                                         first_public_at: first_published_at,
+                                        change_history: [
+                                          {
+                                            note: "First published.",
+                                            public_timestamp: first_published_at,
+                                          },
+],
                                       }),
         )
 
@@ -206,6 +218,12 @@ RSpec.describe WhitehallImporter::IntegrityChecker do
                                       },
                                       details: {
                                         first_public_at: first_published_at,
+                                        change_history: [
+                                          {
+                                            note: "First published.",
+                                            public_timestamp: first_published_at,
+                                          },
+],
                                       }),
         )
 
@@ -262,6 +280,12 @@ RSpec.describe WhitehallImporter::IntegrityChecker do
         state_history: { "1" => "published" },
         details: {
           body: "body text",
+          change_history: [
+            {
+              note: "",
+              public_timestamp: Time.zone.now,
+            },
+          ],
           image: {
             alt_text: "alt text",
             caption: "caption",
@@ -349,6 +373,10 @@ RSpec.describe WhitehallImporter::IntegrityChecker do
 
     it "returns a problem when the body text doesn't match" do
       expect(integrity_check.problems).to include("body text doesn't match")
+    end
+
+    it "returns a problem when the change history doesn't match" do
+      expect(integrity_check.problems).to include("change history doesn't match")
     end
 
     it "returns a problem when the image alt_text doesn't match" do
@@ -565,6 +593,12 @@ RSpec.describe WhitehallImporter::IntegrityChecker do
       schema_name: edition.document_type.publishing_metadata.schema_name,
       details: {
         body: "",
+        change_history: [
+          {
+            note: "First published.",
+            public_timestamp: Time.zone.rfc3339("2020-03-11T12:00:00.000+00:00"),
+          },
+        ],
       },
     }.deep_merge!(override_hash)
   end
--- a/lib/whitehall_importer/integrity_checker.rb
+++ b/lib/whitehall_importer/integrity_checker.rb
@@ -56,6 +56,12 @@ module WhitehallImporter
         publishing_api_content.dig("details", "body"),
       ).sufficiently_similar?
 
+      problems << "change history doesn't match" unless ChangeHistoryCheck.new(
+        proposed_payload.dig("details", "change_history"),
+        publishing_api_content.dig("details", "change_history"),
+        live_edition: edition.live?,
+      ).match?
+
       problems
     end
 


spec/lib/whitehall_importer/integrity_checker_spec.rb
```diff
@@ -31,7 +31,7 @@
       default_publishing_api_item(edition,
                                   details: {
                                     body: GovspeakDocument.new(edition.contents["body"], edition).payload_html,
-                                  },
```
emmabeynon: Looks like indentation has slipped a bit here.

spec/lib/whitehall_importer/integrity_checker/change_history_check_spec.rb
```diff
@@ -0,0 +1,114 @@
...
+      ]
+
+      publishing_api_change_history = [
+        {
+          note: "Updated",
+          public_timestamp: public_timestamp,
+        },
+      ]
+
+      integrity_check = described_class.new(published_edition, proposed_change_history, publishing_api_change_history)
```
emmabeynon: Same comment re: line length

spec/lib/whitehall_importer/integrity_checker/change_history_check_spec.rb
```diff
@@ -0,0 +1,114 @@
...
+      ]
+
+      publishing_api_change_history = [
+        {
+          note: "First published.",
+          public_timestamp: Time.zone.now.rfc3339,
+        },
+      ]
+
+      integrity_check = described_class.new(published_edition, proposed_change_history, publishing_api_change_history)
```
emmabeynon: Same comment re: line length

lib/whitehall_importer/integrity_checker/change_history_check.rb
```diff
@@ -0,0 +1,40 @@
...
+      @edition = edition
+      @proposed_change_history = proposed_change_history
+      @publishing_api_change_history = publishing_api_change_history
+    end
+
+    def match?
+      return false unless compare_length_of_histories
+
+      if edition.draft?
+        compare_histories_excluding_first_timestamp
```
emmabeynon: Same here - maybe something like `history_excluding_first_timestamp_matches?`

lib/whitehall_importer/integrity_checker/change_history_check.rb
```diff
@@ -0,0 +1,40 @@
...
+      else
+        compare_histories(proposed_change_history, publishing_api_change_history)
```
emmabeynon: `history_matches?`?

lib/whitehall_importer/integrity_checker/change_history_check.rb
```diff
@@ -0,0 +1,40 @@
...
+
+  private
+
+    def compare_length_of_histories
+      proposed_change_history.length == publishing_api_change_history.length
+    end
+
+    def compare_histories(proposed, publishing_api)
+      proposed.zip(publishing_api).all? do |proposed_history, publishing_api_history|
+        proposed_history[:note] == publishing_api_history[:note] && proposed_history[:public_timestamp] == publishing_api_history[:public_timestamp]
```
emmabeynon: It would be good to split this over a couple of lines.

```suggestion
        proposed_history[:note] == publishing_api_history[:note]  &&
        proposed_history[:public_timestamp] == publishing_api_history[:public_timestamp]
```

lib/whitehall_importer/integrity_checker/change_history_check.rb
```diff
@@ -0,0 +1,40 @@
...
+      end
+    end
+
+    def compare_histories_excluding_first_timestamp
+      proposed_change_history_first_item, *proposed_change_history_items = proposed_change_history
+      publishing_api_change_history_first_item, *publishing_api_change_history_items = publishing_api_change_history
+
+      proposed_change_history_first_item[:note] == publishing_api_change_history_first_item[:note] && compare_histories(proposed_change_history_items, publishing_api_change_history_items)
```
emmabeynon: To make this a little more readable we could do something like:
```suggestion
      first_note_matches = proposed_change_history_first_item[:note] == publishing_api_change_history_first_item[:note]
      remaining_history_matches = compare_histories(proposed_change_history_items, publishing_api_change_history_items)
      first_note_matches && remaining_history_matches
```

spec/lib/whitehall_importer/integrity_checker/change_history_check_spec.rb
```diff
@@ -0,0 +1,114 @@
+RSpec.describe WhitehallImporter::IntegrityChecker::ChangeHistoryCheck do
+  let(:published_edition) { build(:edition, document_type_id: "news_story", state: "published") }
+  let(:draft_edition) { build(:edition, document_type_id: "news_story") }
+  let(:public_timestamp) { Time.zone.now.yesterday.rfc3339 }
+
+  describe "#match?" do
+    it "returns true if the proposed payload matches" do
```
emmabeynon: Should this be "returns true if the proposed change history matches the Publishing API history for a non-draft edition"?

spec/lib/whitehall_importer/integrity_checker/change_history_check_spec.rb
```diff
@@ -0,0 +1,114 @@
...
+          note: "First published.",
+          public_timestamp: public_timestamp,
+        },
+      ]
+
+      integrity_check = described_class.new(published_edition, proposed_change_history, publishing_api_change_history)
+      expect(integrity_check.match?).to be true
+    end
+
+    it "returns true if the proposed payload matches for a draft edition with mismatched time stamps for first item" do
```
emmabeynon: I would say it's "proposed change history" rather than "proposed payload".

spec/lib/whitehall_importer/integrity_checker/change_history_check_spec.rb
```diff
@@ -0,0 +1,114 @@
...
+          note: "Updated",
+          public_timestamp: public_timestamp,
+        },
+      ]
+
+      integrity_check = described_class.new(published_edition, proposed_change_history, publishing_api_change_history)
+      expect(integrity_check.match?).to be false
+    end
+
+    it "returns false if change history does not match" do
```
emmabeynon: I think it's the public_timestamp that we're testing rather than the whole change history item here?

lib/publishing_api_payload.rb
```diff
@@ -22,9 +22,9 @@ def payload
       links: links,
       access_limited: access_limited,
       auth_bypass_ids: auth_bypass_ids,
-      public_updated_at: history.public_updated_at,
+      public_updated_at: history.public_updated_at.rfc3339,
```
kevindew: We shouldn't need this by the way following: https://github.com/alphagov/content-publisher/pull/1887

Rosa-Fox:
Thanks @emmabeynon and @kevindew. I have made the changes and amended. I am not sure about two things...

1. @emmabeynon, you said to write in the commit as to why we don't check that `public_timestamp` matches for the first item in the change history for draft editions... I think it is due to how it is set in the [History class](https://github.com/alphagov/content-publisher/blob/f7e0c44ca3768ebe11960e29b9f754011b0bba69/lib/publishing_api_payload/history.rb#L23) as the note is hardcoded and the `first_published_at ` method depends on wether `backdated_to` has been set... but I am a bit lost... will speak to you about it tomorrow.

2. Probably related to my above confusion... in order to check that the first mis matched time stamps for draft documents still cause the check to pass, [I have to convert the time to rfc3339](https://github.com/alphagov/content-publisher/blob/ace06ca2c07c0b6aa9700552362e8dc47f6eb3a7/spec/lib/whitehall_importer/integrity_checker/change_history_check_spec.rb#L53)... this doesn't seem right but I haven't quite figured out what is happening there :(

Rosa-Fox:
Thanks @emmabeynon. I have updated the commit message.

spec/lib/whitehall_importer/integrity_checker/change_history_check_spec.rb
```diff
@@ -0,0 +1,138 @@
+RSpec.describe WhitehallImporter::IntegrityChecker::ChangeHistoryCheck do
+  let(:published_edition) {
```
emmabeynon: A small one, when we've got a block spanning more then one line, the convention is to use `do...end` rather than curly braces.

spec/lib/publishing_api_payload_spec.rb
```diff
@@ -156,6 +156,15 @@
...
     end
 
+    it "includes public_updated_at if the edition has backdated_to and is a first edition" do
+      date = Time.zone.now.yesterday
+      revision = build(:revision, backdated_to: date)
+      edition = build(:edition, revision: revision, number: 1)
+      payload = described_class.new(edition).payload
+
+      expect(payload).to match a_hash_including(public_updated_at: date)
+    end
```
kevindew: I'm not sure where this has come from. I wonder if it's removed code that has slipped in from a rebase?

lib/whitehall_importer/integrity_checker/change_history_check.rb
```diff
@@ -0,0 +1,44 @@
...
+    def initialize(edition, proposed_change_history, publishing_api_change_history)
+      @edition = edition
+      @proposed_change_history = proposed_change_history
+      @publishing_api_change_history = publishing_api_change_history
+    end
+
+    def match?
+      return false unless history_length_matches?
+
+      if edition.draft?
```
kevindew: This will only catch editions in the status of draft whereas we'd want it to cover statuses like submitted for review etc. I'd suggest switching to checking the item isn't live. 

I also think we don't actually need to pass in an edition (I think we only use one for this question) so it's probably simpler that we just pass in a boolean of whether the item is live or not and then can simplify tests a bit (we'd want to switch to a keyword arg on initialise for it)

spec/lib/whitehall_importer/integrity_checker/change_history_check_spec.rb
```diff
@@ -0,0 +1,138 @@
...
+  let(:draft_edition) { build(:edition, document_type_id: "news_story") }
+  let(:public_timestamp) { Time.zone.now.yesterday }
+
+  describe "#match?" do
+    it "returns true if the proposed change history matches the Publishing API history for a non-draft edition" do
+      proposed_change_history = [
+        {
+          note: "First published.",
+          public_timestamp: public_timestamp,
+        },
```
kevindew: We might find we can make this test a bit shorter by having a basic method of:

```
def change_note(note, timestamp) 
  { note: note, public_timestamp: timestamp.rfc3339 }
end
```

that we can use each time we need a change note for this

lib/whitehall_importer/integrity_checker/change_history_check.rb
```diff
@@ -0,0 +1,44 @@
...
+  private
+
+    def history_length_matches?
+      proposed_change_history.length == publishing_api_change_history.length
+    end
+
+    def history_matches?(proposed, publishing_api)
+      proposed.zip(publishing_api).all? do |proposed_history, publishing_api_history|
+        proposed_history[:note] == publishing_api_history[:note] &&
+          proposed_history[:public_timestamp] == publishing_api_history[:public_timestamp]
```
kevindew: I reckon we'll want to use https://github.com/alphagov/content-publisher/pull/1875/files#diff-93e7bf043d189b6ad9f28ef07d33176bR5 for this once it's merged.

spec/lib/whitehall_importer/integrity_checker/change_history_check_spec.rb
```diff
@@ -0,0 +1,138 @@
+RSpec.describe WhitehallImporter::IntegrityChecker::ChangeHistoryCheck do
+  let(:published_edition) do
+    build(:edition,
+          document_type_id: "news_story",
+          state: "published")
+  end
+  let(:draft_edition) { build(:edition, document_type_id: "news_story") }
+  let(:public_timestamp) { Time.zone.now.yesterday }
```
kevindew: I imagine we can remove the need for this let - it seems a bit confusingly named as it's not a special timestamp and we can maybe replace it with `Date.yesterday.noon` in each of the tests at a similar number of characters.

spec/lib/whitehall_importer/integrity_checker/change_history_check_spec.rb
```diff
@@ -0,0 +1,138 @@
...
+      ]
+
+      integrity_check = described_class.new(
+        published_edition,
+        proposed_change_history,
+        publishing_api_change_history,
+      )
+      expect(integrity_check.match?).to be false
+    end
+  end
```
kevindew: This is a good thorough collection of tests 👍

spec/lib/whitehall_importer/integrity_checker_spec.rb
```diff
@@ -240,6 +246,12 @@ def default_publishing_api_item(edition, override_hash = {})
...
       details: {
         body: "",
+        change_history: [
+          {
+            note: "First published.",
+            public_timestamp: Time.zone.now,
+          },
+        ],
       },
     }.deep_merge!(override_hash)
```
kevindew: I feel like we could use a basic test in here that checks we are comparing change histories - It'd be fine if it mocks the change history check. I think we just enough that were we to remove https://github.com/alphagov/content-publisher/pull/1900/files#diff-93e7bf043d189b6ad9f28ef07d33176bR47-R52 the tests would fail.

lib/whitehall_importer/integrity_checker/change_history_check.rb
```diff
@@ -0,0 +1,44 @@
...
+      end
+    end
+
+    def history_excluding_first_timestamp_matches?
+      proposed_change_history_first_item, *proposed_change_history_items = proposed_change_history
+      publishing_api_change_history_first_item, *publishing_api_change_history_items = publishing_api_change_history
```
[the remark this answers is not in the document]
Rosa-Fox: 👍

spec/lib/whitehall_importer/integrity_checker_spec.rb
Rosa-Fox: 👍 thanks! This was useful in that I realised I needed to access the attributes of `proposed_change_history` and `publishing_api_change_history` with string keys rather than symbols in `ChangeHistoryCheck` due to the `PublishingApiPayload` being JSON.

lib/whitehall_importer/integrity_checker/change_history_check.rb
```diff
@@ -0,0 +1,47 @@
...
+    def initialize(live_edition, proposed_change_history, publishing_api_change_history)
+      @live_edition = live_edition
+      @proposed_change_history = proposed_change_history
+      @publishing_api_change_history = publishing_api_change_history
+    end
+
+    def match?
+      return false unless history_length_matches?
+
+      if !live_edition
```
emmabeynon: Tiny one, but I think it's an easier read if it's `if live_edition...else`, which will mean a switch around of the methods that are called.

spec/lib/whitehall_importer/integrity_checker/change_history_check_spec.rb
```diff
@@ -0,0 +1,99 @@
...
+      proposed_change_history = [
+        {
+          "note" => "First published.",
+          "public_timestamp" => Date.yesterday.noon.rfc3339,
+        },
+        {
+          "note" => "Updated",
+          "public_timestamp" => Time.zone.now.rfc3339,
+        },
+      ]
```
emmabeynon: Could we use the `change_note` method to construct this?  I think you could call it with the required data twice and add them together.  Same would apply to the other instances of a change history with multiple change notes.

lib/whitehall_importer/integrity_checker/change_history_check.rb
emmabeynon: @Rosa-Fox I think from Kevin's comment here he's expecting `live_edition` to be a keyword arg, so `def initialize(live_edition:...`

lib/whitehall_importer/integrity_checker/change_history_check.rb
emmabeynon: @Rosa-Fox This has been merged now so we can change to use it `IntegrityChecker.time_matches?(...)`

spec/lib/whitehall_importer/integrity_checker/change_history_check_spec.rb
emmabeynon: Can we use the `change_note` method to construct this?  We could create two separate change notes and add them together.  Same would apply to the other test data in this spec which has more than one change note.

Rosa-Fox:
Thank you @emmabeynon :) Have amended the changes.

lib/whitehall_importer/integrity_checker/change_history_check.rb
Rosa-Fox: 👍

PR #1902, opened 2020-03-13
Rename official_document -> official_document_type

benthorner:
https://trello.com/c/HRUXGp3K/1518-support-metadata-for-featured-file-attachments-db-modelling

This renames the attribute to better reflect it's value.

### DIFF
--- a/app/models/file_attachment/metadata_revision.rb
+++ b/app/models/file_attachment/metadata_revision.rb
@@ -2,9 +2,9 @@
 class FileAttachment::MetadataRevision < ApplicationRecord
   belongs_to :created_by, class_name: "User", optional: true
 
-  enum official_document: { unofficial: "unofficial",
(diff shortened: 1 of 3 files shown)


benthorner:
@ChrisBAshton looks like we both ended up at "officially unofficial". The problem with other words is that they don't reflect any terminology we've been using thus far, and ideally we want to be consistent with words that are in use already. I'd say "classification" and "type" are pretty much synonymous, so I'm not seeing any particular advantage there.

At this point, I think we should just go with something and we can always change it if it turns out to be confusing. Do we prefer `official_document` or `official_document_type`?

(By the way, "classification" has a bad rep from Whitehall - I challenge you to work out what [this model](https://github.com/alphagov/whitehall/blob/master/app/models/classification_featuring.rb) is for!)

PR #1903, opened 2020-03-13
Add tracking to reorderable list component

alex-ju:
- Emit events when reordering items (via buttons or drag-and-drop)
- Add reorder listener script to capture these events and send them to GTM

Note: I'm not too opinionated about the structure of these events as long as they are consistent and usable for performance analysis.

[Trello card](https://trello.com/c/H3T7JWtS)

### DIFF
--- a/app/assets/javascripts/components/reorderable-list.js
+++ b/app/assets/javascripts/components/reorderable-list.js
@@ -13,7 +13,10 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
     this.sortable = window.Sortable.create(this.$module, { // eslint-disable-line new-cap
       chosenClass: 'app-c-reorderable-list__item--chosen',
       dragClass: 'app-c-reorderable-list__item--drag',
-      onSort: this.updateOrderIndexes.bind(this)
+      onSort: function () {
+        this.updateOrderIndexes.bind(this)
+        this.triggerEvent(this.$module, 'reorder-drag')
+      }.bind(this)
     })
 
     if (typeof window.matchMedia === 'function') {
@@ -59,6 +62,7 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
         e.target.nextElementSibling.focus()
       }
     }
+    this.triggerEvent(e.target, 'reorder-move-up')
   }
 
   ReorderableList.prototype.onClickDownButton = function (e) {
@@ -76,6 +80,7 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
         e.target.previousElementSibling.focus()
       }
     }
+    this.triggerEvent(e.target, 'reorder-move-down')
   }
 
   ReorderableList.prototype.updateOrderIndexes = function () {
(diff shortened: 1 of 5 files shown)


spec/javascripts/modules/gtm-reorder-listener-spec.js
```diff
@@ -0,0 +1,85 @@
...
+
+      document.body.appendChild(container)
+      window.dataLayer = []
+    })
+
+    afterEach(function () {
+      document.body.removeChild(container)
+    })
+
+    describe('reorder-move-up events', function () {
```
benthorner: Minor: could we combine these two (up/down) into a single spec? We do this [for copy/paste](https://github.com/alphagov/content-publisher/blob/master/spec/javascripts/modules/gtm-copy-paste-listener-spec.js), as the events are so similar.

app/views/components/_reorderable_list.html.erb
```diff
@@ -2,6 +2,8 @@
   items ||= []
   data_attributes ||= {}
   data_attributes[:module] = "reorderable-list"
+  data_gtm = data_attributes[:gtm] || "reorderable-list"
```
benthorner: I think we should avoid GTM attributes here, since they're coupled to our approach to analytics. We could just handle this as part of the listener, much like [we do for topics](https://github.com/alphagov/content-publisher/blob/master/app/assets/javascripts/modules/gtm-topics-listener.js#L3).

app/assets/javascripts/components/reorderable-list.js
```diff
@@ -85,5 +90,19 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
     })
   }
 
+  ReorderableList.prototype.triggerEvent = function (element, eventName, detail) {
```
benthorner: Minor: we don't use the `detail` parameter, so we could lose it, I think.

alex-ju: Removed the detail – not used in the current scenario, indeed

spec/javascripts/modules/gtm-reorder-listener-spec.js
alex-ju: Combined.

app/views/components/_reorderable_list.html.erb
alex-ju: I've simplified the way we send data to GTM (without requiring a data attribute)

app/assets/javascripts/modules/gtm-reorder-listener.js
```diff
@@ -0,0 +1,19 @@
...
+GtmReorderListener.handleReorder = function (eventName) {
+  return function (event) {
+    var element = event.target
+
+    if (!element) {
+      return
+    }
+
+    window.dataLayer.push({
+      event: eventName
```
benthorner: We seem to be missing the `dataGtm` attribute now 😬. I think we should follow [the approach for topics](https://github.com/alphagov/content-publisher/blob/master/app/assets/javascripts/modules/gtm-topics-listener.js#L3), here. We could have two handlers: `handleClick` and `handleDrag`.

Happy to chat some more if that helps!

alex-ju: I think I get what you mean. It felt a bit unnecessary to me to have both `eventName` and `dataGtm` passed to the listener as we can deduce them, but have updated now.

app/assets/javascripts/modules/gtm-reorder-listener.js
```diff
@@ -0,0 +1,20 @@
+var GtmReorderListener = {}
+
+GtmReorderListener.handleReorder = function (eventName, dataGtm) {
+  return function (event) {
+    var element = event.target
+
+    if (!element || !eventName || !dataGtm) {
```
benthorner: Minor: we don't really need this, since the first one doesn't get used anywhere, and the latter two we set ourselves in the same file.

alex-ju: Good call – I removed them.

PR #1914, opened 2020-03-17
Support unautheticated error viewing

kevindew:
While looking at our previous incident I noticed we had a number of 500
responses for users being served a 404 page. These were occurring when
users were requesting a URL that Rails did not recognise and the
`authenticate_user!` action of warden was raising an exception, trace:

```
UncaughtThrowError
uncaught throw :warden
/data/vhost/content-publisher/shared/bundle/ruby/2.6.0/gems/warden-1.2.8/lib/warden/proxy.rb:134:in `throw'
/data/vhost/content-publisher/shared/bundle/ruby/2.6.0/gems/warden-1.2.8/lib/warden/proxy.rb:134:in `authenticate!'
/data/vhost/content-publisher/shared/bundle/ruby/2.6.0/gems/gds-sso-14.3.0/lib/gds-sso/controller_methods.rb:47:in `authenticate_user!'
/data/vhost/content-publisher/shared/bundle/ruby/2.6.0/gems/activesupport-6.0.2.1/lib/active_support/callbacks.rb:429:in `block in make_lambda
```

This suggests that the Rails exception handler is preventing the throw
mechanism that warden uses to require authentication.

To resolve this I've set up tests for the error actions without
authentication and modified the application layout to handle a user not
existing.

### DIFF
--- a/app/controllers/errors_controller.rb
+++ b/app/controllers/errors_controller.rb
@@ -1,6 +1,14 @@
 class ErrorsController < ApplicationController
+  skip_before_action :authenticate_user!
   skip_before_action :verify_authenticity_token
   skip_before_action :check_user_access
+
+  before_action do
+    # retrieves an existing sesssion for users if available. Unlike
+    # `authenticate_user!` this does not redirect a user when they are not
+    # authenticated.
+    warden.authenticate
+  end
 
   def bad_request
     render status: :bad_request, formats: :html
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -6,8 +6,8 @@
 
   <script>
     var dataLayer = [{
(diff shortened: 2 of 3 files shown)


app/controllers/errors_controller.rb
```diff
@@ -1,7 +1,15 @@
 class ErrorsController < ApplicationController
+  skip_before_action :authenticate_user!
   skip_before_action :verify_authenticity_token
   skip_before_action :check_user_access
 
+  before_action do
+    # soft authenticate so we can serve error pages to users who aren't logged
+    # in - the hard `authenticate_user!` approach is incompatible with a rails
+    # error app
```
benthorner: Minor: would be helpful to say this just retrieves any available session, instead of forcing a login.

PR #1917, opened 2020-03-18
Confirm delete attachment

GDSNewt:
When interacting with attachments via the markdown editor the option to "Delete attachment" will now route the user to a confirmation page. Previously clicking this link would have deleted the attachment and displayed a flash message in the modal.  

Screenshot 2020-03-18 at 11 26 08

Adding a confirmation page in this manner brings attachments in line with other delete actions. 

Screenshot 2020-03-18 at 11 25 58

### DIFF
--- a/app/assets/javascripts/modules/inline-attachment-modal.js
+++ b/app/assets/javascripts/modules/inline-attachment-modal.js
@@ -34,7 +34,7 @@ InlineAttachmentModal.prototype.actionCallback = function (item) {
     'insert': function () {
       this.workflow.render(window.ModalFetch.getLink(item))
     },
-    'confirm-delete': function () { 
+    'confirm-delete': function () {
       this.workflow.render(window.ModalFetch.getLink(item))
     },
     'delete': function () {
--- a/app/assets/javascripts/modules/inline-attachment-modal.js
+++ b/app/assets/javascripts/modules/inline-attachment-modal.js
@@ -34,6 +34,9 @@ InlineAttachmentModal.prototype.actionCallback = function (item) {
     'insert': function () {
       this.workflow.render(window.ModalFetch.getLink(item))
     },
+    'confirm-delete': function () { 
+      this.workflow.render(window.ModalFetch.getLink(item))
+    },
     'delete': function () {
       this.workflow.render(window.ModalFetch.postForm(item))
     },
@@ -44,6 +47,9 @@ InlineAttachmentModal.prototype.actionCallback = function (item) {
       this.workflow.render(window.ModalFetch.postForm(item))
     },
     'back': function () {
+      this.workflow.render(window.ModalFetch.getLink(item))
+    },
+    'cancel': function () {
       this.workflow.render(window.ModalFetch.getLink(item))
     },
     'insert-attachment-block': function () {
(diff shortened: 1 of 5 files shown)


app/assets/javascripts/modules/inline-attachment-modal.js
```diff
@@ -34,6 +34,9 @@ InlineAttachmentModal.prototype.actionCallback = function (item) {
     'insert': function () {
       this.workflow.render(window.ModalFetch.getLink(item))
     },
+    'confirm-delete': function () {
```
benthorner: I'd actually switch this around with 'delete'. Thinking about it, the 'confirmation' is really on the next page, and the first action is to 'delete'. Does that make sense?

app/views/file_attachments/confirm_delete.html.erb
```diff
@@ -4,7 +4,7 @@
                       file_attachments_path(@edition.document)
                     end %>
 
-<% content_for :back_link, render_back_link(href: back_link_path) %>
+<% content_for :back_link, render_back_link(href: back_link_path, data_attributes: { "modal-action": "back" }) %>
```
benthorner: Good spot ⭐️

app/views/file_attachments/confirm_delete.html.erb
```diff
@@ -25,7 +29,8 @@
       <div class="govuk-body">
         <%= link_to "Cancel",
                   back_link_path,
-                  class: "govuk-link govuk-link--no-visited-state" %>
+                  class: "govuk-link govuk-link--no-visited-state", 
+                  data: { "modal-action": "cancel" } %>
```
benthorner: Minor: I think we expect 'cancel' to behave the same as 'back'. What do you think about re-using 'back' here?

benthorner: Minor: looking at the other four places we have a 'Cancel' action, we should probably use a slightly more specific name here e.g. `cancel-delete-attachment`.

GDSNewt: Using 'back' did occur to me as it is functionally identical but in the end I thought maybe it was good to be a bit more explicit in order to reinforce the idea that they are two seperate user actions. Having said that, I think i'm more inclined just to remove cancel entirely and use back. At least until we develop a policy regarding cancel.

app/views/file_attachments/_file_attachment.html.erb
```diff
@@ -29,19 +29,14 @@
   },
 ) %>
 
-<% actions << capture do %>
-  <%= form_tag(
-    file_attachment_path(document, attachment.file_attachment_id),
-    method: :delete,
-    class: "app-inline-block",
+<% actions << link_to(
+  "Delete attachment",
```
benthorner: Minor: alignment issue here.

app/views/file_attachments/confirm_delete.html.erb
```diff
@@ -30,7 +30,9 @@
         <%= link_to "Cancel",
                   back_link_path,
                   class: "govuk-link govuk-link--no-visited-state", 
-                  data: { "modal-action": "cancel" } %>
+                  data: { "modal-action": "cancel",
+                  gtm: "cancel-delete-attachment"
```
benthorner: Minor: alignment issue here.

GDSNewt:
> Looks good, thanks @GDSNewt! Ideally we should `fixup` those last two "fix" commits into the previous commits, so the commit history is focussed on what we changed. However, I think this PR has been around for long enough, so that's just a suggestion, and I'm happy for this to go in 👏.

Thanks Ben, I've 'fixuped' the two commits. Should be prettier now 😄

PR #1918, opened 2020-03-18
Add a new feature to reorder featured attachments

benthorner:
https://trello.com/c/7NXYdxZV/1538-user-can-set-desired-featured-attachment-order

This adds a new 'reorder' page/route so a user can set the order of
featured attachments. Ordering for newly uploaded attachments is currently
undefined, and will be addressed in future work. Note that I'm leaving out the 
'Cancel' actions [pending a decision on this](https://trello.com/c/wmJzDSA1/1678-audit-the-use-of-cancel-actions-and-implement-a-policy-for-them).

## JS
Screenshot 2020-03-18 at 13 12 44

## no-JS
Screenshot 2020-03-18 at 13 15 15

### DIFF
--- a/app/interactors/featured_attachments/update_order_interactor.rb
+++ b/app/interactors/featured_attachments/update_order_interactor.rb
@@ -1,0 +1,54 @@
+class FeaturedAttachments::UpdateOrderInteractor < ApplicationInteractor
+  delegate :params,
+           :user,
+           :edition,
+           :attachments,
+           :updater,
+           to: :context
+
+  def call
+    Edition.transaction do
+      find_and_lock_edition
+      reorder_attachments
+      update_edition
+      create_timeline_entry
+      update_preview
+    end
+  end
+
+  def find_and_lock_edition
+    context.edition = Edition.lock.find_current(document: params[:document])
+    assert_edition_state(edition, &:editable?)
+
+    assert_edition_feature(edition, assertion: "supports featured attachments") do
+      edition.document_type.attachments.featured?
+    end
+  end
+
+  def reorder_attachments
+    new_ordering = edition.featured_attachments.sort_by do |attachment|
+      ordering_params[attachment.featured_attachment_id].to_i
+    end
+
+    context.updater = Versioning::RevisionUpdater.new(edition.revision, user)
+    updater.assign(featured_attachment_ordering: new_ordering.map(&:featured_attachment_id))
+    context.fail! unless updater.changed?
+  end
+
+  def update_edition
+    EditDraftEditionService.call(edition, user, revision: updater.next_revision)
+    edition.save!
+  end
+
+  def create_timeline_entry
+    TimelineEntry.create_for_revision(entry_type: :attachments_reordered, edition: edition)
+  end
+
+  def update_preview
+    FailsafeDraftPreviewService.call(edition)
+  end
+
+  def ordering_params
+    @ordering_params ||= params.require(:attachments).permit(ordering: {})[:ordering].to_h
+  end
+end
--- a/app/models/metadata_revision.rb
+++ b/app/models/metadata_revision.rb
@@ -5,6 +5,7 @@
 # This model is immutable.
 class MetadataRevision < ApplicationRecord
   validates :change_history, "metadata_revision/change_history" => true
+  validates :featured_attachment_ordering, "metadata_revision/featured_attachment_ordering" => true
 
   belongs_to :created_by, class_name: "User", optional: true
 
--- a/app/models/metadata_revision.rb
+++ b/app/models/metadata_revision.rb
@@ -4,32 +4,7 @@
 #
 # This model is immutable.
 class MetadataRevision < ApplicationRecord
-  validates_each :change_history do |record, attribute, change_history|
-    uuid_regex = /\A[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}\Z/
-
-    previous_time = nil
-    change_history.each do |change_note|
-      unless change_note.keys.sort == %w(id note public_timestamp)
-        record.errors.add(attribute, "has an entry with invalid keys", strict: true)
-      end
-
-      unless change_note["id"].match?(uuid_regex)
-        record.errors.add(attribute, "has an entry with a non UUID id", strict: true)
-      end
-
-      begin
-        time = Time.zone.rfc3339(change_note["public_timestamp"])
-      rescue ArgumentError
-        record.errors.add(attribute, "has an entry with an invalid timestamp", strict: true)
-      end
-
-      if previous_time && previous_time < time
-        record.errors.add(attribute, "is not in a reverse chronological ordering", strict: true)
-      end
-
-      previous_time = time
-    end
-  end
+  validates :change_history, "metadata_revision/change_history" => true
 
   belongs_to :created_by, class_name: "User", optional: true
 
--- a/app/models/metadata_revision/featured_attachment_ordering_validator.rb
+++ b/app/models/metadata_revision/featured_attachment_ordering_validator.rb
@@ -1,0 +1,15 @@
+class MetadataRevision::FeaturedAttachmentOrderingValidator < ActiveModel::EachValidator
+  ORDER_ITEM_REGEX = /\AFileAttachment\d+\Z/.freeze
+
+  def validate_each(record, attribute, value)
+    value.each do |order_item|
+      unless order_item.match?(ORDER_ITEM_REGEX)
+        record.errors.add(attribute, "has an entry with a malformed ID", strict: true)
+      end
+    end
+
+    unless value.uniq == value
+      record.errors.add(attribute, "has a duplicate entry", strict: true)
+    end
+  end
+end
--- a/app/models/revision.rb
+++ b/app/models/revision.rb
@@ -56,6 +56,7 @@ class Revision < ApplicationRecord
            :backdated_to,
            :document_type,
            :editor_political,
+           :featured_attachment_ordering,
            to: :metadata_revision
 
   delegate :tags,
--- a/app/models/revision.rb
+++ b/app/models/revision.rb
@@ -75,4 +75,10 @@ class Revision < ApplicationRecord
   def assets
     image_revisions.flat_map(&:assets) + file_attachment_revisions.map(&:asset)
   end
+
+  def featured_attachments
+    file_attachment_revisions.sort_by do |attachment|
+      featured_attachment_ordering.find_index(attachment.featured_attachment_id).to_i
+    end
+  end
 end
--- a/app/views/featured_attachments/index.html.erb
+++ b/app/views/featured_attachments/index.html.erb
@@ -24,6 +24,15 @@
         <%= t("featured_attachments.index.featured_attachments.description_govspeak") %>
       </p>
 
+      <p class="govuk-body govuk-!-margin-bottom-8">
+        <%= link_to(
+          "Reorder attachments",
+          reorder_featured_attachments_path(@edition.document),
+          class: "govuk-link",
+          data: { gtm: "reorder-attachments" }
+        ) %>
+      </p>
+
       <% attachments.each do |attachment| %>
         <%= render "featured_attachment", document: @edition.document, attachment: attachment %>
       <% end %>
--- a/app/views/featured_attachments/index.html.erb
+++ b/app/views/featured_attachments/index.html.erb
@@ -13,7 +13,7 @@
       margin_bottom: true,
     } %>
 
-    <% attachments = @edition.file_attachment_revisions %>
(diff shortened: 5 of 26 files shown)


app/models/revision.rb
```diff
@@ -75,4 +75,10 @@ def image_revisions_without_lead
   def assets
     image_revisions.flat_map(&:assets) + file_attachment_revisions.map(&:asset)
   end
+
+  def featured_attachments
+    file_attachment_revisions.sort_by do |attachment|
+      featured_attachment_ordering.find_index(attachment.featured_attachment_id)
```
kevindew: I think this is liable to crash if we have a situation where one of the attachments doesn't exist in the featured_attachment_ordering array (it'd crash when it compares an integer with a nil).

Might need this to fall back to Float::INFINITY so any missing items end up at the end of the list.

app/interactors/featured_attachments/update_order_interactor.rb
```diff
@@ -0,0 +1,45 @@
...
+
+    assert_edition_feature(edition, assertion: "supports featured attachments") do
+      edition.document_type.attachments.featured?
+    end
+  end
+
+  def reorder_attachments
+    new_ordering = params[:ordering].keys.sort_by do |featured_attachment_id|
+      params[:ordering][featured_attachment_id].to_i
+    end
```
kevindew: This seems to assume that a users query included all attachments in it (and doesn't include any junk) which could be a source of subtle and rare bugs.

I'd suggest sorting a duplicate of the existing featured_attachment_ordering and looking up the param of each one so that way it'll contain all the same items only re-ordered and won't allow anything unexpected in there.

app/views/featured_attachments/index.html.erb
```diff
@@ -24,6 +24,18 @@
...
+        <%= link_to(
+          "Reorder attachments",
+          reorder_featured_attachments_path(@edition.document),
+          class: "govuk-link",
+          data: { gtm: "reorder-attachments" }
+        ) %>
+      </p>
+
+      <br/>
+      <br/>
```
kevindew: Are these present to create some additional margin? I'd assume there's a better class based approach we can use for this.

app/views/featured_attachments/reorder.html.erb
```diff
@@ -0,0 +1,36 @@
...
+  <div class="govuk-grid-column-two-thirds">
+    <p class="govuk-body app-js-only">
+      <%= t("featured_attachments.reorder.description") %>
+    </p>
+
+    <p class="govuk-body app-no-js">
+      <%= t("featured_attachments.reorder.nojs_description") %>
+    </p>
+
+    <br/>
```
kevindew: Similar point about margin

app/interactors/featured_attachments/update_order_interactor.rb
```diff
@@ -0,0 +1,45 @@
...
+           :attachments,
+           :updater,
+           to: :context
+
+  def call
+    Edition.transaction do
+      find_and_lock_edition
+      reorder_attachments
+      update_edition
+      create_timeline_entry
```
kevindew: I assume we also want to update the preview here? even though it won't yet have an affect for users.

app/models/metadata_revision.rb
```diff
@@ -4,32 +4,7 @@
...
-      end
-
-      if previous_time && previous_time < time
-        record.errors.add(attribute, "is not in a reverse chronological ordering", strict: true)
-      end
-
-      previous_time = time
-    end
-  end
+  validates_with ChangeHistoryValidator
```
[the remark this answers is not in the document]
benthorner: Ah, interesting. I've updated as you suggested. My conclusion is that it simplified the validators, but complicated the tests - we now have `ATTRIBUTE` there instead! I'm also struggling to like the `validate_each` technique, when we're only validating one attribute.

app/models/metadata_revision/featured_attachment_ordering_validator.rb
```diff
@@ -0,0 +1,12 @@
+class MetadataRevision::FeaturedAttachmentOrderingValidator < ActiveModel::Validator
+  ATTRIBUTE = :featured_attachment_ordering
+  ORDER_ITEM_REGEX = /^FileAttachment\d+$/.freeze
```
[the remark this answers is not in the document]
benthorner: Sure, I've updated with `\A` and `\Z`. My assumption was `$` and `^` are more familiar, but the former pair are more correct.

I'm going to leave the ID format as-is for now, since we're not expecting to be reading it often. I like that it's the shortest thing that works.

app/models/metadata_revision/featured_attachment_ordering_validator.rb
```diff
@@ -0,0 +1,12 @@
...
+
+  def validate(record)
+    record.featured_attachment_ordering.each do |order_item|
+      unless order_item.match?(ORDER_ITEM_REGEX)
+        record.errors.add(ATTRIBUTE, "has an entry with a malformed ID", strict: true)
+      end
```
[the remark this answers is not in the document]
benthorner: That's a good idea - I can see some dodgy code doing an append or something. I've added that in.

app/models/revision.rb
benthorner: That's an interesting point. It looks like `sort_by` copes with this, which is why https://github.com/alphagov/content-publisher/pull/1918/files#diff-da3c22393aa90b5ca4f6c96d0459b297R3 is passing.

app/views/featured_attachments/index.html.erb
benthorner: Yeah, the design calls for it, and it does look worse without them. I'd like to get @alex-ju's opinion on what we can do here. I tried `govuk-!-margin-bottom-8`, but it wasn't enough. I need more margin bottom!

app/interactors/featured_attachments/update_order_interactor.rb
benthorner: This is an interesting one. I actually used your suggested approach originally, and then found this to be slightly less convoluted. Some reasons this could be acceptable:

   - Ordering is 'optional', in that we cope without it
   - If there is dodgy data, we'll store it and be able to debug
   - The validation protects us from obviously dodgy stuff

Would you be more concerned if we continued as-is?

app/interactors/featured_attachments/update_order_interactor.rb
benthorner: So my thinking here was to leave this to be covered by https://trello.com/c/hCrcFkHT/1514-render-featured-attachments-in-the-preview-of-a-publication, which is where it's needed. But it's also conventional to put it in. Updated 👍 .

benthorner:
@kevindew thanks for the suggestions! I think this is ready for another look now, when you have a moment.

app/models/revision.rb
kevindew: this would cope with every item not being in the list (as all nils would be compared to nil) but  would error once you have a mixture (int compared to nil)

app/interactors/featured_attachments/update_order_interactor.rb
kevindew: Yeah, I think this way seems the less ideal approach since it would miss any new items added during submission and any dodgy items would raise a 500 exception rather than be a client error.

app/views/featured_attachments/index.html.erb
kevindew: Yup, good call. Alex should have some ideas for this.

app/models/revision.rb
benthorner: Ah, I understand. Good spot - I've added an extra test and updated 👍

app/interactors/featured_attachments/update_order_interactor.rb
benthorner: Ah, fair enough - I thought it would give a 4xx response. Updated to ensure we only set a valid ordering.

app/views/featured_attachments/index.html.erb
alex-ju: The prototype uses `govuk-body govuk-!-margin-bottom-8` – I don't really like seeing overwrites, but the design calls for it as you mentioned.

app/views/featured_attachments/reorder.html.erb
alex-ju: Same here, `govuk-body govuk-!-margin-bottom-8`

app/views/featured_attachments/index.html.erb
benthorner: Ah, OK. Strange, I thought it wasn't working when I tried it - updated, thanks @alex-ju 👍

benthorner:
@kevindew thanks for following-up. Hopefully this is ready to go now.

app/interactors/featured_attachments/update_order_interactor.rb
kevindew: Thanks - one thing to note is that to_i will mean that anything not submitted will end up at the top of the list whereas Float::INFINITY would push those to the bottom. I feel the latter is slightly more intuitive but not too big a deal as long as we've made an informed choice.

Also I don't think there's anything that ensures this is a hash so this could still crash off unexpected input. Perhaps we can use dig on params to avoid that?

app/models/revision.rb
```diff
@@ -74,4 +75,10 @@ def image_revisions_without_lead
   def assets
     image_revisions.flat_map(&:assets) + file_attachment_revisions.map(&:asset)
   end
+
+  def featured_attachments
+    file_attachment_revisions.sort_by do |attachment|
+      featured_attachment_ordering.find_index(attachment.featured_attachment_id).to_i
```
kevindew: Same point on making an informed choice of the consequences of items not in the list - may just need something in commit message.

app/views/featured_attachments/index.html.erb
```diff
@@ -24,6 +24,15 @@
         <%= t("featured_attachments.index.featured_attachments.description_govspeak") %>
       </p>
 
+      <p class="govuk-body govuk-!-margin-bottom-8">
```
kevindew: yay 👍

app/interactors/featured_attachments/update_order_interactor.rb
benthorner: Indeed, the `to_i` seemed like the simplest thing to ensure we "cope" with that scenario (rather than explaining the need for a specific position). I've updated the commits to explain this more; we could always define the ordering differently if it becomes an issue.

I was about to follow the `dig` suggestion, but then realised `[].dig` would cause a crash. In this case, I think we may be putting too much thought into something that should only affect the user submitting the form, and is likely due to them doing some form of hackery.

Is there a particular scenario you think we should protect against here?

app/models/revision.rb
benthorner: (I've also updated this commit to explain a bit more about the choice.)

app/interactors/featured_attachments/update_order_interactor.rb
kevindew: I think it'll only be an issue if multiple users work on the same thing and are surprised by the ordering, which is hopefully a very low risk.

Rather than particular scenarios we should just always be cautious that user input is handled in the appropriate way - so any case where a user input can cause a 500 error is a smell to me. 

I did some tinkering and realise dig isn't the right approach apologies, it does leave vulnerability for 500s still. Hmm the Rails preferred approach would be using the `params.permit(ordering: {})` which raises a bad request if it's not a hash - however this creates that annoying second warning for unpermitted params - so this would probably better in a context of `params.require(:attachments).permit(ordering: {})` which would be a clean way with bad requests handled.

But yeah without making further modifications this could be safe with `params[:ordering].try(:[], attachment.featured_attachment_id).to_i`.

app/models/revision.rb
kevindew: Great thanks

app/interactors/featured_attachments/update_order_interactor.rb
benthorner: Hmm...that doesn't work if we do this:

```
irb(main):017:0> ActionController::Parameters.new(ordering: [1])[:ordering].try(:[], :foo)      
Traceback (most recent call last):
        2: from (irb):17
        1: from (irb):17:in `rescue in irb_binding'
TypeError (no implicit conversion of Symbol into Integer)
```

I agree it makes sense to have definite knowledge about what our app would do in the case of an invalid input. I've added an extra commit in order to do a `require...permit`.

Could you take another look?

kevindew: Argh good catch! thanks for that. Looks good 👍 I think the `to_h` is redundant in `ordering_params` but I know we do similar elsewhere to break out of Parameters so no biggie.

benthorner:
@kevindew thanks for approving ⚡️. (Now it won't let me comment inline.) Just so you know, it turns out we do need `to_h`, as it can still be `nil` (we'd have to `require` it, instead). Somehow I think this convention could be easier to use!

PR #1920, opened 2020-03-18
Fix flakey reorderable-list tests

alex-ju:
### What
Attach dataLayer object to `window` to fix flakey `reorderable-list` tests

### Why
The events emitted by the reorderable-list component are picked up by listeners that expect a `dataLayer` object to exist.

Fix tested with `jasmine:ci[true,64655]` seed [which currently fails](https://ci.integration.publishing.service.gov.uk/job/content-publisher/job/missing-full-stop/1/console).

### DIFF
--- a/spec/javascripts/helpers/gtm-mock.js
+++ b/spec/javascripts/helpers/gtm-mock.js
@@ -1,0 +1,5 @@
+/* eslint-env jasmine */
+
+beforeEach(function () {
+  window.dataLayer = []
+})


spec/javascripts/components/reorderable-list-spec.js
```diff
@@ -40,6 +40,7 @@ describe('Reorderable list component', function () {
     document.body.classList.add('js-enabled')
     document.body.appendChild(container)
     element = document.querySelector('[data-module="reorderable-list"]')
+    window.dataLayer = [] // the emitted events are picked up by listeners that expect a dataLayer object to exist
```
kevindew: This looks good - just wondering if there is anywhere more global we can put it. Otherwise any future tests that get caught on this will be dependent on the ordering of this.

Perhaps this should be added in a beforeEach in a helper file as per: https://github.com/alphagov/content-publisher/blob/6410299fa6f08d9e45ba2f32ef33d090ac94a2e6/spec/javascripts/helpers/fetch-mock.js

alex-ju: I was reluctant about doing this as our scripts, in general, shouldn't really rely on `dataLayer` being there. I'd see this as an exception. But happy to do that if you think it's valuable.

alex-ju: Moved the dataLayer initialisation in a global helper.

kevindew: Yeah I think it'd be better placed there. This alternative of doing it in one suite of tests causes side effects for any later running tests which isn't ideal.

If we want to do it for one suite of tests it should probably be a mock that is cleared after the tests complete.

PR #1926, opened 2020-03-19
Use mail-notify gem instead of custom ActionMailer implementation

benjamineskola:
The code for integrating GOV.UK Notify with ActionMailer is now being used in several different apps, and ideally we want to reduce duplication.

This replaces the custom implementation with the third-party [mail-notify gem](https://github.com/dxw/mail-notify), which should be feature-equivalent.

https://trello.com/c/9zIaFpS0/1822-3-specialist-publisher-support-api-use-the-gem-for-notify


Gemfile
```diff
@@ -16,7 +16,7 @@ gem "hashdiff", "~> 1.0"
 gem "image_processing", "~> 1"
 gem "interactor", "~> 3"
 gem "kaminari", "~> 1"
-gem "notifications-ruby-client", "~> 5.1"
+gem "mail-notify"
```
kevindew:
```suggestion
gem "mail-notify", "~> 1"
```
Could you pin this to version 1 for consistency with other gems pinning?

app/mailers/application_mailer.rb
```diff
@@ -1,5 +1,9 @@
-class ApplicationMailer < ActionMailer::Base
+class ApplicationMailer < Mail::Notify::Mailer
   def default_url_options
     { host: Plek.new.external_url_for("content-publisher") }
   end
+
+  def template_id
+    @template_id ||= ENV.fetch("GOVUK_NOTIFY_TEMPLATE_ID", "fake-test-template-id")
```
kevindew:
```suggestion
    ENV.fetch("GOVUK_NOTIFY_TEMPLATE_ID", "fake-test-template-id")
```

I don't think we really gain anything from this being a memo'd variable?

kevindew:
@benjamineskola you'll need to re-run bundle to fix the Gemfile.lock

PR #1928, opened 2020-03-19
Enforce 100% I18n coverage from now on

benthorner:
https://trello.com/c/BhRV4Sqy/1679-increase-our-i18n-coverage-to-100

This add an extra step in our internal I18n coverage checker, which will
return a non-zero (error) exit code if the coverage is below 100%. While
this will be the case when running specific tests, it won't interfere with
the test runner e.g.

     Finished in 0.56256 seconds (files took 6.93 seconds to load)
     1 example, 0 failures

     Coverage report generated for I18n to /govuk/content-publisher/coverage/i18n. 7 / 648 keys (1.08%) covered.
     I18n coverage (1.08%) is below the expected minimum coverage (100%)
     I18nCov failed with exit 1
     Coverage report generated for RSpec to /govuk/content-publisher/coverage. 296 / 528 LOC (56.06%) covered.
     SimpleCov failed with exit

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -2,10 +2,9 @@
 # for example lib/tasks/capistrano.rake, and they will automatically be available to Rake.
 
 require_relative "config/application"
-
 Rails.application.load_tasks
 
 # RSpec shoves itself into the default task without asking, which confuses the ordering.
 # https://github.com/rspec/rspec-rails/blob/eb3377bca425f0d74b9f510dbb53b2a161080016/lib/rspec/rails/tasks/rspec.rake#L6
 Rake::Task[:default].clear
-task default: %i(spec jasmine:ci lint brakeman)
+task default: %i(spec i18n_cov:ci jasmine:ci lint brakeman)
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -14,9 +14,7 @@ require "govuk_sidekiq/testing"
 require "json_matchers/rspec"
 
 Dir[Rails.root.join("spec/support/**/*.rb")].sort.each { |f| require f }
-SimpleCov.start
 GovukTest.configure
-I18nCov.start
 WebMock.disable_net_connect!(allow_localhost: true)
 Capybara.automatic_label_click = true
 ActiveRecord::Migration.maintain_test_schema!
@@ -37,6 +35,11 @@ RSpec.configure do |config|
   config.include AuthenticationHelper, type: ->(spec) { spec.in?(%i[feature request view]) }
   config.include BulkDataHelper
   config.include Capybara::RSpecMatchers, type: :request
+
+  unless config.files_to_run.one?
+    I18nCov.start
+    SimpleCov.start
+  end
 
   config.before :suite do
     Rails.application.load_seed
--- a/spec/support/i18n_cov.rb
+++ b/spec/support/i18n_cov.rb
@@ -1,7 +1,7 @@
 require "singleton"
 
 class I18nCov
-  REPORT_PATH = Rails.root.join("coverage/i18n")
+  REPORT_PATH = Rails.root.join("coverage/i18n.json")
   IN_APP_LOCALE_FILES = Dir.glob(Rails.root.join("config/locales/**/*.yml"))
 
   include Singleton
@@ -33,11 +33,10 @@ class I18nCov
   def report
     return if used_keys.blank?
 
-    all = all_locale_keys
-    scoped_used = all.select { |as| used_keys.any? { |us| as.start_with?(us) } }
-
-    write_report(all - scoped_used)
-    print_message(all, scoped_used)
+    unused_keys = check_unused_keys
+    report = generate_report(unused_keys)
+    write_report(report)
+    print_message(report)
   end
 
 private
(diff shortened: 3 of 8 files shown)


benthorner:
@ChrisBAshton thanks for taking a look. That's an interesting idea. Personally I don't find the coverage outputs annoying. For example, we always get the following when a spec fails

```
Coverage report generated for RSpec to /govuk/content-publisher/coverage. 47 / 137 LOC (34.31%) covered.
SimpleCov failed with exit 1
```

If we were to turn on SimpleCov minimum coverage, we'd get a similar response. (I forgot to mention that I was making them consistent.)

```
Coverage report generated for I18n to /govuk/content-publisher/coverage/i18n. 14 / 650 keys (2.15%) covered.
Coverage report generated for RSpec to /govuk/content-publisher/coverage. 182 / 356 LOC (51.12%) covered.
Line coverage (51.12%) is below the expected minimum coverage (100.00%).
SimpleCov failed with exit 2
```

It looks like this [is a non-trivial problem to solve](https://github.com/colszowka/simplecov/issues/429), and I'm cautious about moving coverage checking away, as I can imagine it breaking without anyone noticing. As a compromise, I've tried one of the suggestions in the issue. How does it look now?

spec/spec_helper.rb
```diff
@@ -38,6 +36,11 @@
   config.include BulkDataHelper
   config.include Capybara::RSpecMatchers, type: :request
 
+  unless config.files_to_run.one?
```
ChrisBAshton: This is an improvement 👍 I suppose we could look at only running it on full test suite by doing something like:

```suggestion
  if ARGV.grep(/rake$/)
```

But this is more brittle 🤷‍♂ I'm happy going with this compromise for the time being!

benthorner: Yeah, I also thought it was brittle. Especially as I usually want to look at coverage by running `rspec` 🤔.

kevindew: This looks cool 👍

spec/support/i18n_cov.rb
```diff
@@ -73,6 +74,15 @@ def print_message(all, used)
...
       "#{used.count} / #{all.count} keys (#{percent}%) covered."
   end
+
+  def check_covered(all, used)
+    percent = ((used.count.to_f / all.count) * 100).round(2)
+    return if percent == 100
+
+    warn "I18n coverage (#{percent}%) is below the expected minimum coverage (100%)"
+    warn "I18nCov failed with exit 1"
+    Kernel.exit 1
```
kevindew: I don't think this is the correct place to exit with 1 and raise an alert. It looks like this will conflate the test results with the coverage results. It'd show for instance in Jenkins that your tests have failed when actually it would have been this coverage. 

I'd suggest that you have a stage defined in https://github.com/alphagov/content-publisher/blob/8a9582c738b62181a3d17bc56180a505b9f4cd93/Jenkinsfile#L23-L27 which checks whether the file of unused keys is empty or not and then fail the build accordingly if necessary.

benthorner: @kevindew I'm struggling a bit to see your point of view here. Reading the file is a good idea if we do decide to go this way, as otherwise we'd have to run the tests again.

I'm basing this on the way [SimpleCov would behave](https://github.com/colszowka/simplecov#getting-started) if we set `SimpleCov.minimum_coverage 100` (although it would fail with an exit code of `2`). Are we saying we would do something special for SimpleCov as well? It's also not related to any test failures.

kevindew: Yeah, I wouldn't expect to run tests again

In my experiences of running these things in CI is you use running the tests to generate the information and the CI machine itself is configured to read this information and make decisions on the build.

For example for Simplecov and Jenkins rather than configuring it in Ruby you'd more conventionally hook the report into the [code coverage API and specify that to fail]((https://cdn.jsdelivr.net/gh/jenkinsci/code-coverage-api-plugin@master/images/config.png))

Rakefile
```diff
@@ -4,4 +4,4 @@
 require_relative "config/application"
 
 Rails.application.load_tasks
-task default: %i(jasmine:ci lint brakeman)
+task default: %i(i18n_cov:ci jasmine:ci lint brakeman)
```
ChrisBAshton: Are we confident that the default task gets run after the tests?
If not, I think there's a risk the JSON file might be missing? 🤔

benthorner: So this is because:

   - [RSpec shoves itself into the default task](https://github.com/rspec/rspec-rails/blob/eb3377bca425f0d74b9f510dbb53b2a161080016/lib/rspec/rails/tasks/rspec.rake#L6)
   - We then attempt to redefine the task
   - [Rake merges the prerequisites together](https://github.com/ruby/rake/blob/4fe73ff6b19ea1d8490b0442c75fc3a53815c4cf/lib/rake/task.rb#L116)
   - [Ruby set union appears to append new elements](https://ruby-doc.org/core-2.7.0/Array.html#method-i-7C)

I've raised this, as it doesn't particularly block this PR: https://github.com/alphagov/content-publisher/pull/1951

ChrisBAshton: Thanks for the explanation 👏

PR #1931, opened 2020-03-20
Allow users to add a unique reference to existing attachments

leenagupte:
Trello: https://trello.com/c/QZmkUoTr

# What's changed?

Adds an edit page that allows users to add a unique reference to existing attachments. The work to allow users to add a unique reference to new attachments will follow in the next PR.

This PR focuses on setting up the architecture of the edit details page (controller actions, route, validations and tests) and should hopefully unblock the work to add the ISBN and Command Paper details.

# Expected changes

Screenshot 2020-03-20 at 09 36 33
Screenshot 2020-03-20 at 09 36 48
Screenshot 2020-03-19 at 12 15 56

### DIFF
--- a/app/helpers/file_attachment_helper.rb
+++ b/app/helpers/file_attachment_helper.rb
@@ -14,6 +14,7 @@ module FileAttachmentHelper
       file_size: attachment_revision.byte_size,
       number_of_pages: attachment_revision.number_of_pages,
       url: preview_file_attachment_path(document, attachment_revision.file_attachment),
+      unique_reference: attachment_revision.unique_reference,
     }.compact
   end
 end
--- a/app/helpers/file_attachment_helper.rb
+++ b/app/helpers/file_attachment_helper.rb
@@ -6,7 +6,7 @@ module FileAttachmentHelper
   end
 
   def file_attachment_attributes(attachment_revision, edition)
-    {
+    attributes = {
       id: attachment_revision.filename,
       title: attachment_revision.title,
       filename: attachment_revision.filename,
@@ -14,7 +14,14 @@ module FileAttachmentHelper
       file_size: attachment_revision.byte_size,
       number_of_pages: attachment_revision.number_of_pages,
       url: preview_file_attachment_path(edition.document, attachment_revision.file_attachment),
-      unique_reference: attachment_revision.unique_reference,
-    }.compact
+    }
+
+    if edition.document_type.attachments.featured?
+      attributes.merge!(
+        unique_reference: attachment_revision.unique_reference,
+      )
+    end
+
+    attributes.compact
   end
 end
--- a/app/helpers/file_attachment_helper.rb
+++ b/app/helpers/file_attachment_helper.rb
@@ -5,7 +5,7 @@ module FileAttachmentHelper
     attachment_revision.asset_url + "?" + params
   end
 
-  def file_attachment_attributes(attachment_revision, document)
+  def file_attachment_attributes(attachment_revision, edition)
     {
       id: attachment_revision.filename,
       title: attachment_revision.title,
@@ -13,7 +13,7 @@ module FileAttachmentHelper
       content_type: attachment_revision.content_type,
       file_size: attachment_revision.byte_size,
       number_of_pages: attachment_revision.number_of_pages,
-      url: preview_file_attachment_path(document, attachment_revision.file_attachment),
+      url: preview_file_attachment_path(edition.document, attachment_revision.file_attachment),
       unique_reference: attachment_revision.unique_reference,
     }.compact
   end
--- a/app/interactors/file_attachments/update_interactor.rb
+++ b/app/interactors/file_attachments/update_interactor.rb
@@ -1,0 +1,63 @@
+class FileAttachments::UpdateInteractor < ApplicationInteractor
+  delegate :params,
+           :user,
+           :edition,
+           :file_attachment_revision,
+           to: :context
+
+  def call
+    Edition.transaction do
+      find_and_lock_edition
+      find_file_attachment
+
+      update_file_attachment
+      update_edition
+
+      create_timeline_entry
+      update_preview
+    end
+  end
+
+private
+
+  def find_and_lock_edition
+    context.edition = Edition.lock.find_current(document: params[:document])
+    assert_edition_state(edition, &:editable?)
+  end
+
+  def find_file_attachment
+    context.file_attachment_revision = edition.file_attachment_revisions
+                                              .find_by!(file_attachment_id: params[:file_attachment_id])
+  end
+
+  def update_file_attachment
+    updater = Versioning::FileAttachmentRevisionUpdater.new(file_attachment_revision, user)
+    revision_attributes = attachment_params.slice(:unique_reference)
+    updater.assign(revision_attributes)
+
+    context.file_attachment_revision = updater.next_revision
+  end
+
+  def update_edition
+    updater = Versioning::RevisionUpdater.new(edition.revision, user)
+    updater.update_file_attachment(file_attachment_revision)
+
+    context.fail!(unchanged: true) unless updater.changed?
+
+    EditDraftEditionService.call(edition, user, revision: updater.next_revision)
+    edition.save!
+  end
+
+  def create_timeline_entry
+    TimelineEntry.create_for_revision(entry_type: :file_attachment_updated,
+                                      edition: edition)
+  end
+
+  def update_preview
+    FailsafeDraftPreviewService.call(edition)
+  end
+
+  def attachment_params
+    params.require(:file_attachment).permit(:unique_reference)
+  end
+end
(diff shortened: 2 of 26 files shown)


lib/requirements/file_attachment_metadata_checker.rb
```diff
@@ -0,0 +1,30 @@
+module Requirements
+  class FileAttachmentMetadataChecker
+    UNIQUE_REF_MAX_LENGTH = 255
+
+    attr_reader :unique_reference
+
+    def initialize(unique_reference: nil)
```
benthorner: We usually pass the revision and/or all the params to the checker, and let it decide what data it needs. This makes it easier to add more checks, since we don't need to alter any parameters. The [image checker](https://github.com/alphagov/content-publisher/blob/master/lib/requirements/image_revision_checker.rb) is a good example of how we've done it for something similar, could we follow that approach here?

spec/features/editing_file_attachments/edit_file_attachment_file_spec.rb
```diff
@@ -0,0 +1,35 @@
+RSpec.feature "Edit a file attachment file", js: true do
+  scenario do
+    given_there_is_an_edition_with_featured_attachments
+    when_i_go_to_edit_an_attachment
+    and_i_add_a_unique_reference
+    then_i_see_the_timeline_entry
```
benthorner: Can we add a step here to check we can see the new metadata? We should [now have the component ready](https://trello.com/c/UET4vx6k/1439-update-attachment-component-to-accommodate-new-attachment-design) to do this.

spec/features/editing_file_attachments/edit_file_attachment_file_spec.rb
```diff
@@ -3,7 +3,8 @@
     given_there_is_an_edition_with_featured_attachments
     when_i_go_to_edit_an_attachment
     and_i_add_a_unique_reference
```
benthorner: Minor: we could call this "attachment metadata" to save someone changing it later.

spec/factories/file_attachment_metadata_revision_factory.rb
```diff
@@ -2,5 +2,6 @@
   factory :file_attachment_metadata_revision, class: "FileAttachment::MetadataRevision" do
     association :created_by, factory: :user
     title { SecureRandom.hex(8) }
+    unique_reference { SecureRandom.alphanumeric(10) }
```
benthorner: Minor: it's a bit strange having a non-required attribute in the factory, and could lead to subtle bugs where it's presence is assumed somewhere.

app/views/file_attachments/edit.html.erb
```diff
@@ -0,0 +1,40 @@
+<% content_for :title, t("file_attachments.edit.update.title", title: @edition.title_or_fallback) %>
```
benthorner: Looks like we have a stray key here - should this just be edit.title?

app/helpers/file_attachment_helper.rb
```diff
@@ -14,6 +14,7 @@ def file_attachment_attributes(attachment_revision, document)
       file_size: attachment_revision.byte_size,
       number_of_pages: attachment_revision.number_of_pages,
       url: preview_file_attachment_path(document, attachment_revision.file_attachment),
+      unique_reference: attachment_revision.unique_reference,
```
kevindew: I believe we'll only want to do this for documents with featured attachments as this is a field only allowed in publication assets: https://github.com/alphagov/govuk-content-schemas/blob/ca3f442da137444b246111fb4b3579835cbd2ff7/formats/shared/definitions/asset_links.jsonnet#L32-L40

app/interactors/file_attachments/update_interactor.rb
```diff
@@ -0,0 +1,63 @@
...
+    updater.update_file_attachment(file_attachment_revision)
+
+    context.fail!(unchanged: true) unless updater.changed?
+
+    EditDraftEditionService.call(edition, user, revision: updater.next_revision)
+    edition.save!
+  end
+
+  def create_timeline_entry
+    TimelineEntry.create_for_revision(entry_type: :file_attachment_updated,
```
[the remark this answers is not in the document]
benthorner: Turns out this is actually consistent with [the way we do timeline entries for images](https://github.com/alphagov/content-publisher/blob/master/app/interactors/images/update_crop_interactor.rb#L50). I suppose this makes sense because all this stuff is about updating the attachment, and we could have  it all on one form (as a counter-argument to having different types).

spec/features/editing_file_attachments/edit_file_attachment_file_spec.rb
benthorner: Also spotted the name of this file is inaccurate - we should change it to something like `edit_file_attachment_spec`.

app/views/file_attachments/edit.html.erb
```diff
@@ -0,0 +1,40 @@
...
+<% content_for :back_link, render_back_link(href: featured_attachments_path(@edition.document),
+                                            data_attributes: { "modal-action": "back" }) %>
```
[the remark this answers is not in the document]
benthorner: Updated.

app/views/file_attachments/edit.html.erb
```diff
@@ -0,0 +1,40 @@
...
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <%= form_tag(
+      edit_file_attachment_path(@edition.document, @attachment.file_attachment_id),
+      method: :patch,
+      multipart: true,
+      data: {
+        "modal-action": "update",
```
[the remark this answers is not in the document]
benthorner: Updated.

spec/features/editing_file_attachments/edit_file_attachment_file_spec.rb
benthorner: Updated.

spec/factories/file_attachment_metadata_revision_factory.rb
benthorner: Updated.

app/helpers/file_attachment_helper.rb
benthorner: Updated.

app/views/file_attachments/edit.html.erb
benthorner: Updated.

spec/features/editing_file_attachments/edit_file_attachment_file_spec.rb
benthorner: Updated.

lib/requirements/file_attachment_metadata_checker.rb
benthorner: I'm inclined to leave this for now, as it's not necessary to change it here, and the need for it to change will become clear in future work. For now, it's consistent with the checker for file attachments ([in future](https://trello.com/c/3E2vqBkS/1507-support-official-documents-for-featured-file-attachments), we'll be renaming this to `FileAttachmentUploadChecker`, as part of supporting issues for file attachments on the document summary).

benthorner:
@leenagupte as you're busy working on other things, I've made most of my proposed changes as extra commits on this PR. Could you take a look and let me know if you agree with the changes?

spec/requests/file_attachments_spec.rb
```diff
@@ -279,22 +279,38 @@
...
       patch edit_file_attachment_path(edition.document,
                                       file_attachment_revision.file_attachment_id),
             params: { file_attachment: { unique_reference: "Uniq Ref" } }
 
       expect(response).to redirect_to(featured_attachments_path(edition.document))
     end
+
+    it "returns issues and an unprocessable response when there are requirement issues" do
+      max_length = Requirements::FileAttachmentMetadataChecker::UNIQUE_REF_MAX_LENGTH
+      too_long_unique_reference = "a" * (max_length + 1)
```
ChrisBAshton: Minor, but in some tests we're using `z` and in others we're using `a`. Would be better to stick to one!

spec/features/editing_file_attachments/edit_file_attachment_file_spec.rb
```diff
@@ -0,0 +1,35 @@
+RSpec.feature "Edit a file attachment file", js: true do
```
ChrisBAshton: JS isn't strictly needed for this feature, is it? Omitting `js: true` should make it run faster.

benthorner: Ah, good spot! I've added a commit to remove it.

spec/requests/file_attachments_spec.rb
benthorner: Hmm...this didn't cause any pain for me, really, so I'm going to leave as-is. We can always come back to it if it becomes a source of confusion.

PR #1933, opened 2020-03-20
ADR to record publishing times and change note decisions

kevindew:
Trello: https://trello.com/c/GrVzLvnd/1468-document-the-change-note-decisions-in-content-publisher-adr-and-publishing-api-documentation

### DIFF
--- a/docs/adr/0016-publishing-times-and-change-history.md
+++ b/docs/adr/0016-publishing-times-and-change-history.md
@@ -1,0 +1,210 @@
+# 16. Publishing times and change history
+
+Date: 2020-03-20
+
+## Context
+
+Content that is published on GOV.UK has metadata to indicate the history of
+the content. This is comprised of a time to indicate when the first iteration of
+that content was published (`first_published_at`), a time for when the content
+was last changed in a meaningful way (`public_updated_at`) and a
+change history which records a change note and the publishing time of each of
+these meaningful changes. By convention, the first item in a change history of
+GOV.UK content is provided by the publishing application without publisher
+input. This item has a change note of "First published." and has the time the
+content was initially published on GOV.UK.
+
+To set these metadata fields Content Publisher has relied on the Publishing API
+automatically setting the time fields and using it's
+[ChangeNote][pub-api-change-note] model. This model stores a record for each
+major change published on GOV.UK and can collate them to build a change
+history. This meant that Content Publisher has not needed to store
+`first_published_at`, `public_updated_at` or a full change history and has only
+needed to store a change note for an edition.
+
+We had concerns that continued usage of the Publishing API time and change
+history system would lead to discrepancies when migrating content from Whitehall
+to Content Publisher. This led to us doing a review of the system and problems,
+where some relate to migration and others are broader.
+
+### Problems identified
+
+#### 1. Having accurate times for Whitehall migrated content
+
+When content is migrated from Whitehall to Content Publisher we want all the
+times associated with that content to remain the same. Whitehall complicates
+this by passing two time values to represent the first publishing time: the
+aforementioned `first_published_at` field and a deprecated time value,
+`first_public_at`, which is used [preferentially to
+`first_published_at`][first-public-at] when rendering the content. Content
+Publisher does not populate the `first_public_at` field.
+
+
+Some content items, notably older ones that pre-date Whitehall sending a
+`first_published_at` value, can have different values for
+each of these times - A consequence of these not matching is that migrated
+content would show a different first publishing time.
+
+#### 2. Migrating change notes from Whitehall
+
+Whitehall doesn't make use of the [Publishing API ChangeNote][pub-api-change-note]
+model to build a change history, instead Whitehall creates it's own change
+history for content and includes it when updating content with the Publishing
+API. This has a consequence that the data stored in ChangeNote models in
+Publishing API for Whitehall content cannot be relied upon as there are no
+guarantees of its accuracy. There is also not an expectation that this data is
+accurate as many pieces of Whitehall content were created before the
+ChangeNote model was introduced to Publishing API.
+
+For Content Publisher to migrate change histories for Whitehall content it
+would need to ensure the entire change history is in sync with the Publishing
+API.
+
+#### 3. Removing/editing change notes
+
+It is a relatively common task that GOV.UK 2nd line support engineers are
+[requested to remove or modify a change note][change-note-docs] for a published
+piece of content. To resolve this in Whitehall a developer would modify a past
+edition of a piece of content to either remove or amend the change note.
+
+In Content Publisher it is not ideal to modify a past edition. An
+[edition][adr9-core-concepts] represents a snapshot of the content as it was
+published on GOV.UK and any edits invalidate the accuracy of this snapshot.
+
+#### 4. Backdating content
+
+The backdating feature in Content Publisher supports only backdating the first
+edition of a piece of content and not subsequent ones as is allowed by
+Whitehall. This is because the Publishing API does not have the means to change
+the initial item in the change history for content and thus would cause a
+change history that is inconsistent with the `first_published_at` time. This
+has a consequence for publishers that if they make a mistake with backdating
+or forget to backdate in Content Publisher then they cannot rectify this
+without creating a new document or requesting support from developers.
+
+### Approaches considered
+
+#### 1. Correct Whitehall data in the Publishing API
+
+We considered doing an audit and syncing exercise for times and change
+histories from Whitehall to the Publishing API. This would involve either
+doing a large data modification exercise or creating new endpoints
+in the Publishing API that allow importing this data.
+
+The positive of this approach was that it could mean Content Publisher
+could continue to rely on the Publishing API for time and change history
+management without needing this data stored in Content Publisher.
+
+The negatives to this were that there would be no guarantees that this data
+would remain accurate once the work was complete, as Whitehall continues to
+operate; nor would it be easy to check whether data is still
+accurate during migration as Content Publisher does not store it. This also
+didn't allow any improvements in removing/editing change history or allowing
+backdating past the first edition.
+
+#### 2. Store times and change history in Content Publisher and sync them to Publishing API
+
+We considered changing Content Publisher to be similar in approach to
+Whitehall and be the canonical source of publishing times and change history.
+Each time Content Publisher content is updated the Publishing API would be
+sent the times and change history.
+
+This approach offers a greater degree of control over data, which allows
+Content Publisher to enhance the backdating feature and provide more support for
+change history removal/editing. This also offers a higher degree of
+compatibility with Whitehall for migrating content due to it being a similar
+model.
+
+This negative is that it increases Content Publisher's responsibilities
+to storing time and change history. It also means that prior to publishing
+content Content Publisher would need to update timestamps in the Publishing API
+which adds an additional API call.
+
+#### 3. Adapt Publishing API to support an altering change history
+
+We considered modifying the Publishing API to allow the altering of change
+histories. This would involve change notes having a particular id value that
+could be used to identify them in requests.
+
+As a positive, this allows each change note to be treated as a separate concept
+to content and, compared to other approaches, allows a reduction in duplicate
+data shared between different editions of content.
+
+A negative of this is that it requires additional development and complexity
+for the Publishing API with a potential migration needed for existing content.
+It also presented a number of complications over whether change history would
+be part of a draft/live workflow, since content already has this concern.
+
+## Decision
+
+We decided to take the approach of storing times and change history in Content
+Publisher.
+
+We felt that this was the most pragmatic approach as it didn't require
+additional development to the Publishing API and was the most compatible with
+Whitehall's approach to change history. The increased control this offered
+was also consistent with a potential future feature of allowing change history
+to be modified via a user interface.
+
+We concluded that the recommended approach for change history in the Publishing
+API was ultimately flawed and should be reconsidered as a recommendation. Our
+concerns were:
+
+- all publishing applications, with the exception of Specialist Publisher,
+  use the Publishing API as a syncing target rather than a store of publishing
+  history, which is a reversal from the unrealised intention that Publishing
+  API is the canonical store for content and history;
+- change history is the lone outlier as user visible data on GOV.UK that a
+  publishing application cannot fix in case of any issues;
+- the Publishing API ChangeNote model was designed to be append-only and
+  is therefore incompatible with the ability to change the time that content
+  is marked as first published (as the change history will disagree with this
+  value) or to resolve any mistakes.
+
+## Status
+
+Accepted
+
+## Consequences
+
+Content Publisher will store all the information needed to present
+`first_published_at`, `public_updated_at` and change history to the Publishing
+API. In doing so Content Publisher will be considered the canonical source
+of this data. This data will be reflected by non user editable data of
+`first_published_at` on a Document model and `published_at` on an Edition model,
+which reflects the time that Content Publisher published the content. In
+addition to the existing `change_note` field on a MetadataRevision there
+will also be a `change_history` field that stores a JSON representation of all
+the previous change history entries for the content.
+
+This `change_history` field will be used to store a collection of timestamps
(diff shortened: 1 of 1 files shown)


docs/adr/0016-publishing-times-and-change-history.md
```diff
@@ -0,0 +1,214 @@
...
+
+This offered a negative that it increased Content Publisher's responsibilities
+to storing time and change history. It also means that prior to publishing
+content the Publishing API would need times updating which adds an additional
+API call during publishing.
+
+### Adapt Publishing API to support an altering change history
+
+We considered modifying the Publishing API so that it could provide
+an API means to allow the altering of change histories. This would involve
```
gclssvglx: Maybe `... could provide an API to allow ...`.  `means` doesn't really add anything here.

docs/adr/0016-publishing-times-and-change-history.md
```diff
@@ -0,0 +1,214 @@
...
+change notes having a particular id value that could be used to identify them
+in requests.
+
+This offered a positive that this allows change history to be treated as a
```
gclssvglx: Maybe `As a positive, this allows each change history to be treated as a...`

docs/adr/0016-publishing-times-and-change-history.md
```diff
@@ -0,0 +1,214 @@
...
+## Decision
+
+We decided to take the approach of storing times and change history in Content
+Publisher.
+
+We felt that this was the most pragmatic approach as it didn't require
+additional development to the Publishing API and was the most compatible with
+Whitehall's approach to change history. The increased control this offered
+was also consistent with a potential future feature of allowing change history
+to be modifiable by a user interface.
```
gclssvglx: `modified via` maybe?

docs/adr/0016-publishing-times-and-change-history.md
```diff
@@ -0,0 +1,214 @@
...
+
+We concluded that the recommended approach for change history in the Publishing
+API was ultimately flawed and should be reconsidered as a recommendation. Our
+concerns were:
+
+- many publishing applications tend to sync their data to Publishing API with
+  no regard to maintaining a consistent history, which is a reversal from the
```
gclssvglx: `for maintaining` maybe?

docs/adr/0016-publishing-times-and-change-history.md
```diff
@@ -0,0 +1,214 @@
...
+  intention that Publishing API stores the history (an idea that was never
+  fully realised);
+- change history is the lone outlier as user visible data on GOV.UK that a
+  publishing application cannot fix in case of any issues;
+- the Publishing API change note system is incompatible with the ability to
+  change the time content is marked as first published (as the change history
```
gclssvglx: `...time that content...` maybe?

docs/adr/0016-publishing-times-and-change-history.md
```diff
@@ -0,0 +1,214 @@
...
+  will disagree with this value);
+- the Publishing API is ill suited to provide mechanisms to rectify any issues
```
gclssvglx: `Ill suited` should be hyphenated.

docs/adr/0016-publishing-times-and-change-history.md
```diff
@@ -0,0 +1,214 @@
...
+Accepted
+
+## Consequences
+
+Content Publisher will store all the information needed to present
+`first_published_at`, `public_updated_at` and change history to the Publishing
+API. In doing so Content Publisher will be considered the canonical source
+of this data. This data will be reflected by non user editable data of
+`first_published_at` on a Document model and `published_at` on an Edition model,
+which reflect the times Content Publisher first sent content to the Publishing
```
gclssvglx: `... the times that Content Publisher first sent...` maybe?

docs/adr/0016-publishing-times-and-change-history.md
```diff
@@ -0,0 +1,214 @@
...
+API. In addition to the existing `change_note` field on a MetadataRevision
+there will also be a `change_history` field that stores a JSON representation
+of all the previous change history for the content.
```
gclssvglx: `...all previous change history entries for the content.` maybe?

docs/adr/0016-publishing-times-and-change-history.md
```diff
@@ -0,0 +1,214 @@
...
+
+This `change_history` field will be used to store a collection of timestamps
+with notes. These will not include the fixed "First Published" change note as
+this will automatically appended to the start of a change history with a time
+based on the `first_published_at` value for a document (which can be overriden
+by a contents backdating) and therefore does not have a fixed time. It also
```
gclssvglx: Is `contents` possessive here?

docs/adr/0016-publishing-times-and-change-history.md
```diff
@@ -0,0 +1,214 @@
...
+does not include the current change note of an edition because this not become
+an aspect of change history until the content is published and the update type
+and publishing time are fixed. It was noted that having both `change_note`
+and `change_history` fields risks confusion for developers, however this
+seemed preferable to complicating the `change_history` field with additional
+responsibilities.
+
+Initially the only means to edit items in a change history will be via a rake
+task that developers can perform. This will provide means to add, edit and
```
gclssvglx: `... provide a means to ...` maybe?

docs/adr/0016-publishing-times-and-change-history.md
```diff
@@ -0,0 +1,214 @@
...
+delete items. Eventually we may choose to build the means to edit this into the
+Web UI however at this point in time it is unclear which users would be granted
```
gclssvglx: Feels like the end of a sentence `... Web UI.  However, at this point...`

docs/adr/0016-publishing-times-and-change-history.md
```diff
@@ -0,0 +1,214 @@
...
+the permissions to do this given the risks this editing can have on the
+integrity of GOV.UK content.
+
+We will expand the [IntegrityChecker][] system used by Content Publisher to
+verify that times and change history sent to the Publishing API matches what
```
gclssvglx: `... verify that the times and change history that is sent to the Publishing API matches what is already stored for Whitehall content.` maybe?

docs/adr/0016-publishing-times-and-change-history.md
```diff
@@ -0,0 +1,214 @@
...
+is already stored for Whitehall content. This will be used to ensure that
+we have a confidence these will remain consistent through a migration.
```
gclssvglx: `This will ensure that these remain consistent during migration.` maybe?  Sounds like a much more positive outcome and statement.

docs/adr/0016-publishing-times-and-change-history.md
```diff
@@ -0,0 +1,214 @@
...
+
+We will remove the deprecation of the `change_history` attribute in Publishing
+API. This is to reflect that this field is still in operation and that the
+recommended alternative approach, of using the `change_note` field, does not
+sufficiently meet enough needs to be a full replacement.
```
1pretz1: This sounds to me like we're still using the `change_note` field from this sentence, would it be more fitting to drop the 'full' so '... does not sufficiently meet enough needs to be a replacement.'

docs/adr/0016-publishing-times-and-change-history.md
```diff
@@ -0,0 +1,214 @@
...
+integrity of GOV.UK content.
+
+We will expand the [IntegrityChecker][] system used by Content Publisher to
+verify that times and change history sent to the Publishing API matches what
+is already stored for Whitehall content. This will be used to ensure that
+we have a confidence these will remain consistent through a migration.
+
+We will remove the deprecation of the `change_history` attribute in Publishing
+API. This is to reflect that this field is still in operation and that the
+recommended alternative approach, of using the `change_note` field, does not
```
1pretz1: This is a wider question but is something that occurred to me when reading this - If we're removing the deprecation from change_history, should we be adding one to the change_note field? It feels like if we've decided that change_history offers a lot more then this should be the preferred approach to managing change notes.

docs/adr/0016-publishing-times-and-change-history.md
```diff
@@ -0,0 +1,208 @@
...
+
+When content is migrated from Whitehall to Content Publisher we want all the
+times associated with that content to remain the same. Whitehall complicates
+this by passing two time values to represent the first publishing time: the
+aforementioned `first_published_at` field and a deprecated time value,
+`first_public_at`, which is used [preferentially to
+`first_published_at`][first-public-at] when rendering the content. Content
+Publisher does not populate the `first_public_at` field.
+
+An occasional problem is that the Publishing API can have different values for
```
benthorner, line 42: This could do with some more explanation, as at first it conflicts with the previous statement that Whitehall is "passing two time values to represent the first publishing time", which suggests they should always be the same. I'm guessing there's a way they can differ when they're set by Whitehall, and that's the problem?

docs/adr/0016-publishing-times-and-change-history.md
```diff
@@ -0,0 +1,208 @@
...
+API.
+
+#### 3. Removing/editing change notes
+
+It is a relatively common task that GOV.UK 2nd line support engineers are
+[requested to remove or modify a change note][change-note-docs] for a published
+piece of content. To resolve this in Whitehall a developer would modify a past
+edition of a piece of content to either remove or amend the change note.
+
+In Content Publisher it is not ideal to modify a past edition. An edition
```
benthorner, line 68: Minor: we could link to the modelling ADR to provide clear support for this view.

docs/adr/0016-publishing-times-and-change-history.md
```diff
@@ -0,0 +1,208 @@
...
+doing a large data modification exercise or creating new endpoints
+in the Publishing API that allow importing this data.
+
+The positive of this approach was that it could mean Content Publisher
+could continue to rely on the Publishing API for time and change history
+management without needing this data stored in Content Publisher.
+
+The negatives to this were that there would be no guarantees that this data
+would remain accurate once the work was complete, as Whitehall continues to
+operate; nor would there be an easy way to check whether times are still
```
benthorner, line 98: Minor: I'm not really sure what we mean by 'easy', since we seem to be calling the API for data integrity checks already. The latter argument sways it for me anyway, so I'm not worried if it's too complicated to explain here.

docs/adr/0016-publishing-times-and-change-history.md
```diff
@@ -0,0 +1,208 @@
...
+Whitehall's approach to change history. The increased control this offered
+was also consistent with a potential future feature of allowing change history
+to be modified via a user interface.
+
+We concluded that the recommended approach for change history in the Publishing
+API was ultimately flawed and should be reconsidered as a recommendation. Our
+concerns were:
+
+- all publishing applications, with the exception of Specialist Publisher,
+  sync their data to Publishing API with little regard to maintaining a
```
benthorner, line 152: What do we mean by 'little regard' here?

docs/adr/0016-publishing-times-and-change-history.md
```diff
@@ -0,0 +1,208 @@
...
+Accepted
+
+## Consequences
+
+Content Publisher will store all the information needed to present
+`first_published_at`, `public_updated_at` and change history to the Publishing
+API. In doing so Content Publisher will be considered the canonical source
+of this data. This data will be reflected by non user editable data of
+`first_published_at` on a Document model and `published_at` on an Edition model,
+which reflects the times that Content Publisher first sent content to the
```
benthorner, line 174: Would this actually be "the times that Content Publisher published the content"?

docs/adr/0016-publishing-times-and-change-history.md
1pretz1: Yup my comment is addressed 👍

docs/adr/0016-publishing-times-and-change-history.md
1pretz1: Okay sure that makes sense, feels like we should maybe add something somewhere in the publishing-api to advise future devs looking into this topic that change_history is the field that should ideally be used.

My original thought process was that a deprecation warning on the change_note field would give that impression but maybe there is some other way to convey that intent, what do you think?

kevindew: I don't think it's necessarily ideal though. If you're doing basic publishing the change_note field is simpler to use and a good choice. If you want to support advanced features like backdating or removing a change note you have to use change_history with the added burden of date management.

docs/adr/0016-publishing-times-and-change-history.md
kevindew: Sure, I've updated it a bit. I think the reason for the disparity is mostly wrapped in historical issues (Whitehall didn't used to populate one, now does; Publishing API had tons of date problems) and then lingering issues of there being two fields there is scope for them not to match.

docs/adr/0016-publishing-times-and-change-history.md
kevindew: Added in.

docs/adr/0016-publishing-times-and-change-history.md
kevindew: I've made an amend. 

The reason it's not easy is that the Content Publisher edition wouldn't have the data stored so to integrity check it you'd need to compare the data Whitehall has against the Publishing API. We could of course access that from our Whitehall dump of data but that would also be complex as that is not the subject in use at the point of integrity checking.

docs/adr/0016-publishing-times-and-change-history.md
kevindew: I've made an amend. 

To provide more context: no apps attempt to achieve a parity in what the Publishing API stores in terms of number of editions with what publishers have actually created with all using various means of republishing to achieve syncs.

docs/adr/0016-publishing-times-and-change-history.md
kevindew: Thanks

kevindew:
@benthorner this should be ready for another look.

docs/adr/0016-publishing-times-and-change-history.md
```diff
@@ -0,0 +1,210 @@
...
+
+#### 3. Removing/editing change notes
+
+It is a relatively common task that GOV.UK 2nd line support engineers are
+[requested to remove or modify a change note][change-note-docs] for a published
+piece of content. To resolve this in Whitehall a developer would modify a past
+edition of a piece of content to either remove or amend the change note.
+
+In Content Publisher it is not ideal to modify a past edition. An
+[edition][adr-9-core-concepts] represents a snapshot of the content as it was
```
benthorner, line 71:
```suggestion
[edition][adr9-core-concepts] represents a snapshot of the content as it was
```

docs/adr/0016-publishing-times-and-change-history.md
```diff
@@ -0,0 +1,210 @@
...
+Accepted
+
+## Consequences
+
+Content Publisher will store all the information needed to present
+`first_published_at`, `public_updated_at` and change history to the Publishing
+API. In doing so Content Publisher will be considered the canonical source
+of this data. This data will be reflected by non user editable data of
+`first_published_at` on a Document model and `published_at` on an Edition model,
+which reflects the times that Content Publisher published the content. In
```
benthorner, line 175:
```suggestion
which reflects the time that Content Publisher published the content. In
```

docs/adr/0016-publishing-times-and-change-history.md
```diff
@@ -0,0 +1,210 @@
...
+addition to the existing `change_note` field on a MetadataRevision there
+will also be a `change_history` field that stores a JSON representation of all
+the previous change history entries for the content.
+
+This `change_history` field will be used to store a collection of timestamps
+with notes. These will not include the fixed "First published." change note as
+this will automatically appended to the start of a change history with a time
```
benthorner, line 182:
```suggestion
this will automatically be appended to the start of a change history with a time
```

docs/adr/0016-publishing-times-and-change-history.md
kevindew: Thanks, nice catch.

kevindew:
Thanks @benthorner

PR #1940, opened 2020-03-24
Move non-task into main lib/ directory

benthorner:
https://trello.com/c/BhRV4Sqy/1679-increase-our-i18n-coverage-to-100

While adding a new rake task it confused me to find a non-task file in
this directory. This moves the file into the main 'lib/' directory,
noting that we already have a precedent for this (whitehall import).


1pretz1:
Can you not merge this BTW for like 30 mins, testing on integration

PR #1941, opened 2020-03-24
Check edition state in integrity check

1pretz1:
We now want to check the editions state in the integrity checker to verify it's represented correctly between content-publisher and publishing-api. The states don't always have a 1 to 1 matching as content-publisher can have multiple internal states which are equivalent to a publishing-api state. E.g content-publisher's published_but_needs_2i and published map to publishing-api's published state.

Trello:
https://trello.com/c/nuK7ICOU/1591-integrity-check-for-edition-state

### DIFF
--- a/lib/whitehall_importer/integrity_checker.rb
+++ b/lib/whitehall_importer/integrity_checker.rb
@@ -129,15 +129,7 @@ module WhitehallImporter
     def unpublishing_problems
       problems = []
       return problems unless edition.withdrawn? || edition.removed?
-
-      unless expected_state?
-        problems << problem_description(
-          "edition state isn't as expected",
-          "unpublished",
-          publishing_api_content.dig("publication_state"),
-        )
-        return problems
-      end
+      return problems << "publishing_api_unpublishing is missing" if publishing_api_unpublishing.blank?
 
       check = UnpublishingCheck.new(edition, publishing_api_unpublishing)
 
@@ -219,11 +211,6 @@ module WhitehallImporter
       attribute == "alt_text" &&
         proposed_image_payload.empty? &&
         publishing_api_image[attribute] == "placeholder"
-    end
-
-    def expected_state?
-      %w[withdrawn removed].include?(edition.state) &&
-        publishing_api_content.dig("publication_state") == "unpublished"
     end
 
     def publishing_api_unpublishing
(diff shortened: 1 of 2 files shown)


lib/whitehall_importer/integrity_checker.rb
```diff
@@ -86,6 +86,30 @@ def time_matches?(proposed_time, publishing_api_time)
       self.class.time_matches?(proposed_time, publishing_api_time)
     end
 
+    def state_problems
+      problems = []
+
+      unless publishing_api_content["publication_state"] == expected_state
+        problems << problem_description("our state is not equivalent to publication_state",
```
emmabeynon: I think it would be more helpful in the problem description to pass in what we're expecting to see in the Publishing API, rather than what the edition's state is, since states in Content Publisher and Publishing API don't match each other.  Otherwise we could be in a scenario where the description says that we're expecting a state like 'withdrawn' which doesn't exist in the Publising API.  We could do something like:
```suggestion
        problem_description("publication_state isn't as expected", expected_state, publishing_api_content["publication_state"]
```

spec/lib/whitehall_importer/integrity_checker_spec.rb
```diff
@@ -16,9 +16,10 @@
   end
 
   describe "#valid?" do
+    let(:status) { :published }
     let(:state) { :published }
```
emmabeynon: Should this be removed if we're now using `:status`?

1pretz1: Nope as we need to be able to set the exact state for this test - https://github.com/alphagov/content-publisher/pull/1941/files#diff-8372787ef5224cbd5801d35735fc6dd4R115

lib/whitehall_importer/integrity_checker.rb
1pretz1: I used edition state instead of the expected_state as it seemed to give a more informative message on what state the edition in content-publisher is as using the expected state could mean the edition in cp is in multiple states  (draft => scheduled, submitted_for_review, draft).

I suppose it's a tradeoff and it probably does make more sense for it to be publishing-api's expected state rather than edition state. Change incoming.

1pretz1:
Thanks for the review @emmabeynon. I've updated the PR accordingly.

PR #1942, opened 2020-03-24
Fix change_history bug

1pretz1:
This fixes a bug introduced where by an error is thrown if there is no
change_history for the publishing_api_item, this occurs when the edition
is the first draft. The error is a method error when we call length on
the change_history from within ChangeHistoryCheck.

This commit sets the default change_history to an empty array if it's
not present so length now returns 0 instead of a method error.

The offending line is here - https://github.com/alphagov/content-publisher/blob/master/lib/whitehall_importer/integrity_checker/change_history_check.rb#L24

### DIFF
--- a/lib/whitehall_importer/integrity_checker.rb
+++ b/lib/whitehall_importer/integrity_checker.rb
@@ -58,7 +58,7 @@ module WhitehallImporter
 
       problems << "change history doesn't match" unless ChangeHistoryCheck.new(
         proposed_payload.dig("details", "change_history"),
-        publishing_api_content.dig("details", "change_history"),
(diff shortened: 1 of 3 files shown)


emmabeynon:
Can we add a test for this?

1pretz1:
We can, I was trying to avoid adding more tests to the ever growing integrity_checker_spec...

Let me see if I can nicely move this logic into the ChangeHistoryCheck class and then add a test there.

emmabeynon:
@1pretz1 Sure, I think a test would be good to catch this from happening again.  It'll also help us document that it's possible to receive a payload from Publishing API that has no change history.

PR #1948, opened 2020-03-25
Go to new attachment metadata page after upload

benthorner:
https://trello.com/c/QZmkUoTr/1516-allow-users-to-add-the-unique-reference-as-metadata-for-featured-file-attachments

This changes the upload behaviour for a featured attachment, so that we now
redirect to the additional page for the user to enter metadata.

### DIFF
--- a/spec/views/file_attachments/replace.html.erb_spec.rb
+++ b/spec/views/file_attachments/replace.html.erb_spec.rb
@@ -1,0 +1,22 @@
+RSpec.describe "file_attachments/replace.html.erb" do
+  describe "file attachment replace" do
+    it "shows a 'Save' button by default" do
+      assign(:edition, build(:edition))
+      assign(:attachment, create(:file_attachment_revision))
+      render
+      expect(rendered).to have_button("Save")
+    end
+
+    context "when uploading and navigating back" do
+      it "shows a 'Save and continue' button" do
+        assign(:edition, build(:edition))
+        assign(:attachment, create(:file_attachment_revision))
+
+        render(template: "file_attachments/replace",
+               locals: { params: { wizard: "new" } })
+
+        expect(rendered).to have_button("Save and continue")
+      end
+    end
+  end
+end
--- a/spec/views/file_attachments/replace.html.erb_spec.rb
+++ b/spec/views/file_attachments/replace.html.erb_spec.rb
@@ -13,7 +13,7 @@ RSpec.describe "file_attachments/replace.html.erb" do
         assign(:attachment, create(:file_attachment_revision))
 
         render(template: "file_attachments/replace",
-               locals: { params: { wizard: "new" } })
+               locals: { params: { wizard: "featured-attachment-upload" } })
 
         expect(rendered).to have_button("Save and continue")
       end
--- a/spec/views/images/crop.html.erb_spec.rb
+++ b/spec/views/images/crop.html.erb_spec.rb
@@ -1,0 +1,38 @@
+RSpec.describe "images/crop.html.erb" do
+  describe "image crop" do
+    it "shows a 'Save' button with JS" do
+      assign(:image_revision, create(:image_revision))
+      assign(:edition, build(:edition))
+      render
+      expect(rendered).to have_button("Save")
+    end
+
+    it "shows a 'Continue' button without JS" do
+      assign(:image_revision, create(:image_revision))
+      assign(:edition, build(:edition))
+      render
+      expect(rendered).to have_selector(".app-no-js button", text: "Continue")
+    end
+
+    context "when the image has exact dimensions" do
+      it "shows a 'Continue' button instead" do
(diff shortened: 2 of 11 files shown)


spec/views/images/crop.html.erb_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+      render
+      expect(rendered).to have_button("Save")
+    end
+
+    context "when the image has exact dimensions" do
+      it "shows a 'Continue' button instead" do
+        assign(:image_revision, create(:image_revision, filename: "960x640.jpg"))
+        assign(:edition, build(:edition))
+        render
+        expect(rendered).to have_button("Save")
```
ChrisBAshton: Oops ☝️

spec/views/images/crop.html.erb_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+      end
+    end
+
+    context "when uploading a new image" do
+      it "shows a 'Save and continue' button" do
+        assign(:image_revision, create(:image_revision))
+        assign(:edition, build(:edition))
+        controller.extra_params = { wizard: "featured-attachment-upload" }
+        render
+        expect(rendered).to have_button("Save")
```
ChrisBAshton: And here.

spec/views/file_attachments/replace.html.erb_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+      expect(rendered).to have_button("Save")
+    end
+
+    context "when uploading and navigating back" do
+      it "shows a 'Save and continue' button" do
+        assign(:edition, build(:edition))
+        assign(:attachment, create(:file_attachment_revision))
+        controller.extra_params = { wizard: "featured-attachment-upload" }
+        render
+        expect(rendered).to have_button("Save")
```
ChrisBAshton: And here ^

benthorner:
Oh my, thanks for spotting @ChrisBAshton 🏅. I've updated the tests, which turned out to be more complicated than I realised. Can you take another look?

spec/views/images/crop.html.erb_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+      assign(:edition, build(:edition))
+      render
+      expect(rendered).to have_button("Save")
+    end
+
+    it "shows a 'Continue' button without JS" do
+      assign(:image_revision, create(:image_revision))
+      assign(:edition, build(:edition))
+      render
+      expect(rendered).to have_selector('.app-no-js button[text()="Continue"]')
```
ChrisBAshton: I think this would be cleaner as:

```ruby
      expect(rendered).to have_selector(".app-no-js button", "Continue")
```

I don't think I've come across the `text()` syntax before 🤔 is that an RSpec thing, a Capybara thing or a CSS3 thing?

spec/views/images/crop.html.erb_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+        render
+        expect(rendered).to have_selector('.app-js-only button[text()="Continue"]')
+      end
+    end
+
+    context "when uploading a new image" do
+      it "shows a 'Save and continue' button" do
+        assign(:image_revision, create(:image_revision))
+        assign(:edition, build(:edition))
+        render(template: "images/crop", locals: { params: { wizard: "upload" } })
```
ChrisBAshton: I'm not quite sure what the `template: "images/crop"` bit is doing. I'd have expected something like this instead, consistently with the others:

```
assign(:params, wizard: "upload")
```

spec/views/images/crop.html.erb_spec.rb
```diff
@@ -0,0 +1,38 @@
+RSpec.describe "images/crop.html.erb" do
+  describe "image crop" do
+    it "shows a 'Save' button by with JS" do
```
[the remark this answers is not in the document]
benthorner: Oops - fixed 👍

spec/views/images/crop.html.erb_spec.rb
benthorner: Turns out it's an xpath thing - I'm used to that kind of syntax [from govuk_publishing_components](https://github.com/alphagov/govuk_publishing_components/blob/master/spec/components/file_upload_spec.rb#L53), although your suggestion is a bit nicer. Updated.

spec/views/images/crop.html.erb_spec.rb
benthorner: So the `template` part is necessary in order to override the default agruments of `render` (otherwise you'll get an error, so I'm going to say it's self explaining). We can't use `assign` here, as otherwise we'd need to be using `@params` in the view.

ChrisBAshton: Makes sense. Thanks for explaining!

PR #1949, opened 2020-03-25
Don't compare attachment file sizes

gclssvglx:
We need to exclude the file size of the document from both inline and linked inline attachments (those with an anchor tag associated) when comparing the output from Publishing API with that of the proposed Content Publisher output. The proposed will have a space between the size and
the size type, (i.e. 119 KB) and the output on Publishing API will not (i.e. 119KB).

First we need to find all possible attachment HTML blocks, their classes and markup for inline and linked inline attachments on Content Publisher and Publishing API. From there we can create appropriate tests and remove them from the comparison process.

Trello: https://trello.com/c/j3kBctLj/1543-dont-compare-inline-attachment-file-sizes-cma-dwp-dhsc

### DIFF
--- a/spec/lib/whitehall_importer/integrity_checker/body_text_check_spec.rb
+++ b/spec/lib/whitehall_importer/integrity_checker/body_text_check_spec.rb
@@ -15,25 +15,52 @@ RSpec.describe WhitehallImporter::IntegrityChecker::BodyTextCheck do
       expect(integrity_check.sufficiently_similar?).to be true
     end
 
-    it "returns true even if there is a mismatch in inline atttachment URL filesize" do
+    it "returns true even if there is a mismatch in an attachment link filesize" do
       proposed_body = %(
-        <p>
-          <span class="gem-c-attachment-link">
-            <a class="govuk-link" href="filename.pdf" target="_blank">Test File</a>
-            (<span class="gem-c-attachment-link__attribute"><abbr title="Portable Document Format" class="gem-c-attachment-link__abbr">PDF</abbr></span>,
-              <span class="gem-c-attachment-link__attribute">391 KB</span>, <span class="gem-c-attachment-link__attribute">9 pages</span>)
-          </span>
+        <span class="gem-c-attachment-link">
+          <a class="govuk-link" href="filename.pdf" target="_blank">Test File</a>
+          (
+            <span class="gem-c-attachment-link__attribute">
+              <abbr title="Portable Document Format" class="gem-c-attachment-link__abbr">PDF</abbr></span>,
+            <span class="gem-c-attachment-link__attribute">391 KB</span>,
+            <span class="gem-c-attachment-link__attribute">9 pages</span>
+          )
+        </span>
+      )
+
+      publishing_api_body = %(
+        <span class="attachment-inline">
+          <a href="/filename.pdf">Test File</a>
+          (
+            <span class="type">PDF</span>,
+            <span class="file-size">391KB</span>,
+            <span class="page-length">9 pages</span>
+          )
+        </span>
+      )
+
+      integrity_check = described_class.new(proposed_body, publishing_api_body)
+      expect(integrity_check.sufficiently_similar?).to be true
+    end
+
+    it "returns true even if there is a mismatch in an attachment filesize" do
+      proposed_body = %(
+        <p class="gem-c-attachment__metadata">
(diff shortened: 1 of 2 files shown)


spec/lib/whitehall_importer/integrity_checker/body_text_check_spec.rb
```diff
@@ -15,27 +15,55 @@
...
+            <span class="page-length">9 pages</span>
+          )
+        </span>
+      ).gsub("\n", "").gsub("  ", "")
+
+      integrity_check = described_class.new(proposed_body, publishing_api_body)
+      expect(integrity_check.sufficiently_similar?).to be true
+    end
+
+    it "returns true even if there is a mismatch in an inline atttachment URL filesize" do
```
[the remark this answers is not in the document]
gclssvglx: Whoops!  Will fix.

spec/lib/whitehall_importer/integrity_checker/body_text_check_spec.rb
```diff
@@ -15,27 +15,55 @@
       expect(integrity_check.sufficiently_similar?).to be true
     end
 
-    it "returns true even if there is a mismatch in inline atttachment URL filesize" do
+    it "returns true even if there is a mismatch in a linked inline atttachment URL filesize" do
```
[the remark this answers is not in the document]
gclssvglx: Whoops!  Will fix.  The only need for `a linked` is to distinguish between the two tests.  Otherwise, apart from the body text's they are the same, but testing one with - and one without - an anchor tag.  Hence, linked.  I'll see if I can come up with something more meaningful.

gclssvglx:
> Thanks for this Graham, the logic seems sound to me. I wanted to check the reasoning behind replacing the line breaks and extra spaces in the test body HTML. Does this make it more similar to what we would send to and get back from Publishing API?

@emmabeynon Thanks for the review.  This is interesting - I added this to as the use of newlines and double spaces wasn't consistent within the tests.  However, I'm getting issues with re-migrating and spaces in odd places too.  So, suspect there is a subtle bug in `body_text_check.rb` that is not handled by Sanitize.clean(text).squish.

spec/lib/whitehall_importer/integrity_checker/body_text_check_spec.rb
```diff
@@ -15,25 +15,52 @@
       expect(integrity_check.sufficiently_similar?).to be true
     end
 
-    it "returns true even if there is a mismatch in inline atttachment URL filesize" do
+    it "returns true even if there is a mismatch in an inline attachment URL filesize - with an anchor tag" do
```
emmabeynon: I don't think we need to specify '- with an anchor tag' here - this type of attachment will always have a link, as per this component: https://components.publishing.service.gov.uk/component-guide/attachment_link

To make a distinction between this and the other style of attachment, we could call this 'attachment links' instead of inline attachments.  It would be consistent with what we call it when inserting an attachment:
Screenshot 2020-03-31 at 11 21 15

spec/lib/whitehall_importer/integrity_checker/body_text_check_spec.rb
```diff
@@ -15,25 +15,52 @@
...
+            <span class="page-length">9 pages</span>
+          )
+        </span>
+      )
+
+      integrity_check = described_class.new(proposed_body, publishing_api_body)
+      expect(integrity_check.sufficiently_similar?).to be true
+    end
+
+    it "returns true even if there is a mismatch in an inline attachment URL filesize - without anchor tag" do
```
emmabeynon: Similarly I think we can call this just "attachment" and don't need to specify "without an anchor tag".  In reality, when this attachment is rendered there will be a link to the attachment.  It's just that in order to solve this issue we only need to focus on the metadata part of the component.  You can see the code for this here: https://github.com/alphagov/govuk_publishing_components/blob/master/app/views/govuk_publishing_components/components/_attachment.html.erb#L35

gclssvglx: Thanks @emmabeynon - that helps.  Have updated both tests accordingly.

spec/lib/whitehall_importer/integrity_checker/body_text_check_spec.rb
gclssvglx: Thanks @emmabeynon - tests updated.

PR #1955, opened 2020-03-26
Support ISBN for featured file attachments

ChrisBAshton:
Trello: https://trello.com/c/vs8MEQKB/1519-support-isbn-for-featured-file-attachments

I wasn't too sure of the error message to use, as it's complicated to explain fully what an ISBN should look like. I'll check the error message with a content designer.

### Error state

Screenshot 2020-03-27 at 16 14 43

### Valid saved input, preview

Screenshot 2020-03-27 at 09 50 30

### DIFF
--- a/lib/requirements/file_attachment_metadata_checker.rb
+++ b/lib/requirements/file_attachment_metadata_checker.rb
@@ -1,18 +1,31 @@
 module Requirements
   class FileAttachmentMetadataChecker
     UNIQUE_REF_MAX_LENGTH = 255
+    ISBN10_REGEX = /^(?:\d[\ -]?){9}[\dX]$/i.freeze
+    ISBN13_REGEX = /^(?:\d[\ -]?){13}$/i.freeze
 
-    attr_reader :unique_reference
+    attr_reader :isbn, :unique_reference
 
-    def initialize(unique_reference: nil)
-      @unique_reference = unique_reference
+    def initialize(params)
+      @isbn = params[:isbn]
+      @unique_reference = params[:unique_reference]
     end
 
     def pre_update_issues
-      unique_reference_issues
+      isbn_issues + unique_reference_issues
     end
 
   private
+
+    def isbn_issues
+      issues = CheckerIssues.new
+
+      unless isbn.blank? || ISBN10_REGEX.match?(isbn) || ISBN13_REGEX.match?(isbn)
+        issues.create(:file_attachment_isbn, :invalid)
+      end
+
+      issues
+    end
 
     def unique_reference_issues
       issues = CheckerIssues.new
--- a/spec/requests/file_attachments_spec.rb
+++ b/spec/requests/file_attachments_spec.rb
@@ -320,7 +320,7 @@ RSpec.describe "File Attachments" do
 
       patch edit_file_attachment_path(edition.document,
                                       file_attachment_revision.file_attachment_id),
-            params: { file_attachment: { unique_reference: "Uniq Ref" } }
+            params: { file_attachment: { isbn: "", unique_reference: "Uniq Ref" } }
 
       expect(response).to redirect_to(featured_attachments_path(edition.document))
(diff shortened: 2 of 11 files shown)


lib/requirements/file_attachment_metadata_checker.rb
```diff
@@ -1,19 +1,32 @@
 module Requirements
   class FileAttachmentMetadataChecker
     UNIQUE_REF_MAX_LENGTH = 255
+    ISBN10_REGEX = /^(?:\d[\ |-]?){9}[\d|X]$/i.freeze
```
benthorner: Apologies this wasn't clear on the card: we're going to drop the `|` separator:

> @kevindew2 we could make our own custom validator using just those two regexes. I've checked, and we can lose the | in the separators.

ChrisBAshton:
Thanks Ben. I believe I deliberately left `|` support in as it's part of the ISBN standard and it doesn't hurt to keep it, it's almost more risk removing it and having to touch the regexes. But also a nice feeling deleting code we don't need, so have done that.

Will hold off on merging until this has had a content review. (If that ends up taking too long I may end up merging anyway and it should be a small thing we can tweak post product review).

benthorner:
Thanks @ChrisBAshton - appreciate that! I suspect @kevindew was coming from a readability perspective: `9788|700|6316|25` might be confusing if it's unusual - judging by the fact that no one has used this capability in Whitehall!

benthorner:
Turns out there's a small problem with this and the previous PR: we're not highlighting any issues. Can you fix it for both fields before this goes in, to save another PR?

screencapture-content-publisher-staging-publishing-service-gov-uk-documents-ad466cbd-7149-48c3-94e7-21ad0e4b862e-en-file-attachments-65-edit-2020-03-27-13_22_49

spec/requests/file_attachments_spec.rb
```diff
@@ -298,18 +298,27 @@
...
-            params: { file_attachment: { unique_reference: too_long_unique_reference } }
+            params: { file_attachment: { isbn: "foo", unique_reference: too_long_unique_reference } }
 
       expect(response).to have_http_status(:unprocessable_entity)
+
+      expect(response.body).to have_content(
+        I18n.t!("requirements.file_attachment_isbn.invalid.form_message"),
+      )
+      expect(response.body)
+        .to have_selector("input[name='file_attachment[isbn]'].govuk-input--error")
```
benthorner: Ooh, this is interesting: we don't actually test for highlighted errors anywhere else in the app. I notice that we also don't check that user input is preserved after such an error.

I was wondering if these checks should be in the request test, and it seems reasonable, because otherwise we're not covering [this line](https://github.com/alphagov/content-publisher/blob/master/app/controllers/file_attachments_controller.rb#L134). However, they could get quite verbose.

Since these checks are inconsistent with all the other tests for unprocessable responses, would you be willing to write a pain card about these (two) issues, in favour of these checks?

ChrisBAshton: Yes, it was difficult to know where to put these tests as there wasn't really a precedent. Are you suggesting keeping the tests I've added here (and indeed adding to them, to test that the incorrect values are retained), then writing a pain card so that similar tests can be applied everywhere else? Or are you suggesting removing the tests from this PR?

benthorner: I think we should remove the tests for now, as opposed to putting more effort into making them complete in this one place. Since none of these behaviours - the additional highlighting, the preservation of values - is critical for users, I'm not too worried if such things continue to slip through for the time being, provided we have a card to address the issue in general. Would you like to write a pain card, if you agree?

ChrisBAshton: Here is the pain card: https://trello.com/c/hccZ9cuf/1715-we-dont-test-highlighted-error

I've now removed the tests 👍

PR #1957, opened 2020-03-27
Remove 'TODO' to change RSpec/InstanceVariable

benthorner:
https://trello.com/c/3B7k3YN1/1680-make-our-linting-config-for-rspec-the-standard-for-govuk

Previously we disabled this cop for feature specs with the intention to
explore using 'let' or 'given' instead of instance variables.

This removes the 'TODO', noting instance variables in our feature tests
are assigned *during the test*, so cannot be replaced by 'let' or
'given'. The documentation for this cop also indicates it's meant to be
specific to instance variables assigned in a 'before' block, which could
be replaced with 'let' or 'given'.

```
# bad
describe MyClass do
  before { @foo = [] }
  it { expect(@foo).to be_empty }
end

# good
describe MyClass do
  let(:foo) { [] }
  it { expect(foo).to be_empty }
end

# necessarily 'bad'
describe MyClass do
  it { <feature_steps> }
  def step_x; ... @foo = input_in_form; ... end
  def step_y; ... expect(result).to include @foo; end
end
```

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -73,7 +73,6 @@ Capybara/FeatureMethods:
     - 'spec/features/**/*.rb'
 
 # Part of the GOV.UK feature spec style involves instance variables
-# TODO: Investigate if we can embrace this rule and use let/given methods
 RSpec/InstanceVariable:
   Exclude:
     - 'spec/features/**/*.rb'


kevindew:
Thanks for looking into this 👍 

Out of curiosity I've had a bit of a play around and didn't seem to find any truly dynamic instance variables on the tests I looked at: https://github.com/alphagov/content-publisher/commit/ebfc8dda51ee0fafcdbc8ae119bc67774e063a63?diff=split. I think the `given(` approach works pretty well when there aren't many things to set, for example [edit edition spec](https://github.com/alphagov/content-publisher/blob/9a6c8a71b05e2fc861f139f2bb1e963e9d57d1fd/spec/features/editing_content/edit_edition_spec.rb#L2-L14),  but can be a bit of a mouthful when there are a number of items, for example [access limit spec](https://github.com/alphagov/content-publisher/blob/ebfc8dda51ee0fafcdbc8ae119bc67774e063a63/spec/features/editing_content_settings/enforce_access_limit_spec.rb)

benthorner:
@kevindew ah, I see: so we were proposing to change the structure of the test itself. That makes more sense now. I agree it does get a bit awkward, especially with lower-level variables [like this one](https://github.com/alphagov/content-publisher/blob/536142ec8e21513ece6e6030119ab317e6b7eb44/spec/features/editing_file_attachments/edit_file_attachment_spec.rb#L27). My underlying assumption here was that using instance variables in this way is part of our 'house style'. Is that a reasonable assertion?

PR #1958, opened 2020-03-27
Add import documents migration task

gclssvglx:
This change adds the ability to import multiple Whitehall documents via a single rake task.

This is especially useful when re-running imports that have failed for debugging purposes.

The associated `import-from-whitehall.md` documentation has also been updated to include this rake task.

Also, removed a un-required environment variable from the `import-from-whitehall.md`.

Trello: https://trello.com/c/RGNEM7u3/1632-add-whitehall-import-rake-task-for-multiple-documents

### DIFF
--- a/lib/tasks/import.rake
+++ b/lib/tasks/import.rake
@@ -6,7 +6,7 @@ namespace :import do
     include Rails.application.routes.url_helpers
     organisation_content_id = GdsApi.publishing_api.lookup_content_id(
       base_path: "/government/organisations/#{args.organisation_slug}",
-     )
+    )
     document_subtypes = args.document_subtypes ? args.document_subtypes.split(",") : []
     whitehall_migration = WhitehallImporter::CreateMigration.call(
       organisation_content_id, args.document_type, document_subtypes
@@ -25,11 +25,30 @@ namespace :import do
       whitehall_document_id: args.document_id,
       whitehall_migration: WhitehallMigration.create!,
       state: "pending",
-     )
+    )
 
     WhitehallDocumentImportJob.perform_later(whitehall_import)
     puts "Added whitehall document with ID:#{args.document_id} to the import queue"
 
     puts whitehall_migration_url(whitehall_import.whitehall_migration_id, host: Plek.new.external_url_for("content-publisher"))
+  end
+
+  desc "Import multiple documents from Whitehall Publisher using Whitehall's internal document ID's e.g. import:whitehall_documents[\"123 456 789\"]"
+  task :whitehall_documents, [:document_ids] => :environment do |_, args|
+    include Rails.application.routes.url_helpers
+    whitehall_migration = WhitehallMigration.create!
+
+    args.document_ids.split(" ").map(&:to_i).each do |document_id|
+      whitehall_import = WhitehallMigration::DocumentImport.create!(
+        whitehall_document_id: document_id,
+        whitehall_migration: whitehall_migration,
+        state: "pending",
+      )
+
+      WhitehallDocumentImportJob.perform_later(whitehall_import)
+      puts "Added whitehall document with ID:#{document_id} to the import queue"
+    end
+
+    puts whitehall_migration_url(whitehall_migration.id, host: Plek.new.external_url_for("content-publisher"))
   end
 end
--- a/spec/lib/tasks/import_spec.rb
+++ b/spec/lib/tasks/import_spec.rb
@@ -52,4 +52,38 @@ RSpec.describe "Import tasks" do
       )
     end
   end
+
+  describe "import:whitehall_documents" do
+    before do
+      allow($stdout).to receive(:puts)
+      Rake::Task["import:whitehall_documents"].reenable
+    end
+
+    it "creates pending whitehall migration document imports" do
+      allow(WhitehallDocumentImportJob).to receive(:perform_later)
+
+      expect {
+        Rake::Task["import:whitehall_documents"].invoke("123 456")
+      }.to change {
+        WhitehallMigration::DocumentImport.pending.exists?(whitehall_document_id: 123)
+      }.to(true).and change {
+        WhitehallMigration::DocumentImport.pending.exists?(whitehall_document_id: 456)
+      }.to(true)
+    end
+
+    it "queues a job for each document to be imported" do
+      Rake::Task["import:whitehall_documents"].invoke("123 456 789")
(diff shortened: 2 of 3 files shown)


emmabeynon:
@gclssvglx Could you add a test for this please?

gclssvglx:
> @gclssvglx Could you add a test for this please?

Whoops!  Tests added.

lib/tasks/import.rake
```diff
@@ -32,4 +32,12 @@ namespace :import do
 
     puts whitehall_migration_url(whitehall_import.whitehall_migration_id, host: Plek.new.external_url_for("content-publisher"))
   end
+
+  desc "Import an array of documents from Whitehall Publisher using Whitehall's internal document ID e.g. import:whitehall_documents[\"123 456 789\"]"
```
emmabeynon, line 36: I think we should change "an array of" to "multiple", since the argument we're passing to the task is a string.

lib/tasks/import.rake
```diff
@@ -32,4 +32,12 @@ namespace :import do
...
+  task :whitehall_documents, [:document_ids] => :environment do |_, args|
+    args.document_ids.split(" ").map(&:to_i).each do |document_id|
+      Rake::Task["import:whitehall_document"].reenable
+      Rake::Task["import:whitehall_document"].invoke(document_id)
```
emmabeynon, line 40: I think instead of invoking the single document rake task for each document ID, which creates a `WhitehallMigration` for each document, I think it would be helpful to create just one `WhitehallMigration` and associate that with each `DocumentImport`.  By doing this, we can quickly and easily access all document imports for a migration, rather than having to access each one individually.

gclssvglx: > I think instead of invoking the single document rake task for each document ID, which creates a `WhitehallMigration` for each document, I think it would be helpful to create just one `WhitehallMigration` and associate that with each `DocumentImport`. By doing this, we can quickly and easily access all document imports for a migration, rather than having to access each one individually.

I like the idea of this, but doesn't it imply that either all documents share the same organisation and/or document_type - which is not necessarily the case as the document id's can be chosen at random by the user, or we'd need to concat these values into fields that are only intended to hold a single value (e.g. "news_story | press_release")?

lib/tasks/import.rake
gclssvglx: Thanks - will fix

lib/tasks/import.rake
gclssvglx: I like the idea, but I do have one reservation with this approach - both `organisation_content_id` and `document_type` will remain `nil`.  

I totally get it that `WhitehallMigration` is being used as a kind of 'Composite', but perhaps if `organisation_content_id` and `document_type` were arrays and populated during the document loop I'd feel happier.  However, that does seem like quite a bit of work for little (if any) value given that we aren't really using those fields for anything meaningful - except for organisation/type/subtype(s) imports where they describe the input params to the rake task - and it doesn't change the UI at all!

emmabeynon: It's not necessary to populate `organisation_content_id` and `document_type` for `WhitehallMigration` - we don't do that when running the migration for a single document either.  I feel the value comes in the UI where we can get a link to one migration, and can monitor all the documents being imported in one place, rather than having to look at a bunch of separate migration pages.  Similarly, we can access all the document imports from one migration in the console, rather than having to look them up individually.

gclssvglx: @emmabeynon - Great thanks.  Please see code changes addressing the requested change.

lib/tasks/import.rake
```diff
@@ -25,11 +25,30 @@ namespace :import do
...
-     )
+    )
 
     WhitehallDocumentImportJob.perform_later(whitehall_import)
     puts "Added whitehall document with ID:#{args.document_id} to the import queue"
 
     puts whitehall_migration_url(whitehall_import.whitehall_migration_id, host: Plek.new.external_url_for("content-publisher"))
   end
+
+  desc "Import multiple documents from Whitehall Publisher using Whitehall's internal document ID e.g. import:whitehall_documents[\"123 456 789\"]"
```
1pretz1, line 36: V minor but should `ID` be plural as we're referencing importing multiple docs and so multiple ID's?

spec/lib/tasks/import_spec.rb
```diff
@@ -52,4 +52,38 @@
...
+      Rake::Task["import:whitehall_documents"].reenable
+    end
+
+    it "creates pending whitehall migration document imports" do
+      allow(WhitehallDocumentImportJob).to receive(:perform_later)
+
+      expect {
+        Rake::Task["import:whitehall_documents"].invoke("123 456")
+      }.to change {
+        WhitehallMigration::DocumentImport.pending.exists?(whitehall_document_id: 123)
```
1pretz1, line 68: Didn't know you could query like this, looks good! 👍

spec/lib/tasks/import_spec.rb
```diff
@@ -52,4 +52,38 @@
...
+        WhitehallMigration::DocumentImport.pending.exists?(whitehall_document_id: 456)
+      }.to(true)
+    end
+
+    it "queues a job for each document to be imported" do
+      Rake::Task["import:whitehall_documents"].invoke("123 456 789")
+      expect(WhitehallDocumentImportJob).to have_been_enqueued.exactly(3).times
+    end
+
+    it "creates a single WhitehallMigration given a single document" do
```
1pretz1, line 79: I think this test and the one [below](https://github.com/alphagov/content-publisher/pull/1958/files#diff-9c116fd10c62f7c423ebfcf6c4a21937R84) can be merged into one test checking that a single WhitehallMigration has been created as there is no logic here we need to test regarding multiple / single doc imports.

spec/lib/tasks/import_spec.rb
gclssvglx: Me neither, found it by accident - cool 'eh!

lib/tasks/import.rake
gclssvglx: Good spot Peter - thanks!

spec/lib/tasks/import_spec.rb
gclssvglx: Yeah, I did want them in one test, but it kept failing!  Found the issue though - needed to `reenable` the task between calls.  Testing rake tasks eh!

PR #1963, opened 2020-03-31
Fix backdate bug

1pretz1:
The previous code for backdating was doing some unexpected things and wasn't fully functional leading to cryptic errors in other parts of the import. This PR addresses those issues. 

More info in the commits ->

Trello:
https://trello.com/c/yR8ES4fq/1628-fix-backdate-bug

### DIFF
--- a/spec/lib/whitehall_importer/create_revision_spec.rb
+++ b/spec/lib/whitehall_importer/create_revision_spec.rb
@@ -30,20 +30,53 @@ RSpec.describe WhitehallImporter::CreateRevision do
       expect(revision.base_path).to eq("/a-path")
     end
 
-    it "sets backdated_to when edition has been backdated" do
+    it "sets backdated_to when the first edition has been backdated" do
       backdated_to = Time.zone.now.yesterday.rfc3339
       whitehall_edition = build(:whitehall_export_edition,
                                 first_published_at: backdated_to)
+      document_import = build(:whitehall_migration_document_import,
+                              document: document,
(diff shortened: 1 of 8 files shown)


spec/lib/whitehall_importer/create_revision_spec.rb
```diff
@@ -30,22 +30,59 @@
       expect(revision.base_path).to eq("/a-path")
     end
 
-    it "sets backdated_to when edition has been backdated" do
-      backdated_to = Time.zone.now.yesterday.rfc3339
-      whitehall_edition = build(:whitehall_export_edition,
-                                first_published_at: backdated_to)
+    context "when an edition has been backdated" do
```
emmabeynon, line 33: We might need to think about how we phrase these tests.  For example, this currently reads as "when an edition has been backdated it sets backdated_to when the first edition has been backdated" which doesn't make a lot of sense.  Perhaps the contexts could be "when an edition is a first edition" and "when an edition is not a first edition"?  Then the tests could be something like "sets backdated_to when it has been backdated" and "does not set backdated_to when it has not been backdated"?

1pretz1: Good point, they didn't make much sense! I've removed the context's all together because as I was rewording the tests it became apparent that they were not adding much, if anything. Hopefully the tests are now clearer!

1pretz1:
Thanks for the review @emmabeynon, I've updated the tests accordingly!

PR #1966, opened 2020-04-02
Import superseded editions with archived history events

emmabeynon:
For https://trello.com/c/RfpyAbs1/1536-investigate-missing-superseded-event-cma-dwp-dhsc

While building the status for a number of imported superseded editions, an
error has been raised as there is no event with a "superseded" state in the
edition's revision history. After some investigation into the source data in
Whitehall, it transpires that we have instances where superseded editions have
a final event with the state "archived".  For example:

```
"document_id"=>124314,
"state"=>"superseded",
 ...
"revision_history"=>
  [{"event"=>"create",
    "created_at"=>"2013-03-06T09:37:56.000+00:00",
    "state"=>"imported",
    "whodunnit"=>256},
    ...
   {"event"=>"update",
   "created_at"=>"2013-09-23T13:27:42.000+01:00",
   "state"=>"archived",
   "whodunnit"=>817}],
```

"Archived" looks to be an old edition state which is no longer in use. It was
renamed "Withdrawn" in May 2015:
https://github.com/alphagov/whitehall/commit/95e11e3
However, the documents that we have investigated weren't withdrawn but were
simply superseded by another edition.

The last usage of the "archived" state for edition events was in June 2015, so
we suspect it may be a legacy state, though its unclear why it was being used
instead of 'superseded' in these cases: `Version.where(state:
"archived").maximum(:created_at) => Tue, 02 Jun 2015 16:20:31 BST +01:00`

In order to handle the import of these unusual events for superseded editions,
we have modified the `Create Edition` importer class to look up the last
"archived" event for superseded editions without a "superseded" event.

### DIFF
--- a/lib/whitehall_importer/edition_history.rb
+++ b/lib/whitehall_importer/edition_history.rb
@@ -96,6 +96,10 @@ module WhitehallImporter
       end
     end
 
+    def last_event
+      revision_history.last
+    end
+
   private
 
     attr_reader :revision_history
--- a/lib/whitehall_importer/create_edition.rb
+++ b/lib/whitehall_importer/create_edition.rb
@@ -160,7 +160,7 @@ module WhitehallImporter
 
     def create_edition(status:, edition_number:, current:, revision:, create_event: nil, last_event: nil)
       create_event ||= history.create_event!
-      last_event ||= whitehall_edition["revision_history"].last
+      last_event ||= history.last_event
 
       editor_ids = history.editors.map { |editor| user_ids[editor] }.compact
       published_at = history.last_state_event!("published")["created_at"] if status.live? || status.superseded?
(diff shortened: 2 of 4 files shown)


lib/whitehall_importer/edition_history.rb
```diff
@@ -96,6 +96,10 @@ def imported_entry_type(event, edition_number)
       end
     end
 
+    def last_event
```
1pretz1, line 99: I wonder if we need this method, it fits into the idea that this class should handle all edition history logic but as the only logic is `revision_history.last` which is pretty self explanatory I think it'd be fine to have it inline in the create_edition file which would be a nice code reduction. What do you think?

1pretz1:
I'm now picking up this work as Emma is on AL, I've gone back on my recommendation to remove the `last_event` method as we can reuse it in another location. Have added a commit to reflect this.

PR #1970, opened 2020-04-03
Enable change history check to check non first drafts

1pretz1:
Previously this code assumed that publishing-api would return
change_history for draft editions. This has turned out to be incorrect
as publishing-api only returns change_history for live editions. This
has lead to some discrepancies when checking the state of change history
between cp and pub-api.

The difference in behaviour is that content-publisher can return
change_history for non-live editions if the draft edition is a major
change with a change_note and is not the first edition. See ->
https://github.com/alphagov/content-publisher/commit/72b6556e218d66d49e4fd21f14743ea49926d36f#diff-4e6bc683d583eff1852c25cce42073d2R26

The above means that content_history can be different when comparing
drafts so we need to trim the latest draft change_note in cp to reflect
the state in publishing-api.

## Importing stuff 
```
# in whitehall
irb(main):007:0> Document.find(440172).editions.pluck(:state, :minor_change,  :change_note)
=> [["published", false, nil], ["draft", false, "Second Major change note"]]

# Importing doc with second major draft edition using check-non-first-drafts-change-history branch
irb(main):007:0> WhitehallMigration::DocumentImport.find_by_whitehall_document_id(440172).state
=> "imported"

# Importing doc with second major draft edition using master branch
irb(main):001:0> WhitehallMigration::DocumentImport.find_by_whitehall_document_id(440172).state
=> "import_aborted"
irb(main):002:0> WhitehallMigration::DocumentImport.find_by_whitehall_document_id(440172).integrity_check_problems
=> ["change history doesn't match"]

# after running NDA import locally with staging data using check-non-first-drafts-change-history branch (one aborts due to unrelated foreign locale error)
irb(main):029:0> WhitehallMigration::DocumentImport.group(:state).count
=> {"imported"=>131, "import_aborted"=>1}
```
Trello:
https://trello.com/c/1lgwEe6n/1636-update-change-history-checker-to-cater-for-non-first-drafts

### DIFF
--- a/lib/whitehall_importer/integrity_checker/change_history_check.rb
+++ b/lib/whitehall_importer/integrity_checker/change_history_check.rb
@@ -2,27 +2,34 @@ module WhitehallImporter
   class IntegrityChecker::ChangeHistoryCheck
     attr_reader :edition, :proposed_change_history, :publishing_api_change_history
 
-    def initialize(proposed_change_history, publishing_api_change_history, edition)
-      @proposed_change_history = proposed_change_history
-      @publishing_api_change_history = publishing_api_change_history
+    def initialize(edition, proposed_change_history, publishing_api_change_history)
       @edition = edition
+      @proposed_change_history = trim_proposed_change_history(proposed_change_history)
+      @publishing_api_change_history = publishing_api_change_history
     end
 
     def match?
-      if publishing_api_change_history.empty?
+      if publishing_api_change_history.empty? && edition.live?
         return proposed_history_has_first_published_change_note?
       end
 
       return false unless history_length_matches?
 
-      if edition.live?
-        history_matches?(proposed_change_history, publishing_api_change_history)
-      else
-        history_excluding_first_timestamp_matches?
-      end
+      history_matches?(proposed_change_history, publishing_api_change_history)
     end
 
   private
+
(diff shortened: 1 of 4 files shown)


lib/whitehall_importer/integrity_checker/change_history_check.rb
```diff
@@ -2,28 +2,32 @@ module WhitehallImporter
...
-      else
-        history_excluding_first_timestamp_matches?
-      end
+      history_matches?(proposed_change_history, publishing_api_change_history)
     end
 
   private
 
+    def trim_proposed_change_history(proposed_change_history)
+      if edition.major? && edition.change_note && !edition.live? || !edition.live? && edition.first?
```
gclssvglx, line 24: It maybe just me, and you guys may not agree... Anyway, just a suggestion, but I'd prefer some brackets in this - tends to make the logic more obvious and that whole thing easier to read and parse, e.g.

```
if ((edition.major? && edition.change_note && !edition.live?) || (!edition.live? && edition.first?))
```

1pretz1: Good point about it being a bit hard to parse, I don't normally use brackets as it doesn't make it any easier for me personally to read. I have taken your point on board though and created variables to better describe the conditional arguments which I think it makes it much clearer.

PR #1972, opened 2020-04-03
[DO NOT MERGE] Add publication groups 1 and 2

pollygreen:
https://trello.com/c/t1OsZPiN/1494-identify-changes-to-how-to-guidance-based-on-publications-workflow-and-draft-them

We need to tell users that they can create publications with file attachments in Content Publisher when we've released them.


benthorner:
Adding a "DO NOT MERGE" flag as these shouldn't be merged until we can release Publications.

benthorner:
Decided to close this since it's no longer active work, due to the project being paused.

PR #1975, opened 2020-04-03
Make new document types accessible to pre-release

GDSNewt:
We want to enable Group 2 Publications for pre-release users so that they can publish them if they onlyneed to add file attachments.

In this PR I have made the following document types 'publishable' by pre-release users 

- Guidance 
- Statutory Guidance
- Decision
- Forms
- Maps
- Regulation
- Promotional material
- Notice

'Non-statutory guidance' is referred to as guidance in the Gov.uk schema, so i've had to rename any reference to 'non_statutory_guidance' to 'guidance.'

Some existing references to 'guidance' have had to be changed to 'guidance_title" in order not to
conflict with my changes made to the 'non_statutory_guidance' id

### DIFF
--- a/config/document_type_selections.yml
+++ b/config/document_type_selections.yml
@@ -28,7 +28,7 @@
       hostname: whitehall-admin
       path: /government/admin/speeches/new
 
-- id: guidance_title
+- id: guidance_publications
   options:
     - id: detailed_guide
       type: managed_elsewhere
@@ -57,7 +57,7 @@
 
 - id: guidance_and_regulation
   options:
-    - id: guidance_title
+    - id: guidance_publications
       type: document_type_selection
     - id: form
       type: document_type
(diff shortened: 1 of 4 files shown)


config/locales/en/document_types.yml
```diff
@@ -265,3 +417,13 @@ en:
               [Writing for GOV.UK](https://www.gov.uk/guidance/content-design/writing-for-gov-uk#writing-to-govuk-style){:target="_blank"}
 
               [The style guide](https://www.gov.uk/guidance/style-guide){:target="_blank"}
+
```
benthorner, line 420: We should cleanup these extra lines.

config/document_type_selections.yml
```diff
@@ -28,17 +28,21 @@
       hostname: whitehall-admin
       path: /government/admin/speeches/new
 
-- id: guidance
+- id: guidance_title
```
benthorner, line 31: Ah, this is annoying. Thanks for persevering with a solution. I think we should try and use a different ID if possible, as "title" suggests it's not an ID. How about "guidance_publications"?

PR #1976, opened 2020-04-03
Firm things up after incident

JonathanHallam:
Trello - https://trello.com/c/xkQzBiFE/1614-content-publisher-document-filter-incident-follow-up

This introduces testing for some incident related fixes. It also does some validation on a database level to stop potentially incident causing data to make its way in in the first place.


benthorner:
Closing in favour of https://github.com/alphagov/content-publisher/pull/1988.

PR #1977, opened 2020-04-03
Use 'top_level_description' to render view specs

benthorner:

This replaces the duplication of the partial to render with the
'top_level_description', which couples the partial rendered to the
described text of the view spec. Using 'top_level_description' is
more consistent with our other specs, where we use 'described_class'.

One caveat is that the description of the view specs is no longer
compatible with the direct use of 'render' (they now lack the '_'),
which seems reasonable, since we should be passing locals to partials,
instead of relying on instance variables in the parent view.

### DIFF
--- a/spec/views/documents/history/_content_publisher_entry.html.erb_spec.rb
+++ b/spec/views/documents/history/_content_publisher_entry.html.erb_spec.rb
@@ -1,15 +1,15 @@
-RSpec.describe "documents/history/_content_publisher_entry.html.erb" do
+RSpec.describe "documents/history/content_publisher_entry.html.erb" do
   it "shows a timeline entry without an author" do
     timeline_entry = create(:timeline_entry,
                             created_by: nil)
-    render partial: "documents/history/content_publisher_entry", locals: { entry: timeline_entry }
+    render partial: self.class.top_level_description, locals: { entry: timeline_entry }
     expect(rendered).not_to have_content(I18n.t!("documents.history.by"))
     expect(rendered).to have_content(I18n.t!("documents.history.entry_types.created"))
   end
 
   it "shows a timeline entry with an author" do
     timeline_entry = create(:timeline_entry)
-    render partial: "documents/history/content_publisher_entry", locals: { entry: timeline_entry }
+    render partial: self.class.top_level_description, locals: { entry: timeline_entry }
     expect(rendered).to have_content(I18n.t!("documents.history.by") + " John Smith")
     expect(rendered).to have_content(I18n.t!("documents.history.entry_types.created"))
   end
@@ -18,7 +18,7 @@ RSpec.describe "documents/history/_content_publisher_entry.html.erb" do
     timeline_entry = create(:timeline_entry,
                             entry_type: :internal_note,
                             details: create(:internal_note))
-    render partial: "documents/history/content_publisher_entry", locals: { entry: timeline_entry }
+    render partial: self.class.top_level_description, locals: { entry: timeline_entry }
     expect(rendered).to have_selector(".app-timeline-entry__content",
                                       text: "Amazing internal note")
     expect(rendered).to have_content(I18n.t!("documents.history.entry_types.internal_note"))
@@ -30,7 +30,7 @@ RSpec.describe "documents/history/_content_publisher_entry.html.erb" do
       timeline_entry = create(:timeline_entry,
                               entry_type: :removed,
                               details: removal)
-      render partial: "documents/history/content_publisher_entry",
+      render partial: self.class.top_level_description,
              locals: { entry: timeline_entry }
       expect(rendered).to have_content("My note")
     end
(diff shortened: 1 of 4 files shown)


kevindew:
@benthorner Argh you merged before I got my review in 😞 

This feels like it creates rather odd looking render calls. What's the motivation behind this? just duplication of the view name or have actual problems occurred?

benthorner:
There were a couple of problems, which I experienced while writing another view spec for https://github.com/alphagov/content-publisher/pull/1981:

   - Duplication of the partial path. For most views we can just call `render`. Having to type `render partial: "long_and_repetitious_path_even_though_its_in_the_description"` is a bit pointless

   - We had inconsistency between `_partial` and `partial`, and `render "path"` and `render partial: "path"`, and it's now the second time I've debugged which one to use

So the idea of this is to reduce the duplication and adopt a consistent approach whereby we can render partials for testing in a way that also works with `locals`.

kevindew:
@benthorner ok, whats the partial/_partial concern I can't see it on the link you've done? I'm not sure I understand that and I'm also a bit unsure what you mean about "a way that also works for locals"?

benthorner:
Ah, sorry, I realise I'm coming out of a well of knowledge. Here are some examples that show the problem, when compared to most view specs, which just do `render()`.

```
# doesn't work ('_' not valid)
render "file_attachments/index/_featured_attachment", locals: { a: "b" }

# doesn't work (need 'partial:')
render "file_attachments/index/featured_attachment", locals: { a: "b" }

# doesn't work ('_' not valid)
render partial: "file_attachments/index/_featured_attachment", locals: { a: "b" }

# doesn't work ('_' not valid in description)
render self.class.top_level_description, locals: { a: "b" }

# doesn't work (doesn't work, as assigns != locals, weirdly needs '_' in description)
assign(:a, "b")
render
```

There seems to be relatively little online about view testing for partials (or even view testing!), so I thought it would be helpful to adopt something that's easy to copy.

Since some of our partial specs don't have locals, it was confusing to find an example to copy from, when they were inconsistent (e.g. the filters partial spec). It's also confusing to be describing one kind of view, but then rendering another (with/out '_'). My thinking here is that we should always be passing locals when testing partials (otherwise, we're relying on instance variables), so this seemed like a helpful pattern to adopt.

Do you see it being a problem in some way?

kevindew:
Ah sure I gotcha, thanks for explaining. Yeah there's basically only [this page](https://relishapp.com/rspec/rspec-rails/v/4-0/docs/view-specs/view-spec). 

I see where you're going. Mostly I'm concerned as I feel like self.class.top_level_description seems really unusual - it also seems a bit of a hack since you have a `top_level_description` method but it won't work in this context.  I also find the change of description a bit odd too where say: "documents/history/content_publisher_entry.html.erb" references a file that doesn't really exist.

I can see the pains of the duplication so ideally I'd want to work towards working out if we can get rid of the self.class.top_level_description.

I've had a look myself in rails-rspec and notice they recommend using that to stub locals: https://github.com/rspec/rspec-rails/blob/9fb6c9c482da29b8803dd6f17b34d37a6367b8f2/lib/rspec/rails/example/view_example_group.rb#L76

So I wonder if these could be refactored to be like:

```
RSpec.describe "documents/history/_content_publisher_entry.html.erb" do
  it "shows a timeline entry without an author" do
    timeline_entry = create(:timeline_entry,
                            created_by: nil)
    view.stub(entry: timeline_entry)
    expect(rendered).not_to have_content(I18n.t!("documents.history.by"))
    expect(rendered).to have_content(I18n.t!("documents.history.entry_types.created"))
  end
```

benthorner:
@kevindew that's an interesting suggestion - thanks for tracking it down! I was in the process of trying it out, when I encountered a problem, which I think means we can't use it.

```
docker exec -it govuk-docker_content-publisher-lite_1 bundle exec rspec spec/views/documents/history
...............

Deprecation Warnings:

Using `stub` from rspec-mocks' old `:should` syntax without explicitly enabling the syntax is deprecated. Use the new `:expect` syntax or explicitly enable `:should` instead. Called from /govuk/content-publisher/spec/views/documents/history/_whitehall_entry.html.erb_spec.rb:7:in `block (2 levels) in <top (required)>'.
```

I guess the alternative is to do `allow(view).to receive(:entry) { timeline_entry }`. I did consider this, and it seemed an even more bizarre deviation from the normal `render partial: ..., locals: ...` call in the real code. Arguably `stub` has this same issue; it's just less verbose. There is [a slight alternative we could use](https://github.com/rails/rails/blob/157920aead96865e3135f496c09ace607d5620dc/actionview/lib/action_view/test_case.rb#L203), which is closer to how we do views:

```
# old
render partial: self.class.top_level_description, locals: { a: "b" }

# new
render self.class.top_level_description, a: "b"
```

I would have gone with that, had I known. Both use the non-underscore form though. It may be helpful to note that the `html.erb` extensions are also unnecessary in `RSpec.describe`. Does that allay your concern about the form of the descriptions, perhaps?

PR #1978, opened 2020-04-03
Allow official document type to be null

benthorner:

Previously we defaulted the official document type to 'unofficial',
but this is inconsistent with the design/intention to force users to
make an explicit selection. If the user aborts the upload workflow
for an attachment, it's also unclear if a selection was made or not.
This changes the column to default the official doc type to null.

### DIFF
--- a/db/migrate/20200403165044_allow_official_document_type_to_be_null.rb
+++ b/db/migrate/20200403165044_allow_official_document_type_to_be_null.rb
@@ -1,0 +1,9 @@
+class AllowOfficialDocumentTypeToBeNull < ActiveRecord::Migration[6.0]
+  def up
+    change_column :file_attachment_metadata_revisions, :official_document_type, :string, null: true, default: nil
+  end
+
+  def down
+    change_column :file_attachment_metadata_revisions, :official_document_type, :string, null: false, default: "unofficial"
+  end
+end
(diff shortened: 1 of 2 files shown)


db/schema.rb
```diff
@@ -139,7 +139,7 @@
     t.string "unique_reference"
     t.string "paper_number"
     t.string "parliamentary_session"
-    t.string "official_document_type", default: "unofficial", null: false
+    t.string "official_document_type", default: "unofficial"
```
ChrisBAshton, line 142: This still has a default of `"unofficial"`.

benthorner:
Thanks for spotting - should by fixed now 🌤

db/migrate/20200403165044_allow_official_document_type_to_be_null.rb
```diff
@@ -0,0 +1,9 @@
+class AllowOfficialDocumentTypeToBeNull < ActiveRecord::Migration[6.0]
+  def up
+    change_column :file_attachment_metadata_revisions, :official_document_type, :string, null: true, default: nil
+  end
+
+  def down
+    change_column :file_attachment_metadata_revisions, :official_document_type, :string, null: false, default: "unofficial"
+  end
```
ChrisBAshton, line 8: Well done for defining the `down` method - we quite often don't bother!

benthorner: It may have forced me 😜

PR #1982, opened 2020-04-06
Remove redundant requirements checks

benthorner:

This makes a few different changes to remove pre-preview and pre-publish
checks that are redundant: they could never occur. Making these changes
will help simplify the work to checks requirements for file attachment
metadata, where the complexity of the parameters made it difficult to
adopt a consistent approach to images.


app/models/document_type/multi_tag_field.rb
```diff
@@ -16,8 +16,4 @@ def pre_update_issues(_edition, _params)
   def pre_preview_issues(_edition)
     Requirements::CheckerIssues.new
   end
-
-  def pre_publish_issues(_edition)
-    Requirements::CheckerIssues.new
-  end
```
ChrisBAshton, line 22: >  Since the edition must satisfy all pre-preview issues before
> publishing, this removes the redundant pre-publish check.

Where is this in the code? I didn't know that.

spec/services/failsafe_draft_preview_service_spec.rb
```diff
@@ -21,7 +21,13 @@
     end
 
     context "when there are pre-preview issues" do
-      let(:edition) { create(:edition, summary: "new\nline", revision_synced: true) }
+      let(:document_type) do
+        build :document_type, contents: [DocumentType::TitleAndBasePathField.new]
+      end
+
+      let(:edition) do
+        create(:edition, document_type: document_type, title: "", revision_synced: true)
```
ChrisBAshton, line 29: Looks like `document_type` can be a variable scoped to this block, rather than a `let` in its own right.

benthorner: Oh yeah, good point - updated.

app/models/document_type/multi_tag_field.rb
benthorner: I guess it kind of makes sense, right? I can see why we'd want them to 'cascade', but they are actually separate, primarily because they are presented separately.

There's no actual code that 'enforces' this, but the user action to "Publish" isn't available unless the document can be previewed, and it can't be previewed unless the pre-preview requirements are met, which you can see in the following code:

   - https://github.com/alphagov/content-publisher/blob/master/app/services/failsafe_draft_preview_service.rb#L8
   - https://github.com/alphagov/content-publisher/blob/master/app/views/documents/show/_actions.html.erb#L25

Since the user probably doesn't want to publish a document that's out-of-date in the content store, it seems safe to assume they won't. There's [nothing technically stopping them](https://github.com/alphagov/content-publisher/blob/master/app/interactors/publish/publish_interactor.rb#L28), though.

PR #1984, opened 2020-04-06
Show featured attachment metadata on summary & reorder pages

ChrisBAshton:
Summary page:

Screenshot 2020-04-06 at 11 45 54

Reorder page*:

|Current|Design|
|-------|--------|
|Screenshot 2020-04-06 at 11 48 30|Screenshot_2020-03-19_at_12 15 51|

\* This doesn't quite match the designs, which I assume have changed since it was created. But I believe it's a reasonable MVP we can iterate later. Here is the design:

Trello: https://trello.com/c/a5SwDzjX/1440-show-featured-attachment-metadata-when-reordering-and-on-the-document-summary


app/views/components/docs/attachment_meta.yml
```diff
@@ -0,0 +1,81 @@
+name: Attachment meta
+description: Renders attachment metadata
+body: |
+  Renders only the metadata associated with an attachment (e.g. file type, size, reference numbers).
+  This duplicates the view and logic of the centralised 'Attachment' component, but is intended for
+  use where the thumbnail, 'request an accessibile version' and 'Order a copy' links, and title are
+  not desired.
+part_of_admin_layout: true
+shared_accessibility_criteria:
+  - link
```
benthorner, line 10: I'm pretty sure this isn't relevant here.

spec/views/components/_attachment_meta.html.erb_spec.rb
```diff
@@ -0,0 +1,68 @@
+RSpec.describe "components/_attachment_meta.html.erb", type: :view do
+  it "fails to render when no attachment is given" do
+    assert_raises do
+      render "components/attachment_meta", {}
+    end
+  end
+
+  it "can include attribute metadata" do
+    render "components/attachment_meta", attachment: {
+      title: "Attachment",
```
benthorner, line 10: Can we avoid the `title` and `url` for this component?

app/views/components/_attachment_meta.html.erb
```diff
@@ -0,0 +1,42 @@
+<%
+  attachment = GovukPublishingComponents::Presenters::Attachment.new(attachment)
+  attributes = []
+
+  if attachment.content_type_name
```
benthorner, line 5: Slightly outside of the scope of this PR: could we extract this into the presenter in the main components gem, so we could simplify the component and the code here?

All it seems to do is defined a bunch of textual attributes (the spans/classes don't add anything, individually), and then join them together with a comma [later on](https://github.com/alphagov/content-publisher/pull/1984/commits/c9bf7e60b561acf6de76acaa0baa7f8755a95ccf#diff-bb4f2ef4fecbbfab06aea98c17b6d0b3R40).

app/views/components/_attachment_meta.html.erb
```diff
@@ -29,11 +29,11 @@
 %>
 <%= tag.section class: "gem-c-attachment-meta" do %>
   <% if attachment.reference.present? %>
-    <%= tag.p "Ref: #{attachment.reference}", class: "gem-c-attachment-meta__metadata gem-c-attachment-meta__metadata--compact" %>
+    <%= tag.p "Ref: #{attachment.reference}", class: "gem-c-attachment-meta__metadata govuk-!-margin-bottom-0" %>
```
benthorner, line 32: Could we fix this in the main components gem, by using a `last-of-type` selector? It seems like the problem is due to the `> p` applying to all `p` elements, instead of the last one.

app/views/documents/show/_featured_attachments.html.erb
```diff
@@ -7,5 +7,15 @@
       { href: featured_attachments_path(@edition.document),
         data_attributes: { gtm: "edit-attachments" } }
     end
+  ),
+  items: (
+    @edition.featured_attachments.each_with_index.map do |attachment, index|
+      {
+        field: index + 1,
+        value: sanitize(
+          "<h4 class='govuk-heading-s govuk-!-margin-bottom-0'>#{attachment.title}</h4>#{render('components/attachment_meta', { attachment: file_attachment_attributes(attachment, @edition) })}"
```
benthorner, line 16: Nice one for working out how to get this in - I think I would have struggled to find the right combination! I wonder if this suggests a smell about the `summary_list` component: we're using `field` to mean a number here; perhaps a cleaner approach (involving less manipulation here), would be for the component to support a "numbered" attribute.

What do you think?

app/views/components/_attachment_meta.html.erb
ChrisBAshton: I’m afraid it looks like it would require a change to govuk-frontend, and then (presumably?) a bump in the govuk-frontend dependency in govuk_publishing_components.
But worse than that, it looks as though the last-of-type suggestion goes against the intended behaviour of the component (i.e. everything should have a 10px margin, except the last child which should have no margin: https://github.com/alphagov/govuk-frontend/blob/master/src/govuk/components/summary-list/_summary-list.scss#L90-L96)

And if we attempt to fix this in Content Publisher by using stronger selectors (`p.gem-c-attachment-meta__metadata--compact`, to beat the `.govuk-summary-list__value > p`), we get a lint error: `app/assets/stylesheets/components/_attachment-meta.scss:16:1 [W] QualifyingElement: Avoid qualifying class selectors with an element.`

I suspect it's better to use the utility classes provided.

app/views/documents/show/_featured_attachments.html.erb
ChrisBAshton: We do already [use a number as a key in collections-publisher](https://github.com/alphagov/collections-publisher/blob/896b0a145e25d4f68926b39724656290ac4102ac/app/views/step_by_step_pages/show.html.erb#L30), although sometimes it's a word like "AND" or "OR", so not quite as simple.

Are you proposing this sort of structure?

```erb
{
  numbered: true,
  field: "<h4 class='govuk-heading-s govuk-!-margin-bottom-0'>#{attachment.title}</h4>",
  value: render('components/attachment_meta', { attachment: file_attachment_attributes(attachment, @edition) })}
}
```

I wonder how this should be presented in the underlying [description list/terms/descriptions](https://www.w3schools.com/tags/tag_dd.asp) markup. For reference, this is our current output:

```html
<dl>
  <dt>1</dt>
  <dd><h4>Foo</h4>Unique ref</dd>
</dl>
```

Looking over this, I think the implementation is probably incorrect. We're using the summary list because it's styled the way we want it, but semantically this should really be an ordered list instead.

```html
<ol>
  <li><h4>Foo</h4>Unique ref</li>
</ol>
```

I think a `numbered` attribute is probably fixing the wrong problem, and suggest we mark up our own `<ol>` to look like the summary list instead. (If/when we have time, of course...!)

benthorner: Yes, that's right. I also had a quick look at the component, and although I can't see any huge semantic issues with describing the items this way, there is at least room to improve the interface so we don't have to fake a numbered list ourselves. And perhaps the enhanced interface would be implemented as an ordered list, like you suggest.

I think there is one thing we could do here: would you be willing to run this by [the design system team](https://github.com/alphagov/govuk-design-system-backlog/issues/182)? Perhaps we can incorporate this need into the design system.

app/views/components/_attachment_meta.html.erb
ChrisBAshton: Hi - you're correct except for the [`gem-c-attachment__abbr` class](https://github.com/alphagov/govuk_publishing_components/blob/master/app/views/govuk_publishing_components/components/_attachment.html.erb#L11-L13), which does have some associated styling & behaviour. If we didn't have to worry about this, we could indeed move this out to the presenter and just call `attachment.attributes` later on.

It would still be possible to refactor by, say, returning an array of hashes of the form `{ text: "foo", class: "gem-c-attachment__abbr", title: "Expanded foo" }`, where the majority of hashes would just contain a `text` key/value pair. It means the view needs to know what to do with those hashes, though. I can kind of see why it was just left in the view now.

benthorner:
Thanks for all the effort on the comments, @ChrisBAshton! I'll try and summarise where I think we are. There a couple of things we ought to do before this goes in:

- We should consider [raising the numbering issue with the design system team](https://github.com/alphagov/content-publisher/pull/1984#discussion_r404644154), so the semantic issue with the component doesn't get lost when this gets merged.

- We should document the work left to do here, or add an extra commit if it's not too hard. I'm referring to the comment in the description that this doesn't match the designs.

app/views/documents/show/_featured_attachments.html.erb
ChrisBAshton: Now done: https://github.com/alphagov/govuk-design-system-backlog/issues/182#issuecomment-610387202

ChrisBAshton:
I've now [left a comment in the Design System](https://github.com/alphagov/govuk-design-system-backlog/issues/182#issuecomment-610387202).
The frontend work to get it matching the designs is not trivial, so we'll have to leave it for the time being. I don't know where best to document that it doesn't yet match the design - perhaps I could write up a quick Trello card showing the 'current' vs 'desired' screenshots?

benthorner:
Thanks @ChrisBAshton. Can we add some screenshots to this PR to clarify the comment in the description (this PR vs. the design)? Since I'm not sure what the discprepancies are, it's hard for me to make a suggestion about what to do next. We probably don't need another card, though: we could include this information in the current one, tick off what's been done, and move it back to 'Todo'.

ChrisBAshton:
Done @benthorner 👍

app/assets/stylesheets/components/_attachment-meta.scss
```diff
@@ -1,31 +1,19 @@
-.app-c-attachment-meta {
+.gem-c-attachment-meta {
```
benthorner, line 1: Minor: we should add a comment to this file to explain why we're using `gem`, as it's inconsistent with all our other 'custom' styles. Sorry I didn't spot this before 😬.

spec/views/components/_attachment_meta.html.erb_spec.rb
```diff
@@ -0,0 +1,58 @@
+RSpec.describe "components/_attachment_meta.html.erb", type: :view do
+  it "fails to render when no attachment is given" do
```
benthorner, line 2: Minor: this seems like an unusual test, compared to other components I've seen. We just shouldn't be using the component in this way, so it's OK for it's behaviour to be undefined in the tests.

ChrisBAshton: This is a [pretty common test in govuk_publishing_components](https://github.com/alphagov/govuk_publishing_components/blob/master/spec/components/contents_list_spec.rb#L39-L41). I think it's to avoid breaking apps that accidentally don't pass everything that's expected to a component.

benthorner: Hmmm...I'm not sure that one's a great precedent. I guess you mean *silently* breaking the app. That still seems the fault of the app. I'm content for it to stay in 👍.

app/assets/stylesheets/components/_attachment-meta.scss
ChrisBAshton: I've pushed up one last commit which renames to `app-c` - could you take a look?

benthorner: Looks good to me 👍. 

(I suspect the build failures are because you've also renamed `app/gem-c-attachment`.)

PR #1986, opened 2020-04-06
Manage featured attachment ordering

benthorner:
https://trello.com/c/UDEgwkWF/1556-apply-featured-attachment-order-everywhere

This extends the updater module for file attachments to manage the featured
attachment ordering when file attachments are added or removed, so that it
is always representative of the file attachments that are currently
associated with a revision.

We have arbitrarily decided to append new file attachments to the existing
ordering. Please see the commits for more details.


benthorner:
Thanks for reviewing @ChrisBAshton! I realised that there was an issue with this, and I've added an extra commit. Could you take another look?

PR #1987, opened 2020-04-06
Remove redundant interactor specs

benthorner:
https://trello.com/c/UDEgwkWF/1556-apply-featured-attachment-order-everywhere

This turned out to be wasted work, as I realised I was adding the specs in
the wrong place to complete the above card. However, I think it's an overall
improvement, and worth a review. Please see the commits for more details.

### DIFF
--- a/spec/interactors/file_attachments/create_interactor_spec.rb
+++ b/spec/interactors/file_attachments/create_interactor_spec.rb
@@ -18,10 +18,6 @@ RSpec.describe FileAttachments::CreateInteractor do
     end
 
     context "when input is valid" do
-      it "is successful" do
-        expect(described_class.call(**args)).to be_success
-      end
-
       it "creates a new file attachment revision" do
         expect { described_class.call(**args) }
           .to change { FileAttachment::Revision.count }.by(1)
@@ -34,11 +30,16 @@ RSpec.describe FileAttachments::CreateInteractor do
         described_class.call(**args)
       end
 
-      it "delegates generating a unique filename to GenerateUniqueFilenameService" do
+      it "generates a unique filename for the attachment" do
+        other_attachment = create :file_attachment_revision
+        edition = create :edition, file_attachment_revisions: [other_attachment]
+        args[:params].merge!(document: edition.document.to_param)
+
         expect(GenerateUniqueFilenameService).to receive(:call)
-          .with(existing_filenames: edition.revision.file_attachment_revisions.map(&:filename),
+          .with(existing_filenames: [other_attachment.filename],
                 filename: file.original_filename)
           .and_call_original
+
         described_class.call(**args)
       end
 
--- a/spec/interactors/file_attachments/update_file_interactor_spec.rb
+++ b/spec/interactors/file_attachments/update_file_interactor_spec.rb
@@ -1,0 +1,39 @@
+RSpec.describe FileAttachments::UpdateFileInteractor do
+  describe ".call" do
+    let(:user) { create(:user) }
+    let(:attachment) { create(:file_attachment_revision) }
+    let(:file) { fixture_file_upload("files/13kb-1-page-attachment.pdf") }
+
+    let(:edition) do
+      create :edition, file_attachment_revisions: [attachment]
+    end
+
+    let(:params) do
+      ActionController::Parameters.new(
+        file_attachment: {
+          file: file,
+          title: "My title",
+        },
+        document: edition.document.to_param,
+        file_attachment_id: attachment.file_attachment_id,
+      )
+    end
+
+    before do
+      allow(FailsafeDraftPreviewService).to receive(:call)
+    end
+
+    it "generates a unique filename for the attachment" do
+      other_attachment = create :file_attachment_revision
+      edition = create :edition, file_attachment_revisions: [attachment, other_attachment]
+      params.merge!(document: edition.document.to_param)
+
+      expect(GenerateUniqueFilenameService).to receive(:call)
+        .with(existing_filenames: [other_attachment.filename],
+              filename: file.original_filename)
+        .and_call_original
+
+      described_class.call(params: params, user: user)
+    end
+  end
+end
--- a/spec/interactors/images/create_interactor_spec.rb
+++ b/spec/interactors/images/create_interactor_spec.rb
@@ -14,10 +14,6 @@ RSpec.describe Images::CreateInteractor do
     end
 
     context "when input is valid" do
-      it "is successful" do
-        expect(described_class.call(**args)).to be_success
-      end
-
       it "creates a new image revision" do
         expect { described_class.call(**args) }
           .to change { Image::Revision.count }.by(1)


benthorner:
I was actually thinking the same thing a few weeks ago. Then I changed my mind. You may have set something off with that article you shared about "necessary and sufficient testing". That said, the motivation for this PR really came from the pain of adding [a brand new interactor test](https://github.com/alphagov/content-publisher/pull/1981/commits/7f3f284ea0bd102fd648d32a9676e264c474b0b1#diff-d8a25217959e5a65699a660ecc245b22) just to cover one thing. What do you think are the downsides of not having a complete unit test?

kevindew:
> I was actually thinking the same thing a few weeks ago. Then I changed my mind.

I was a little surprised as I recall you evangelising in the past that a unit test should be full or nothing

> You may have set something off with that article you shared about "necessary and sufficient testing".

Sure, in that article they demonstrate going beyond necessary as testing scenarios in the unit test that are not coded for in the code under test. 

> That said, the motivation for this PR really came from the pain of adding [a brand new interactor test](https://github.com/alphagov/content-publisher/pull/1981/commits/7f3f284ea0bd102fd648d32a9676e264c474b0b1#diff-d8a25217959e5a65699a660ecc245b22) just to cover one thing.

Yeah I can understand that pain point. It is annoying and tedious.

Outside of this context this is something I can find challenging in other apps. You change something slightly go to adapt the tests and then discover that class does not have unit tests, then you face deciding do you: a) add full tests for the class b) add a unit test for this one condition or c) find the closest high level test that touches the code and contrive some example that exercises.

When I'm those situations I'm nearly always disappointed that the unit tests didn't exist and frustrated at the thought I have to put in to test the thing

> What do you think are the downsides of not having a complete unit test?

To rephrase this a bit, my understanding of the upsides and downsides of having complete interactor tests are:

Upsides

- It's clear by precedent how to test any changes you make to interactors or what tests you add when you create a new one, you don't need to know exactly what is tested at integration tests or understand a subtle strategy
- Interactors are tested in a conventional way to other unit tested classes in the application
- When adapting any integration tests you don't need to be aware of the interactor tests to identify what is tested or not (for example when we decided previews needn't be covered in feature tests)
- Subtleties and nuances at a low level are asserted that wouldn't make sense in the context of an integration test (for example user attribution: https://github.com/alphagov/content-publisher/blob/e3bb61f60a426461b52b862b3062addc44135658/spec/interactors/file_attachments/create_interactor_spec.rb#L51-L59)

Downsides:

- Whenever we go to make changes to an existing interactor without tests someone has to write all the tests which isn't fun, particularly as we expect all the code works correctly
- Interactors in their declarative, easy to understand, nature are quite repetitive which is compounded in writing the tests which result in a number of tests that feel low value
- They add to the build time

I did ponder about whether I felt that coverage repeition was a downside but I feel that since the coverage repetition is done by higher level integration tests (request and feature) and is mostly incidental (for example we test that text is shown rather than say persisted to a database) that I felt this didn't concern me. I'd feel differently if we were doing the same tests at a unit level though.

benthorner:
> I was a little surprised as I recall you evangelising in the past that a unit test should be full or nothing

I'm the living embodiment of that "strong opinions" maxim 😆. Thanks for weighing up the pros and cons. One thing I noticed is that they're phrased in a future world where all interactors are fully tests, which is quite a long way from where we are now (only a few are (partially) tested).

At the moment, none of the tests in this PR are complete. It would be good to keep the new and modified tests here, so I'm hoping there's a way we can reach an agreement that's a net improvement by looking at the removed tests individually.

### Covered by the feature (happy path)
- "delegates saving the file to the CreateFileAttachmentBlobService"
- "sets the title of the File attachment"
- "is successful"
- "creates a new image revision" (sort of, immutability covers the rest)
- "creates a new file attachment revision" (sort of, immutability covers the rest)

### Covered by the request spec
- "fails with issues returned"
- "raises a state error"

### Somewhat subtly covered
- "creates a timeline entry" (covered by feature spec)

### Not commonly tested
- "attributes the various created file attachment models to the user"
- "updates the preview"

### Should probably keep (on reflection)
- "normalises the uploaded image and delegates saving it to CreateImageBlobService"

I'm not convinced that we need to test the basic happy path twice, except when there are hard-to-feature-test subtleties like normalisation going on. Here are a few suggestions for how to change this PR (or at least a plan for how to follow it), that may help to reach a conclusion:

- Move all timeline entry tests into (new) interactor tests (out of features), with a corresponding (single) view test that they can (all) be shown in the associated tab.

- Add a test that each interactor calls the failsafe preview service. Although this isn't a critical piece of code, it's surprising that it's not tested (almost everywhere).

- Remove the tests around the user ID. I'm not entirely sure why we store this nullable data at so many levels, so it seems odd to expend effort testing it.

My overall thinking here is that it's unlikely all devs will expend the effort to write a 'complete' unit test, but we can still make things better by promoting their use for 'useful happy path things' (without going as far as "is successful"). Does that address some of your concerns?

kevindew:
> One thing I noticed is that they're phrased in a future world where all interactors are fully tests, which is quite a long way from where we are now (only a few are (partially) tested).

Sure, I'm hoping we get to that point gradually though. To be honest, I don't really mind putting the leg work in to adding them all for the sake of consistency so long as their contents aren't too contentious.

> At the moment, none of the tests in this PR are complete. It would be good to keep the new and modified tests here, so I'm hoping there's a way we can reach an agreement that's a net improvement by looking at the removed tests individually.

Right, I think the simplest route forward to wrapping things in a "agree to disagree" approach would be to not delete existing tests but just add new or amended one. I don't think either of us would be satisfied with that but it remains an available option.

---

So of the things flagged I've got some thoughts, though I think we may still end up at disagreement if your pre-requisite is that integration test usage renders a unit test no longer necessary, anyway here goes:

- is successful: very little value since we don't make use of it, mostly wrote for completeness sake since it's a key property of the object, mostly feels something good to have to counter tests that check failures.
- delegates the saving to *Service: doesn't add much value, added for consistency with other approaches, could be dropped
- creates a db record test: I think these are only subtly tested in other places so I actually think this is a good test
- sets title (and similar metadata): tricky here, I see this as in conjunction with the record test (maybe they could be combined) in that nothing else would really exercise that we created a db record with the expected data, feels like they should be used to cover all the fields whereas a feature test doesn't need to.
- user attribution - maybe we disagree on the value of this but as this is a part of our data integrity I think it should be tested and ideally you just want this tested at the lowest level.
- fails with issues - I see this as having a different role at integration request vs unit test level, at a unit level we should be testing that this interactor uses the various requirements checker a request may deal with. At a request test level however it should just be that we can return the right response no matter what request test issues there are.
- state error - this one I do find a bit of a pain since it's so frequently same logic in controller and interactor which are tested differently and thus seems one of the most annoying for duplicate testing (and logic in app). I wonder if there's a way we can inject a controller lambda into interactors so this information can not be duplicated so much, maybe that'd be confusing :-/ Digression aside there are occasions where there is more going on with this in interactors compared to controllers so some conditions end up not tested.

On the other items I think we're in agreement about what can be tested.

I'm happy with idea to move timeline entry stuff out of feature tests.

It seems a bit of a shame to be having this conversation in the midst of the pausing point of this project as it'd be quite fun to explore some options to see whether there are techniques to resolve both our concerns. I'm wondering if at the heart of it we have a challenge is that interactors have a lot of repetitive and relatively simple code which offer good advantages for readability and ease of understanding but then bite us with a similar repetition in tests in order to be consistent with wider testing strategy.

benthorner:
I think we might actually be able to agree to agree 😸. I see the argument around requirements tests a bit more clearly now: we're covering that a particular set of requirements checks are being done, instead of just one (potentially) arbitrary check in a request test.

Trying to summarise (I think we could do with a pain card!), we have:

- A timeline entry test (including the values stored)
- A requirements checker test (non-stubbed)
- A general 'stored data' test (including user attributes)
- A preview test (stubbed to avoid extra API stubs)

This would be the 'minimum set' of tests for many/most interactors, and we may want others and to structure them differently. (We also have a consistency issue in how we structure the tests, and we could address that as part of the general pain card work.)

How about the following next steps?

- See if we can get this (amended) PR in
- Write a pain card to document/continue/plan the wider discussion

kevindew:
@benthorner yup next steps sound good 👍

benthorner:
@kevindew so we now have this: https://trello.com/c/NmORzdGR/1752-our-approach-to-testing-interactors-is-incomplete-and-inconsistent. Do you think this PR can go in as-is, or should we defer it to the point when we address the pain card?

PR #1988, opened 2020-04-06
Add data integrity validations for JSON columns

benthorner:
https://trello.com/c/xkQzBiFE/1614-content-publisher-document-filter-incident-follow-up

This adds a custom validator around the 'contents' and 'tags'
columns, to ensure that we don't store syntactically invalid
data in these otherwise flexible JSON columns. With this change,
we now apply this schema-like approach to all JSON columns.

Note that, although we could use the document type to detect which
tag fields are permitted, this approach doesn't extend to validating
'contents', since the document type isn't accessible from the
associated ContentRevision.

### DIFF
--- a/app/models/content_revision.rb
+++ b/app/models/content_revision.rb
@@ -3,6 +3,8 @@
 #
 # This model is immutable.
 class ContentRevision < ApplicationRecord
+  validates :contents, "content_revision/contents" => true
+
   belongs_to :created_by, class_name: "User", optional: true
 
   def readonly?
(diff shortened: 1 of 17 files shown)


benthorner:
@kevindew thanks for reviewing - I did wonder if this might be a tad strict, and we can always relax it if it becomes an issue. This warrants another look, as I've now added fixes for the tests, and an extra couple of commits to finish [the original incident fix](https://github.com/alphagov/content-publisher/pull/1899/files).

PR #1994, opened 2020-04-08
Address a number of consistency issues in view specs

kevindew:
This is worth reviewing commit by commit as the overall diff looks a lot more dramatic than the actual changes made and there are full explanations of each change.

Mostly these changes continue the work and discussion from: https://github.com/alphagov/content-publisher/pull/1977

Happy to drop off the last commit if that is contentious.

### DIFF
--- a/spec/views/documents/history/_whitehall_entry.html.erb_spec.rb
+++ b/spec/views/documents/history/_whitehall_entry.html.erb_spec.rb
@@ -4,7 +4,7 @@ RSpec.describe "documents/history/_whitehall_entry.html.erb" do
                             :whitehall_imported,
                             whitehall_entry_type: :new_edition,
                             created_by: nil)
-    render template: self.class.top_level_description, locals: { entry: timeline_entry }
+    render template: described_template, locals: { entry: timeline_entry }
     expect(rendered).not_to have_content(I18n.t!("documents.history.by"))
     expect(rendered).to have_content(I18n.t!("documents.history.entry_types.whitehall_migration.new_edition"))
   end
@@ -14,7 +14,7 @@ RSpec.describe "documents/history/_whitehall_entry.html.erb" do
                             :whitehall_imported,
                             whitehall_entry_type: :new_edition)
 
-    render template: self.class.top_level_description, locals: { entry: timeline_entry }
+    render template: described_template, locals: { entry: timeline_entry }
     expect(rendered).to have_content(I18n.t!("documents.history.by") + " John Smith")
     expect(rendered).to have_content(I18n.t!("documents.history.entry_types.whitehall_migration.new_edition"))
   end
@@ -24,7 +24,7 @@ RSpec.describe "documents/history/_whitehall_entry.html.erb" do
                             :whitehall_imported,
                             whitehall_entry_type: :internal_note)
 
-    render template: self.class.top_level_description, locals: { entry: timeline_entry }
+    render template: described_template, locals: { entry: timeline_entry }
     expect(rendered).not_to have_selector(".app-timeline-entry--highlighted")
   end
 
@@ -35,7 +35,7 @@ RSpec.describe "documents/history/_whitehall_entry.html.erb" do
                             whitehall_entry_type: :internal_note,
                             whitehall_entry_contents: { body: note })
 
-    render template: self.class.top_level_description, locals: { entry: timeline_entry }
+    render template: described_template, locals: { entry: timeline_entry }
     expect(rendered).to have_content(note)
     expect(rendered).to have_selector(".app-timeline-entry--highlighted")
   end
@@ -45,7 +45,7 @@ RSpec.describe "documents/history/_whitehall_entry.html.erb" do
                             :whitehall_imported,
                             whitehall_entry_type: :fact_check_request)
 
-    render template: self.class.top_level_description, locals: { entry: timeline_entry }
+    render template: described_template, locals: { entry: timeline_entry }
     expect(rendered).not_to have_selector(".app-timeline-entry--highlighted")
   end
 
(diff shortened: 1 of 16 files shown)


spec/views/documents/history/_whitehall_entry.html.erb_spec.rb
```diff
@@ -57,13 +55,12 @@
...
     email = "someone@somewhere.com"
     instructions = "Do something, then do something else"
-    timeline_entry = create(:timeline_entry,
-                            :whitehall_imported,
-                            whitehall_entry_type: :fact_check_request)
-    timeline_entry.details[:contents] = {
-      email_address: email,
-      instructions: instructions,
-    }
+    timeline_entry = create(
```
benthorner, line 55: Much better! I wondered about this myself.

spec/views/featured_attachments/index.html.erb_spec.rb
```diff
@@ -1,29 +1,27 @@
 RSpec.describe "featured_attachments/index.html.erb" do
-  describe "featured_attachments index" do
-    it "shows file attachments that exist on the edition" do
-      file_attachment_revision = create(:file_attachment_revision)
-      edition = create(:edition,
-                       file_attachment_revisions: [file_attachment_revision])
-      assign(:edition, edition)
+  it "shows file attachments that exist on the edition" do
```
benthorner, line 2: Nice! I think this was something I introduced 😬.

spec/views/featured_attachments/index.html.erb_spec.rb
```diff
@@ -4,24 +4,31 @@
...
-    expect(render).to have_content(I18n.t!("featured_attachments.index.no_attachments"))
+    render
+    expect(rendered).to have_content(I18n.t!("featured_attachments.index.no_attachments"))
   end
 
-  it "only shows a reorder action for multiple attachments" do
-    edition = create(:edition)
-    edition.file_attachment_revisions << build(:file_attachment_revision)
-
+  it "shows a reorder action for a multiple attachments" do
```
benthorner, line 17:
```suggestion
  it "shows a reorder action for multiple attachments" do
```

spec/views/featured_attachments/index.html.erb_spec.rb
```diff
@@ -4,24 +4,31 @@
...
+                     file_attachment_revisions: file_attachment_revisions)
     assign(:edition, edition)
-    expect(render).not_to have_content("Reorder attachments")
+    render
+    expect(rendered).to have_content("Reorder attachments")
+  end
 
-    edition.file_attachment_revisions << build(:file_attachment_revision)
-    expect(render).to have_content("Reorder attachments")
+  it "doesn't show a reorder action for a single attachments" do
```
benthorner, line 26:
```suggestion
  it "doesn't show a reorder action for a single attachment" do
```

spec/views/featured_attachments/index.html.erb_spec.rb
benthorner: I kind of preferred having this as one test, as we are really testing "one thing".

kevindew:
Thanks for the review @benthorner. I ended up putting this up in a rush and meant to add some more context about my experiments with a single render call so I'll jot this up now (as it is a very quiet working Saturday 😄).

So I made really good progress with setting all view variables with stubbing and was feeling quite happy with myself (see: https://github.com/alphagov/content-publisher/commit/16a5f0c2fcd56af7def98ee428570a7cfc3babb3) and then all I had to do was rebase against master to the changes from https://github.com/alphagov/content-publisher/pull/1984 and everything went to hell.

It turns out that the subtle difference between methods and variables is exposed [where `attachment` undefined variable overrides the `attachment` variable](https://github.com/alphagov/content-publisher/blob/a9f0f9fab2096520954aac5219e7060fbfb1ac1c/app/views/components/_attachment_meta.html.erb#L2)  so needs [naming differently](https://github.com/alphagov/content-publisher/blob/e8864754c11b051368d133effc7cc3862f20b219/app/views/components/_attachment_meta.html.erb#L2). I then ventured into govuk_publishing_components and realised that the [very common `||=`](https://github.com/alphagov/govuk_publishing_components/blob/19f17b858c939ec82a3de910751db8e482e65dcd/app/views/govuk_publishing_components/components/_attachment.html.erb#L2-L7) approach wouldn't work with stubs either so retreated. 

> Perhaps something we could do to address the con here is to raise an issue with [rspec-rails](https://github.com/rspec/rspec-rails). 

That's a good idea. I'm not sure how into that they'd be as they are sometimes inconsistent with what they use as description for view tests in their examples (for example https://relishapp.com/rspec/rspec-rails/v/4-0/docs/view-specs/view-spec#view-specs-can-explicitly-render-templates) 

---

On a tangent once I finished putting this together I was quite pleased how all the conventions in the describe names matched the approach across the [rails-rspec docs](https://relishapp.com/rspec/rspec-rails/v/2-2/docs/view-specs/view-spec) (yay 🎉 ) and then just after I opened this PR I realised I'd been looking at 2.2 of the docs and they've used more Rails style naming in the [newer versions](https://relishapp.com/rspec/rspec-rails/v/4-0/docs/view-specs/view-spec#view-specs-can-have-before-block-and-nesting). So I'm going to have a quick experiment with that for greater consistency with rspec-rails conventions

spec/views/featured_attachments/index.html.erb_spec.rb
kevindew: Thanks for catching this and https://github.com/alphagov/content-publisher/pull/1994/files/136572f9073baec61c29245d8edf544e185bb85d..eb91e079fd7213c996d0ed0e4c05c409d5e7fd20#diff-60cb707cf57eb05421be54da245b4243R17 - they were rather careless mistakes :(

spec/views/featured_attachments/index.html.erb_spec.rb
kevindew: I'm not massively fussed either way but going down a single test route means we have to treat the rendered method different as it [appends to it with each render](https://github.com/rails/rails/blob/4d0947207a54db89b646b6277b9f8c8f6dcf53d6/actionview/lib/action_view/test_case.rb#L122) (which I just learnt while trying to move this back) and thus move the expectations back to being `expect(render).to have_content()`. 

I'm not particularly opposed to using render in that way it's just a tricky balance of consistency across the view tests.

benthorner: Oh, yuk. That's unfortunate - let's keep it like this!

PR #2002, opened 2020-04-11
Switch time helper methods to use Rails time formats

kevindew:
ActiveSupport provides a system to specify named time formatting and has a
convention to do this through a config/initializers/time_formats.rb file
[1]. By using this we can remove the need to have helper functions to
provide the time and rely on passing a symbol to the `to_s` method of a
time. This change was inspired by https://andycroll.com/ruby/using-strftime-in-a-rails-view-is-probably-a-mistake/
that I'd been meaning to try out and finally found myself working on a
rather quiet Saturday shift.

As always with something like this, a painful part of this was deciding what to
name the formatters. I felt it was best to go with very similar to
naming as what we'd had with the helper. This has a downside that it
clobbers the existing formatting of `Time::DATE_FORMATS[:time]` but
since we make no use of that it has no affect on the application.

I also resolved some trailing comma issues in hashes so they were
consistent with our linting rules (which aren't enforced on erb files)

[1]: https://github.com/rails/rails/blob/b2eb1d1c55a59fee1e6c4cba7030d8ceb524267c/activesupport/lib/active_support/core_ext/time/conversions.rb#L43-L50


config/initializers/time_formats.rb
```diff
@@ -0,0 +1,4 @@
+Time::DATE_FORMATS[:date] = "%-d %B %Y"
+Time::DATE_FORMATS[:time] = "%-l:%M%P"
```
leenagupte, line 2: You could call this "time_12_hour_clock" ?

kevindew: I wondered about time_with_period but that level of detail for time wouldn't be on other helpers.

kevindew:
> This is an interesting change 🤔. I'm struggling to see it as an improvement, and the arguments in the blog seem pretty weak/circumstantial. The confusing aspect of this change for me is that a presentation concern is now in a (somewhat obscure) initializer. While it is more concise and possibly more conventional, I wonder if it will lose more in confusion than we gain in brevity.

Oh boo, you seem rather negative on this 🙁

More conventional and concise both seem like plus points for the approach. Are you concerns based around lack of familiarity with it? The convention seems to be used already by a [bunch](https://github.com/alphagov/content-data-admin/blob/9f93f0c7013b7eb2654427645d5f699b8abe3c3e/config/initializers/time_formats.rb) [of](https://github.com/alphagov/whitehall/blob/174677869d949f59523848869aaa0a20bc93c392/config/initializers/formats.rb) [govuk](https://github.com/alphagov/frontend/blob/bc692b6bebc9e29f1a4fcc113630914cc1aa8634/config/initializers/date_formats.rb) [projects](https://github.com/alphagov/publishing-api/blob/4028a4e690d1ac5b1ca0b9fce268b2f0bd892d03/config/initializers/datetime.rb) so it doesn't feel like introducing special knowledge to this app.

> The existing helpers aren't great either, due to the overloaded naming.
> 
> I think there's a separate issue around naming. I'm afraid `date_at_time` threw me until I saw the "at" in the definition. Perhaps the simplest improvement we could make to this is to adopt `time_on_date` as `date_time` everywhere, noting that the `date_at_time` variant is only used in one non-debug scenario, which seems like an oversight, rather than something we need to preserve.

Yeah, there's a few issues with the existing ones. I was also surprised to find we had the date_at_time one and agree it could be removed. We seem to be in a muddle with the time_on_date one where we have the helper available for the whole thing but tons of locale strings that use time and date variables individually (I don't think we actually have an instance of `format_time` being used outside of a date on time string, so there is opportunity to simplify those

It'd be good to have better names that were more precise, which I think ends up being slightly easier with the Rails approach since you get to avoid some of the overhead of needing to describe what you are formatting (the format_date part) but it's a tough naming challenge without a wider context. Wordings that describe our most commonly used date formatting seem to be friendly, conversational or long.

> This is nice! I may have to "borrow" it!
> Just one naming suggestion, but nothing blocking.

Thanks Leena 😄

config/initializers/time_formats.rb
leenagupte: I think the `%P` means that it always shows "am" or "pm" right? So the name would be consistent for this format. I'm old, so I still call it "12 hour clock" 😄

kevindew: Yup yup - that's officially called a period though no-one uses that term in colloquial language. 12 hour clock is a classic phrase 😁

benthorner:
@kevindew sorry, I should say it's always useful to learn about a new technique. My concerns are around maintenance and consistency. Although I'm comfortable using the technique, the following sequence of events seems quite likely:

1. Developer wants to format a date/time in a view
1. Developer can't find a helper to do it, thinks it's a new problem
1. Developer writes a helper because that's conventional too

So while I agree it is more conventional, the problem for me is that there are two competing conventions here. Of the examples you gave, it looks like [two](https://github.com/alphagov/publishing-api/blob/4028a4e690d1ac5b1ca0b9fce268b2f0bd892d03/config/initializers/datetime.rb) [are](https://github.com/alphagov/frontend/blob/bc692b6bebc9e29f1a4fcc113630914cc1aa8634/config/initializers/date_formats.rb) unused. So we could put this in, and I wouldn't be surprised if we still had time/date helpers in the future.

Perhaps I'm being overcautious?

kevindew:
> @kevindew sorry, I should say it's always useful to learn about a new technique. My concerns are around maintenance and consistency. Although I'm comfortable using the technique, the following sequence of events seems quite likely:
> 
> 1. Developer wants to format a date/time in a view
> 2. Developer can't find a helper to do it, thinks it's a new problem
> 3. Developer writes a helper because that's conventional too
> 
> So while I agree it is more conventional, the problem for me is that there are two competing conventions here. Of the examples you gave, it looks like [two](https://github.com/alphagov/publishing-api/blob/4028a4e690d1ac5b1ca0b9fce268b2f0bd892d03/config/initializers/datetime.rb) [are](https://github.com/alphagov/frontend/blob/bc692b6bebc9e29f1a4fcc113630914cc1aa8634/config/initializers/date_formats.rb) unused. So we could put this in, and I wouldn't be surprised if we still had time/date helpers in the future.
> 
> Perhaps I'm being overcautious?

I'm not too concerned about that scenario myself as that implies no-one involved is aware of existing conventions. Similarly the inverse of that situation could happen if someone expects a time formatter not aware there are helpers. So that should just seems to be a question of ongoing care, knowledge and consistency.

kevindew:
I'll merge this in now and do some clean up after.

PR #2004, opened 2020-04-14
Show message when there are no attachments

benthorner:
https://trello.com/c/a5SwDzjX/1440-show-featured-attachment-metadata-when-reordering-and-on-the-document-summary

I've made the copy consistent with 'No topics', instead of using
what's in the prototype.

### DIFF
--- a/app/views/documents/show/_featured_attachments.html.erb
+++ b/app/views/documents/show/_featured_attachments.html.erb
@@ -1,4 +1,4 @@
-<%= render "govuk_publishing_components/components/summary_list", {
+<% attributes = {
   id: "attachments",
   title: t("documents.show.featured_attachments.title"),
   borderless: true,
@@ -7,9 +7,12 @@
       { href: featured_attachments_path(@edition.document),
         data_attributes: { gtm: "edit-attachments" } }
     end
-  ),
-  items: (
-    @edition.featured_attachments.each_with_index.map do |attachment, index|
+  )
+} %>
+
+<% if @edition.featured_attachments.any? %>
+  <%= render "govuk_publishing_components/components/summary_list", attributes.merge(
+    items: @edition.featured_attachments.each_with_index.map do |attachment, index|
       {
         field: index + 1,
         value: sanitize(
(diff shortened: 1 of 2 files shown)


app/views/documents/show/_featured_attachments.html.erb
```diff
@@ -7,15 +7,22 @@
...
           "<h4 class='govuk-heading-s govuk-!-margin-bottom-0'>#{attachment.title}</h4>#{render('components/attachment_meta', { attachment: file_attachment_attributes(attachment, @edition) })}"
         ),
       }
     end
-  )
-} %>
+  ) %>
+<% else %>
+  <%= render "govuk_publishing_components/components/summary_list", attributes.merge(
+    block: t("documents.show.featured_attachments.none")
```
kevindew, line 26: Hmm it seems quite strange of this component to accept a block argument that can be a string 🤔

benthorner: A string is valid HTML, no? 😉

kevindew: Ha no I meant in the sense that you expect this to be a Ruby block. It seems strange that it can be something yielded and passed as an argument.

In comparison to Rails content_for helper where you have an argument of content that is the alternative to passing a block.

benthorner: Oh, I see what you mean. The component [supports both approaches](https://github.com/alphagov/govuk_publishing_components/blob/master/app/views/govuk_publishing_components/components/_summary_list.html.erb#L7). I was reading it as 'HTML block', and I can now see why it might be confusing. Really, I just [nicked it from here](https://github.com/alphagov/content-publisher/blob/master/app/views/documents/show/_lead_image.html.erb#L3).

kevindew: Yeah sorry, I was commenting on the component. I get the impression it was done with the `||=` as that is the way most component variables are defined but does lead to odd syntax.

It seems to be the only component that has it so might be worth changing it, though it is rather nitpicky of me to introduce a breaking change over 🤔

benthorner: I'm on the fence here, really. I agree it gives unnecessary flexibility. We could avoid using it, but then it's also quite convenient. I'm going to leave it as-is for now.

kevindew: Sorry I don't think I expressed myself well. I think functionality is good but argument name is bad - should probably be an argument called content.

PR #2005, opened 2020-04-14
Fix allowing empty primary orgs

benthorner:
Previously we added validations on the JSON stored for the 'tags',
which has uncovered a bug where '[""]' was being stored when the
user had no primary publishing organisation (or cleared it). This
fixes the field to properly cleanup the input and also changes it
to behave similarly to a MultiTagField when the param isn't there.

### DIFF
--- a/app/models/document_type/primary_publishing_organisation_field.rb
+++ b/app/models/document_type/primary_publishing_organisation_field.rb
@@ -8,7 +8,8 @@ class DocumentType::PrimaryPublishingOrganisationField
   end
 
   def updater_params(_edition, params)
-    { primary_publishing_organisation: params[:primary_publishing_organisation] }
+    value = params[id] && params[id].map(&:presence).compact
+    { id.to_sym => value }.compact
   end
 
   def pre_update_issues(_edition, params)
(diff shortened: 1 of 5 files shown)


benthorner:
@kevindew while the new JSON validations are working as expected, we didn't get a Sentry issue for this. I know you modified the allowed exceptions recently. Would we expect to see an issue?

app/models/document_type/primary_publishing_organisation_field.rb
```diff
@@ -8,7 +8,8 @@ def payload(edition)
   end
 
   def updater_params(_edition, params)
-    { primary_publishing_organisation: params[:primary_publishing_organisation] }
+    value = params[id] && params[id].map(&:presence).compact
+    { id.to_sym => value }.compact
```
kevindew, line 12: Maybe not for today - but could this be easier if the form didn't submit an array?

benthorner: Well, we store it as an array, so probably not on the whole. It's another one of those situations where we (can) only partially trust the submitted params.

PR #2008, opened 2020-04-14
Add requirements checks for official docs

benthorner:

In doing the work for this PR, I noticed that the [EditionChecker](https://github.com/alphagov/content-publisher/blob/master/lib/requirements/edition_checker.rb) has no spec, despite not being (fully) covered at a higher level. It's difficult to add one due to the lack of a common interface for the checkers it delegates to. For the purpose of this PR, I've avoided making any changes to the structure of checkers, and will look at how to write a test for this in a subsequent PR.

- [ ] Work out how to test the [EditionChecker](https://github.com/alphagov/content-publisher/blob/master/lib/requirements/edition_checker.rb)

Screenshot 2020-04-17 at 16 37 07
Screenshot 2020-04-17 at 16 37 27
Screenshot 2020-04-17 at 16 37 43

### DIFF
--- a/app/interactors/file_attachments/create_interactor.rb
+++ b/app/interactors/file_attachments/create_interactor.rb
@@ -25,8 +25,9 @@ private
   end
 
   def check_for_issues
-    issues = Requirements::FileAttachmentChecker.new(file: params[:file], title: params[:title])
-                                                .pre_upload_issues
+    issues = Requirements::FileAttachmentUploadChecker
+      .new(file: params[:file], title: params[:title]).pre_upload_issues
+
     context.fail!(issues: issues) if issues.any?
   end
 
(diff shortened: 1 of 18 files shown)


benthorner:
> Only thing I'm a bit concerned about is the loops of items to create tests (I have déjà vu from the early days of this project when this was a hot debate).

@kevindew that's a fair comment, as we don't do it elsewhere. I think the main concern with this technique is the complexity of the test being repeated (low in this case, higher in previous discussions) and the data used for the repetition (single strings). I also agree that other ways (I can think) of doing this are less good:

- Using a shared example (more abstraction, it's not really shared)
- Putting all the checks in one `it` (similarly unconventional, harder to structure)
- Writing out all the tests (tired fingers)

PR #2017, opened 2020-04-20
Use a common interface for requirements checkers

benthorner:

This is a proposal to reorganise the requirements namespace to improve consistency between the checkers, which then makes the [(former) EditionChecker](https://github.com/alphagov/content-publisher/blob/master/lib/requirements/edition_checker.rb) easier to test:

- Split and reorganise the checkers by use case ("preview", "publish" or "form")
- Introduce a top-level "Checker" class, common to all checkers

Please see the commits for further details. Sorry about the volume of the first commit: it's a big change, and splitting it across commits would just make any comments harder to address, if they involve a fundamental change to the proposal.


kevindew:
Thanks for raising Ben.

So If I understand correctly all issues will have the `.issues` interface (and ones without such as [FileAttachmentMetaChecker](https://github.com/alphagov/content-publisher/pull/2017/files#diff-0e718e913fac5eca78c3b20049e563b8R8) are just part of the WIP aspect). Namespacing is used for different contexts so you've got a form, preview and publish namespaces.

Would we expect a form checker to perhaps call a preview one as a means of sharing, do you anticipate rules/conventions? I'm just wondering how boxed in we might find ourselves for sharing logic?

I like the direction this is going in :+1:

Suggestion wise, I seem to remember early discussions involving using a `.call` interface - I think that would feel more idiomatic than that the `.issues` one. It may also offer some advantages for mocking/code re-use than a checker class can be substituted by something callable (but not too sure we'll see value there)

Suggestion wise I'd suggest that this could be an opportunity to lose the repetitive `issues = CheckerIssues.new` that we have. The base class could be something like this:

```
class Requirements::Checker
  private_class_method :new

  def self.call(*args)
    instance = new(*args)
    instance.check
    instance.issues
  end

  def issues
    @issues ||= CheckerIssues.new
  end
  
  def check
    raise "implement me"
    # implementation could be something like:
    # return if edition.access_limit.nil?
    #
    # if user.organisation_content_id.blank?
    #   issues.create(:access_limit, :user_has_no_org)
    #   return
    # end
    #
    # if user_is_not_in_access_limit_orgs?
    #   issues.create(:access_limit, :not_in_orgs)
    # end
  end
end
```

benthorner:
@kevindew thanks for taking a look! I was thinking along the same lines with DRYing up the checkers, which is now done in subsequent commits. I've also updated the commits/PR.

> Sorry about the volume of the first commit

Just thought I'd emphasise that bit. Could you take a look? 😊

> Would we expect a form checker to perhaps call a preview one as a means of sharing, do you anticipate rules/conventions? I'm just wondering how boxed in we might find ourselves for sharing logic?

We could do. I haven't needed to as part of this PR, and it might be more trouble than it's worth if the form checker needs specific arguments. Time will tell.

lib/requirements/form/content_checker.rb
```diff
@@ -6,13 +6,9 @@ def initialize(edition, params)
     @params = params
   end
 
-  def issues
-    issues = Requirements::CheckerIssues.new
-
+  def check
     edition.document_type.contents.each do |field|
-      issues += field.pre_update_issues(edition, params)
+      self.issues += field.form_issues(edition, params)
```
kevindew, line 11: This seemed a little surprising. How would you feel about a merge method on issues that combined a collection? I assume we don't ever need to actually reset the issues to an empty collection.

lib/requirements/preview/edition_checker.rb
```diff
@@ -0,0 +1,23 @@
+class Requirements::Preview::EditionChecker < Requirements::Checker
+  attr_reader :edition
+
+  CHECKERS = [
+    Requirements::Preview::ContentChecker,
+    Requirements::Preview::TagsChecker,
+    Requirements::Preview::ImagesChecker,
+  ].freeze
```
kevindew, line 8: I wonder if we might end up seeing some autoloading problems with this approach at some point - feels like the sort of thing that may lead to a cyclic dependency.

No problem with it going in now but may need to consider using strings or an instance method if we hit it.

lib/requirements/form/content_checker.rb
benthorner: Yes, it was a bit annoying to have to use `self`. I did try out a `merge!` method on CheckerIssues, so the above line would become

```
issues.merge!(field.form_issues(edition, params))
```

Is that what you were thinking?

There wasn't much in it for me: not using a mutable `!` method seemed clearer, and it meant I didn't need to modify the CheckerIssues class.

lib/requirements/preview/edition_checker.rb
benthorner: Sure, I see what you mean. I do prefer using the constant name, as then we get a clearer error if it's been mis-typed. Like you say, we can always fix it if need be.

kevindew: Yup, that's a good point about catching the undefined constant. Lets just cross our fingers no dependencies become cyclic 🤞

lib/requirements/form/content_checker.rb
kevindew: Yup - just like that but without the bang (since there isn't a non-bang merge method). 

I think it'd be worth the small change to resolve the subtle wtfs the assignment approach caused (for example, I was quite surprised we had an attr_accessor)

benthorner: I'd like to suggest an alternative: `push`. The reason I avoided `merge` is that I'm used to its immutable behaviour for hashes, so although there's only one method, it would be surprising to me. Whereas `push` already has mutable semantics for `Array`.

Any particular concern with using `push`?

kevindew: I'm happy with push 👍

benthorner: I've added a commit for that: https://github.com/alphagov/content-publisher/pull/2017/commits/0ae2c8883b973a77dbd265dc5c8ecd348f2e3959. Can you take a look?

PR #2019, opened 2020-04-21
Expand on time formatting work

kevindew:
Following on from https://github.com/alphagov/content-publisher/pull/2002 this makes a number of improvements to time formatting:

- it changes the time handling in the debug view to use the more common approach of timestamp formatting of time_on_date;
- it removes the common repetition of passing time and date into translation strings only for them to be rendered with the time_on_date formatting within the translation string;
- it removes the date_at_time formatting rule and applies that manually in the one instance it is used (timeline entries) to remove it as a competing rule.

More information about each in the commits

I had hoped to entirely remove the date_at_time style of formatting but when considering it I felt this wasn't a decision that should be made without a designers involvement as it changes the precedence of date when reading this page (which felt like a bad thing).

I'd also hoped that I could get rid of the time formatting rule since it only seemed to be used as part of a time_on_date rule - however, I found out it is used to render times individually for scheduling.

### DIFF
--- a/app/views/debug/index.html.erb
+++ b/app/views/debug/index.html.erb
@@ -19,7 +19,7 @@
       Revision <%= revision.number %><%= edition_text %>
     </h2>
     <p class="govuk-caption-m govuk-!-margin-top-0">
-      Created at <%= revision.created_at.to_s(:date_at_time) %>
+      Created at <%= revision.created_at.to_s(:time_on_date) %>
       <% if revision.created_by %>
         by <%= name_or_fallback(revision.created_by) %>
       <% end %>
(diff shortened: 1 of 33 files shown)


kevindew:
> Looks good to me. Thanks for following-up!
> 
> I do think we could just change the history tab, as it's never actually been fully designed. The order of time vs. date doesn't make a lot of difference if you're visually parsing something.
> 
> ```
> Attachment uploaded
> 23 April 2020 at 4:16pm by User A
> 
> Document updated
> 4 March 2020 at 12:03pm by User B
> ```
> 
> vs.
> 
> ```
> Attachment uploaded
> 4:16pm on 23 April 2020 by User A
> 
> Document updated
> 12:03pm on 4 March 2020 by User B
> ```
> 
> Do you think anyone would notice?

The way I see it is mostly you care about the dates if you're looking at this info so it's a lot easier to read them on the left side and then scan to time if you care to a higher degree.

As for if anyone would notice, no i doubt it. But I'd still not want to make it seem slightly worse for it's purpose off my own back.

PR #2026, opened 2020-04-24
Update previews when changing schedule proposal

kevindew:
Since 9392d8dcd5a0524bfd2c21c8e8c485b57cd84cb3 we now mark all editions
with revision_synced of false when using the EditDraftEditionService.
This means that when we change proposed publish time we are shown the
actions view to indicate a preview was not successful.

Previously I believe that we didn't update the preview on these actions
as these changes didn't actually affect the Publishing API. However it's
probably best not to treat this as something special.

### DIFF
--- a/spec/requests/schedule_proposal_spec.rb
+++ b/spec/requests/schedule_proposal_spec.rb
@@ -16,6 +16,8 @@ RSpec.describe "Schedule Proposal" do
   end
 
   describe "POST /documents/:document/schedule-proposal" do
+    before { stub_any_publishing_api_put_content }
+
     let(:edition) { create(:edition) }
     let(:schedule_params) do
       tomorrow = Time.zone.tomorrow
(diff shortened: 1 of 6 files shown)


spec/requests/schedule_proposal_spec.rb
```diff
@@ -72,6 +74,10 @@
   end
 
   describe "DELETE /documents/:document/schedule-proposal" do
+    before { stub_any_publishing_api_put_content }
+
+    let(:edition) { create(:edition) }
```
benthorner, line 79: Can you help me understand where this has come from?

kevindew: Argh sorry I think that's just a copy and paste error.

kevindew: Sorted

PR #2028, opened 2020-04-24
Restrict requirements to featured attachments

benthorner:
Previously we added new requirements checks that should only apply
to featured attachments, as defined by the type of document they're
associated with. Although the checks primarily run in the context
of the form, which is only available for featured attachments, we
also run these checks at the 'edition level' before publish. This
fixes the edition level check to reflect the type of attachment.

### DIFF
--- a/spec/lib/requirements/file_attachment_metadata_checker_spec.rb
+++ b/spec/lib/requirements/file_attachment_metadata_checker_spec.rb
@@ -1,14 +1,16 @@
 RSpec.describe Requirements::FileAttachmentMetadataChecker do
-  let(:checker) { described_class.new }
-
   describe "#pre_publish_issues" do
+    let(:document_type) { build :document_type, attachments: "featured" }
+
     it "returns no issues if there are none" do
+      checker = described_class.new(build(:edition, document_type: document_type))
       attachment_revision = build(:file_attachment_revision, official_document_type: "unofficial")
       issues = checker.pre_publish_issues(attachment_revision)
       expect(issues).to be_empty
     end
 
     it "returns an issue when the official document type is blank" do
+      checker = described_class.new(build(:edition, document_type: document_type))
       attachment_revision = build(:file_attachment_revision)
       issues = checker.pre_publish_issues(attachment_revision)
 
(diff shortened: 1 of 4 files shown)


spec/lib/requirements/file_attachment_metadata_checker_spec.rb
```diff
@@ -1,14 +1,23 @@
...
       issues = checker.pre_publish_issues(attachment_revision)
       expect(issues).to be_empty
     end
 
+    it"returns no issues unless the document type supports featured attachments" do
+      checker = described_class.new(build(:edition))
+      attachment_revision = build(:file_attachment_revision)
+      issues = checker.pre_publish_issues(attachment_revision)
+      expect(issues).to be_empty
+    end
```
ChrisBAshton, line 17: I'd move this test below the next one, so that it's clearer that the absence of `official_document_type` would ordinarily raise issues.

benthorner: That wouldn't make much difference to me. I think I used to put effort into making tests tell a story, but then I realised I seldom read them top-to-bottom. Since this is causing a bug on the master branch, could we postpone this discussion?

spec/lib/requirements/file_attachment_metadata_checker_spec.rb
```diff
@@ -1,14 +1,23 @@
...
+    let(:document_type) { build :document_type, attachments: "featured" }
+
     it "returns no issues if there are none" do
+      checker = described_class.new(build(:edition, document_type: document_type))
       attachment_revision = build(:file_attachment_revision, official_document_type: "unofficial")
       issues = checker.pre_publish_issues(attachment_revision)
       expect(issues).to be_empty
     end
 
+    it"returns no issues unless the document type supports featured attachments" do
```
ChrisBAshton, line 12:
```suggestion
    it "returns no issues unless the document type supports featured attachments" do
```

spec/lib/requirements/file_attachment_metadata_checker_spec.rb
ChrisBAshton: Just spotted a formatting issue too - perhaps you could reorder at the same time as fixing that? 😇 
It's not a deal-breaker though, I appreciate wanting to fix this quickly.

spec/lib/requirements/file_attachment_metadata_checker_spec.rb
benthorner: Sure, updated.

PR #2054, opened 2020-05-14
Explicitly specify which gems can update

benthorner:
Previously we specified versions of each dependency in the Gemfile,
and relied on DependaBot to automatically update them. This changes
the specified versions to explicitly state which dependencies can
freely update, and which remain relatively static.

We are proposing to adopt a new usage of DependaBot, which will not
raise a PR to update a dependency if that would require a change in
the Gemfile. The aim of this change is to limit the time we spend
updating dependencies, by only updating the following:

- Our internal dependencies
- Rails
- PostGreSQL
- RSpec

We can always update any dependency manually, at the point of need.
The reason for auto-updating some dependencies are:

- We write a lot of code for/with them. Postponing updates could
make it impractical to apply those updates in the future.

- We frequently benefit from the using the latest releases of these
dependencies.

- We frequently refer to the documentation for these dependencies,
which is usually limited or defaulted to the latest version.


kevindew:
Thanks for raising this 👍 So does this correspond with something we turn on in dependabot and do we know how that GitHub Security dependabot interacts with that?

A few thoughts that come to mind are:

It does feel a bit arbitrary why some have version constraints when others don't, perhaps it'd make sense to group them together and have a line break (and perhaps a comment):

```
gem "gds-api-adapters"
gem "gds-sso"
gem "govspeak"
gem "govuk_app_config"
gem "govuk_publishing_components"
gem "govuk_sidekiq"
gem "mail-notify"
gem "pg"
gem "plek"
gem "rails"

gem "aws-sdk-s3", "~> 1.64.0"
gem "bootsnap", "~> 1.4.0"
...
```
By the way I don't think mail-notify fits the criteria here, as it's a third party gem.

It does sit a little wrong for me specifying the minor versions. Makes me feel like I'll miss being able to run a straight `bundle update` and have it bump minor versions, admittedly - since dependabot - it's been ages since I've done that. I imagine it'll actually be fine to work with.

benthorner:
@kevindew thanks for taking a look 😄. This is just an example, which I'm not intending to merge, so we could figure out the details of layout in some documentation, if we get to that point.

> does this correspond with something we turn on in dependabot

Yes, we're proposing to turn on the "Only lockfile updates (ignore updates that require Gemfile changes)" option.

> I don't think mail-notify fits the criteria here

Oh yeah. I thought it was one of our own. I've updated that.

> Makes me feel like I'll miss being able to run a straight bundle update and have it bump minor versions

This is the point, though: we're saying we don't want to update these dependencies (bar patches). Or if we do, we'd be doing it specifically, at the point of need. That's the premise here, anyway.

> do we know how that GitHub Security dependabot interacts with that?

I wasn't sure what you meant initially. Then I found [this doc](https://help.github.com/en/github/managing-security-vulnerabilities/configuring-automated-security-updates#about-automated-security-updates), but the feature appears to be [somewhat broken](https://github.com/alphagov/whitehall/network/alert/Gemfile.lock/actionview/open/update-errors/25918240). I'm assuming we're covered by our own DependaBot for (patch) security updates.

kevindew:
> @kevindew thanks for taking a look 😄. This is just an example, which I'm not intending to merge, so we could figure out the details of layout in some documentation, if we get to that point.
> 
> > does this correspond with something we turn on in dependabot
> 
> Yes, we're proposing to turn on the "Only lockfile updates (ignore updates that require Gemfile changes)" option.

Cool cool, are we going to apply the same with package.json?

> > Makes me feel like I'll miss being able to run a straight bundle update and have it bump minor versions
> 
> This is the point, though: we're saying we don't want to update these dependencies (bar patches). Or if we do, we'd be doing it specifically, at the point of need. That's the premise here, anyway.

Yeah, just feels a bit more arduous to try bring everything back up to date manually. But I understand that's half the trade off

> > do we know how that GitHub Security dependabot interacts with that?
> 
> I wasn't sure what you meant initially. Then I found [this doc](https://help.github.com/en/github/managing-security-vulnerabilities/configuring-automated-security-updates#about-automated-security-updates), but the feature appears to be [somewhat broken](https://github.com/alphagov/whitehall/network/alert/Gemfile.lock/actionview/open/update-errors/25918240). I'm assuming we're covered by our own DependaBot for (patch) security updates.

Sure, it can be a bit cryptic. There are two different dependabot users: [dependabot](https://github.com/alphagov/content-publisher/pulls?q=is%3Apr+author%3Aapp%2Fdependabot+is%3Aclosed) and [dependabot-preview](https://github.com/alphagov/content-publisher/pulls?q=is%3Apr+author%3Aapp%2Fdependabot-preview+is%3Aclosed) where the former is github that will raise PRs on (I think) any repo and the latter is our configured dependabot.

I don't think we'd be quite covered by our dependabot for patch updates as, if we're a few versions behind, a gem maintainer may not make a patch release for our dependency. But hopefully GitHub would swoop in and bump the Gemfile. Although I'm saying this without knowing whether our regular dependabot would do this anyway.

benthorner:
@kevindew I hadn't thought about our JS repos, thought I don't see why we can't apply the same heuristic to them. There are a couple of risks to this approach:

- A locked gem may restrict the version of an unlocked gem. I've checked this for Rails, and all the version specs are upgrade-friendly. Don't think this will be a problem.

- A locked gem may fall out of support for security patches. GitHub DependaBot [doesn't cope with conflicting dependencies](https://talk.jekyllrb.com/t/dependabot-cannot-update-to-the-required-version/2974); and I'm not sure if DependaBot would override the lockfile restriction for security updates.

Options I can see:

- Ask DependaBot how they reconcile security updates with lockfile restrictions.

- Move to config files in each repo, instead of making changes in the Gemfiles

```
version: 1
update_configs:
  - package_manager: "ruby:bundler"
    directory: "/"
    update_schedule: "daily"
    allowed_updates:
      - match:
          update_type: "security"
      - match:
          dependency_name: "rails"
      - match:
          dependency_name: "pg"
      - match:
          dependency_name: "rspec-rails"
```

Although the config files would mean a lot of duplication, that's arguably already there in the UI we're currently using. The config file approach is also more flexible.

What do you think?

kevindew:
> @kevindew I hadn't thought about our JS repos, thought I don't see why we can't apply the same heuristic to them. 

Sounds good

There are a couple of risks to this approach:
> 
> * A locked gem may restrict the version of an unlocked gem. I've checked this for Rails, and all the version specs are upgrade-friendly. Don't think this will be a problem.
> * A locked gem may fall out of support for security patches. GitHub DependaBot [doesn't cope with conflicting dependencies](https://talk.jekyllrb.com/t/dependabot-cannot-update-to-the-required-version/2974); and I'm not sure if DependaBot would override the lockfile restriction for security updates.

Sure, I think the GitHub dependabot would (as I don't think the configuration applies to it) but would still have the conflicting problem, as you say, and the whole thing is rather mysterious.

> Options I can see:
> 
> * Ask DependaBot how they reconcile security updates with lockfile restrictions.

That sounds like something good to find out 👍 

> * Move to config files in each repo, instead of making changes in the Gemfiles
> 
> ```
> version: 1
> update_configs:
>   - package_manager: "ruby:bundler"
>     directory: "/"
>     update_schedule: "daily"
>     allowed_updates:
>       - match:
>           update_type: "security"
>       - match:
>           dependency_name: "rails"
>       - match:
>           dependency_name: "pg"
>       - match:
>           dependency_name: "rspec-rails"
> ```
> 
> Although the config files would mean a lot of duplication, that's arguably already there in the UI we're currently using. The config file approach is also more flexible.
> 
> What do you think?

Yeah, I quite like a config file approach since it's managed through code. It feels like something where https://github.com/alphagov/govuk-saas-config could help us out in monitoring/maintaining too (I guess we'd want everything at a minimum having security updates for instance)

I also like that it'd be more explicit than the gemfile style approach and does offer some flexibility (ie, when Publishing Workflow team is off pause it's not really much of a bother keeping every dependency up-to-date, compared to say the misery of that approach on an app that is rarely touched)

benthorner:
@kevindew I've had a chat with @thomasleese, who's already done similar changes in [some of our other repos](https://github.com/alphagov/content-tagger/pull/1099/files), primarily as an experiment. We agreed that we want to revisit this with the aim of applying a common policy to every app on GOV.UK, using GOV.UK SaaS Config to enforce it. @thomasleese is going to write an RFC to propose this.

In the meantime, we agreed that I would supersede this PR with one that can, if approved, act as a showcase for the RFC. Could you review that, when you have time?

Superseded by: https://github.com/alphagov/content-publisher/pull/2062

PR #2072, opened 2020-05-22
Add rake task to repopulate scheduled jobs

kevindew:
This can be used to add any schedulings for publishing to Sidekiq and
has been prepared in anticipation of losing Redis data during a server
migration.

If there are already jobs scheduled this will add them again, this won't
cause any actual problems since the publishing job is idempotent.
However, if we were to run this a lot we'd create rather a lot of noise
in Redis so this isn't advised.

I've included some documentation for this which includes info of our
common risk with scheduled publishing that has mostly lived in status
docs before. I've included some rather cumbersome instructions on how to
fix a failed scheduled publishing were one of those to occur. It seemed
more prudent to provide cumbersome UI steps than to suggest running code
since the task can be achieved with the UI and the usage is expected to
be rare.

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,6 +52,7 @@ bundle exec rake jasmine
 - [History mode](docs/history-mode.md)
 - [Importing documents from Whitehall](docs/import-from-whitehall.md)
 - [Removing documents](docs/removing-documents.md)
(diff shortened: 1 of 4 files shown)


kevindew:
@benthorner thanks and yup, downtime is part of their plan.

PR #2082, opened 2020-06-02
Use array contains operator in contains query

kevindew:
This changes the query used to determine if a tag appears inside an
array of tags to use the rather cryptic "@>" operator that Postgres
provides.

My motivation for changing this is I had planned to explain a bug we
experienced based on the previous query and I realised I very much
struggled to understand how that worked, in particular where "value"
came from in the `array_to_json(array[value])->>0 = ?`.

I realise this is all rather subjective as it replaces a cryptic query
with another cryptic query but I hope the intent is clearer by use of
the contains operator.


kevindew:
D'oh I see we can't use this as we're still on Postgres 9.3 (I thought it was 9.4). I'll close this until we're on 9.6 everywhere.

PR #2093, opened 2020-06-15
Revert "Bump govuk_publishing_components from 21.55.0 to 21.55.3"

barrucadu:
Reverts alphagov/content-publisher#2091


barrucadu:
I'll be able to hop on the VPN and massage this through CI in a little while

PR #2106, opened 2020-07-01
Backfill backwards compatible auth_bypass_id

kevindew:
Trello: https://trello.com/c/IQ26eiaW/240-make-content-publisher-use-changing-auth-bypass-tokens

This is marked as draft until https://github.com/alphagov/content-publisher/pull/2105 has been merged and deployed to production.

This PR backfills the editions table to populate auth_bypass_id values for every row. Once this migration is complete all entries will have an auth_bypass_id so we can also set it with a NOT NULL db constraint.

In the next PR for this I'll make use of this column in code.


kevindew:
Thanks @benthorner 👍

PR #2108, opened 2020-07-02
Enhance auth-bypass approach

kevindew:
Trello: https://trello.com/c/IQ26eiaW/240-make-content-publisher-use-changing-auth-bypass-tokens

This builds on the work of https://github.com/alphagov/content-publisher/pull/2105 and https://github.com/alphagov/content-publisher/pull/2106 to add in a UUID that is generated for each edition as an auth_bypass_id. If we have a situation where a token is leaked we can reset this value to generate new tokens for an edition.

This also updates the approach to auth_bypass tokens to reflect the guidance in [authenticating-proxy](https://github.com/alphagov/authenticating-proxy#generating-a-auth-bypass-token) and approach used in [collections-publisher](https://github.com/alphagov/collections-publisher/blob/622a6e1f50ca1e8ce337f5e447e649d9ea05ed4e/app/helpers/application_helper.rb#L14-L25). Where tokens expire in a month and include the content id.

I went through a few changes and personal debates about whether to include the user_uid at all, in plain text or encrypted. I decided to go with not including it for now and be something to consider adding if it becomes some important information in a future incident.


kevindew:
@benthorner after opening this I'm wondering if it might be better to get rid of the PreviewAuthBypass class and just have an auth_bypass_token method on edition since it's only essentially one methods worth of code.

PR #2111, opened 2020-07-07
Don't LoadError for i18n task in production

benthorner:
Since Rails will load this task automatically, we need to prevent
a LoadError in production, where the 'i18n-coverage' gem is not
available. This wraps the whole task definition in a conditional,
since it only makes sense to define the task in dev/test.

### DIFF
--- a/lib/tasks/i18n_cov.rake
+++ b/lib/tasks/i18n_cov.rake
@@ -1,3 +1,5 @@
+return unless defined?(I18n::Coverage)
+
 require "i18n/coverage/printers/file_printer"
 
 namespace :i18n_cov do


lib/tasks/i18n_cov.rake
```diff
@@ -1,16 +1,18 @@
-require "i18n/coverage/printers/file_printer"
+unless Rails.env.production?
```
kevindew, line 1: You can do `return unless Rails.env.production?` to save nesting the whole file.

I'd suggest doing `unless defined?(I18n::Coverage)` rather than the more circumstantial environment (assuming that at this point bundler has already required files)

benthorner: Sure, I've updated and run a test deploy in Integration.

PR #2113, opened 2020-07-10
Upgrade rails to 6.0.3

benbaumann95:


Gemfile
```diff
@@ -2,53 +2,53 @@ ruby File.read(".ruby-version").strip
 
 source "https://rubygems.org"
 
-gem "rails", "~> 6.0"
+gem "rails", "~> 6.0.3", ">= 6.0.3.2"
```
kevindew, line 5: How come we need this extra level of specificity for Rails? If we just had `gem "rails"` we'd end up with the same result of 6.0.3.2 so not sure why this needs to be treated different to every other gem.

Gemfile
```diff
@@ -2,53 +2,53 @@ ruby File.read(".ruby-version").strip
...
-gem "pg", "~> 1"
-gem "plek", "~> 4"
-gem "rinku", "~> 2"
-gem "rubyzip", "~> 2", require: "zip"
-gem "sanitize", "~> 5"
-gem "sass-rails", "< 6"
-gem "sidekiq-scheduler", "~> 3"
-gem "uglifier", "~> 4"
-gem "with_advisory_lock", "~> 4"
+gem "aws-sdk-s3"
```
kevindew, line 7: It'd be useful to have the commit message that is removing the pins to explain why they're being removed.

Gemfile
benbaumann95: Hi there. Please see https://github.com/alphagov/service-manual-frontend/pull/749#discussion_r454243642 as why we have taken this approach.

PR #2114, opened 2020-07-10
Remove unnecessarily strict time Cop

benthorner:
https://github.com/alphagov/rubocop-govuk/issues/73

Previously we added this config on the basis that "Time.current"
was not a supported mode [1], but it is. This removes the custom
config, noting that this Cop is enabled by default [2].

[1]: https://github.com/alphagov/content-publisher/pull/1820
[2]: https://docs.rubocop.org/rubocop-rails/cops_rails.html#railstimezone

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -16,9 +16,5 @@ Rails/DynamicFindBy:
 Rails/SaveBang:
   Enabled: true
 
-Rails/TimeZone:
-  Enabled: true
-  EnforcedStyle: "strict"
-
 Style/DateTime:
   Enabled: true


kevindew:
Thanks for opening, it's nice that we're getting these consistent. I don't think your comment captures the scenario around adding this accurately. Our previous problem was that we had inconsistent usage of `Time.zone.now` and `Time.current` dotted around interchangeably as the linter was happy with either, setting this to strict meant only one was allowed, `Time.zone.now`.

As for `Time.current` not being a supported mode, that is in reference to the fact that the linter only allows you to specify a preference for `Time.zone.now` whereas you cannot specify the alternative.

It's bit of an annoying default rule, given it leaves developers with a choice. It might make sense to be ported over to rubocop-govuk to avoid customisation in this project alone.

benthorner:
Hmm...OK. Closing in favour of https://github.com/alphagov/rubocop-govuk/pull/98.

PR #2118, opened 2020-07-16
Add Dependabot configuration

ChrisBAshton:
In accordance with [RFC 126], we're configuring Dependabot to only
raise Pull Requests for the following three categories:

1. Security updates
2. Internal libraries
3. Framework libraries

Things have moved on slightly since the RFC; we're now using
[GitHub native] configs rather than the legacy `.dependabot`
approach. The syntax is slightly different to the example config
provided in the RFC but most of it is easily transferable.

We've handled (2) and (3) in config, and have set it to only raise
PRs for direct updates, i.e. updates to subdependencies of the
named dependencies will not raise PRs.

We've chosen a `govuk*` wildcard to cover things like
`gds-api-adapters` and `govuk_app_config` and other unknown future
dependencies. Having `govuk` as a [suffix is not encouraged][wildcards]
so instead of a `*govuk` wildcard we've named `rubocop-govuk`
explicitly. We explicitly name dependencies beginning with `gds`
[for the same reason][discussion].

Trello: https://trello.com/c/uPoriyfJ/2049-add-dependabot-configuration-to-each-repo-blitz-pair

[discussion]: https://github.com/alphagov/content-store/pull/750#discussion_r454208436
[GitHub native]: https://docs.github.com/en/github/administering-a-repository/enabling-and-disabling-version-updates
[security]: https://github.community/t/how-to-get-dependabot-to-trigger-for-security-updates-only/117257/5
[RFC 126]: https://github.com/alphagov/govuk-rfcs/blob/master/rfc-126-custom-configuration-for-dependabot.md
[wildcards]: https://github.com/alphagov/content-store/pull/750#discussion_r453693082


.github/config.yml
```diff
@@ -0,0 +1,47 @@
...
+        dependency-type: direct
+      - dependency-name: rails
+        dependency-type: direct
+      - dependency-name: rspec-rails
+        dependency-type: direct
+      - dependency-name: sass-rails
+        dependency-type: direct
+  - package_manager: npm
+    directory: /
+    update_schedule: daily
```
kevindew, line 37: this doesn't seem to match the schedule definition in bundler.

ChrisBAshton: Whoops, this is what happens when making error-prone manual changes to 70 apps 🤦 thanks for catching!

ChrisBAshton:
Yes, the filename change is hinted at in the commit message. It will still override the Dependabot UI config.

`update_type: security` is not listed anymore, that is now handled on the repo's [security settings tab](https://github.com/alphagov/content-publisher/security). I think it's enabled by default - it's enabled on this repo at any rate.

Unfortunately `dependency-type: direct` is required, otherwise [every dependency in the lockfile will be updated]https://docs.github.com/en/github/administering-a-repository/configuration-options-for-dependency-updates#allow).

kevindew:
> Yes, the filename change is hinted at in the commit message. It will still override the Dependabot UI config.

Super 👍 
> 
> `update_type: security` is not listed anymore, that is now handled on the repo's [security settings tab](https://github.com/alphagov/content-publisher/security). I think it's enabled by default - it's enabled on this repo at any rate.

Oh cool that might be something we can manage with govuk-saas-config

> Unfortunately `dependency-type: direct` is required, otherwise [every dependency in the lockfile will be updated](https://docs.github.com/en/github/administering-a-repository/configuration-options-for-dependency-updates#allow).

Hmm I can see in the example there they have one without `dependency-type: direct` (NPM) and one one with python. Do we know that we get a bad outcome by not specifying this? From my reading it doesn't sound like it'd do all of the dependencies, maybe all of the ones that matched the pattern, but then that's what we want? Sorry to be a pain, just seems the sort of thing that'd be annoying to undo once applied everywhere if it turns out not to be necessary.

ChrisBAshton:
Tom & I were looking at this a few days ago. It's a shame that the docs aren’t clearer…! This was the thought process I had on Slack:

> In the absence of an allow list, Dependabot will keep everything in the lockfile up to date (i.e. dependencies and subdependencies)

> When an allow list is provided, *I don’t know whether it overrides the entire list* (i.e. only update what you have explicitly put in the allow list) or if it just gives you *tighter control over the packages you name* (i.e. it will still update any packages you haven’t named) …although it does say if you provide an allow and an ignore and an item appears on both lists, it will be ignored (so it’s NOT a case of having to ignore everything and then ‘opt-in’ the things we want to allow).

> I’m assuming that if we provide an allow list, any items not named are ignored. 🤞 

> **But even with these, it doesn’t say what the default behaviour is (e.g. direct/indirect/all…). I’m assuming direct just because that’s the first option documented in the docs.**

So I think I arrived at '`direct` is _probably_ default, but just to be safe let's be explicit about it'.

I'm not too worried about a slightly noisy config, as I'm anticipating we'll go in and delete all of these configs once [org-wide support](https://github.com/dependabot/dependabot-core/issues/2193) is implemented. But I'm happy to try omitting `dependency-type` and see what Dependabot does.

kevindew:
Sure, yeah the docs have a frustrating lack of "this is the default".

Yeah happy to go for that then until we learn otherwise, can always delete the lines later if it transpires they're superfluous.

ChrisBAshton:
Thanks @kevindew - I pushed up an extra commit that removes those lines, just before you approved. Do you think we should proceed with the lines removed and see what happens, or add the lines back in until we know more?

kevindew:
> Thanks @kevindew - I pushed up an extra commit that removes those lines, just before you approved. Do you think we should proceed with the lines removed and see what happens, or add the lines back in until we know more?

Happy to go with whichever way would be more consistent with the rest of GOV.UK repos. I'd prefer the shorter version, but would prefer the more verbose definitions if that is the consistent way.

PR #2129, opened 2020-07-29
Use stylelint for scss-linting and switch Ruby Sass to Sassc

kevindew:
This continues the trial of [stylelint-config-gds](https://github.com/alphagov/stylelint-config-gds) by applying this to a GOV.UK application. For further details on this see https://github.com/alphagov/govuk_publishing_components/pull/1611. I chose to use Content Publisher because:

- it's one of our more modern applications so the expectation was that there wouldn't be too much SCSS to change 
- it's not under active development so any [sassc govuk-docker performance issues](https://trello.com/c/H9wrGrwQ/161-investigate-if-libsass-rails-performance-is-bad-with-govuk-docker) are unlikely to be experienced widely
- It's my favourite GOV.UK app ❤️

The most significant change in this PR is switching the approach to importing CSS files into the SCSS. Previously we were using the native CSS `@import` statement and triggering an additional request. This instead imports the file via SCSS so it's all part of the same singular css file that is served to a user. Details are in this commit: https://github.com/alphagov/content-publisher/commit/25ab60cba2eb3cbc0a3e6a977685dffbbb1613dc

Next steps for this stream of work is to explore resolving the govuk-docker performance issues and to talk more with the design system about how we can move stylelint-config-gds beyond a pre-release. I'll probably also try doing similar on an older GOV.UK application to see the scale of SCSS changes needed.


app/assets/stylesheets/components/_autocomplete.scss
```diff
@@ -1,4 +1,4 @@
-@import url(asset-path("accessible-autocomplete/dist/accessible-autocomplete.min.css"));
```
alex-ju, line 1: The old sass compiler interpreted these as CSS `@import` statements, which is why we had to use this twisted approach. The new one seems to interpret the import in an expected way. Great to see this hack removed.

app/assets/stylesheets/components/_autocomplete.scss
```diff
@@ -13,7 +14,7 @@ $app-focus-colour: $govuk-focus-colour;
 .app-c-autocomplete--search {
   .autocomplete__input {
     // scss-lint:disable PlaceholderInExtend
-    @extend .gem-c-input--search-icon;
+    @extend .gem-c-input--search-icon; // stylelint-disable-line scss/at-extend-no-missing-placeholder
```
alex-ju, line 16: Hope to have a placeholder for this, so we can remove the exception.

app/assets/stylesheets/components/_reorderable-list.scss
```diff
@@ -98,23 +98,21 @@
...
+  .app-c-reorderable-list__actions .govuk-form-group {
+    display: none;
+  }
 
-    .gem-c-button {
-      display: inline-block;
-      margin-left: govuk-spacing(3);
-      width: 80px;
-    }
+  .app-c-reorderable-list__actions .gem-c-button {
```
alex-ju, line 105: The nesting level on these is a sign we should try to refactor/simplify this component a bit in the near future.

PR #2136, opened 2020-08-07
Remove the word "placeholder" from image checker

danacotoran:
The Whitehall to Content Publisher migration code uses the presence of the word "placeholder" to determine if we're using a default image.

Whitehall has [recently been updated](https://github.com/alphagov/whitehall/pull/5756) to remove "placeholder" as the
default for image alt text as this was an accessibility fail. The alt text field should now be blank by default

### DIFF
--- a/lib/whitehall_importer/integrity_checker.rb
+++ b/lib/whitehall_importer/integrity_checker.rb
@@ -241,7 +241,7 @@ module WhitehallImporter
     def default_image?(proposed_image_payload, publishing_api_image, attribute)
       attribute == "alt_text" &&
         proposed_image_payload.empty? &&
-        publishing_api_image[attribute] == "placeholder"
+        publishing_api_image[attribute].empty?
     end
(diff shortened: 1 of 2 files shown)


lib/whitehall_importer/integrity_checker.rb
```diff
@@ -241,7 +241,7 @@ def empty_caption?(proposed_image_payload, publishing_api_image, attribute)
     def default_image?(proposed_image_payload, publishing_api_image, attribute)
       attribute == "alt_text" &&
         proposed_image_payload.empty? &&
-        publishing_api_image[attribute] == "placeholder"
+        publishing_api_image[attribute] == ""
```
brucebolt, line 244:
```suggestion
        publishing_api_image[attribute].empty?
```

danacotoran: Have updated!

PR #2155, opened 2020-09-09
Update edit.yml

Tobiogunsina:
I'm updating this to be consistent with the new guidance for images - https://www.gov.uk/guidance/content-design/images

### DIFF
--- a/config/locales/en/images/edit.yml
+++ b/config/locales/en/images/edit.yml
@@ -10,7 +10,7 @@ en:
         credit: Image credit
         lead_image: Select as lead image
       guidance:
-        alt_text_govspeak: A simple and specific description of what the image shows. This helps screen-readers and search engines.
+        alt_text_govspeak: Optional. A short and specific description of what the image shows. This helps screen-readers and search engines. Use this only when the image provides a reader with additional information.
         caption_govspeak: Optional. A description of what the image shows. Caption appears next to the image.
         credit_govspeak: |
           Images licensed from a 3rd party must be credited. Government images do not need a credit.


config/locales/en/images/edit.yml
```diff
@@ -10,7 +10,7 @@ en:
         credit: Image credit
         lead_image: Select as lead image
       guidance:
-        alt_text_govspeak: A simple and specific description of what the image shows. This helps screen-readers and search engines.
+        alt_text_govspeak: If your image conveys information,  describe the information in the main body text and leave the alt text field blank. Do not describe decorative images.
```
kevindew, line 13: It might be useful to explain when someone should use alt text? With this guidance update it doesn't sound like I should ever use the field.

How does the guidance interact with caption?

pollygreen:
I agree, I'd suggest leaving in the original text and adding to it if
there's space.

Polly Green
Senior content designer
GOV.UK
Government Digital Service

config/locales/en/images/edit.yml
Tobiogunsina: The guidance is encouraging publishers not to use alt text at all but to put the image description in the body image with the aim that if all publishers gradually do that, then we can [retire the field and figure out a workflow for ensuring image descriptions](https://docs.google.com/document/d/163q3Ys_pRbKgVMqPgwBld6MTDTm5A1aQ_QoSnNOXo-M/edit#heading=h.v5nhdg8g3owu) are added in. 

Caption are still optional, they are not read reliably by screen readers and are not generally for describing the image' information, more like title or something else related to the image

kevindew: Ok sure, so if we don't want people to use the field should it just be removed? It feels a little counter-intuitive to provide an optional field but then recommend people don't use it. Are there any situations where we'd want people to fill this in?

Tobiogunsina: There isn't currently. It's the long term plan to remove it. We can't just remove it though, We need to encourage departments to put their image descriptions in content first and have some way to encourage them to do so in the workflow. If we remove it now, then we may lose the alt text that's currently available before there's a description in the content

kevindew: I see, hmm this seems a new unprecedented pattern. After all you'd only receive this guidance if you were to look at optional field. It also feels rather strange than we'd be asking for the description in a completely different part of the app.

Is this something we should have a designer get involved with?

Should the alt text field be replaced with a description that can be output with the image instead of needing people to do different things in different places? You'd have to manage both things together :-/

Tobiogunsina: Yes. Exactly. The workflow needs to be designed, especially for Content Publisher where the image workflow is different. There were some draft plans for next steps in the kick off but we don't have the capacity to do it before the deadline so it'll need to be after. There were also some considerations that meant we can't just replace it with a description that can be output. It needs some thinking

My concern was that the instructions in Content publisher is now diverging from the guidance but maybe we should leave it in until we've worked out a better workflow. Lots of images in News stories are probably decorative so may be we just change it to "**A simple and specific description of what the image shows. This helps screen-readers and search engines. Do not describe decorative images**." :-/

kevindew: Great @Tobiogunsina sounds good to meet this in the middle while there isn't time for the wider work.

I wonder if it's possible to move away from a "do not" to a when to "use this" as I think that would feel more positive and easier to understand (I personally struggle with what the definition of a decorative image is - it has to have different meanings). Though I appreciate that isn't easy so I've tried to have a go myself 🙈 .

I did also wonder if we could copy [captions](https://github.com/alphagov/content-publisher/blob/db8f92bd36d5b9aeca313950287bf0cae2fc5439/config/locales/en/images/edit.yml#L14) and prefix with "Optional".

"Optional. A short and specific description of what the image shows. This helps screen-readers and search engines. Use this only when the image provides a reader with additional information."

Does that help at all?

alex-ju: Oh, difficult topic – according to standards (W3C and WHATWG) the alt text is a must (the only exception being made only for decorative images where a blank string should be used as value). Now, 'decorative' is subjective and we may consider most images on GOV.UK as decorative if they don't convey information. Technically, not providing an alt text (which results in us presenting an ``) is ignored by _most_ screen readers, but a small combination of assistive tech and browsers would still read out: 'image' when they encounter it. I feel this is more of a policy decision than a technical check.

Tobiogunsina: Great Idea @kevindew , I like the idea of adding optional like captions and I like the text you've proposed too. 

@alex-ju The theory is that since Alt text is only available for screen reader users, image descriptions in body text will make them available to more users (everybody including with users with other accessibility needs who might not be able to view the image properly) and encourage better image descriptions. 

We think with proper descriptions where needed to convey the same information, screen reader users should not need to engage with the images at all. That way it's fine for alt to be left blank because having image description both in content and in alt text becomes auditory clutter for them

It was endorsed by accessibility specialists in GDS and Anika has checked [what happens in different screen readers](https://docs.google.com/spreadsheets/d/1j7EXXUBg0TWg5Xidc-mTRA0fC0SFWRuiidmRCR8QfKk/edit#gid=437542015)

That said, it seems to be a controversial approach and given more time, we should have tested it specifically with screen reader users so that we don't make it worse for them in the bit to make it better for everyone

Tobiogunsina: @kevindew My github use is rusty, do I just create a new pull request with the updated text?

kevindew: @Tobiogunsina I've updated your commit with the different text. You can amend it further by editing https://github.com/alphagov/content-publisher/blob/Tobiogunsina-patch-1/config/locales/en/images/edit.yml as a new commit which will get added to this PR.

Text I've amended looks like this in app:

Screenshot 2020-09-11 at 15 58 58

Tobiogunsina: Thank you! @kevindew Looks fine to me

alex-ju (approved):

PR #2156, opened 2020-09-09
Ensure we check coverage for all code in tests

benthorner:
As specified by the SimpleCov docs [1].

[1]: https://github.com/simplecov-ruby/simplecov#getting-started

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -1,12 +1,15 @@
 ENV["RAILS_ENV"] ||= "test"
 ENV["GOVUK_APP_DOMAIN"] = "test.gov.uk"
 
+# Must go at top of file
+require "simplecov"
+SimpleCov.start "rails"
+
 require File.expand_path("../config/environment", __dir__)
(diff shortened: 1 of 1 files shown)


spec/spec_helper.rb
```diff
@@ -1,12 +1,15 @@
 ENV["RAILS_ENV"] ||= "test"
 ENV["GOVUK_APP_DOMAIN"] = "test.gov.uk"
 
+# Must go at top of file
+require "simplecov"
+SimpleCov.start "rails"
```
kevindew, line 6: Ooh what does the "rails" do?

kevindew (approved):

spec/spec_helper.rb
benthorner: It appears to set [a custom profile](https://github.com/simplecov-ruby/simplecov#profiles) that does a bit of filtering. Moreover, it also seems to modify the way it collects coverage - I got very different results without it in Smart Answers!

PR #2172, opened 2020-10-07
Fix bug where you can't preview scheduled documents

kevindew:
This resolves a bug where we asserting that an edition had to be
editable to be previewable, this is incorrect as you can preview
scheduled editions. This is easy enough to fix.

A bit more painful was deciding what process to take to add a regression
test for this - as this seemed a relatively easy mistake to make as most
work we do asserts an edition is editable rather than it being pre-live.
We don't really have a pattern to exhaustively test state assertions.

I added the interactor test for this action since there wasn't one
already and validated that the edition assertion test returned the
appropriate message. This seemed sufficient to me as you couldn't change
this without re-considering the logic of the condition.

I then added an explicit test at the request point for the GET endpoint
since this isn't tested at a unit level and this makes the need for this
feature explicit.

### DIFF
--- a/app/interactors/preview/create_interactor.rb
+++ b/app/interactors/preview/create_interactor.rb
@@ -18,7 +18,7 @@ private
 
   def find_and_lock_edition
     context.edition = Edition.lock.find_current(document: params[:document])
-    assert_edition_state(edition, &:editable?)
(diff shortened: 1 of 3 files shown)


1pretz1 (approved):
Looks good to me! 👍

PR #2189, opened 2020-10-27
Fix test always failing on a Tuesday

benthorner:
Previously the test would change behaviour if the beginning of the
week coincided with yesterday. This ensures the backdated date is
always older than the original publishing date.


1pretz1 (approved):
Gblame to find me at the helm 🙈 Thanks for sorting this!

PR #2203, opened 2020-11-19
Extend healthcheck to optionally check S3

benthorner:
https://trello.com/c/ILpSOFhq/201-enable-cd-for-quick-win-and-possible-quick-win-apps

Please see the commit for more details.

I think the argument to make this an optional part of the healthcheck
is relatively weak, but I think it's worth highlighting the potential
concerns so we can explicitly decide if we want to check S3 connectivity
more frequently. Arguments for treating S3 like the other checks:

- Attachments (will) represent a relatively high proportion of user
journeys, so it's simpler to mark the app as unhealthy than let users
continue to use it and experience random errors.

- S3 request pricing is *very* cheap: with 3 backend machines, it amounts
to about £4pa, per app.

- We treat S3 as a robust service and do not handle any failure associated
with it; therefore, we should not expect it to fail, ever.

It's also worth noting the S3 request limit of 3500r/s per bucket means we
are in no danger whatsoever of going over the limit.

### DIFF
--- a/app/controllers/healthcheck_controller.rb
+++ b/app/controllers/healthcheck_controller.rb
@@ -2,11 +2,13 @@ class HealthcheckController < ApplicationController
   skip_before_action :authenticate_user!
 
   def index
-    healthcheck = GovukHealthcheck.healthcheck([
+    checks = [
       GovukHealthcheck::SidekiqRedis,
       GovukHealthcheck::ActiveRecord,
       Healthcheck::GovernmentDataCheck,
-    ])
-    render json: healthcheck
+    ]
+
+    checks << Healthcheck::ActiveStorage if params[:storage]
+    render json: GovukHealthcheck.healthcheck(checks)
   end
 end
--- a/lib/healthcheck/active_storage.rb
+++ b/lib/healthcheck/active_storage.rb
@@ -1,0 +1,14 @@
+module Healthcheck
+  class ActiveStorage
+    def name
+      :active_storage
+    end
+
+    def status
+      ::ActiveStorage::Blob.service.exist?("does-not-exist")
+      GovukHealthcheck::OK
+    rescue StandardError
+      GovukHealthcheck::CRITICAL
+    end
+  end
+end
--- a/spec/lib/healthcheck/active_storage_spec.rb
+++ b/spec/lib/healthcheck/active_storage_spec.rb
@@ -1,0 +1,14 @@
+RSpec.describe Healthcheck::ActiveStorage do
+  describe "#status" do
+    it "returns OK when connected to the storage service" do
+      expect(described_class.new.status).to eq GovukHealthcheck::OK
+    end
+
+    it "returns CRITICAL when the storage connection fails" do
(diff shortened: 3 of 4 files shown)


app/controllers/healthcheck_controller.rb
```diff
@@ -2,11 +2,13 @@ class HealthcheckController < ApplicationController
...
-    healthcheck = GovukHealthcheck.healthcheck([
+    checks = [
       GovukHealthcheck::SidekiqRedis,
       GovukHealthcheck::ActiveRecord,
       Healthcheck::GovernmentDataCheck,
-    ])
-    render json: healthcheck
+    ]
+
+    checks << Healthcheck::ActiveStorage if params[:storage]
```
kevindew, line 11: What's the purpose of the param here?

benthorner: Oh, maybe it's not clear from the test/commit: this PR is a suggestion for how we could implement an optional check for S3 connectivity as part of the existing `/healthcheck` endpoint.

I've for this in the commit, and against on the PR description, which I guess is a bit confusing! Hopefully it makes sense now, though. I've made this a draft PR, and will tidy up if we reach an agreement on the approach.

kevindew: Thanks, it didn't really answer my query I'm afraid. I'm a bit confused why this would be something optional as ActiveStorage should work in all environments (even those without S3) so I don't know why we'd need it to be optional. Maybe this concern is different for other apps that directly communicate with S3?

benthorner: Ah, so this is because there are multiple uses of the `/healthcheck` endpoint:

- LB healthcheck, which seems to be roughly every 10s, per machine.
- Alert healthcheck (every 5mins, per machine).
- (New) Smokey healthcheck, as part of enabling CD.

We need to add the check as part of enabling CD, but it's not clear to me if we should use it as a signal for the app being unavailable, which how the healthcheck gets interpreted when requested by the LB.

So I've suggested a compromise of making it an optional check on the endpoint. In the description I've argued why it could also be just like any other healthcheck.

Does that make more sense? (I can update the PR/commit to explain better.)

kevindew: Ok thanks, I think I understand. 

I'm not particularly keen on params to turn on different aspects of the healthcheck if we don't want to care about whether they pass or not in some situations. This feels to me like a conflation of the three states of healthcheck: ok, warning and critical. It feels like if we're not sure we care about this then it should be a warning rather than a critical?

benthorner: Fair enough - I wasn't set on the approach, and was also considering a WARNING status. Given the problems with the description and approach, I've raised [a fresh PR](https://github.com/alphagov/content-publisher/pull/2204) with that approach.

Could you take a look at that one?

benthorner:
Closing in favour of https://github.com/alphagov/content-publisher/pull/2204.

PR #2257, opened 2021-02-10
Favour GOV.UK Docker for running the app

benthorner:
https://trello.com/c/Psf7EzPB/204-move-away-from-startupsh

This removes legacy "startup.sh" files from this project in favour
of using GOV.UK Docker, which provides more functionality. In order
to keep the README environment-agnostic, we avoid prefixing commands
like "bundle exec rake" with "govuk-docker-run". However, any setup
commands are removed as these are covered by GOV.UK Docker Makefiles,
and are seldom sufficient for comprehensive local development.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,16 +11,13 @@ A unified publishing application for content on GOV.UK
 
 ## Technical documentation
 
-This is a Ruby on Rails application. You can use the GOV.UK Docker environment to run the application and its tests with all the necessary dependencies; **see the [README](https://github.com/alphagov/govuk-docker#usage) for more details**.
+This is a Ruby on Rails app, and should follow [our Rails app conventions](https://docs.publishing.service.gov.uk/manual/conventions-for-rails-applications.html).
 
-### Dependencies
+You can use the [GOV.UK Docker environment](https://github.com/alphagov/govuk-docker) to run the application and its tests with all the necessary dependencies. Follow [the usage instructions](https://github.com/alphagov/govuk-docker#usage) to get started.
 
-- [postgresql][] - provides a backing database
-- [redis][] - used as a storage layer for asynchronous job processing
-- [yarn][] - package manager for JavaScripts
-- [imagemagick][] - image manipulation library
+**Use GOV.UK Docker to run any commands that follow.**
 
-### Running the application
+### Before running the app
 
 The first time you run this application for development, enable `debug` and `pre_release_features` permissions:
 
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ A unified publishing application for content on GOV.UK
 
 ## Technical documentation
 
-This is a Ruby on Rails application.
+This is a Ruby on Rails application. You can use the GOV.UK Docker environment to run the application and its tests with all the necessary dependencies; **see the [README](https://github.com/alphagov/govuk-docker#usage) for more details**.
 
 ### Dependencies
 
@@ -33,8 +33,6 @@ To enable them for your GOV.UK account add them to your account in [Signon](http
 ### Running the test suite
 
 ```
-yarn install
-
 # ruby tests
 bundle exec rspec
 
--- a/Procfile.dev
+++ b/Procfile.dev
@@ -1,2 +1,0 @@
-web: BYEBUG_PORT=3237 bundle exec rails s -p 3221
-worker: BYEBUG_PORT=3238 bundle exec sidekiq -C ./config/sidekiq.yml
--- a/startup.sh
+++ b/startup.sh
@@ -1,5 +1,0 @@
-#!/bin/bash
-
-bundle install
-gem install foreman --conservative
-foreman start -f Procfile.dev


kevindew (approved):

README.md
```diff
@@ -11,7 +11,7 @@ A unified publishing application for content on GOV.UK
 
 ## Technical documentation
 
-This is a Ruby on Rails application.
+This is a Ruby on Rails application. You can use the GOV.UK Docker environment to run the application and its tests with all the necessary dependencies; **see the [README](https://github.com/alphagov/govuk-docker#usage) for more details**.
```
kevindew, line 14: This sounds good however I wonder how it contrasts with the headers of "Running the application" and "Running the test suite". It seems to be ok in this app because we seem to just have supplementary information under those headers but may need to decide on a different approach for other apps - so may need to wonder whether the [underlying template headers differ](https://docs.publishing.service.gov.uk/manual/setting-up-new-rails-app.html#replace-the-default-readmemd)

kevindew (commented):
This looks good and resolves long running discrepancies 👍

README.md
benthorner: Yes, it's a bit awkward. I think we actually have a mixture here:

- "Running the app" is supplementary.
- "Running the tests" is more directly about what it says.

I assume by "other apps" you were thinking more like [this](https://github.com/alphagov/email-alert-api#running-the-application)? 

I'd like to avoid peppering the READMEs with `govuk-docker-*`. On the other hand, it could cause confusion unless someone's fully embedded the idea that `govuk-docker-*` must prefix everything.

Perhaps a middle ground is to add a comment e.g.

```
# prefix "govuk-docker-run" if using GOV.UK Docker
bundle exec rake

# or just "govuk-docker-up" if using GOV.UK Docker
bundle exec rails server
```

I did also wonder if the headings are necessary in general, now that we have `rake` setup properly for everything. We'll still have the `startup.sh` scripts in frontend apps, so arguably not much.

Did you have anything particular in mind?

kevindew: > Yes, it's a bit awkward. I think we actually have a mixture here:
> 
> * "Running the app" is supplementary.
> * "Running the tests" is more directly about what it says.
> 
> I assume by "other apps" you were thinking more like [this](https://github.com/alphagov/email-alert-api#running-the-application)?

Yup exactly, since this app seems a bit special.

> Did you have anything particular in mind?

Nope, haven't got anything particular. I'm trying to ponder what can be relatively minimal with either maintaining these headers as per the template - or whether the structure and templates should be adjusted. However adjusting the template would probably increase the scope so might not be great if there isn't a slam dunk approach.

It might be worth having a quick think about what we'd want the whole "Technical documentation" to say if we were to be starting today given that we have govuk-docker.

I'm not sure we get a huge amount of value from "dependencies" as they tend to be incomplete and arbitrary on many apps - and for govuk-docker apps that information is already there (albeit not super clear). There's a delegated command to run the app, `govuk-docker-up`, so that may not need documentation. It might still be useful to know how to run particular tests - though `rake` is ubiquitous.

A quick pass at what I think might be sufficient to be helpful could be:

> ## Technical Documentation
> 
> You can use the [GOV.UK Docker environment](https://github.com/alphagov/govuk-docker) to run the application and its tests with all the necessary dependencies. Once [configured](https://github.com/alphagov/govuk-docker#usage), the app can be run with `govuk-docker up content-publisher-app` and the test suite can be run with `govuk-docker run content-publisher-lite bundle exec rake`.
> 
> ### Configure permissions in development
> 
> To have all functionality available in development the default user account needs extra permissions. These can be added with:
> 
> ```
> govuk-docker run content-publisher-lite bundle exec rake development_permissions
> ```

and perhaps for Email Alert API:

> ## Technical Documentation
> 
> You can use the [GOV.UK Docker environment](https://github.com/alphagov/govuk-docker) to run the application and its tests with all the necessary dependencies. Once [configured](https://github.com/alphagov/govuk-docker#usage), the app can be run with `govuk-docker up email-alert-api-app` and the test suite can be run with `govuk-docker run email-alert-api-lite bundle exec rake`.

It seems tricky trying to reach a spot where the information is sufficient to be helpful, lacks ambiguity and to have as few duplications as possible. 

Anyway I hope this comment has helped move us towards a solution rather than further into a can of worms 🤞

benthorner: Thanks for the suggestions. I'm also cautious about this project, as it could turn into "rewriting the README", and while that would be nice, it's more effort than necessary. Things I'd like to be true (for non-frontend apps):

- The README points to GOV.UK Docker for running code.
- The README doesn't specify govuk-docker commands directly.

The reasons for the latter are:

- To be friendly to people _not_ using GOV.UK Docker.
- To make it easier to change GOV.UK Docker commands.
- To avoid confusing people who do/n't use the shortcuts (like `govuk-docker-run`).

Something more like this would be compatible with that:

> # Technical Documentation
>
> You can use the [GOV.UK Docker environment](https://github.com/alphagov/govuk-docker) to run the application and its tests with all the necessary dependencies. Follow [the usage instructions](https://github.com/alphagov/govuk-docker#usage) to get started.
>
> [vague intros removed ([example](https://github.com/alphagov/publishing-api#technical-documentation)), supplementary info retained (e.g. links, docs)]
>
> **Use GOV.UK Docker to run any commands that follow.**
>
> ## Running the application
>
> [only supplementary info]
>
> ## Running the tests
>
> [unchanged]

This is something of a hybrid of what you were suggesting:

- Remove "dependencies" section. It's often out-of-date, and not very informative. It's also an easy change to incorporate.

- Remove "running the app" in general. Otherwise someone might run `govuk-docker-run bundle exec rails` or similar, which doesn't work (port not exposed).

- Otherwise, preserve the sections as they are now. Given the scale of the change, this reduces the effort and should help to avoid pitfalls.

The dev doc could change to make the "technical documentation" section less rigid. "running the tests" could be a suggested section, as could others like "architecture and docs". Repos I've mentally run this against:

- [smart-answers](https://github.com/alphagov/smart-answers)
- [email-alert-api](https://github.com/alphagov/email-alert-api)
- [publishing-api](https://github.com/alphagov/publishing-api#technical-documentation) (headings are a mess)
- [whitehall](https://github.com/alphagov/whitehall)
- [bouncer](https://github.com/alphagov/bouncer) (this will be a bit of an exception)
- [hmrc-manuals-api](https://github.com/alphagov/hmrc-manuals-api)
- [content-data-api](https://github.com/alphagov/content-data-api)

How does that sound? On the whole I think it's a net improvement, but also still a light touch one. Whoever applies the changes can of course do more tweaking, but I'd like to keep the scope tight if possible.

kevindew: > Thanks for the suggestions. I'm also cautious about this project, as it could turn into "rewriting the README", and while that would be nice, it's more effort than necessary. Things I'd like to be true (for non-frontend apps):
> 
> * The README points to GOV.UK Docker for running code.
> * The README doesn't specify govuk-docker commands directly.
> 
> The reasons for the latter are:
> 
> * To be friendly to people _not_ using GOV.UK Docker.
> * To make it easier to change GOV.UK Docker commands.
> * To avoid confusing people who do/n't use the shortcuts (like `govuk-docker-run`).

Yup, make sense 👍  - seems reasonable goals, I'm not too concerned about the "not using GOV.UK Docker" one as I imagine this is mostly basic guides for people new to a project. However I guess we have to deal with that anyway with frontend apps and startup scripts so may as well be consistent.

> Something more like this would be compatible with that:
> 
> > # Technical Documentation
> > You can use the [GOV.UK Docker environment](https://github.com/alphagov/govuk-docker) to run the application and its tests with all the necessary dependencies. Follow [the usage instructions](https://github.com/alphagov/govuk-docker#usage) to get started.
> > [vague intros removed ([example](https://github.com/alphagov/publishing-api#technical-documentation)), supplementary info retained (e.g. links, docs)]
> > **Use GOV.UK Docker to run any commands that follow.**
> > ## Running the application
> > [only supplementary info]
> > ## Running the tests
> > [unchanged]

This sounds good, you might want to change the "Running the application" one to be "Before running the application" since the section doesn't actually tell you how to run the app.

> This is something of a hybrid of what you were suggesting:
> 
> * Remove "dependencies" section. It's often out-of-date, and not very informative. It's also an easy change to incorporate.
> * Remove "running the app" in general. Otherwise someone might run `govuk-docker-run bundle exec rails` or similar, which doesn't work (port not exposed).
> * Otherwise, preserve the sections as they are now. Given the scale of the change, this reduces the effort and should help to avoid pitfalls.

Sounds great.

> The dev doc could change to make the "technical documentation" section less rigid. "running the tests" could be a suggested section, as could others like "architecture and docs". 

Yup, that's a good idea. I think the section is probably too rigid now anyway

I feel like it could be something like:

```
## Technical documentation

Provide the required technical information for someone to get started with running this software. Provide guidance on what dependencies are needed, how to start the software, how to run its tests and any unusual quirks someone may need to consider.
```

If you're happy with that I don't mind raising a PR for developer docs?

benthorner: Great, thanks for reviewing. I agree the "not using GOV.UK Docker" goal isn't particularly important, but I mentioned it based on previous discussions that we should keep the magic to a minimum (e.g. [here](https://github.com/alphagov/govuk-developer-docs/pull/2897#pullrequestreview-551745745)).

I've added another commit with the changes we've been discussing, so I think this is ready for another look. 

> If you're happy with that I don't mind raising a PR for developer docs?

Thanks, that would be helpful, although I think it would be more useful to just give snippets of suggested text/headings/pointers of what good looks like here, to avoid people writing boilerplate.

kevindew: Ok that sounds fair - I will save that PR until you've applied this in a few more places in case we end up tweaking things a little during rollout. 

This all looks good to me now and I'm happy to approve.

kevindew (approved):

README.md
```diff
@@ -11,16 +11,11 @@ A unified publishing application for content on GOV.UK
 
 ## Technical documentation
 
-This is a Ruby on Rails application. You can use the GOV.UK Docker environment to run the application and its tests with all the necessary dependencies; **see the [README](https://github.com/alphagov/govuk-docker#usage) for more details**.
```
ChrisBAshton: Whilst it's common knowledge to GOV.UK developers, I don't think it hurts to explicitly state that a given repo is a Ruby on Rails app. We do after all have some special cases like [router](https://github.com/alphagov/router) or [govuk-dependencies](https://github.com/alphagov/govuk-dependencies), which are not Rails apps, so somebody knew to GOV.UK might appreciate the distinction. For the sake of seven words, I'd probably just leave this in.

README.md
```diff
@@ -11,16 +11,11 @@ A unified publishing application for content on GOV.UK
...
+You can use the [GOV.UK Docker environment](https://github.com/alphagov/govuk-docker) to run the application and its tests with all the necessary dependencies. Follow [the usage instructions](https://github.com/alphagov/govuk-docker#usage) to get started.
 
-### Dependencies
+**Use GOV.UK Docker to run any commands that follow.**
 
-- [postgresql][] - provides a backing database
```
ChrisBAshton, line 18: I agree that this section is often incomplete or out of date, but I do find it useful to know what database an application is using, as it affects the app console syntax I'm able to use (Postgres vs Mongo). I'll sometimes check the repo to figure out which one it's using before I query (though more often than not, I'll just use trial and error in my queries).

It's a shame that the `gds govuk connect app-console` command doesn't expose what the DB is, which would mitigate this concern.

ChrisBAshton (approved):
LGTM, though I have a couple of minor comments below.
Sorry it took so long to get around to reviewing!

README.md
benthorner: I was going to disagree:

- This is actually (7 words) * (most of our repos), which is a classic instance of DRY.
- Crucially, if we're going to have boilerplate, it should be actionable boilerplate.

However, I thought of some documentation this could usefully link to, so I've amended the last commit with it.

Does that address your concern?

README.md
benthorner: Yes, it was tempting to leave this section in. But it's really a symptom of a bigger problem: if there's a need to know about these dependencies, then the repo should have actual documentation about their use. In the case you describe, it sounds like there's a need for "common DB queries" documentation. Simply calling out "it's a mongo thing" only provides a relatively small piece of information on the journey to getting the required data.

I don't think it's going to cause a major issue: someone can easily check `config/` to see what DB is in use. By removing it, we make room for having useful, actionable documentation about it.

README.md
ChrisBAshton: Yes, that looks great. Thanks.

README.md
ChrisBAshton: I think the right place to make a change would be in [govuk-connect](https://github.com/alphagov/govuk-connect), making it state which DB is in use when you open a console. So happy to approve this section being removed 👍

README.md
ChrisBAshton: Did we want to apply those changes to https://github.com/alphagov/email-alert-service/pull/437 and https://github.com/alphagov/email-alert-service/pull/437 too?

benthorner: Sure can, if you'd like to merge and deploy (otherwise Slack will start whining). The line only makes sense for Email Alert API, so I've just updated that (Email Alert Service already has blurb explaining how it works).

PR #2278, opened 2021-03-11
Replace reorderable component with new gem version

DilwoarH:
This change removes the local reorderable component in favour of the global gem version.

This is done to bring the styling and functionality in line with the rest of the publishing components.

This had to be done in one commit as the original component was causing the gem component JS to fail.

There are some styling changes made:
- The up/down buttons is now secondary solid grey where it was secondary outlined before
- The reorderable list items now have a border around them
- The numbering has been removed.

https://trello.com/c/9CQm5A0k/251-update-publishing-tools-to-use-new-reorderable-component

### Before

Screenshot 2021-03-11 at 17 19 31

### After

Screenshot 2021-03-11 at 17 19 12


alex-ju (commented):
Thanks for raising this, Dilwoar! We'd probably want to remove [sortable.js](https://github.com/alphagov/content-publisher/blob/master/package.json#L50) from this repo and use it via GOV.UK Publishing Components.

DilwoarH:
> Thanks for raising this, Dilwoar! We'd probably want to remove [sortable.js](https://github.com/alphagov/content-publisher/blob/master/package.json#L50) from this repo and use it via GOV.UK Publishing Components.

Done thanks for flagging !

kevindew (approved):
Looks great - thanks 👍 . I've checked and the gtm-reorder analytics still works with this too.

PR #2293, opened 2021-03-29
Upgrade config for Rails 6.1

kevindew:
I found all but one of the new config defaults to work with Content
Publisher.

The problematic one is a change to activestorage which looks a pain for
us to fix, so I've deferred fixing it for now. I wouldn't be too
surprised if it eventually becomes a deprecation (Rails 7) so I don't
think I'm free of it forever.

The problem is that we generate variants of images dynamically when
we're updating Asset Manager within a transaction. The new Rails
behaviour doesn't upload the variant files to the third party service
until after a transaction is committed. Thus, a failure occurs where the
file doesn't exist during the transaction.

It seems like our use case is pretty quirky and I imagine the Rails
dev's were only anticipating these variants to be used to display images
and not as part of an API pipeline. The approach that seemed reasonable
to fix this would be to cut out the variant system for generating these
resized images and produce the data for the resized images on demand,
when they are needed. This would save an unnecessary round trip to S3.
However, this doesn't sound super simple so turning off the functionality
seems sufficient for now.


ChrisBAshton (approved):
Agreed, our use case for variants sounds quite niche, and we should defer the problem until it becomes a deprecation.

thomasleese (approved):

PR #2299, opened 2021-04-15
Update rubocop-govuk to 4.0.0.pre.1

kevindew:
Trello: https://trello.com/c/pqXbKZiC/215-test-rubocop-govuk-400pre1-prior-to-400-release

This bumps to the 4.0.0 pre release of rubocop-govuk which is part of [trialling the new rubocop-govuk release](https://github.com/alphagov/rubocop-govuk/issues/129). The gem pin to 4.0.0.pre.1 is intended to be removed once the final release of rubocop-govuk 4.0.0 is out.

The most interesting commits in this are https://github.com/alphagov/content-publisher/commit/5b668b8d94def0bbc07ac810c2b1fb7afce08f2e and https://github.com/alphagov/content-publisher/commit/c85025e769cac28a5b0bf70b1e08aee004a09dec where a class inheritance patterns is replaced with a mixin approach. 


lib/callable.rb
```diff
@@ -0,0 +1,14 @@
+# This mixin is to provide the boilerplate for classes that implement a single
+# public class method of `.call` - typically used by objects that are entirely
+# to perform a business transaction.
+module Callable
+  extend ActiveSupport::Concern
+
+  included do
+    def self.call(*args, **kwargs, &block)
+      new(*args, **kwargs, &block).call
```
ChrisBAshton, line 9: Nitpick: might be better to rename the second parameter to `keyword_args` for clarity. EDIT: I see the same parameter names in `Requirements::Checker`, which I assume you copied this from.

Also there are a couple of lines that seem to have snuck into the commit message and can probably be deleted:

```
Taking inspiration from the Ruby
standard library there is an in
```

ChrisBAshton (approved):
Don't see anything controversial here 👍 LGTM

lib/callable.rb
kevindew: Yeah `kwargs` seems to also be a near ubiquitous convention for generic keyword args naming so I think it's the most idiomatic choice. I don't particularly like it though and I'm very much looking forward to [brakeman gaining support](https://github.com/presidentbeef/brakeman/issues/1449) for [Ruby 2.7 shorthand syntax](https://rubystyle.guide/#arguments-forwarding) so these things can be removed.

Ooh thanks about the commit message, I'll fix that.

PR #2304, opened 2021-04-23
Check ActiveStorage in /healthcheck/ready

barrucadu:
This is a critical dependency, so it should be part of the readiness
healthcheck.

---

[Trello card](https://trello.com/c/tl6RV1el/723-improve-govuk-healthchecks)


kevindew (commented):
Thanks for this - I notice the healthcheck is set to return warning should that be critical?

barrucadu:
Good spot, I'll update that

kevindew (approved):

PR #2355, opened 2021-06-30
Tweak guidance on running JS tests

benthorner:
In response to: https://github.com/alphagov/govuk_publishing_components/pull/2060#discussion_r631761457

This makes the format consistent with other repos. I've also changed
the overall test command to "rake", which is what we have in all our
other repos. If we need to tell people how to check which framework
is being used, we should do this centrally.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,12 +30,17 @@ To enable them for your GOV.UK account add them to your account in [Signon](http
 ### Running the test suite
 
 ```
-# ruby tests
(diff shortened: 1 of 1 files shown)


alex-ju (approved):

benthorner:
Thanks for the review @alex-ju!

Would you like to merge? (I no longer have access to do so.)

PR #2460, opened 2022-02-23
Migrate from Jasmine gem to jasmine-browser-runner

ollietreend:
The Jasmine Ruby gems have been [deprecated][1] and are no longer supported. And whilst they still work with Rails 6 apps, they [aren't compatible with Rails 7][2] which is now becoming a blocker.

This commit implements the `jasmine-browser-runner` npm package, which is now the recommended approach for running Jasmine tests in the browser, and removes the deprecated Ruby gems.

This follows the example set by https://github.com/alphagov/govuk_publishing_components/pull/2625

[1]: https://github.com/jasmine/jasmine-gem#discontinued
[2]: https://github.com/alphagov/govuk_publishing_components/issues/2621#issuecomment-1048007995


spec/javascripts/helpers/fetch-mock.js
```diff
@@ -1,5 +1,8468 @@
 /* global fetchMock */
 
-//= require fetch-mock/es5/client-bundle.js
```
kevindew: This file seems to contain both the vendored code and a customisation (`beforeEach`) which seems a bit dangerous.

I'd suggest that we have separate files. I think it'd be a good idea to make use of a  `spec/javascripts/vendors` for vendored dependencies. I think that we could also move jasmine-jquery there.

Other thought is that it'd be good to record what version of the dependency we have (filename might be best way to express that - i.e. fetch-mock-9.11.0.js. Ideal would be a weblink we can download from that we can reference in a comment that people can use for future updates.

As an aside a whacky counter idea could be to pull the dependency in from a JS module e..g `import fetchMock from "view-source:https://ga.jspm.io/npm:fetch-mock@9.11.0/es5/client-bundle.js"` but rabbit hole / can of worms / dodgy assumptions

spec/support/jasmine-browser.json
```diff
@@ -0,0 +1,23 @@
...
+  ],
+  "helpers": [
+    "helpers/*.js",
+    "vendor/*.js"
+  ],
+  "env": {
+    "stopSpecOnExpectationFailure": false,
+    "stopOnSpecFailure": false,
+    "random": false
+  },
```
kevindew, line 21: I imagine we can remove these.

spec/support/jasmine-browser.json
```diff
@@ -0,0 +1,23 @@
...
+  "cssFiles": [
+    "application-*.css"
+  ],
+  "specDir": "spec/javascripts",
+  "specFiles": [
+    "**/*[sS]pec.js"
+  ],
+  "helpers": [
+    "helpers/*.js",
+    "vendor/*.js"
```
kevindew, line 15: If we do use vendors for deps I imagine it might need to come first in this list (I wonder why they don't have it as first in their examples)

kevindew (commented):
Looking great - I've added a few suggestions.

spec/javascripts/helpers/fetch-mock.js
ollietreend: I did consider the `import` approach – as I know this is something that can be done with Skypack: https://www.skypack.dev/view/fetch-mock

But if I understand it correctly, we'd need the script to be loaded as `<script type="module">`, which I don't think is the case. Very happy to be proven wrong on that, though.

kevindew: I understand it does do something different if you have an .mjs file - presumably turns into a module import.

I’d suggest having a quick test - but I’m expecting we’d find some sort of reason for a dead end. (I’m expecting if things did work we might still hit a scoping issue)

ollietreend: @kevindew what do you think about this approach instead? 51d638b5d6ed52a1359bad7c57cfbf8b500b4ac2

spec/support/jasmine-browser.json
```diff
@@ -0,0 +1,19 @@
...
+  "cssFiles": [
+    "application-*.css"
+  ],
+  "specDir": "spec/javascripts",
+  "specFiles": [
+    "**/*[sS]pec.js"
+  ],
+  "helpers": [
+    "https://unpkg.com/jasmine-jquery@2.1.1/lib/jasmine-jquery.js",
+    "https://unpkg.com/fetch-mock@9.11.0/es5/client-bundle.js",
```
kevindew, line 15: I like this, let's see if anyone has any problems with it.  Nice work sussing this out.

package.json
```diff
@@ -7,7 +7,10 @@
   "scripts": {
     "lint": "yarn run lint:js && yarn run lint:scss",
     "lint:js": "standardx 'app/assets/javascripts/**/*.js' 'spec/javascripts/**/*.js'",
-    "lint:scss": "stylelint app/assets/stylesheets/"
+    "lint:scss": "stylelint app/assets/stylesheets/",
+    "jasmine:prepare": "RAILS_ENV=test bundle exec rails assets:clobber assets:precompile",
+    "jasmine:ci": "yarn run jasmine:prepare && bundle exec rake webdrivers:chromedriver:update && PATH=~/.webdrivers:$PATH yarn run jasmine-browser-runner runSpecs",
```
kevindew, line 12: Shall we go for: 
```suggestion
    "jasmine:ci": "yarn run jasmine:prepare && yarn run jasmine-browser-runner runSpecs",
```
?

I guess we'll get a fail in CI until we install a chromedriver bin there.

kevindew:
@ollietreend I think this might need a rebase? I gave it a kick on CI and it complained about the Gemfile.lock being corrupt.

ollietreend:
@kevindew now rebased & passing in CI 🙂

spec/support/jasmine-browser.json
```diff
@@ -0,0 +1,19 @@
...
+    "helpers/*.js"
+  ],
+  "browser": "headlessChrome"
+}
```
kevindew, line 19: Nit: missing new line

kevindew (approved):
Super, looks great.

spec/support/jasmine-browser.json
ollietreend: Visual Studio Code insists on removing the final newline from JSON files – [apparently deliberately](https://github.com/microsoft/vscode/issues/96112). But I'll use a different editor to add that newline!

PR #2462, opened 2022-03-01
Allow attachments to show accessible form link in preview

edwardkerry:
In order for attachments to render a link to the request accessible format form, we need to supply the attachment_id and the owning_document_content_id. This commit adds these to the attachment_attributes hash so that the correct link may be displayed in the in-app preview.

Note: this change is dependant on [this update being made to govuk_publishing_components](https://github.com/alphagov/govuk_publishing_components/pull/2636)

[trello](https://trello.com/c/rKvWEhgJ/1095-accessible-format-request-content-publisher-add-parent-document-contentid-to-the-attachment-model)


KludgeKML (approved):
Looks okay to me.

PR #2516, opened 2022-05-20
Clarify difference between the two removal tasks

ollietreend:
This change clarifies the difference between the two `remove:` rake tasks in Content Publisher.

Whilst removing and redirecting a document, we found the instructions didn't clearly explain that the `remove:redirect` task would both remove AND redirect the document. It wasn't clear if `remove:gone` needed to be run first or not.

It turns out both tasks will remove the document – so I've tried to tweak the wording to make that more obvious.


1pretz1 (approved):
Awesome, much clearer!

PR #2571, opened 2022-08-30
Remove auto-tagging to legacy-taxons (aka specialist topics)

hannako:
## Old behaviour

When a user tags their document to the topic taxonomy, sometimes the document is auto-tagged to a specialist topic. This happens when the topic taxonomy tag (`taxon`) has been provided with a mapped legacy_taxon value in Content Tagger.

## New behaviour

Do not automatically tag documents to the specialist topic tree.

## No user facing changes

**Related PR and full explanation of why we are removing auto-tagging**: https://github.com/alphagov/whitehall/pull/6646

**Trello**: https://trello.com/c/u5dnc3HW/1285-remove-autotagging-to-legacy-taxons-from-content-publisher-m

### DIFF
--- a/app/models/document_topics.rb
+++ b/app/models/document_topics.rb
@@ -42,7 +42,6 @@ class DocumentTopics
       document.content_id,
       links: {
         taxons: leaf_topic_content_ids + unknown_taxon_content_ids,
-        topics: legacy_topic_content_ids,
       },
       previous_version: version,
     )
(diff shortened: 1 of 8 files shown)


Tetrino (approved):

kevindew (commented):
Changes look good. I think we can also remove the legacy taxons fixture code from: https://github.com/alphagov/content-publisher/blob/2c8caf972938d2bd6d66dcc5aadaf5b05015a3b0/spec/support/topics_helper.rb#L25-L62 as I presume that was only to serve this feature.

PR #2601, opened 2022-09-20
Update to Rails 7

kevindew:

This updates Content Publisher to use Rails 7. It will require a subsequent commit once this is deployed to remove transitory configuration.

It is easiest reviewed commit by commit as all the changes are quite simple.


config/application.rb
```diff
@@ -8,8 +8,8 @@
 require "active_storage/engine"
 require "action_controller/railtie"
 require "action_mailer/railtie"
-require "action_mailbox/engine"
-require "action_text/engine"
+# require "action_mailbox/engine"
+# require "action_text/engine"
```
beccapearce, line 12: Is it worth deleting these rather than commenting them out?

beccapearce (approved):
Looks good just one nitpicky comment but happy to approve as is

config/application.rb
kevindew: That's a good point. In most cases I'd agree but Rails apps seem to have a convention of using comments here rather than deleting lines.

PR #2705, opened 2023-01-05
Add container image build for running on Kubernetes.

sengi:
- Add Dockerfile and .dockerignore
- Fix Rails assets path to include the app name so that assets remain namespaced when uploaded to S3.
- Have Sidekiq log to stdout by default.

Content Publisher is similar to Whitehall in that it has an image (JPEG, PNG etc.) upload feature and expects to be able to exec Imagemagick binaries but doesn't check for them on startup. It also uses Bootsnap and Yarn.

Tested: builds locally on Docker Desktop and runs as far as Puma startup (needs external dependencies to go further so makes sense to test on the integration cluster beyond that point).

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,13 @@
+.dockerignore
+.git
+.gitignore
+Dockerfile
+Jenkinsfile
+Procfile
+README.md
+coverage
+docs
+log
+node_modules
+spec
+tmp
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,34 @@
+ARG ruby_version=3.1.2
+ARG base_image=ghcr.io/alphagov/govuk-ruby-base:$ruby_version
+ARG builder_image=ghcr.io/alphagov/govuk-ruby-builder:$ruby_version
+
+FROM $builder_image AS builder
+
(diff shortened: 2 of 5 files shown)


Dockerfile
```diff
@@ -0,0 +1,35 @@
+ARG ruby_version=3.1.2
```
nsabri1, line 1: Thats good call

nsabri1 (approved):
LGTM

.dockerignore
```diff
@@ -0,0 +1,19 @@
...
+carrierwave-tmp
+coverage
+docs
+features
+log
+node_modules
+spec
+test
+tmp
+uploads
```
kevindew, line 19:
```suggestion
coverage
docs
log
node_modules
spec
tmp
```

It looks like you've picked up a few Whitehall specific quirks in this list, so just I've just suggested reducing it to actual paths this app uses.

sengi: Ah good shout - thanks

sengi:
FWIW I checked manually that this rolled out ok to EC2 and didn't break, just in case the test coverage isn't all that 😅

PR #2984, opened 2023-12-12
Remove Whitehall asset

yuetylauiris:
We are removing whtehall_asset endpoint (and the stubs) from gds-api-adapter. The migration script doesn't seems to be in use anymore so we're removing them as well.

[Trello card](https://trello.com/c/Kp9nquOr/71-task-remove-whitehallasset-concept-from-asset-manager-and-gds-api-adapters)


lib/whitehall_importer/sync.rb
```diff
@@ -22,7 +22,6 @@ def call
       GdsApi.whitehall_export.document_migrated(
         whitehall_import.whitehall_document_id,
       )
-      MigrateAssets.call(whitehall_import)
```
ryanb-gds, line 25: I'm not very comfortable with us deleting this line. In the unlikely event that we ever wanted to do a future migration the migrator might not realise they're missing the assets. I've added a comment to the slack thread about this.

PR #2987, opened 2023-12-13
Remove usage of whitehall asset

yuetylauiris:
We are removing the whitehall_asset endpoint from gds-api-adapters so the script will not be able to run anymore.

To keep most of the code in case it would be needed for future migration, we raise an error and ask to be re-implemented instead. A lot of tests are broken now due to removal of key functionality in migrate assets so we decided to skip them for now so that they can be easily modified and work again if needed.

[Trello card](https://trello.com/c/Kp9nquOr/71-task-remove-whitehallasset-concept-from-asset-manager-and-gds-api-adapters)

### DIFF
--- a/lib/whitehall_importer/migrate_assets.rb
+++ b/lib/whitehall_importer/migrate_assets.rb
@@ -42,11 +42,8 @@ module WhitehallImporter
       whitehall_asset.update!(state: "removed")
     end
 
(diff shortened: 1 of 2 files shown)


ryanb-gds (approved):
Thanks team, I think is the best way to go for now

PR #3019, opened 2024-01-17
Upgrade to Rails 7.1.2

davidgisbey:
## Description

This follows the guidance set out in the EPIC [Trello card](https://trello.com/c/PhGra6VR/1809-epic-rails-712-upgrade) for Rails 7.1.2 upgrades 

I've deployed the changes to integration and it seems to be working as fine when i take all the happy paths.

## Trello card

https://trello.com/c/MvQVHuDI/1811-upgrade-content-publisher-to-rails-712


config/environments/production.rb
```diff
@@ -74,28 +80,23 @@
...
-    config.logger    = ActiveSupport::TaggedLogging.new(logger)
-  end
-
   # Do not dump schema after migrations.
   config.active_record.dump_schema_after_migration = false
 
   config.action_mailer.delivery_method = :notify
   config.action_mailer.notify_settings = {
-    api_key: Rails.application.secrets.notify_api_key,
+    api_key: Rails.application.credentials.notify_api_key,
```
ryanb-gds, line 88: You might find that this doesn't work and that you have to use ENV directly here, as I think this gets loaded before the initialisers. See https://github.com/alphagov/whitehall/blob/b0180e9fb01c7d5168047e457402bdbefb2bbc18/config/application.rb#L72

spec/lib/govspeak_document/in_app_options_spec.rb
```diff
@@ -14,7 +14,7 @@
                              caption: "An optional caption",
                              credit: "An optional credit",
                              filename: "filename.png")
-      edition = build(:edition, image_revisions: [image_revision])
+      edition = create(:edition, image_revisions: [image_revision])
```
ryanb-gds, line 17: This seems... odd that it should be necessary, but I'm not that surprised 🤣

ryanb-gds (changes requested):
I think we'll need to tweak the notify API config to use ENV directly

config/environments/production.rb
davidgisbey: ah right that makes sense. Thanks!

ryanb-gds (approved):
Thanks for sorting this out

PR #3046, opened 2024-02-07
Attempt to fix the Google Analytics

minhngocd:
Making Content Publisher consistent with a few other publishing apps by rendering the partial outside `content_for` method. Assumption is that problems might have been caused by having partials rendered inside a method block which might not work as expected? Testing in integration


minhngocd:
Tested in integration - can see the tag manager scripts being included. This should fix it.

beccapearce (approved):
⭐

PR #3138, opened 2024-05-20
Update feedback link for content publisher

richardTowers:
Content publisher no longer has it's own feedback form, following https://github.com/alphagov/support/pull/1354 so these links were going to 404 pages.

This updates the links to use the General feedback form, which is intended to be used for issues that don't fit into other categories. We probably don't get much feedback through this route, but broken links look bad. So let's fix 'em.

### DIFF
--- a/app/views/layouts/_phase_banner_elements.html.erb
+++ b/app/views/layouts/_phase_banner_elements.html.erb
@@ -7,7 +7,7 @@
 </span>
 <span class="gem-c-phase-banner__banner-item">
   <a class="govuk-link govuk-link--no-visited-state"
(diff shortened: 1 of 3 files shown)


ChrisBAshton (approved):
Thanks - seems a reasonable quick fix, but I'll also investigate removing this link entirely, because the point of removing the content publisher feedback form was that we're no longer really prioritising any of the feedback requests, so we don't really want them filling in feedback via a different route either 😅

ChrisBAshton:
Raised https://github.com/alphagov/content-publisher/pull/3148

PR #3148, opened 2024-05-29
Remove feedback form links from Content Publisher

ChrisBAshton:
In https://github.com/alphagov/support/pull/1354, the Content Publisher specific feedback form was removed, and in #3138 the form link was updated to point to the General support form (not specifically designed for feedback).

Content Publisher is no longer actively maintained and it is [destined for archival](https://github.com/alphagov/govuk-rfcs/blob/main/rfc-161-technical-direction-in-publishing-2023.md) at some point. We are therefore not open to feedback requests and should not be pretending we are - so this commit removes references to giving feedback.

Users can still report technical problems via the linked "Raise a support request" technical fault form.


ryanb-gds (approved):
Good spot, cheers Chris 🙇

PR #3149, opened 2024-05-29
Fix footer links

ChrisBAshton:
These footer links have been broken since version 37.1.1 of govuk_publishing_components, which automatically converts all footer links to absolute URLs, using `Plek.new.website_root`: https://github.com/alphagov/govuk_publishing_components/pull/3699

That corresponds to `https://www.gov.uk` on Production, but needs to be `https://content-publisher.publishing.service.gov.uk`.

No other publishing app passes custom links to the footer like Content Publisher does, so this appears to be the only app affected. With that in mind, the simplest fix is to specify the correct absolute URL rather than deferring to the component (as opposed to, say, changing the value of `Plek.new.website_root`, which may have unintended side-effects).

Tested in a Rails console on Production:

```
"#{Plek.external_url_for('content-publisher')}#{app.guidance_path}"
=> "https://content-publisher.publishing.service.gov.uk/documents/publishing-guidance"
```

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -125,7 +125,7 @@
             attributes: { "data-gtm": "footer-content-publishing-guidance" }
           },
           {
-            href: guidance_path,
+            href: "#{Plek.external_url_for('content-publisher')}#{guidance_path}",
             text: "What to publish on GOV.UK",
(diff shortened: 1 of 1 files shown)


ryanb-gds (approved):
🔗

PR #3150, opened 2024-05-30
Updated political organisation config

ryanb-gds:
This is part of preparation for the July 2024 general election.

This list was generated by using the below command in the Whitehall integration environment, as per the documentation at the top of the config file.

```ruby
puts Organisation.where(political: true).order(:name).map { |o| "- '#{o.content_id}' # #{o.name}" }.join("\n")
```

Trello: https://trello.com/c/q90F4zsa


ryanb-gds:
Tested on integration and it seems to do the trick. Once the document has been resynced the political status is updated

ChrisBAshton (approved):

PR #3260, opened 2024-09-20
Get tests to run locally

minhngocd:
The Github pipelines are checking out helm charts, whereas it is not doing so locally, so tests are failing locally.
 
Changing the test to look at a relative path version of helm-charts so it passes both locally and in CI. Unfortunately this means that we are tied to having the helm charts repo checked out to get it passing locally.


lauraghiorghisor-tw:
I wonder if it would be good to note down in the documentation somewhere, that you'd need to check out the helm charts.

ChrisBAshton (approved):
LGTM 👍 agree with Laura we could add a quick note about the requirement, in https://github.com/alphagov/content-publisher?tab=readme-ov-file#running-the-test-suite

PR #3288, opened 2024-10-18
Fix hosts regex

ChrisBAshton:
See https://github.com/alphagov/contacts-admin/pull/1483#issuecomment-2421660720

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -94,7 +94,7 @@ Rails.application.configure do
 
   # Enable DNS rebinding protection and other `Host` header attacks.
   config.hosts = [
-    /content-publisher\..*gov.uk?/,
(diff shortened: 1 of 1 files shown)


config/environments/production.rb
```diff
@@ -94,7 +94,7 @@
 
   # Enable DNS rebinding protection and other `Host` header attacks.
   config.hosts = [
-    /content-publisher\..*gov.uk?/,
+    /content-publisher\..*\.gov.uk?/,
```
brucebolt, line 97:
```suggestion
    /content-publisher\..*\.gov.uk$/,
```
I'm not sure why I put the `?` originally, but a `$` is safer here as it'll restrict to only `.gov.uk` TLDs and not allow domains without the suffix.

brucebolt (approved):

PR #3310, opened 2025-08-29
Inline markdown-toolbar-element (WHIT-2020)

ChrisBAshton:
markdown-toolbar-element is only used by Content Publisher and
Collections Publisher. Content Publisher is due for imminent
retirement, at which point it will only be being used by
Collections Publisher. We're therefore inlining the module in
these publishing apps, so that we can retire the repo and have
one less component to maintain.

We're only copying over the required compiled JS for Content
Publisher, as it's being retired soon - so there's little value
in copying over the source code, tests etc.

JIRA: https://gov-uk.atlassian.net/browse/WHIT-2020

---


ChrisBAshton:
Actually, let's just continue to pull the module from NPM.

